PR #2, opened 2020-07-12
Don't spend build time optimizing build-time-only crates

joshtriplett:
Crates used exclusively at build time, such as proc-macro crates and
their dependencies, don't benefit substantially from optimization; they
take far longer to optimize than time saved when running them during the
build. No machine code from these crates will appear in the final
compiled binary.

Use the new profile-overrides mechanism in Rust 1.41
(https://doc.rust-lang.org/cargo/reference/profiles.html#overrides) to
build such crates with opt-level 0.

This substantially speeds up build time.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,6 +77,9 @@ panic = 'abort'
 codegen-units = 1
 incremental = false
 
+[profile.release.build-override]
+opt-level = 0
+
 [workspace]
 members = [
     "gitoxide-core",


joshtriplett:
On a 96-way system, this reduced compilation time from:
```
real	1m11.120s
user	3m19.392s
sys	0m7.647s
```
to
```
real	0m44.304s
user	2m22.186s
sys	0m7.739s
```

Byron:
Thank you **so** much for this contribution! You see I have tried to control dependencies during every step I make, but this kind of optimization is new to me and so impactful!
And it appears this flag should be on by default, so I will try to remember to turn that on from hereon out!

Byron:
@joshtriplett Thread limiting is now available in `main`, and it can be used like so: `giop -t 2 verify-pack <pack>` - this command would only use up to two cores.

PR #13, opened 2020-08-14
Cargo.toml: Use SPDX license syntax ("OR") in license field

joshtriplett:
https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields
> Previously multiple licenses could be separated with a /, but that
> usage is deprecated.


Byron:
Thanks a lot!
I think I copied that from the flate2 crate.
Furthermore I wholeheartedly admit that I have no clue why people want a dual license option or which difference it makes, but just know that many popular projects do that these days 😅.

joshtriplett:
@Byron 

https://internals.rust-lang.org/t/rationale-of-apache-dual-licensing/8952/3 explains the rationale for the dual license commonly used in the Rust ecosystem.

Byron:
Thanks a lot!
Amazing you are involved there as well.

PR #14, opened 2020-08-14
Fix binary install example

bombsimon:
Seems like the repository has been renamed so the installation for the binary release didn't work. Updated to the current name.


Byron:
Great catch, thank you!

Issue #17, opened 2020-08-20
commit-graph parser


avoidscorn:
I have a project that would benefit from being able to make reachability queries using only commit-graph files, so I'm interested in adding support for reading commit-graph files to this project.

The README mentions commit graphs would be part of `git-repository`, but commit graphs feel more like `git-odb`'s pack indices to me. In particular, I think it could share the fan-out and oid-bisect code from `git-odb/pack/index`. The reachability code could still be in `git-repository`.

So:
1. Do you have a strong preference where commit graph I/O should go?
2. I assume adding read-only support is acceptable for now. Is this correct?

Byron:
Thanks for your interest! 

Indeed, the git-repository is just a catch-all, and I generally organize code by its dependencies to other crates. From [the first looks](https://git-scm.com/docs/commit-graph#:~:text=The%20commit-graph%20file%20is%20a%20supplemental%20data%20structure,or%20in%20the%20info%20directory%20of%20an%20alternate.) it seems that it's a fairly well separated concept. Strictly optional, and really a file format to help speed up certain kinds of commit graph queries.

Reading the [fanout](https://github.com/Byron/gitoxide/blob/1c2bcbd3f510dddfc43e3f02a7987890306d8db7/git-odb/src/pack/index/init.rs#L84-L90) and [bisect](https://github.com/Byron/gitoxide/blob/584350af91f533db4cf980327d530445384c6b5a/git-odb/src/pack/index/access.rs#L115-L125) code is quite trivial, and I clearly prefer putting everything into its own crate, also because there is no other relation to an object database. There will be quite some support code needed for (_eventually_) creating and maintaining the format, and a crate appears like an optimal container for that.

To answer the questions more succinctly:

1. > Do you have a strong preference where commit graph I/O should go?

Their own crate, see the `git-commitgraph` crate I just created.

2. > I assume adding read-only support is acceptable for now. Is this correct?

Absolutely. It has to start somewhere and reading is usually easier than writing.

How do you imagine contributing to this project?

avoidscorn:
A separate crate sounds great!

> How do you imagine contributing to this project?

I'm not sure what you're asking exactly, but my answer is yes!

Byron:
I think I tried to ask about how that collaboration would look like.

Here are my suggestions on how I imagine the contribution process to look like, please chime in if there is something to discuss or alter.

* create a draft PR to allow early feedback
* use `quick-error` for errors exclusively
[...]

PR #21, opened 2020-09-15
[commitgraph] implement basic, low-level read API

avoidscorn:
This is just the first pass at ensuring we can read commit-graph
files correctly. It does not support anything interesting like
reachability, merge-base, or ahead-behind queries.


avoidscorn:
TODO before merging:
1. Add tests for non-happy-path cases.
2. Fill in docs.
3. Integrate into CI.

I wanted to settle on the test fixtures before making a PR, as I wanted to avoid polluting git history with too many fixture changes. But I can still split this into multiple PRs if desired.

Byron:
(Related to #17 )

Thanks a lot! This is a ton of work and a massive, much appreciated contribution!

> I wanted to settle on the test fixtures before making a PR, as I wanted to avoid polluting git history with too many fixture changes. But I can still split this into multiple PRs if desired.

Let's keep things as easy for you as possible, and keeping everything in one PR is absolutely alright with me.
Regarding the TODOs:

* Is it truly necessary to test for non-happy cases? If there are no `unwrap()` calls, I believe we are sure the program will not panic unexpectedly. Maybe an opportunity to save time.
* Docs are nice, and I usually write them this early only if I have to keep intricate, difficult knowledge (relative to myself at the time) somewhere. Maybe the tests are already doing this, but I do recommend waiting with excessive docs until there is a tool using and validating the current API (something that could be added in future PRs).
* For CI integration, have a look at the makefile. I would expect commit-graph tests to be picked up already, but there is where  you would setup anything special.

In any case, I would hope we can focus on making this PR mergable soon, and not extend its scope too much, rather the opposite.
I would also take a more thorough look in the next days if you are interested in suggestions and questions as they come up.

git-commitgraph/tests/fixtures/octopus_merges.sh
```diff
@@ -0,0 +1,37 @@
...
+set -eu -o pipefail
+
+export GIT_AUTHOR_DATE="2000-01-01 00:00:00 +0000"
+export GIT_AUTHOR_EMAIL=author@example.com
+export GIT_AUTHOR_NAME=author
+export GIT_COMMITTER_DATE="2000-01-02 00:00:00 +0000"
+export GIT_COMMITTER_EMAIL=committer@example.com
+export GIT_COMMITTER_NAME=committer
+
+mkdir -p "$1"
```
Byron, line 11: It really is wonderful to have a few shell scripts that use git itself to generate commit-graphs for common cases.

git-commitgraph/src/graph/init.rs
```diff
@@ -0,0 +1,104 @@
...
+        Self::from_single_file(object_dir.as_ref()).or_else(|_| Self::from_split_chain(object_dir.as_ref()))
+    }
+
+    pub fn from_single_file(object_dir: impl AsRef<Path>) -> Result<Self, Error> {
+        let single_graph_file = object_dir.as_ref().join("info").join("commit-graph");
+        let file = GraphFile::at(&single_graph_file).map_err(|e| Error::GraphFile(e, single_graph_file.clone()))?;
+        Self::new(vec![file])
+    }
+
+    pub fn from_split_chain(object_dir: impl AsRef<Path>) -> Result<Self, Error> {
```
Byron, line 61: With the disclaimer of having only partial knowledge of the format itself, let me share some thoughts.

I think all the init functions are a little smarter than they have to be, as the only two (or more) files we seem to be interested in is the `commit-graph*` ones in _some_ directory. Thus it would be enough to provide a directory path to a method, which will figure things out from there.

In these low-level crates I would refrain from enforcing structures usually seen in an actual git repository. This would also help with accessing fixtures in tests, which right now need nested directory structures including a pack directory even though commit-graphs (as far as I could tell) have no business with them.

git-commitgraph/src/graph/init.rs
```diff
@@ -0,0 +1,104 @@
...
+        let commit_graphs_dir = object_dir.as_ref().join("info").join("commit-graphs");
```
Byron, line 62: Pertaining to the comment above, this would probably mean that we wouldn't want to enforce the presence of the 'info' subdirectory here, but expect the caller to know where to find the files we are interested in.

Byron (commented):
I had a very rough look at a few pieces, and really like that the implementation tries to follow the overall style of what's already there in the other crates. If you have any feedback about that of things you would rather want to do differently, please let me know.

When I saw the V1 mentioned in the fixture directories, I was a little confused as there seems to be only one version of the commit-graph format. Maybe this was done akin to the fixtures in git-protocol, which does have to known formats right now (_unfortunately_).

I would be very interested to learn if you see interesting applications even in the context of a 'plumbing' tool, something like statistics maybe.

git-commitgraph/src/graph/init.rs
avoidscorn: Well sometimes there is a single `objects/info/commit-graph` file and sometimes there is an `objects/info/commit-graphs` directory of files. Are you saying passing the `objects/info/` directory instead of `objects/` is preferred? I'm fine with that.

git-commitgraph/tests/access/octopus_merges.rs
```diff
@@ -0,0 +1,140 @@
...
+const COMMIT3_INDEX: GraphPosition = GraphPosition(5);
+const COMMIT4: &[u8] = b"2ec748404455c0b0451b186851a815da13918ebb";
+const COMMIT4_INDEX: GraphPosition = GraphPosition(1);
+const THREE_PARENTS: &[u8] = b"7070cd0b59cd46a0f0fa0c50ca9148390ae1e7c7";
+const THREE_PARENTS_INDEX: GraphPosition = GraphPosition(4);
+const FOUR_PARENTS: &[u8] = b"5ca46a72d91817add3a7218f3eaf8b2b6c1caa13";
+const FOUR_PARENTS_INDEX: GraphPosition = GraphPosition(3);
+
+#[test]
+fn v1() -> Result<(), Box<dyn std::error::Error>> {
```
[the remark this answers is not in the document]
avoidscorn: Sure, I can remove the V1 stuff. Shall I keep the HashKind stuff around? It doesn't provide any value right now either.

git-commitgraph/tests/fixtures/octopus_merges.sh
avoidscorn: I committed the generated repos directly just to ensure we had v1 fixtures available even if git moved to v2. If we shouldn't worry about that yet, I'd just as soon have the tests run these scripts every time instead of pre-generating them.

avoidscorn:
I guess I should follow up with some terminology/style questions:
1. Are you okay with fully-spelled out type names like `GraphPosition`/`LexPosition`? Would you prefer `GraphPos`/`LexPos`?
2. I've been ordering struct fields/impl methods/enum variants in alphabetical order and putting private methods in separate impls. Any preference on this?
3. Do you have guidance on which symbols the top-level module should re-export?

git-commitgraph/tests/access/octopus_merges.rs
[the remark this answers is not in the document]
Byron: Thanks! I think HashKind might provide value since [the canonical implementation](https://github.com/git/git/blob/master/commit-graph.c#L47:L47) also seems to have abstracted over the used hash algorithm.
It seems this is something we should do as well, if not done already. Feel free to add what's needed to `HashKind` to support that.

git-commitgraph/tests/fixtures/octopus_merges.sh
Byron: Thank you! I wonder if there is a way to keep the generator scripts around, since it's usually a good idea to document where fixtures are coming from if feasible. The makefile has a [maintenance section](https://github.com/Byron/gitoxide/blob/791c05e08d9c6148fa0f89894c9b2abdadf3f503/Makefile#L158-L157), some of whose targets already regenerate fixtures. That should be a good fit.

git-commitgraph/src/graph/init.rs
Byron: Yes, that's what I meant. Thank you!

Byron:
Great! Let's do this.

> Are you okay with fully-spelled out type names like GraphPosition/LexPosition? Would you prefer GraphPos/LexPos?

I prefer fully spelled out names, but am fine if the shortened name is canonical. Notable example might be the term `ref`, which is also used in this codebase. That probably doesn't apply to `GraphPos` though.

> I've been ordering struct fields/impl methods/enum variants in alphabetical order and putting private methods in separate impls. Any preference on this?

No. I find myself using multiple impls as well, not for private/public separation, but to (arbitrarily) put related methods into a 'category' impl block which can even carry its own documentation.

> Do you have guidance on which symbols the top-level module should re-export?

I am usually keeping everything in their own modules until code that uses the crate really 'asks' for some top-level exports for convenience. An example for this was `git_packetline::PacketLine`, which is a rename of `git_packetline::Borrowed`. Maybe that's another hint at how the gitoxide crates name things - they try not to repeat their parent crate name or their module name.

I hope that helps!

git-commitgraph/tests/fixtures/octopus_merges.sh
avoidscorn: I meant I plan on keeping the `tests/fixtures/*.sh` scripts around but not commit their generated data to `tests/fixtures/v1`. The tests would run these scripts on demand.

Byron: That sounds good, great :).

Byron:
I saw some changes were made, and besides the windows build failures, is there anything I should look at in particular?

avoidscorn:
I believe this PR is ready for real review now, although it looks like the odb tests are bamboozling me.

I'd prefer these commits to get squashed when or before merging, but I don't know if you are okay doing that via github or whether I should do it after review is complete. I don't plan on signing commits anytime soon, so any merge strategy is okay with me.

Byron:
Alright, I will be working on it, pushing some changes to make the PR mergeable. I will let you know when done for a chance to hear what you think about the changes.

Byron (approved):
I finally managed to have a thorough look at each fo the ~1000 lines of code you are generously contributing, and am absolutely amazed about the level of quality I believe to see even with my limited understanding of the commit-graph data as a whole.

I dared to make minor refactorings mostly around `Position` and `LexPosition`.
The only question remaining is if there is anything you can imagine doing about the commented-out code in `file::init`. My suggestion is to remove it, while possibly keeping the note.

If there is nothing you want to change, I am happy to merge this PR and release a v0.1 to crates.io.

Hopefully you will find the time to continue your work on this crate or gitoxide as a whole! Something I would love to see is a stress-test which runs some plumbing command on big commit-graphs on real-world repositories, like the linux or rust repositories. Infrastructure for that exists already.

<details>

Below is my notes

  * [x] adjust tests to disable gpgsignatures
  * [ ] ~~do graph results need a reference to their owning file?~~
     * Yes, as it allows to obtain additional information related to the item in the file itself, like `File::commit_at(…)`
  * [ ] feature-toggled support for serde
  * [ ] ~~make tests depend on checked-in fixtures, instead of generating them (and depend on git on CI), making it easy to recreate them~~
     * the tests currently rely on calling git, see `inspect_refs(…)`
  * **Questions**
    * ~~How can `Commit` return Graph positions? It doesn't seem to learn about an offset.~~
      * Parent IDs are indeed specified as graph positions, not file positions, as they may be in previous commit graph files.
    * What to do with the ['extra-garbage'](https://github.com/Byron/gitoxide/blob/6f90beeb418480f9cd8bb7ae3b5db678b24103cb/git-commitgraph/src/file/init.rs#L248), 
      some code is commented out.
  * **Future Work**
    * A plumbing command to extract some value from the current implementation, maybe statistics, or verification
    * Application of the command above in a stress test

avoidscorn:
All your changes look good!

Plumbing for verification sounds like a good next step. I'll try to be more incremental next time, e.g. stubs in `gitoxide-core`, then a naive verifier, then a parallelized verifier, then a verifier with progress.

PR #27, opened 2020-10-03
Update error message for type name

bnbeckwith:
This change updates the error message in the `expect` call to display the correct message. If the `from_utf8` fails to unwrap it will be the result of _invalid_ utf8.

### DIFF
--- a/git-object/src/types.rs
+++ b/git-object/src/types.rs
@@ -91,7 +91,7 @@ impl Kind {
 
 impl fmt::Display for Kind {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_str(std::str::from_utf8(self.to_bytes()).expect("valid utf8 in kind name"))
+        f.write_str(std::str::from_utf8(self.to_bytes()).expect("invalid utf8 in kind name"))
     }
 }
 
--- a/git-object/src/types.rs
+++ b/git-object/src/types.rs
@@ -91,7 +91,7 @@ impl Kind {
 
 impl fmt::Display for Kind {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_str(std::str::from_utf8(self.to_bytes()).expect("invalid utf8 in kind name"))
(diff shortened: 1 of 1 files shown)


Byron:
Thank you! It's the question what to optimize for: Should the code read like plain English, or should the panic read correctly?  (_In this particular case it must never occour or it's a very strange bug - I use 'expect(…)' like an assertion_)

Whenever I read these 'inverse' expect statements in code, my brain goes into 'denial' because the statement is incorrect. "No, I am pretty sure we are expecting _valid_ UTF8 here, not _invalid_ one".

With that argument made, I hope you can see why I would pick code readability over panic readability. If so, please close this PR. Otherwise I would be interested to hear your thoughts on it.

But in any case, please keep perusing the code base and sending these PRs as you come across something fishy - it's much appreciated :)!

bnbeckwith:
I understand your point of view and sympathize because I often feel the same way. But it makes for a confusing error message as it was written.

So, I went back to the documentation on `expect` to see what to _expect_.  And I see that my solution wasn't correct either. It seems that the preferred way to use `expect` is to say what you expected to do, not what error you expect. So in this case, I changed the message to be the action we are expecting to take.

I think that solves the two issues. There are no mental hurdles to jump when reading the code, and the resulting panic message from any errors will read in a correct (and helpful) way.

Check out the [documentation on expect](https://doc.rust-lang.org/std/result/enum.Result.html#method.expect) to see if you agree. If not, I am happy to close the PR.

Byron:
Thanks so much for this discourse - I learned something today _and_ the solution we end up here is superior to the previous one :). The new phrasing reads correctly, and provides more information at the same time. Win - Win!

Also I am quite impressed you were able to extract such a conclusion from the linked documentation, because to me it doesn't exactly jump out. Now I wonder if you see a chance to put these learnings into words and improve the std documentation accordingly. I myself would have benefitted from a more graspable example, and probably spelling it out like you did in the surrounding text as well. If such a PR would be accepted, it could have a lasting positive effect for the entire community.

bnbeckwith:
Thank you for the kind words -- you may have motivated me to pursue a clearer explanation of `expect` at the source!

PR #29, opened 2020-10-15
Make `stress` target runnable in more general circumstances.

avoidscorn:
* If the `out/` directory already exists, delete it. Needed to run
  `stress` multiple times.
* If the gitoxide repo contains zero or multiple packs, repack it to
  contain exactly one pack. Needed when the gitoxide repo is not a fresh
  clone.


Byron:
Thanks a lot!

PR #30, opened 2020-10-16
Add stress tests for `gixp commit-graph-verify` command.

avoidscorn:
The new `tests/fixtures/commit-graphs` fixtures are built using the
exact same set of commits each time even as their sources repos in
`tests/fixtures/repos` get newer commits. My hope there is to make
upcoming commit-graph benchmarks more consistent.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,6 @@
 # will have compiled files and executables
 target/
 
+/tests/fixtures/commit-graphs
 # repositories used for local testing
 /tests/fixtures/repos
--- a/Makefile
+++ b/Makefile
@@ -129,6 +129,50 @@ $(linux_repo):
 	mkdir -p $@
 	cd $@ && git init --bare && git remote add origin https://github.com/torvalds/linux && git fetch
 
+test_many_commits_1m_repo = tests/fixtures/repos/test-many-commits-1m.git
+$(test_many_commits_1m_repo):
+	mkdir -p $@
+	cd $@ && git init --bare && git remote add origin https://github.com/cirosantilli/test-many-commits-1m.git && git fetch
+
+## get all non-rc tags up to v5.8, oldest tag first (should have 78 tags)
+## -> convert to commit ids
+## -> write a new incremental commit-graph file for each commit id
+tests/fixtures/commit-graphs/linux/long-chain: $(linux_repo)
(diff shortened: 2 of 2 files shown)


avoidscorn:
On a side node, I usually associate `$(MAKE)` being used to recurse into sub-Makefiles, not to explicitly make other targets in the same Makefile. Is there something I'm missing as to why the `stress` rule uses `$(MAKE)` instead of listing the other targets as prerequisites?

Byron:
The purpose of this is to control concurrency. The preparation step should be done in parallel, cloning all repositories at the same time, but the actual stress test should run sequentially.

If you see a more idiomatic or simple way to achieve this, I would be glad to see this as part of this PR. It might be time to refactor the stress testing facilities, after all they have grown organically and it might be time to factor them out.

Byron:
I have ran the stress test directly using `make stress-commitgraph` and got this result. Have you seen that before?

```
mkdir -p tests/fixtures/commit-graphs/linux/long-chain
rm -rf tests/fixtures/repos/linux.git/objects/info/*graph*
set -x && cd tests/fixtures/repos/linux.git && \
                for tag in $(git tag --list --merged v5.8 --sort=version:refname | grep -Fv -- -rc); do \
                        git show-ref -s "$tag" | git commit-graph write --split=no-merge --stdin-commits; \
                done
+ cd tests/fixtures/repos/linux.git
++ git tag --list --merged v5.8 --sort=version:refname
++ grep -Fv -- -rc
fatal: malformed object name v5.8
mv -f tests/fixtures/repos/linux.git/objects/info/*graphs* tests/fixtures/commit-graphs/linux/long-chain
mv: rename tests/fixtures/repos/linux.git/objects/info/*graphs* to tests/fixtures/commit-graphs/linux/long-chain/*graphs*: No such file or directory
make: *** [tests/fixtures/commit-graphs/linux/long-chain] Error 1
```

**fatal: malformed object name v5.8**

Byron:
In any case, thanks a lot! I believe this could be some local issue as I had a version of the linux repository locally, but possibly with only a partial/old pack that I use specifically for pack testing (1.3GB in size, more compressed than what github provides).

Thus it should work once I get a fresh clone of the repo, and CI agrees.

PR #39, opened 2021-01-24
README.md: Add a few items

joshtriplett:
Ran into this as a difference between libgit2 and git, and wanted to
document that it does provided a performance win.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,6 +252,7 @@ become available.
   * diffing of git-object::Tree structures
   * diffing, merging, working with hunks of data
   * find differences between various states, i.e. index, working tree, commit-tree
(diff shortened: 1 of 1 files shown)


Byron:
Thank you! I noticed this too and the one here will definitely be parallel!

After having read the VFS posts I am very curious to one day test it out with gitoxide, which should be faster there.

Some more information I am throwing your way while parting: the packaged git distributed with XCode on ARM is ridiculously fast and performs much better than what I have seen so far. Gitoxide may need some tweaks for Apple Silicon to be competitive. See also: https://github.com/seanmonstar/num_cpus/issues/104

PR #40, opened 2021-02-01
[WIP] Config parsing

Byron:
Generally there will be many small commits and a lot of 'sketching' in an effort to try find good ways of doing things, and iterate until it feels right, before going into all kinds of depths and details.

Thus there should be the first steps of basic parsing along with some pieces of editing and (re-)serialisation of config files to jiggle the API into something ergonomic and idiomatic.

### DIFF
--- a/git-config/src/lib.rs
+++ b/git-config/src/lib.rs
@@ -41,6 +41,8 @@ impl Span {
 pub mod file;
 pub use file::File;
 
+pub(crate) mod parse;
+
 /// A module with specialized value types as they exist within git config files.
 pub mod value;
 
--- a/git-config/src/lib.rs
+++ b/git-config/src/lib.rs
@@ -30,6 +30,15 @@ impl From<Span> for Range<usize> {
     }
 }
 
+impl From<Range<usize>> for Span {
+    fn from(Range { start, end }: Range<usize>) -> Self {
+        Span {
+            start,
+            end_inclusive: end - 1,
+        }
+    }
+}
+
 impl Span {
     /// Convert a span into the standard library range type.
     fn to_range(&self) -> Range<usize> {
--- a/git-config/src/parse.rs
+++ b/git-config/src/parse.rs
@@ -1,7 +1,7 @@
 use crate::file;
 use dangerous::{BytesReader, Error};
 
-fn read_config<'i, E>(r: &mut BytesReader<'i, E>) -> Result<Vec<file::Token>, E>
+fn config<'i, E>(r: &mut BytesReader<'i, E>) -> Result<Vec<file::Token>, E>
 where
     E: Error<'i>,
 {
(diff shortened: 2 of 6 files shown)


Byron:
Hi @avitex,

I finally pulled myself together to face my fears: Write the git config parser using `dangerous` 😅. Of course this is a bit of a dramatisation as I hope you may find a moment to take a cursory look to call out my blunders if there are any :).

Thus I am super interested to hear what you think, and now I will go and point out a few places where I think I might be doing something wrong.

Thanks in advance.

git-config/src/lib.rs
```diff
@@ -30,6 +30,15 @@ impl From<Span> for Range<usize> {
     }
 }
 
+impl From<Range<usize>> for Span {
```
Byron, line 33: Since `dangerous` uses a `Range<usize>` I might just use this instead of an own version of `..=`. Probably I had a reason to use an inclusive range in the first place, but it's not entirely clear to me right now what that was 🤦‍♂️.

git-config/src/parse.rs
```diff
@@ -0,0 +1,139 @@
...
+
+fn config(bytes: &[u8]) -> Result<Vec<file::Token>, Expected<'_>> {
+    fn config<'i, E>(input: Bytes<'i>, r: &mut BytesReader<'i, E>) -> Result<Vec<file::Token>, E>
+    where
+        E: Error<'i>,
+    {
+        let mut tokens = Vec::new();
+        skip_whitespace_or_comment(r, ConsumeTo::NextToken).map(|section| {
+            tokens.push(spanned::Comment(
+                dangerous::input(section)
```
Byron, line 12: In theory this conversion would have to be done for each token we parse here. Certainly this can be sugared up a bit more to feel less cluttered, but might also be a case for just using `zc`.
After all, the `File` type really is an `zc` type, owning data and keeping offsets into the data. The data will always be read into memory I think, so that would match too (as opposed to memory mapping them).

git-config/src/parse.rs
```diff
@@ -0,0 +1,139 @@
...
+                .len();
+            if last == current {
+                break;
+            }
+            last = current;
+        }
+    }
+    let parsed = r
+        .take_consumed(|r| skip_whitespace_or_comment(r, to_where))
+        .as_dangerous();
```
Byron, line 60: Here I thought there could be an `as_dangerous_opt()`, as I specifically avoided doing `as_dangerous_non_empty().ok()` probably due to a very premature optimization. Generally I like the `_opt()` versions of methods a lot.

git-config/src/parse.rs
```diff
@@ -0,0 +1,139 @@
...
+    input.read_all(|r| config(dangerous::input(bytes), r))
+}
+
+enum ConsumeTo {
+    NextToken,
+    EndOfLine,
+}
+
+#[must_use]
+fn skip_whitespace_or_comment<'a, E>(r: &mut BytesReader<'a, E>, to_where: ConsumeTo) -> Option<&'a [u8]> {
```
Byron, line 30: Yes, this was a little lazy, and this should really be refactored to not need the inner function. One would probably end up with something cloning the input, but… I don't actually know how to do this properly.

avitex:
Hey @Byron, awesome to see this! I'll check this out tonight and give you any feedback!

Byron:
Note that this branch will be replaced with #44 when it exits that draft status.

PR #43, opened 2021-02-11
[commitgraph] Tweak and expand documentation.

avoidscorn:
* Use "number of" phrasing instead of "amount of."
* Describe the graph as an index instead of a cache.
* Add module-level documentation for `file::commit`, `file::verify`, and
  `graph::verify`.

Style-wise, I like to start new sentences on new lines to reduce the amount of reflow needed when updating comments, but I don't know whether that's acceptable in this project.

### DIFF
--- a/git-commitgraph/src/file/access.rs
+++ b/git-commitgraph/src/file/access.rs
@@ -9,7 +9,7 @@ use std::{
 
 /// Access
 impl File {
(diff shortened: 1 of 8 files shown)


Byron:
Thanks so much!

Please feel free to avoid reflow as you see fit now and in future:).

I will make a new release shortly.

PR #44, opened 2021-02-18
Config updates

avitex:
@Byron 

I'm just starting to have a look now.

- [`dangerous::Span`](https://docs.rs/dangerous/0.9.0/dangerous/struct.Span.html) is now used instead. This actually is less error prone because ranges could be taken from the wrong offset. `dangerous::Span` internally holds pointer references to solve this!
- I've removed for now `PartialOrd/Ord` derives as I'm not sure what is semantic in the order, maybe you can help me here. `dangerous::Span` doesn't implement it because it doesn't make too much sense to me, which prevents derives. On a quick search, neither does `Range` https://github.com/rust-lang/rust/issues/75607 

I'll be looking a little more though this soon I promise!

On another note, I'm thinking of making a recommendation of only using `dangerous::input()` for the top level and not within sub parsers. I'll be explaining my reasoning soon :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -466,8 +466,9 @@ dependencies = [
 
 [[package]]
 name = "dangerous"
-version = "0.7.0"
-source = "git+https://github.com/avitex/rust-dangerous#3b6863117353d3ee17113a99c1db19f6460f1731"
+version = "0.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e95a55c47c2c5ce469fd39840be3ec8554765184a119cbcd20149dafaa3d75b4"
 
 [[package]]
 name = "dashmap"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -467,8 +467,7 @@ dependencies = [
 [[package]]
 name = "dangerous"
 version = "0.7.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


Byron:
If this is 'just looking' then I feel the thorough version of this will be a complete implementation - I wouldn't mind that at all ;).
Joking aside, thanks so much for taking a look, I am super happy with everything I see.

Especially the `Span` handling is so much nicer on the eyes - mine felt a bit lacking, not to say 'hacky just to make it work'. `PartialOrd/Ord` is nothing I am particularly interested in, but had it because I _could_ have it. If equality would be interesting one day I am sure one could cook up an implementation by hand as well.

> On another note, I'm thinking of making a recommendation of only using dangerous::input() for the top level and not within sub parsers. I'll be explaining my reasoning soon :)

My guess is that using `input()` multiple times allows to put in the _wrong_ input, and thus needs some care. That's what I felt if I remember correctly.

I won't take my eyes of this PR and have it replace my branch as soon as you give the go!

Thanks again, your work is _invaluable_!

Byron:
I have to close this PR as it was superseded by a contributed implementation which is now available in `main`. It was originally implemented in `nom` but it's possible to rewrite the parser in `dangerous` if any roadblocks are hit.

Thanks a million for your help here ❤️, and even though I am glad to have a `git-config` parser available for `gitoxide` I would have loved to see it realized in `dangerous` maybe even by myself 😅.

PR #48, opened 2021-03-09
Merge serde-git-config

edward-shen:
For the sake of eventually getting this somewhere (instead of just infinitely working on a 0.1.0), I've decided to merge in the current work I have now, which I think is a sufficient stopping point for an initial 0.1.0 release. Right now, a user should be able to completely create a new config from scratch or modify an existing one to their effective desired output (but perhaps not at an ergonomic level that we'd like).

I'm not sure if you'd rather have one large commit or the full history (which admittedly my commit messages haven't been the best). I can merge all the commits down into one if you'd rather have a more conformant history.

### Merging strategy

The primary method of merging was:
 - Remove everything in the git-config subdir
 - apply all commits from my repo to the git-config subdir

I've verified that `cargo build` works on the root repo without any errors.

### What's in this release?

For users:
 - zero-copy parser of git-config files
 - `GitConfig` wrapper, with support for the following:
   - Direct reading vars and multivars
   - Verified 1-1 reconstruction of git-config files (through `Into<u8>`, `to_string` only works when only utf-8 values are in the file)
   - Section API
      - Removing, pushing and renaming sections
      - Hashmap-like API for key-value pair manipulation
        - Push, pop, set, remove, value, values (multivar)
   - Raw value manipulation (e.g. value-oriented API rather than section oriented)
     - editing all multivars and values
   - errors with line number and last thing it tried to parse
   - Comments and whitespace are preserved
   -  Rust wrappers for git-config types
  
Note that this implementation attempts to be correct as possible. This includes:
 - Support for legacy and modern subsection
 - Case insensitivity for section and key names
 - validation of section and key names
 - All key values must be inside a section

For development:
 - 80+% test coverage, including doc tests
 - fully documented public items
 - Initial benches and fuzzing
 - Forbidden unsafe code, missing docs denied by default, and warn on the more pedantic clippy lints

### What's not in this release

 - Serde support (feature is disabled, no way of enabling it right now)
 - Tiered config support (system, global, local querying, although this can be very roughly achieved with 3 separate file instances)


Byron:
Hi @edward-shen, my apologies for the late reply and…I am super happy to see this PR from the summerly heat of my vacation home!

To cause no further delays, I have invited you to become a contributor for this repository. That would allow you to merge this PR yourself if you chose to accept. The idea would be to enable you low-friction work directly in `main` without having to create PRs every time (which is easy enough using the `gh` tool, but still…), and whenever you are interested in specific feedback the comments system of github could be used on each commit. For bigger chunks of work, you are invited to create a PR and I will be happy to follow along.

After merge I will hopefully soon find the time to read every line from start to end, allowing myself to apply minor fixes myself as I find them. Everything else would go through a PR of my own creation so we can collaborate :).

Once done, I will move forward and publish the first version of `git-config` and make you the co-owner, allowing you to create all future releases, too. From my point of view, you are the owner and maintainer of the crate and I see myself as contributor, even though we both have full permissions to this repository.

As with everything, please see all of the above as an invitation for a conversation, I will be glad to hear what you think and adjust everything to the point where it feels right.

Finally, please feel free to merge this PR :).

edward-shen:
Ah, please never feel like you're obliged to respond as soon as you see a ping from me. and especially since you're on vacation right now!

The review process sounds good to me. Thank you for letting me become part of `gitoxide`! :tada:

happybeing:
Watching from the sidelines it is great to see gitoxide ticking along. Thank you both. 🙏

Byron:
@edward-shen Maybe something of interest for you: I have recently implemented the `gix organize` sub-command which calls out to `git config` to obtain information about remotes. It should be trivial to use `git-config` instead and sever the ties to the `git` program. I would expect that to greatly increase performance, too.

(See 80d26267dfb3080b047c863778459c027f6d03d3 for an updated tasks.md file which I use to keep track of everything to be done to reach the next milestone, and as a scratch pad :D)

Byron:
@edward-shen I just released version 0.1 of git-config (admittedly without reading every single line from top to bottom, but I will get there). Also you have been invited to become an owner of the crate to allow you make releases at will.

PR #49, opened 2021-03-12
Have gix organize use git-config

edward-shen:
This PR implements the suggestion from [this comment](https://github.com/Byron/gitoxide/pull/48#issuecomment-797167436).

The parsing test was removed from gitoxide-core and a related test has been added to the git config tests instead.

### DIFF
--- a/gitoxide-core/src/organize.rs
+++ b/gitoxide-core/src/organize.rs
@@ -1,5 +1,8 @@
 use bstr::ByteSlice;
+use git_config::file::GitConfig;
 use git_features::progress::Progress;
+use std::convert::TryFrom;
+use std::io::Read;
 use std::path::{Path, PathBuf};
 
 #[derive(Copy, Clone, Eq, PartialEq)]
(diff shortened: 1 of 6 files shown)


gitoxide-core/src/organize.rs
```diff
@@ -72,21 +75,13 @@ where
...
-            "git invocation failed with code {:?}: {}",
-            out.status.code(),
-            out.stderr.as_bstr()
-        )
-    }
+    let mut config_bytes = vec![];
+    let config = {
+        let mut file = std::fs::File::open(repo.join("./.git/config"))?;
+        file.read_to_end(&mut config_bytes)?;
+        GitConfig::try_from(&config_bytes).map_err(|e| e.to_owned())?
```
edward-shen, line 82: Idk if we want the error type to just always own the bytes or not. If we do then we don't need to have the `map_err` call... On the other hand if we do we're guaranteeing a performance difference between a normal case or on an error, since we might alloc a potentially very large string.

edward-shen:
Not quite too sure where why the CI is failing. I tried running `gix organize` and got:
```
edward $ pwd
/home/edward/Documents/repos/gitoxide
edward $ cargo run -- organize
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `target/debug/gix organize`
 19:06:40 organize WOULD move . to /home/edward/Documents/repos/gitoxide/github.com/Byron/gitoxide.git
edward $ echo $?
0
```

Byron:
Thanks a lot! Did you try running `make journeytests` locally to reproduce the issue? It looks like it changed its exit code.

edward-shen:
That gave the hint needed to debug the test. Thanks--CI should pass now.

gitoxide-core/src/organize.rs
Byron: That looks good to me, but if the alternative would be more readable it could be preferable.

Byron:
Great 👍! Please feel free to merge when CI is happy.
I can’t wait to try it and see how much faster it has become.

Byron:
This is soooo amazing, as my invocation of `gix organize` is now instantaneous, taking only ~80ms to run and check all 200 local clones 🙌. 

Maybe if you find the time you could have  a look at the `git-config` section in the README and update the crate status there. It's useful to see where we are and what's to come in future.

And in case you are interested in implementing more pieces of `gitoxide`, the `tasks.md` file currently contains the path towards being able to clone, the next milestones being git-index handling as well as basic checkouts of working trees (even though that could be delayed if one starts with bare repositories).

edward-shen:
For context, what was the speed before? Sounds like a big improvement!

I'll go and update the readme-- I thought I already did but it looks like I only updated the tasks.md, oops!

Byron:
Thanks a bunch!

Previously it must have been between 3 and 5 seconds even if no work was done has one git invocation was needed even without doing anything to get the remote url.

Massive improvement indeed!

(Independently of that, gix also has a subcommand to find all git repositories in a directory which helps to build automations on top, also quite useful with `skim`).

PR #50, opened 2021-03-15
Use zlib-ng by default

edward-shen:
This PR removes `miniz_oxide` and instead opts into using `zlib-ng` by default. See #47 for a discussion on this.

In doing so, some of the code base was refactored to remove redundant structs.

Note that this also opts `git-transport`'s http-client dependency to use `curl` with `zlib-ng-compat`.


git-odb/src/loose/object/decode.rs
```diff
@@ -94,7 +94,9 @@ impl loose::Object {
             })?;
             self.compressed_data = SmallVec::from(buf);
         }
-        self.decompressed_data = SmallVec::from(decompress_to_vec_zlib(&self.compressed_data[..]).unwrap());
+        let mut vec = Vec::with_capacity(std::cmp::min(self.compressed_data.len() * 2, usize::MAX));
+        Decompress::new(true).decompress_vec(&self.compressed_data[..], &mut vec, flate2::FlushDecompress::None)?;
+        self.decompressed_data = SmallVec::from(vec);
```
edward-shen, line 99: This is what `decompress_to_vec_zlib` does behind the scenes, if my understanding of their source code is correct.

On that note--do we actually use `SmallVec` properly here? Originally and on L95 we're instantiating an instance of `SmallVec` from a `Vec`, which means that `Vec` already performs the heap allocation if it's non-empty, and so we never actually inline anything?

Byron: I agree, there it would by necessity reuse the given `Vec` and if anything add a layer that does nothing for us.

If there is no other, correct, use of that smallvec I think it would better be removed. Please note that I didn't look into it yet but will try soon.

Byron:
This is awesome work and [with tremendous impact](https://github.com/Byron/gitoxide/discussions/47#discussioncomment-488330).

I will take a proper look at this PR soon.

edward-shen:
Hmm, looks like there's a test failure... but as far as I can tell this should've been a drop in replacement. Any tips?

Byron:
I think this might be a situation where there should not be a debug assert, but an actual error. The test is operating on a truncated, partial pack and I am sure that zlib stream is incomplete. For some reason the previous zlib backend managed to return the expected buffer, however, maybe that was partially unwritten to cause a decode failure later.

My suggestion: Turn that debug assert into returning an error and potentially adjust the test.
Then we can see how the stress tests are fairing, but I would expect no issue there as the kernel pack can already be decoded properly.

edward-shen:
Looks like the only change needed was to return a new error variant, which would mean that your insight was correct!

Byron:
In a few hours I will (finally) take a look, can’t wait to merge!

On another note: do you mind me mentioning your name on a little status update in Reddit to give you credit for the git-config and performance work you have done?

edward-shen:
Sure, feel free to mention my name or my reddit username (/u/master7432)!

Byron (approved):
It's all great, thanks a lot!
I am not quite sure if `curl` needs zlib-compat-ng to be meaningfully faster, so maybe it might be better to remove it if it for some reason increases file size. Will test that in a moment.

edward-shen:
All dependences need `zlib-compat-ng`, since the `libz-sys` crate will default to not using the `-ng` variant unless all crates that use it enable `zlib-compat-ng`. See [here](https://github.com/rust-lang/libz-sys/blob/main/README.md) for more info on that.

Byron:
If you would like to get into the crazy complexity of feature toggles as employed in gitoxide, you could try to configure `gitoxide` so that the `small` version uses the `miniz-oxide` backend, whereas the fast versions use the zlib-ng one. That way one can get a 'pure' Rust build.

I will put that into tasks.md in a moment as I update some files.

PR #64, opened 2021-04-05
git-odb: Avoid double-lookup in packs without polonius

joshtriplett:
Split object lookup into two steps: looking up the object index, and
looking up the object itself given the index. This avoids passing in the
buffer (and thus looking like an unconditional borrow to non-polonius)
until we're committed to returning from the loop.

### DIFF
--- a/git-odb/src/compound/locate.rs
+++ b/git-odb/src/compound/locate.rs
@@ -33,18 +33,8 @@ impl compound::Db {
             }
         }
         for pack in &self.packs {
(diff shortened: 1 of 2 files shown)


joshtriplett:
This passes `cargo check`, and it should substantially improve the benchmark numbers.

Inspired by the HashMap entry API.

I think it'd probably be a good idea to create a wrapper newtype for the index, as an extra precaution to prevent passing in an index that didn't come from a lookup. But since the functions are only accessible inside the crate, and they're not unsafe, that may be overkill.

PR #65, opened 2021-04-05
git-odb: Only check alternates for objects not found in packs or loose

joshtriplett:
This matches the behavior of git.

Based on https://github.com/Byron/gitoxide/pull/64 , to avoid conflicts.

### DIFF
--- a/git-odb/src/compound/locate.rs
+++ b/git-odb/src/compound/locate.rs
@@ -19,6 +19,15 @@ impl compound::Db {
         id: borrowed::Id<'_>,
         buffer: &'a mut Vec<u8>,
     ) -> Result<Option<compound::Object<'a>>, Error> {
+        for pack in &self.packs {
+            if let Some(idx) = pack.internal_locate_index(id) {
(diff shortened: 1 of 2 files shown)


Byron:
_(I am answering here also for #64)_.
All looks wonderful, thanks! Thinking about it, this lookup order is certainly the more performant one in the common case and even when alternates are used.

And regarding #64, that indeed works without polonius now and at the same speed as before. Indeed, doing the same for alternates is probably a bit more involved as the current algorithm is plain recursive, but I'd leave that optimization for when there is a usecase and hoping that polonius lands before that, too :D.

```
    Finished release [optimized] target(s) in 34.22s
19468 objects (collected in 46.608625ms
libgit2:  confirmed 52497587 bytes in 30.427708ms (639812 objects/s)
parallel libgit2:  confirmed 52497587 bytes in 139.6845ms (139371 objects/s)
gitoxide: confirmed 52497587 bytes in 269.653208ms (72196 objects/s)
parallel gitoxide: confirmed 52497587 bytes in 51.661083ms (376841 objects/s)
```

PR #80, opened 2021-05-08
Use Seek to skip large objects during indexing.

taralx:
It's a relatively small optimization, but nice when dealing with packs with huge blobs.

### DIFF
--- a/git-odb/src/pack/tree/from_offsets.rs
+++ b/git-odb/src/pack/tree/from_offsets.rs
@@ -4,8 +4,9 @@ use git_features::{
     progress::{self, Progress},
 };
 use std::{
+    convert::TryFrom,
     fs, io,
-    io::{BufRead, Read},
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot, great catch! I don't have repos with big blobs for testing but can see how this could improve performance. I remember trying to avoid seeks because they are actually slower than reading, and can see now how there may be a break-even point where it's faster to seek than to read.

In my tests the performance on the linux kernel pack was exactly the same, creating an index in about 1.18s, probably seeks don't really happen there as the entries are too small.

taralx:
FWIW my go to is https://android.googlesource.com/platform/frameworks/base. There's also some prebuilt repos with even larger stuff.

Byron:
Great, thanks for the hint. I will add this to my set of repositories to test with as something with big blobs.

PR #86, opened 2021-05-20
Add support for cargo deny

edward-shen:
As this project focuses on security, I think it's reasonable to enable `cargo-deny` checks throughout the repo to ensure that we don't run into trouble in the future. 

Specifically, this PR adds conservative checking for the following:
 - Deny on security advisories (I've discovered that `prodash` is in a tough spot from this)
 - Deny on unlicensed code

With allowed exceptions added for:
 - duplicate crates in dependencies
 - copyleft licenses
 - Two advisories blocked on pinning `prodash` to a specific version of `time`.

This should hopefully make it easier to find new vulnerabilities as they're discovered in our dependency tree.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -8,7 +8,6 @@ on:
 
 jobs:
   build-lint-test-benchmark:
-
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
@@ -60,3 +59,20 @@ jobs:
         with:
           command: install
           args: gitoxide
+
+  cargo-deny:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        checks:
+          - advisories
+          - bans licenses sources
+
+    # Prevent sudden announcement of a new advisory from failing ci:
+    continue-on-error: ${{ matrix.checks == 'advisories' }}
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: EmbarkStudios/cargo-deny-action@v1
+      with:
+        command: check ${{ matrix.checks }}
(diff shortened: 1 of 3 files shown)


.github/workflows/rust.yml
```diff
@@ -60,3 +59,20 @@ jobs:
         with:
           command: install
           args: gitoxide
+
+  cargo-deny:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
```
Byron, line 66: I am loving this! There is so much I don't know about how GitHub actions can be used. This is definitely going into my toolbox :).

Byron (approved):
Thanks so much for this contribution! When I was working on prodash the other day I noticed what's really going on with the `time` crate and that it is a sour spot in the dependency chain for sure. With that I knew that configuring something like `cargo-deny` or `cargo-audit` isn't going to be straightforward and moved on.

This PR shows how it can be done nonetheless and is a helpful start to work on making the situation better. I will find a way to run it locally in order to not just ignore-and-forget this one forever.

Let's see how that goes!

edward-shen:
> I will find a way to run it locally in order to not just ignore-and-forget this one forever.

I think you should get a failed pipeline email if a new vuln ever gets discovered? So it might be okay to just keep it in CI. /shrug

Byron:
Even better - I think I misread the lint-level configuration for 'everything is warn'. That's clearly not the case though.
Now I also see how the `time` dependency is ignored for now.

Regarding `time` I have a strong feeling that the only way around it will be to have a well-working cross-platform `local time offsets` crate that is able to read system resources properly, and I truly hope that something like it arrives before `time` becomes an issue for `gitoxide`.

PR #88, opened 2021-05-22
Support pruning subgraphs from ancestor traversal.

avoidscorn:
This is meant to allow an upcoming `commit-graph write` command use the
existing traversal code to find new commits that haven't already been
indexed in an existing commit graph file.

It may also be useful for incremental pack generation.

I'm not exactly happy with the whole `OidPredicate` business, but I couldn't think of a better way.

### DIFF
--- a/git-traverse/src/commit.rs
+++ b/git-traverse/src/commit.rs
@@ -1,5 +1,6 @@
 ///
 pub mod ancestors {
+    use crate::commit::ancestors::predicates::AlwaysTrue;
     use git_hash::{oid, ObjectId};
     use git_object::immutable;
     use quick_error::quick_error;
@@ -41,12 +42,13 @@ pub mod ancestors {
     }
 
     /// An iterator over the ancestors one or more starting commits
-    pub struct Ancestors<Find, StateMut> {
+    pub struct Ancestors<Find, Predicate, StateMut> {
         find: Find,
+        predicate: Predicate,
         state: StateMut,
     }
 
-    impl<Find, StateMut> Ancestors<Find, StateMut>
+    impl<Find, StateMut> Ancestors<Find, predicates::AlwaysTrue, StateMut>
     where
         Find: for<'a> FnMut(&oid, &'a mut Vec<u8>) -> Option<immutable::CommitIter<'a>>,
         StateMut: BorrowMut<State>,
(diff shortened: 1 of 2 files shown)


git-traverse/src/commit.rs
```diff
@@ -109,6 +167,63 @@ pub mod ancestors {
...
+        /// 2. The default case (include all ancestors) should not increase the size of the
+        ///    [`Ancestors`] struct, nor should it require extra branches at runtime.
+        ///    So don't use boxed closures, `Option<impl FnMut...>`, or `fn(&oid) -> bool`.
+        ///    Instead, use a zero-sized type ([`AlwaysTrue`]) to handle the default case.
+        /// 3. Directly implementing `FnMut` is not yet supported in stable Rust.
+        /// 4. I could not get the compiler to accept [`Ancestors::new`] using a closure or static
+        ///    function as the default predicate. I don't think this is possible without using
+        ///    something like `typeof!(always_true)` (where `always_true` is a static function) as a
+        ///    type parameter.
+        pub trait OidPredicate {
```
edward-shen, line 189: So from what I can tell, `OidPredicate` is effectively a reimplementation of `FnMut`?

I'm a little confused w.r.t the 3rd point: what limitation are we running into with trying to directly use `FnMut`? We do something similar with the `Find` type (e.g. a `for<'a> FnMut`) so I would expect something similar to work?

git-traverse/src/commit.rs
```diff
@@ -109,6 +167,63 @@ pub mod ancestors {
...
+                true
+            }
+        }
+
+        pub struct FnMutWrapper<P>
+        where
+            P: FnMut(&oid) -> bool,
+        {
+            pub predicate: P,
+        }
```
edward-shen, line 215: Regarding point 2 of `OidPredicate`, isn't this effectively a function pointer? There's still a layer of indirection with this, but instead of using a fn pointer you're instead using a struct. New type pattern instead, maybe?

git-traverse/src/commit.rs
```diff
@@ -109,6 +167,63 @@ pub mod ancestors {
...
+        use super::*;
+
+        /// `FnMut`-like trait used for pruning commit subgraphs from the ancestors iterator.
+        ///
+        /// This trait only exists because:
+        /// 1. Predicate usage should be optional.
+        ///    The user should opt-in to using predicates via an alternate constructor or via the
+        ///    builder pattern.
+        /// 2. The default case (include all ancestors) should not increase the size of the
+        ///    [`Ancestors`] struct, nor should it require extra branches at runtime.
```
edward-shen, line 181: I agree with the second point, but is there significant harm in increasing the size? The difference should be at most a `usize`, in the case of using a function pointer, right? Are there regressions with a size increase?

edward-shen (commented):
Hey @avoidscorn, hope you don't mind someone else giving their opinions on this PR!

> I'm not exactly happy with the whole `OidPredicate` business, but I couldn't think of a better way.

Looks solid, but I agree that this doesn't feel the best. Should `OidPredicate` just be a subtrait of `FnMut` while we're at it (or perhaps put a blanket `impl` maybe)?

git-traverse/src/commit.rs
avoidscorn: Yes, `OidPredicate` is a reimplementation of `FnMut`.

I could not get a type parameter like `where Predicate: FnMut(&oid) -> bool` to work in an `impl Ancestors` block where `Ancestors::new` itself tried to supply the default value for `predicate`. The compiler complained that the default value supplied by `Ancestors::new` was not guaranteed to fit the caller-chosen type for `Predicate`.

From that error, I gather that the `impl Ancestors` block containing `Ancestors::new` needs to force a specific type for `Predicate`. I don't know of a way to get the exact type of a closure or static function to use in `impl<...> Ancestors<Find, ???, StateMut>` that doesn't rely on non-zero-cost trait objects or function pointers, so I went for a struct that could impl `FnMut` directly.

But then I hit the issue where trying to make a struct impl `FnMut` directly requires using `#![feature(fn_traits)]`, and I figured this was a non-starter. So `OidPredicate` was born.

`Find` should hit the same problem if you tried to provide a constructor that supplied a default value for it.

The closest discussion I found find related to this is https://stackoverflow.com/a/37313554.

git-traverse/src/commit.rs
avoidscorn: I didn't verify this, but I assumed that `P` and `FnMutWrapper` would be a zero-sized types with static dispatch when `AlwaysTrue` was used.

I think newtype would give the same results.

git-traverse/src/commit.rs
avoidscorn: Correct, if we can live with a `usize` increase, then we can use `impl<...> Ancestors <Find, fn(&oid) -> bool, StateMut>` and avoid all of the `OidPredicate` stuff. I'd also need to check whether that approach destroys the compiler's ability to inline the default predicate.

Byron (changes requested):
Thanks a lot for implementing filtering and for cause for foreshadowing your intend to create commit-graphs :)! I am quite confident that `gitoxide` will create them faster than `git` or at least be very much en par: Commit traversal is significantly faster if compared to `libgit2` at least.

When looking at the explanation of why the workaround exists it looks like it was motivated by performance optimizations or the wish for zero overhead and an issue with defaulting existing constructor with `always_true`.

Here are my thoughts on the performance implications.

* increasing the size of the iterator is no issue, there won't be millions of them.
* inlining the predicate is something that may or may not happen depending on what the compiler can do. The cost of this not happening can be neglected knowing that on my machine I only get about 200k commits per second, a laughable number showing that the CPU is busy with literally anything else (like DB lookups).

The other issue with defaulting the predicate without breaking existing code I think I have had before.

This is what worked for me then, could this be applicable here as well?

https://github.com/Byron/gitoxide/blob/51c63c081df4bd26adef7b8336034aee74237a86/git-packetline/src/read/async_io.rs#L142-L169

Thanks a lot for taking a look - if you want I can also give it a shot to get it to work with a plain `FnMut` stored in the iterator struct.

PS: I love the tests!

git-traverse/src/commit.rs
edward-shen: Looks like you're correct! https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9b7dc00fe4aaf269357b4b3c37f5091d

avoidscorn:
I replaced the `OidPredicate` stuff with just using `fn(&oid) -> bool` as the type parameter for the default include-all-ancestors case.

Abridged `traversal` numbers (disabled tree traversal and libgit2 steps) show no noticeable changes in performance on my system:
* `main`:
  ```
  $ command time -v cargo run --release --package traversal -- ~/src/linux/.git master
  [...]
  gitoxide (uncached): collect all 1013596 commits in 9.306947981s (108907 commits/s)
  gitoxide (cache = 57MB): confirmed 1013596 commits in 9.255322714s (109515 commits/s)
  gitoxide (static cache = 64 entries): confirmed 1013596 commits in 9.370033327s (108174 commits/s)
  gitoxide (uncached): confirmed 1013596 commits in 9.254398583s (109526 commits/s)
  [...]
	Maximum resident set size (kbytes): 846692
  ```
* With `OidPredicate`:
  ```
  $ command time -v cargo run --release --package traversal -- ~/src/linux/.git master
  [...]
  gitoxide (uncached): collect all 1013596 commits in 9.268462226s (109360 commits/s)
  gitoxide (cache = 57MB): confirmed 1013596 commits in 9.25482569s (109521 commits/s)
  gitoxide (static cache = 64 entries): confirmed 1013596 commits in 9.451119544s (107246 commits/s)
  gitoxide (uncached): confirmed 1013596 commits in 9.307447228s (108902 commits/s)
  [...]
	Maximum resident set size (kbytes): 846784
  ```
* With `fn(&oid) -> bool`:
  ```
  $ command time -v cargo run --release --package traversal -- ~/src/linux/.git master
  [...]
  gitoxide (uncached): collect all 1013596 commits in 9.268462226s (109360 commits/s)
  gitoxide (cache = 57MB): confirmed 1013596 commits in 9.25482569s (109521 commits/s)
  gitoxide (static cache = 64 entries): confirmed 1013596 commits in 9.451119544s (107246 commits/s)
  gitoxide (uncached): confirmed 1013596 commits in 9.307447228s (108902 commits/s)
  [...]
	Maximum resident set size (kbytes): 846784
  ```

Byron:
Thanks a lot!
Would you like to become a member of the team similar to @edward-shen? Nothing changes except that as a collaborator one can perform direct pushes to main to ease the overhead when moving `git-commitgraph` forward.

avoidscorn:
I'd just as soon keep using PRs for all of my changes. But if that is a burden for you, I'm open to direct pushes for `git-commitgraph`-only changes.

Byron:
Great, I invited you! Welcome to `gitoxide` 🎉🙌! 

Please feel free to do PRs for `git-commitgraph` related changes if you want feedback - it's just that I do trust you delivering great software on your own and want to show that by inviting you to the collaborator group.

I myself will also do only minor changes to `git-commitgraph` without consulting you, but for anything akin to API changes or feature additions, I would start a discussion and/or create PRs to get feedback early. Everything else we can certainly flesh out as we go :D.

PR #108, opened 2021-06-24
Fix skim link in README

kidonng:
Apparently the reference was removed for good in 013ecdc1482ff87248eed11aa4f52cd68451c526.


Byron:
Great catch, thanks!

PR #111, opened 2021-06-25
[git-transport]: make capabilities parsing public

kim:
Good to have for custom `Transport` impls.

See also: https://github.com/Byron/gitoxide/discussions/110


Byron:
Thanks a lot, appreciated! I particularly like the docs!

PR #113, opened 2021-06-28
support ref-in-want

kim:
Exercise to understand better how things interact with the `Delegate` pattern.

Tbh, I'm not entirely convinced -- it seems a bit confusing and potentially error-prone to handle this while having to conform to v2 paradigms at the same time. Perhaps it would be preferable to create a separate, v2-only path, and leave the `Delegate`-as-is to clients which _want_ to be backwards-compatible?


git-protocol/src/fetch/response/mod.rs
```diff
@@ -108,10 +119,25 @@ impl Acknowledgement {
...
+impl WantedRef {
+    /// Parse a `WantedRef` from a `line` as received from the server.
+    pub fn from_line(line: &str) -> Result<WantedRef, Error> {
+        match line.trim_end().split_once(' ') {
+            Some((id, path)) => {
+                let id =
+                    git_hash::ObjectId::from_hex(id.as_bytes()).map_err(|_| Error::UnknownLineType(line.to_owned()))?;
+                Ok(WantedRef { id, path: path.into() })
+            }
+            _ => unreachable!("cannot have an entirely empty line"),
```
kim, line 131: Ehm yes. I'm not sure about those `unreachable!`s here -- should they not better be `Err` results, just in case you're talking to something which is not actually a git server?

git-protocol/tests/fetch/v2.rs
```diff
@@ -50,3 +50,33 @@ async fn ls_remote() -> crate::Result {
     );
     Ok(())
 }
+
+
+#[maybe_async::test(feature = "blocking-client", async(feature = "async-client", async_std::test))]
+async fn ref_in_want() -> crate::Result {
+    let out = Vec::new();
+    let delegate = CloneRefInWantDelegate::default();
```
kim, line 58: Haha testing against fixtures :D

git-protocol/src/fetch/response/mod.rs
Byron: Agreed, since it's untrusted input it's probably a better idea to not crash the client process because somebody isn't conforming to the protocol.

git-protocol/src/fetch/delegate.rs
```diff
@@ -29,12 +29,16 @@ pub trait DelegateBlocking {
     /// Note that some arguments are preset based on typical use, and `features` are preset to maximize options.
     /// The `server` capabilities can be used to see which additional capabilities the server supports as per the handshake which happened prior.
     /// Note that this is called only if we are using protocol version 2.
+    ///
+    /// XXX: `Action::Close` here actually means "skip" for v2. Perhaps that should be a separate
```
Byron, line 33: Agreed, there should be a specialized enum to not 'overload' the meaning of `Close`, which in itself is too specific to really have another meaning.

Byron:
Thanks a lot for contributing! Great work!

> Tbh, I'm not entirely convinced -- it seems a bit confusing and potentially error-prone to handle this while having to conform to v2 paradigms at the same time. Perhaps it would be preferable to create a separate, v2-only path, and leave the Delegate-as-is to clients which want to be backwards-compatible?

I see the point and was thinking the same. Then I realized that besides the response parsing, both paths are very very similar after all. Maybe I just have trouble imagining how much simpler the V2 only path would be though. Assuming it would be considerably simpler for a moment then it would be another code paths x2 for blocking and async io, quite an effort for maintenance. For these cases I hoped one would use `git-transport` directly with the existing implementation(s) as reference.

Something I like about the current implementation is that it feels very much like V2 all around even though V1 is also supported - the Delegate doesn't have to know beyond not using capabilities which aren't supported.

In any case I am glad to have a fresh pair of eyes look at it more closely, there will be many improvements coming out of this PR beyond wanted-refs, for sure.

kim:
Ok yeah, I guess the tradeoff is between controlling the entire fetch flow and requiring `Delegate` impls to take care of _all_ state bookkeeping themselves. I'll see if any patterns emerge, maybe a higher-level interface can be layered on top at some point.

Byron:
> I'll see if any patterns emerge, maybe a higher-level interface can be layered on top at some point.

Thank you, appreciated! I totally agree that there should one day be a higher level implementation that makes this easier. All crates below `git-repository` are plumbing and they will expose all the details. `git-repository` one day will condense that into an API that is much more straightforward to use and more similar to what `git2` already offers.

PR #114, opened 2021-06-29
Make negotiation fallible

kim:
Packfile negotiation needs to access the odb in order to figure out `have`s, which can fail.

### DIFF
--- a/git-protocol/src/fetch/delegate.rs
+++ b/git-protocol/src/fetch/delegate.rs
@@ -1,7 +1,10 @@
 use crate::fetch::{Arguments, Ref, Response};
 use bstr::BString;
 use git_transport::client::Capabilities;
-use std::ops::{Deref, DerefMut};
+use std::{
+    io,
+    ops::{Deref, DerefMut},
+};
 
 /// Defines what to do next after certain [`Delegate`] operations.
 #[derive(PartialEq, Eq, Debug, Hash, Ord, PartialOrd, Clone, Copy)]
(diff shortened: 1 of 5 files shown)


kim:
I would also lean towards making the `prepare_*` methods fallible. 

The reason is that the server's capabilities are passed there, and it could very well be that the client wants to _abort_ the whole interaction if a feature is not supported by the server (eg. shallow fetches, object-format). `Action::Cancel` can be returned in this case, but the _reason_ for cancelling needs to be obtained through a side-channel. Maybe an `Action::Abort(Box<dyn std::error::Error>)` variant could be considered.

Byron:
Thanks a lot for yet another contribution! I do have a the feeling that you will be writing negotiation code before I get the chance, very exciting for sure!

> I would also lean towards making the prepare_* methods fallible.

I can definitely see that this would be useful, a way to not only cancel but to say why. From the perspective of, say, a user interface that would certainly be useful to have nice errors, even though I can't see the need to differentiate kinds of errors beyond`std::io::Result<Action>` and the `ErrorKind::Other` + downcast. The latter should allow to communicate anything as long as the Delegates are controlled by the application that needs distinguished errors.

Thus I would hope that upgrading `prepare_*` with `-> std::io::Result<Action>` would be sufficient for now, potentially in addition to state in the Delegate with additional information to be picked up by the caller (now passing `&mut delegate` is possible).

Only if all that isn't working at all we should see how `Action` can incorporate a side-channel, I like its current simplicity a lot.

kim:
> prepare_* with -> std::io::Result<Action> would be sufficient for now

Yes that works, too. I just like the fact that a signature `fn prepare_x() -> Action` conveys that you're _not_ supposed to have side-effects in your implementation (or at least, no I/O). Matter of taste, I guess.

Byron:
> Matter of taste, I guess.

I hope not, meaning that ideally the signature can represent real world usage. What about the following:

* `prepare_ls_refs(…)` stays as is but adds `LsRefsAction::Abort<Box<dyn Error + 'static>>` to allow aborting if something is wrong about the presented capabilities. This is the first and only time when the client should do that. Here side-effects are discouraged.
* `prepare_fetch(…)` return `std::io::Result<Action>` as it does see the Refs from the step before which gives it information that it might want to look up in the repository which is fallible. Side-effects are anticipated here.

Since the above sounds so right I will sketch it and you can take a look here.

kim:
> I could want to query git-configuration

Good point 👍

PR #125, opened 2021-07-16
Implement Parser::into_iter without extra allocation

In-line:


Byron:
Awesome, thanks!

edward-shen:
Thanks! TIL `std::iter::once` exists. Good to know!

PR #127, opened 2021-07-16
Remove unnecessary unsafe code

In-line:


Byron:
Thanks a lot, great work! I did do performance testing on the linux kernel pack via…

```
mkdir -p out && cargo build --release --bin gixp && /usr/bin/time -lp  ./target/release/gixp --verbose pack-index-from-data -p tests/fixtures/repos/linux.git/objects/pack/pack-3ee05b0f4e4c2cb59757c95c68e2d13c0a491289.pack out
```

…both on `main` and this PR without any degradation in performance. This is a little bit surprising as the code doesn't explicitly buffer `out` anywhere so I would have expected a lot of additional `write` calls to slow thing down considerably.

To share some numbers, without the trailing `out` argument in the line above it achieves optimal performance by writing to an in-memory sink at 550MB/s, whereas the version writing to disk achieves 440MB/s throughput for index writing.

It's still a bit of a riddle to me as from my experience, the amount of write calls and whether IO is buffered or unbuffered makes a huge difference.

In-line:
@Byron simple cases like this are in the domain of optimizer.

In-line:
@Byron in fact, safe version produces even better code.

Simple safe version does just `memcpy`.
```
playground::write_simple:
	pushq	%r15
	pushq	%r14
	pushq	%rbx
	subq	$1056, %rsp
	movq	%rsi, %r15
	movq	%rdi, %rsi
	movq	$256, 24(%rsp)
	leaq	32(%rsp), %rdi
	movl	$1024, %edx
	callq	*memcpy@GOTPCREL(%rip)
	xorl	%ebx, %ebx
	leaq	12(%rsp), %r14
```

While unsafe version calls `write_all`, which in fact calls `flush_buf`, which goes into more complicated code.
```
playground::write_unsafe:
	pushq	%rax
	movq	%rdi, %rax
	movl	$1024, %edx
	movq	%rsi, %rdi
	movq	%rax, %rsi
	callq	<std::io::buffered::bufwriter::BufWriter<W> as std::io::Write>::write_all
	movq	%rax, %rcx
	andq	$-256, %rcx
	xorl	%esi, %esi
	cmpb	$4, %al
	cmoveq	%rsi, %rdx
	cmoveq	%rsi, %rcx
	movzbl	%al, %eax
	orq	%rax, %rcx
	movq	%rcx, %rax
	popq	%rcx
	retq
```

[Playground link](https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=b495d1c354b8f388660a5b698a00614f)

Byron:
That's amazing! Such a good case for not ever doing premature optimization in disbelief that a loop could be optimized to such an extend.

Thanks for sharing.

PR #145, opened 2021-07-30
[protocol] remove misleading documentation about ref-in-want

kim:
It turns out that ref-in-want is simpler than previously thought:
`want-ref` can come any time in the negotiation process, and will result
in a packfile being sent immediately. Hence, remove some docstrings and
simplify the test delegate.

### DIFF
--- a/git-protocol/tests/fetch/mod.rs
+++ b/git-protocol/tests/fetch/mod.rs
@@ -80,27 +80,12 @@ impl fetch::DelegateBlocking for CloneRefInWantDelegate {
         Ok(Action::Continue)
     }
 
-    fn negotiate(
-        &mut self,
-        _refs: &[Ref],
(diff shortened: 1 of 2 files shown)


git-protocol/tests/fetch/mod.rs
```diff
@@ -175,9 +160,9 @@ mod blocking_io {
             mut input: impl io::BufRead,
             _progress: impl Progress,
             _refs: &[Ref],
-            previous: &Response,
+            response: &Response,
```
Byron, line 163: Sometimes it's called `previous_result`, sometimes `previous`, I think I will call it `previous_response` everywhere 😁.

Byron (approved):
Thanks a lot, simplifications like these are the best PRs I can think of!

PR #149, opened 2021-08-07
Canonicalize path when discovering repositories

jeffparsons:
Canonicalize the given path so that `Path::parent` _actually_ gives
us the parent directory. (`Path::parent` just strips off the last
path component, which means it will not do what you expect when
working with paths that contain '..'.)

### DIFF
--- a/git-repository/tests/discover/mod.rs
+++ b/git-repository/tests/discover/mod.rs
@@ -6,7 +6,11 @@ mod existing {
     fn from_bare_git_dir() -> crate::Result {
         let dir = repo_path()?.join("bare.git");
         let path = git_repository::path::discover::existing(&dir)?;
-        assert_eq!(path.as_ref(), dir, "the bare .git dir is directly returned");
+        assert_eq!(
+            path.as_ref().canonicalize()?,
+            dir.canonicalize()?,
+            "the bare .git dir is directly returned"
+        );
         assert_eq!(path.kind(), Kind::Bare);
         Ok(())
     }
@@ -17,8 +21,8 @@ mod existing {
         let dir = git_dir.join("objects");
         let path = git_repository::path::discover::existing(&dir)?;
         assert_eq!(
-            path.as_ref(),
-            git_dir,
+            path.as_ref().canonicalize()?,
+            git_dir.canonicalize()?,
             "the bare .git dir is found while traversing upwards"
         );
         assert_eq!(path.kind(), Kind::Bare);
@@ -31,8 +35,8 @@ mod existing {
         let path = git_repository::path::discover::existing(&dir)?;
         assert_eq!(path.kind(), Kind::WorkingTree);
         assert_eq!(
-            path.into_repository_directory(),
-            dir,
+            path.into_repository_directory().canonicalize()?,
+            dir.canonicalize()?,
             "the .git dir is directly returned if valid"
         );
         Ok(())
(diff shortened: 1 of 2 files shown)


jeffparsons:
Note that previously `discover::existing` would also fail to find the repo if your process's current working directory was somewhere nested under the git repo and you pass the path to discover from as `"."`. (That's how I actually found it.)

edward-shen (commented):
As general feedback on the tests, can we not just directly compare the `Result` instead of `.expect()`ing the output? 

e.g. 
```rust
assert_eq!(path.as_ref().canonicalize(), dir.canonicalize(), "desc")
```
instead of 
```rust
assert_eq!(path.as_ref().canonicalize().expect("..."), dir.canonicalize().expect("..."), "desc")
```

jeffparsons:
@edward-shen I don't think so. Unfortunately `std::io::Error` doesn't impl `Eq`, so you can't just throw the `Result` into `assert_eq!` as-is.

I thought about adding a `canonicalize_unwrap` "extension method" just for the tests, or something like that, but I think it may actually serve to obscure what's going on overall rather than helping readability.

jeffparsons:
The failing job looks unrelated to my changes, and I think it's failing on _main_ just the same.

```
Run cargo fmt --all -- --check
Error: `/home/runner/work/gitoxide/gitoxide/git-config` is a directory
```

edward-shen:
You're right that `std::io::Error` doesn't impl those traits. Looks like we return a `Result` in those tests, can we just use `path.as_ref()?` then?

As for CI errors, yeah that's on me, need to investigate.

jeffparsons:
> Looks like we return a `Result` in those tests, can we just use `path.as_ref()?` then?

I'm not sure I follow. We're comparing two paths to each other in each test, so I don't see what `Result` I could return from the test.

git-repository/tests/discover/mod.rs
```diff
@@ -31,8 +35,10 @@ mod existing {
         let path = git_repository::path::discover::existing(&dir)?;
         assert_eq!(path.kind(), Kind::WorkingTree);
         assert_eq!(
-            path.into_repository_directory(),
-            dir,
+            path.into_repository_directory()
+                .canonicalize()
+                .expect("should be a directory"),
+            dir.canonicalize().expect("should be a directory"),
```
edward-shen, line 41: Something like this:

```suggestion
            path.into_repository_directory().canonicalize()?,
            dir.canonicalize()?,
```

jeffparsons: Ah, yep, sure. Happy to make that change.
I'm on mobile now; will do it when I get a chance a bit later today.

And thanks for looking into the CI issue. 🙂

jeffparsons:
@edward-shen I've made that change and rebased on main. Should be good to go I think, but blocked on the workflow approval thing for now.

edward-shen:
Thanks for the contribution!

Byron:
Many thanks also from my side. Having just returned from vacation I am a bit late to the party and took the liberty to make a few adjustments. They should show up in this PR in a moment, but in short paths will now only be canonicalized when needed.

jeffparsons:
@Byron Awesome! I thought about doing something similar but worried it might be considered overkill. 😅

This definitely feels like having our cake and eating it, too.

Byron:
> @Byron Awesome! I thought about doing something similar but worried it might be considered overkill. 😅

Hehe, I think quality will always be appreciated, and I agree that often enough it's not exactly clear what tradeoffs to make.

If you are using `gitoxide` for a project of yours you are welcome to start a discussion to see how feasible this is right now or which contributions are needed to make it feasible.

PR #156, opened 2021-08-14
rustfmt: Reformat imports and module declarations more aggressively.

avoidscorn:
This requires rustup nightly.

This is more of a "run it up the flagpole" PR, but it fits how I would like to organize `git-commitgraph` at least.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,1 +1,5 @@
+group_imports = "StdExternalCrate"
+imports_granularity = "Crate"
 max_width = 120
+reorder_imports = true
+reorder_modules = true


Byron:
I would love to have these turned on for sure.
Would that mean I should work with nightly to avoid [these warnings](https://github.com/Byron/gitoxide/pull/156/checks?check_run_id=3327436901#step:4:4)?

I would be ok doing that but see the risk of formatting code into a shape that the CI running on stable won’t like.
Alternatively, CI fmt could only run for PRs.

What’s your thoughts, are there other scenarios that prevent me from having to see these warnings while reaping the benefits of nightly?

avoidscorn:
That's odd...I could have sworn the current version of `.github/workflows/rust.yml` had nightly rustfmt selected, but now I see that it was quite an old revision that did. Past and future, so confusing sometimes.

Since stable rustfmt should be [99% happy](https://github.com/rust-lang/rustfmt#limitations) with the output of nightly rustfmt, I'm fine keeping this config out of the repo/CI and just overriding it locally.

Byron:
Maybe you could create a PR that applies rustfmt nightly and run stable afterwards to fix things. That could be repeated from time to time.

Is that feasible?

Byron:
﻿Maybe you could create a PR that applies rustfmt nightly and run stable afterwards to fix things. That could be repeated from time to time.

Is that feasible?

avoidscorn:
Sure, that sounds good!

PR #158, opened 2021-08-14
Add `make rustfmt` target.

avoidscorn:
This is intended to be run on occasion to let us take advantage of
rustfmt features available only in nightly while not requiring all
contributors and CI to use nightly rustfmt.

If the output of nightly rustfmt is not accepted by stable rustfmt, then
the recipe will intentionally fail. For now, we prefer to detect when
the two versions disagree on formatting rather than let nightly and
stable rustfmt stomp on each other.

Replaces #156.


Byron:
Lovely, thanks a lot!

PR #161, opened 2021-08-15
Apply nightly rustfmt rules.

avoidscorn:
Applied via `make rustfmt`, which enables these nightly-only rules:
```
group_imports = "StdExternalCrate"
imports_granularity = "Crate"
```

### DIFF
--- a/cargo-smart-release/src/command/release/cargo.rs
+++ b/cargo-smart-release/src/command/release/cargo.rs
@@ -1,7 +1,9 @@
-use super::{utils::will, Options};
+use std::process::Command;
+
 use anyhow::bail;
 use cargo_metadata::Package;
-use std::process::Command;
+
+use super::{utils::will, Options};
 
 pub(in crate::command::release_impl) fn publish_crate(
     publishee: &Package,
--- a/cargo-smart-release/src/command/release/git.rs
+++ b/cargo-smart-release/src/command/release/git.rs
@@ -1,6 +1,5 @@
-use super::{Context, Options};
-use crate::command::release_impl::tag_name_for;
-use crate::command::release_impl::utils::will;
+use std::{convert::TryInto, process::Command};
+
 use anyhow::{anyhow, bail};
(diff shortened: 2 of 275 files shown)


edward-shen:
I am not a fan of crate level granularity, as it generally problematic with merge conflicts -- can we do module level instead?

Byron:
@avoidscorn Thanks a lot for this PR! @edward-shen I assume the request stems from the multi-file-config branch which would have some conflicts to resolve then. It is mere luck we are having this conversation as I was about to do exactly this last night but forgot (🤦‍♂️) to get back to it as I had to wait for the nightly installation for more than three seconds and moved to the next task 😅.  Please note that I encourage all kinds of refactorings while there are no PRs to consider as we do trunk-based development, and long-lived branches aren't necessary. Collaborators are of course invited to push to master often (`make tests check-size && gp` does it for me), no PRs necessary, when they work on their own crates.
Hence my strong recommendation is to rebase the branch onto `main`, work there and push working code often.

In the meantime, I have adjusted the commit above to exclude `git-config` and avoid merge conflicts, this time. Note that I plan to run `make rustfmt` often from now on as it will automate making `use` lines consistent.

edward-shen:
@Byron, thanks for working with me. To be clear, this is just a personal style choice, having a branch in the works was pure coincidence as well. If you're strongly preferring a `imports_granularity = "Crate"` over `imports_granularity = "Module"` across the entire project `gix` project, I don't mind doing that to maintain consistency. I just prefer `Module` over `Crate`. :laughing:

Byron:
Haha, so my assumption was incorrect 😂, and thanks for the clarification.

Screenshot 2021-08-16 at 07 57 21

The above is what happens with the proposed import granularity, and it's indeed very different from what I tried to do manually with intelliJs 'nest imports' function. The reason I apparently want `Crate` import granularity is that it is how I think about imports, with the added benefit of not looking like java. Even the resemblance of it gives me nightmares and drives up that therapist bill.

Issue #164, opened 2021-08-16
Transform radicle-link integration test setup (as much as possible) to gitoxide


Byron:
* [x] #171    
* [x] https://github.com/radicle-dev/radicle-link/pull/776
* [x] #185 
* [x] #188 
* [x] #190
* [x] `refresh_odb()` to show how pack changes are picked up.
* [x] look into being able to write ObjectRefs exclusively, and convert owned ones into shared ones.
* [x] add remaining git-repository docs
* [x] https://github.com/radicle-dev/radicle-link/pull/780

---- 

<details><summary>Archive</summary>

* **refactorings**
    * finally find a good solution for immutable vs mutable naming (which doesn't always work nicely. `Ref` suffix is a not 
      candidate for those with borrows inside. Candidates for massive changes are `git-object` and `git-ref` and `git-actor`. `borrow()` would then rather be `to_ref()`
    * ~~rename `oid` to `id` like in git2~~ - depends on the context, maybe another time
    * rename `find_existing()` to `find()` and provide the optional version under `try_find()`, both for objects and refs in plumbing crates
   * #165
   * #171 
   * #173
   * #175
   * #177
   * #178
   * #179
   * #180

</details>

PR #165, opened 2021-08-19
[repository] Sketch design of 'Easy' mode

Byron:
### Motivation

`git2` is actually quite easy to use, with a relatively high-level API that hides a lot of the plumbing that actually has to happen under the hood. Currently `gitoxide` has no such interface which requires a lot of knowledge to use and can be cumbersome at times.

This PR aims to provide a draft of a new `Easy` mode that provides a similarly simple API while hiding all the details. A major goal is to keep the lower level accessible for those who need it to eek out a few more percent of performance while offering a higher level that doesn't sacrifice too much  performance for convenience, striking just the right balance.

We will use `cargo smart-release` as a driver and validator for the new API as it touches references, objects and trees right now.

Note that this PR was created 'late' in the development process as it's the first to try out a new workflow to help receiving comments on major features or changes.

### Tasks

* [x] be cargo-release compatible when tagging versions
*  experiment with ways of handling threaded and non-threaded high-level APIs
   * [x] convenient sharing of mutable state in threaded and unthreaded contexts
   * [x] a way to detach extra state to create transfer objects to pass thread boundaries, and to reattach it
   * [x] See if non-panicking could be an option for this API. Soon both make no difference in terms of backtraces
   * [x] Write down and possibly practice how caches are supposed to be handled. What happens with new packs being created by a gc, or with changing config files?
* [x] streamline ergonomics of finding and peeling refs
* [x] make `cargo smart-release` use the `git_easy` API exclusively
* [x] proof of concept that GATs are working - YES, see 7f56dbd82db2abc18b8e6d228c8a5f54b3dbf32a
* [x] make clear what can and can't be stored in the cache - it looks like `packed::Buffer` doesn't really belong there as it _may_ contain a system resource in the form of a memory map. This means it should rather be shared. Finish the packed-refs needs update story, essentially.
    * To assure consistency, it's now behind an `Arc<RwLock>` even within the `State` structure that aims to be more like thread-local.
* ~~finally pull in git-config to see where it should live.~~
* [x] light documentation to get the main ideas across

Note that it should be reasonable stable, so probably dogfood it for a while, see #164 .

### Out of Scope - separate PR

* finally find a good solution for immutable vs mutable naming (which doesn't always work nicely. `Ref` suffix is a not 
      candidate for those with borrows inside. Candidates for massive changes are `git-object` and `git-ref` and `git-actor`. `borrow()` would then rather be `to_ref()`
* rename `oid` to `id` like in git2
* rename `find_existing()` to `find()` and provide the optional version under `try_find()`, both for objects and refs in plumbing crates


Byron:
@kim Here is [a high-level overview](https://github.com/Byron/gitoxide/blob/983d11a1f46c1ad21dbf2d57b63ecf979fab48b9/git-repository/src/lib.rs#L21-L34) of what the `Easy` mode entails especially in the trade-off department. However, I believe it's very powerful despite it needing awareness when implementing functionality or when working with `ObjectRef`s.

To my mind, the biggest shortcoming of the current implementation is that it accepts the potential for runtime panics due to incorrect RefCell access when using `ObjectRef` instances in particular. This is to leverage reusing object memory, so there can be only one at a time, over owning the memory by default. As long as the code runs at least once, these panics will always be visible so I think they wouldn't evade testing in the common case. It's indisputable though that now there is the potential for panics without these being obvious (e.g. because there is `unwrap()` or `expect(…)`) which is definitely surprising.

Maybe the type name could signal this more strongly than `ObjectRef` or the `try_borrow*()` API could be used to avoid panics and transform these into errors instead. The latter is so appealing that I will look into it.

git-repository/src/object/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+    mod error {
+        use quick_error::quick_error;
+
+        use crate::{object, object::find};
+
+        quick_error! {
+            #[derive(Debug)]
+            pub enum Error {
+                FindExisting(err: find::existing::Error) {
+                    display("A non existing object was encountered during object peeling")
```
kim, line 145: Tangent, because I came across `find_existing` before: this error means that the repository is probably corrupted, right? Like, we found a reference to the oid, but the actual object doesn't exist. Should this then scream very loudly?

kim:
> incorrect RefCell access when using ObjectRef

Can you help me understand how? If I try to use it as the wrong kind? Sorry, maybe too early over here... :innocent:

git-repository/src/object/mod.rs
Byron: I think with sparse checkouts everything is possible, and blobs can usually be missing. Would anybody point a tag to a blob? Probably not. Could they? Probably.
I think libraries shouldn't scream in panics, if this means loudly :), and so far in plumbing crates I found myself at least degenerate the error in the case that objects that should be there aren't to keep error handling easier.

With `git_repository::Easy` panicking on incorrect usage, it's definitely worth thinking about what that means for servers for example and generally, who may panic and when.

Byron:
> Can you help me understand how? If I try to use it as the wrong kind? Sorry, maybe too early over here... 😇

Let me try, it's already afternoon over here 😅. The `Cache` uses `RefCells` to allow `ObjectRef` instances to not only use their parent repository's facilities, but also avoid allocations and reuse pack caches.

https://github.com/Byron/gitoxide/blob/0f13104375216ccf099ebc2fcf0d180ed0de5237/git-repository/src/easy.rs#L6-L16

Even though it's quite straightforward to program with this limitation in mind and possible to protect from some surprises by consuming the instance…

https://github.com/Byron/gitoxide/blob/fc703937a078937840ea1c254f11e64aaf31de90/git-repository/src/object/tree.rs#L13

… nothing prevents the user from trying to obtain multiple `ObjectRef` instances and hold onto them which will panic.

It's quite likely this will happen the first time you work with it, judging from my own experience, and certainly a reduction in usability compared to `git2`. The latter as we know suffers from other scalability problems in return which won't be an issue for `gitoxide`.

Right now it looks like you can bake a cake and eat it, but the cherry on top is strictly hands off ;).

Byron:
PS: By following the issue link one can get back to the overarching goal (akin a tracking issue). All this should beef up transparency of what's going on and greatly facilitate influencing what's done to the better. This should lead to more stability, too, as one will get it _more right_ at first release.

git-repository/src/object/mod.rs
kim: Ah right, haven't thought about sparse checkouts. I was just thinking about the message text, not panicking. I might suggest one once this actually happens to me :)

Byron: Yes, please do!
Sparse-Checkouts are a killer feature that I got to have, or else `gitoxide` won't ever be the fastest thing around ever growing mono-repos. Recently git learned to checkout repos with multiple threads, a planned USP now relegated to maintaining parity.

kim:
Interesting, that's a little non-obvious indeed. Could it not be made safe by requiring the `Cache` to be mutably borrowed? I wouldn't even expect that there are huge gains by sharing it across threads, but I suppose you have better data on that.

Either way, I would certainly take the inconvenience of `Result` returns over accidentally not noticing a panic in a concurrent setup.

Byron:
> Interesting, that's a little non-obvious indeed. Could it not be made safe by requiring the Cache to be mutably borrowed? I wouldn't even expect that there are huge gains by sharing it across threads, but I suppose you have better data on that.

The cache is never meant to be (and can't be, thanks to RefCell) shared across threads. Instead, `EasyArc` is there for each thread to have their own cache. All limitations apply to a single thread, and nothing prevents the user from cloning `Easy` or `EasyArc` to support more concurrent `ObjectRef` instances if that's needed for their algorithm, or to call `to_owned()` or `into_owned()` in order to get rid of the contained `Ref`.

> Either way, I would certainly take the inconvenience of Result returns over accidentally not noticing a panic in a concurrent setup.

And that's where I am a little unsure, as running into such a panic is considered (by me :D) a logic bug, and bugs never… right, it's all about assuring these are visible. What about `panic = "abort"` when running dev instances in setups that swallow panics entirely? Does that even exist in a server setup?

An alternative certainly would be to provide panicking and non-panicking versions of some methods, and maybe that's the way to go and the most expected when looking at `std`.

kim:
>    The cache is never meant to be (and can't be, thanks to RefCell) shared
>    across threads. Instead, EasyArc is there for each thread to have their
>    own cache.

Right. Although it is still not so clear to me why the interior mutability
cannot be, you know, interior. Having to borrow an Easy mutably in order to get
at an ObjectRef seems perfectly fine to me.


>    And that's where I am a little unsure, as running into such a panic is
>    considered (by me :D) a logic bug

I like it when the compiler shows me a proof that my logic is flawed :grimacing:

`panic` is a loophole in the type system, so if my coverage _accidentally_ falls
below 100% I won't even get a warning. So I see deliberate use of `panic` the
same as `unsafe` -- if you really think you need to avoid that extra allocation,
please go ahead, but bring your own pen and paper for the proof. Offering both
panicking and safe functions when it's not obvious how to make invalid states
unrepresentable is the next best thing, I think.

Byron:
> Right. Although it is still not so clear to me why the interior mutability
cannot be, you know, interior. Having to borrow an Easy mutably in order to get
at an ObjectRef seems perfectly fine to me.

I tried that first :/. The problem is that these `Easy` infused objects need mutable access to its `State` (formerly `Cache`) in order to do most useful things. And with actual compiler-provable mutability the compiler will absolutely prevent you from passing on such mutable reference to another object to avoid aliasing. I think this is most akin to having a linked tree or list that these objects form with mutable access to something shared. RefCell's are the only way to get such a nice API while sharing state mutably. However, I am happy to try again with some sort of consuming API that prevents mutable refs to exist in more than one object at a time, but… I really think it won't work or be too restrictive it partially does.

One can rightfully argue that saving this one allocation, say for the buffer backing for an object, isn't worth the trouble. It's only part of the issue though as we need a pack cache as well, a packed refs buffer, and in future probably some config file caches as well.

> `panic` is a loophole in the type system, so if my coverage _accidentally_ falls
below 100% I won't even get a warning. So I see deliberate use of `panic` the
same as `unsafe` -- if you really think you need to avoid that extra allocation,
please go ahead, but bring your own pen and paper for the proof. Offering both
panicking and safe functions when it's not obvious how to make invalid states
unrepresentable is the next best thing, I think.

Point taken, I also can't live with writing library code that casually panics and thinks it's OK - I probably felt lazy when even arguing for it, let's just say I played the devils advocate pretty well there.

In the most recent commits, all RefCell access that could previously panic will now produce an Error. Most methods already returned a `Result` so it's really not a big change for the API.

Byron:
Besides the possibility for servers to create `Repository` update schemes to allow them to on-demand create a new repository which will be handed out to subsequent connections, such a scheme comes with the disadvantage that for a short amount of time about twice the  amount of system resources will be needed as two `Repository` instances are present - the old and the new.

Could there be a way to allow the existing `Repository` to be updated? For example once a pack was received, it should update itself to learn about the new pack only to make these objects available to subsequent connections.

Right now it seems the answer is _yes_, there is a generic way to do this even though for now it only works for `EasyArcExclusive` which uses an `Arc<RwLock<Repository>>`. In theory `EasyExclusive` is possible via `Rc<RefCell>` but would require GAT's as a `RefMut<'_, Repository>` has a lifetime. The big question will be if GATs will be compatible to things that don't have a lifetime like `ArcRwLockWriteGuard<parking_lot::RawRwLock, Repository>`, but maybe that's just `'static` then.

I will re-read the [RFC](https://github.com/rust-lang/rfcs/blob/master/text/1598-generic_associated_types.md) and see if that could actually work.

If it does, this might be it here as a proof of concept.

kim:
I can see now where this is going ... and yeah, I guess I'd need the mutexed variant in most cases.

Thanks for making it easier to follow your progress!


Nit: I think to should be `too_easy` instead of `to_easy`. No, don't listen to me.

Byron:
> I can see now where this is going ... and yeah, I guess I'd need the mutexed variant in most cases.

Great! And I think there is something there for everyone, each of the use case only paying what they actually need. And here is the proof that GATs will indeed work - on nightly they already do. Please don't look at the details of what follows, it's very hacky but shows what it it supposed to.

https://github.com/Byron/gitoxide/blob/7f56dbd82db2abc18b8e6d228c8a5f54b3dbf32a/git-repository/src/easy/impls.rs#L143-L175

> Thanks for making it easier to follow your progress!

And thanks for motivating this change, it's super valuable to have another set of eyes (at least) on the changes and it's the way to go from now on, for me.

> Nit: I think to should be too_easy instead of to_easy. No, don't listen to me.

I thought the same, each time that function is called people will subconsciously think it's indeed `too_easy()` to use `gitoxide`, that's the bar for sure :).

The plan is to write some docs, then work on the outscoped items which are all some breaking refactorings in plumbing crates you don't yet use and that I think will make everything more consistent. This should pave the way for writing down a guide for stability so I know the rules when creating that PR that brings `gitoxide` to the tests of the radicle-link flavored git-protocol.

Byron:
> I can see now where this is going ... and yeah, I guess I'd need the mutexed variant in most cases.

I thought up a limitation of `EasyArcExclusive` that probably is particularly relevant in servers:

https://github.com/Byron/gitoxide/blob/7578f1e2e578010eee087a9176d53a5862ec8862/git-repository/src/lib.rs#L181-L195

Of course there are workarounds, and I feel that avoiding to keep the `Repository` around too long is the way to go. Even though it would also be possible to re-acquire the repository read lock each time one needs to access an object (when creating the pack) that will certainly be costly quickly.

Lastly, a server is expected to have many repositories and keeping them in memory also holds on to their system resources, so having a way to release these (soon) after use is probably preferable. Using `EasyArcExclusive` for sharing such resources is certainly valuable for busy repos with a lot of read-accesses, but writers would then get into trouble. Should they wait for the write lock on the shared repo to eventually make the update? If so the result is consistent but pushes take longer than they have to. Or should they offload the repository data structure update into a concurrent task but open the system to consistency  issues - the write could clone what they just pushed and potentially get a pre-clone copy (assuming refs are updated after the objects are sound and safe).

All this makes me believe that a custom mechanism is needed to access repositories on servers, maybe along the lines of the following:

> Writers will always use their own repository (and object database with it) and can skip updating the in-memory data structure entirely. Readers can share a repository _as an optimization_, but there must be a way to indicate it as 'changed' so that new connections will get a freshly instantiated version of the repository, while existing connections may keep using the old one until they are done.

In the world described above servers would use `EasyShared`, `Easy` or `EasyArc`. But the world above would scale less well with many writers as these never share resources. An extension could be the following:

> Writers share repositories but only lock the repository on a per-object basis. That should be fine as they are expected to that only when resolving thin packs. This leaves enough time to acquire write-locks swiftly for (now) and to keep their data structures up to date with what's on disk. (For the most part, this would be changes to the number of packs).

The writers above would use `EasyArcExclusive`. Even though getting a read-lock per object is relatively costly, the typical push doesn't require too many lookups to resolve the received pack with a few thousand objects at most.

Now we share loaded repositories among readers and writers, and thus consume more system resources with an increasing amount of hosted repositories. Now, if this hypothetical `RespositoryStore` would have some sort of LRU mechanism so that when an IO error happens due to lack of resources, it can unload some in-memory repositories to free resources. That could be aided by the `Repository` being able to estimate resource usage even so IO errors due to resource constraints can be entirely avoided.

Such a system should scale nicely, and I imagine a beefy server machine that can host repositories for an entire region, with tasks like `GC` also being schedule by the server process itself to not only do it smartly (i.e. --geometric) but to do it when there is free resources, like at night, based on which repositories have changed recently.

It's quite amazing to think that this work here eventually leads to this 🤤.

Byron:
@edward-shen Now with `git-repository` starting to take shape, what's really missing for what seems like completeness is a `git-config` instance. I have seen work on a combined `git-config` has already started, and I'd love to put it in there when it's possible at all.

The reason I am pinging you is the hope that knowing where it's used helps to provide an API that works with that. My thought is that it would need the path to a git repository config file and can work out the rest from there.

https://github.com/Byron/gitoxide/blob/d287a4aec70e5dd33976a25d9a849c44d62d77c9/git-repository/src/lib.rs#L127-L137

edward-shen:
Thanks for the heads up. There is a  blocker to multi-file git config due to platform differences (_especially Windows_) that I think I should open up a discussion on for feedback.

As for the use case for this, it looks like our intended APIs are aligned. In the process of making a more ergonomic interface for multi-file `git-configs` I've been writing a wrapper (`ResolvedGitConfig`) that properly converts the config into a HashMap-like structure for a conceptually easier-to-use API, trading off a higher initialization cost and the ability to have non-destructive writes (e.g. comments and whitespace is preserved). For what it's worth, it should also have faster lookup performance as we actually parse the values during initialization, opposed to a lazily parsed value during lookup.

As an API user, would you prefer a multi-config struct to use `ResolvedGitConfig`, `GitConfig`, or the ability to have both? It looks like the goals of `Easy` mode seem to align with a multi-config struct to use `ResolvedGitConfig`, but my understanding may be incorrect.

(Also names are subject to change, I'm not entirely happy with them :laughing:)

Byron:
> As an API user, would you prefer a multi-config struct to use ResolvedGitConfig, GitConfig […]?

Independently of the initialization costs I see `Repository` as a general purpose utility. As such, it should be able to support writing changed `git-config` back to disk for those commands and tools that need it. From what I read, `GitConfig` is the type to support it as it keeps all the necessary information around, like comments and whitespace.

Personally I like less initialization time to let users pay when actually using the feature, which coincidentally is also what `GitConfig` allows.

Lastly, if there is a use-case where a lot of lookups have to be made, `ResolvedGitConfig` could certainly be helpful, but if it's not necessary for the upcoming API, I probably wouldn't even provide it until there is a use-case or user for it. Or said differently, it would be great to have `GitConfig` first :).

>  It looks like the goals of Easy mode seem to align with a multi-config struct to use ResolvedGitConfig, but my understanding may be incorrect.

`Easy` is mostly used for state that has to change, and most of the time that's probably not going to happen. Each thread has their own `Easy` state as well, which is nothing most application would want for `git-config` I presume.

> […], or the ability to have both?

Supporting a choice upon initialization seems interesting as `ResolveGitConfig` would be guaranteed to not have errors when accessing the values. This is probably going to simplify some applications as they won't have to deal with this cause of errors. On the other hand, most might use `anyhow` and the likes so it ultimately might not matter.

Having said that, right now I firmly believe the `GitConfig` would be the right choice for the `Repository` type.

> (Also names are subject to change, I'm not entirely happy with them 😆)

If `file::GitConfig` corresponds to a single file, I'd like to think of `fs::Config` as `git_config::CascadedConfig` or `git_config::Cascade` or `git_config::Stack`. It's a bit like 'cascading style sheets', so using there word `cascade` somewhere in there seems like a fitting choice.

kim:
>> I thought up a limitation of EasyArcExclusive that probably is particularly relevant in servers:

Those are good points. I _think_ what libgit2 is doing amounts to an LRU cache, which gets invalidated when there is a miss: ie. if an object is not found, only then it is determined if something changed on disk. That seems to be a bit prone to fd leaks IME because a. the setting for the upper bound is well-hidden, and b. pack bundles may get unlinked during GC while still being open (no idea how that can possibly work on windows).

Nevertheless, I think that’s a good approximation for the general case, as packfiles are immutable.

Invalidating a refdb cache is much less obvious, but it would be very desirable to avoid IO for reads. In our case, we could probably live with eventually consistent views in some cases, for example by only considering the packed-refs. Filesystem events may also be an option.

Lastly, I think configs could almost always be considered immutable (or if not, reloading can be application-controlled), and thus be shared globally. Important would be to be able to control if and when the different file locations are searched — in our case, we really only want the repo-local config, and never global nor system.

Byron:
> Those are good points. I _think_ what libgit2 is doing amounts to an LRU cache, which gets invalidated when there is a miss: ie. if an object is not found, only then it is determined if something changed on disk.

I wonder how they deal with misses due to sparse checkouts. As far as I know blobs could be entirely missing for parts of the history. Maybe they just avoid trying to access these in the first place, but that might be involved. Otherwise I would expect quite some cache trashing to be happening or at least some IO just to see that nothing changed.

> Nevertheless, I think that’s a good approximation for the general case, as packfiles are immutable.

Do you mean what `libgit2` is doing, or what `gitoxide` thinks it will do?

> Invalidating a refdb cache is much less obvious, but it would be very desirable to avoid IO for reads. In our case, we could probably live with eventually consistent views in some cases, for example by only considering the packed-refs. Filesystem events may also be an option.

I think there is great power in knowing or assuring that only the server process has access to a certain repository. That way consistency can be handled internally and relatively cheaply, for the typical fetch and push cases. Of course there may be other types of data that are expected to change or should be refreshable by using some external trigger like signals or fs-events, and with an optionally mutable borrow to the `Repository` all that should be possible too.

> Lastly, I think configs could almost always be considered immutable (or if not, reloading can be application-controlled), and thus be shared globally.

That's fair, even though for `git-repository` globally can only ever mean to be a field on `Repository`.

>  Important would be to be able to control if and when the different file locations are searched — in our case, we really only want the repo-local config, and never global nor system.

I think that will be possible, from what I have seen thus far. Even though right now there aren't many options regarding Repository discovery and opening, there will be added once there is a little more to configure.

----

Something that I would like to add to the `RepositoryStore` idea is the possibility to close write repositories after some time X and read repositories after some time Y. This should be tuneable so that the likelihood of getting an already opened repository is high. That way one may even keep the amount of time where a single on-disk repository has both a read and write instances loaded small to avoid such double-allocation of system resources.

All in all I am little bit too much in love with this idea.

kim:
>    I wonder how they deal with misses due to sparse checkouts.

That's what .promisor files are for, no? Which libgit2 doesn't support yet:
https://github.com/libgit2/libgit2/pull/5996

It also seems like multi-pack indices are quite crucial for preventing IO hell
on servers: https://github.com/libgit2/libgit2/blob/main/src/odb_pack.c#L41-L158

>      Nevertheless, I think thatâs a good approximation for the general
>      case, as packfiles are immutable.
>
>    Do you mean what libgit2 is doing, or what gitoxide thinks it will do?

_Some_ LRU strategy.

>    I think there is great power in knowing or assuring that only the
>    server process has access to a certain repository.

I would assume that it is fairly common that repository access is _not_ gatekept
through a single process: e.g. SSH access is most likely not handled by the same
server process which handles HTTP(s). Also load-balanced deployments will want
to signal cache invalidation by some means.

So I think it is important that in-memory state for the refdb can be invalidated
programmatically. It is probably less important to do this magically.

Byron:
> That's what .promisor files are for, no? Which libgit2 doesn't support yet: libgit2/libgit2#5996

A fun read, thanks for the pointer. I found interesting that the pack files are loaded lazily, `gitoxide` is quite generally eager because of mutation. This certainly makes `git` a little more efficient when starting up.

> So I think it is important that in-memory state for the refdb can be invalidated programmatically. It is probably less important to do this magically.

Maybe this can be rephrased from having one blessed gatekeeper process to various ones who collaborate. For instance, programs starting up from SSH could connect to the server process and make use of it this way, instead of doing the work themselves, leveraging its caches. And even if they have to handle the repository themselves, they could probably manage to send a message to the server to inform about invalidation.

Maybe `gitoxide` can also find a way to optionally use fs-events to auto-refresh itself. However, I think it's important to have some guarantees about ordering of events. For example, what if a write via SSH claims to be successful and a fetch comes in via HTTP right after with the expectation to clone with the objects that were just written. The fs-event wasn't yet fired or processed in the server handling the HTTP event and thus it lacks behind, presenting an outdated view to the client.

When processes are forced to cooperate, ordering can be enforced.

Since everything discussed here can be implemented with what's there and some servers might be OK with eventual consistency, moving forward should be alright.

> So I think it is important that in-memory state for the **refdb** can be invalidated… [emphasis mine]

Right now, the loose file reference database does have a cache, but it's forcefully updated each time one accesses it. I think git does it similarly. With `reftables` the world might look different, and I think the way this works and code is organized will change a little to abstract over this as well. Git does it that way, and when `reftable` becomes a topic for `gitoxide` there will certainly be changes because of that.

Byron:
Regarding servers and how they gain access to a repository via `Easy*`: A positive side-effect of the access pattern described [above](https://github.com/Byron/gitoxide/pull/165#issuecomment-903087949) certainly is that caches are strictly per request, and are released right after, while being configurable depending on what's expected to maximize performance or minimize memory usage, or somewhere in between.

PR #171, opened 2021-08-24
A draft for a stability guide

Byron:
It's goal is to explain what crate and application users can expect in terms of stability.

Having it should help building trust and encourage more people to rely on the crates provided
here.

* ~~pin all intra-workspace dependencies to their specific version while in pre-release mode~~
     - This allows us to have breaking changes in crate A with a minor version change while only bumping the patch level of B which uses A, without breaking C which uses B and A as long as C also pins A and B.
     - But it's probably not worth protecting against that outside of ALPHA releases as this also means upgrades for patch levels don't happen automatically.
* ~~make `cargo smart-release` handle changing pinned versions when bumping versions~~
* [x] ~~introduce `git-ref-base` as foundation crate~~ make `git-ref` available.
* [x] `cargo smart-release` shouldn't have to use `unstable` git-repository
* [x] which crates would be eligble to a first release? I am looking at `git-lock` and `git-tempflie` - to practice ST2 at least.
* [x] Sort out how to handle omnipresent primitive types, like `git-hash::ObjectId`
    - it's ST1 now
* [x] What about the `bstr` crate? Do we ultimately have to remove it from the API?
   - A 1.0 release [is planned](https://github.com/BurntSushi/bstr/issues/40#issuecomment-905143364)

[Rendered](https://github.com/Byron/gitoxide/blob/stability/STABILITY.md)


kim:
Do you also have a MSRV policy?

Byron:
Great catch! My intuition is to require the latest stable release by default until ST1 is reached by at least one application crate. From there the MSRV can be updated to whatever is current during the next release IF there are any benefits to do so.

What do you think about that? I think it should also be mentioned that the stability guide isn't stable in itself and it can be adjusted if better ways are discovered.

kim:
SGTM. I think it's more of a "good to know" at this stage :)

Byron:
Great, while writing it down I changed the MSRV back to: Let's use the latest stable to keep it simple, but be open to requirements to keep the Rustc version stable for a while or reduce it to a given version.

What I wrote down makes sense to me and seems good enough for a first iteration. It's definitely good to have an outlook and a vision. One thing is for sure: for ST2 and ST1 having some additional tooling support would be great to be sure embargos are considered during these mostly automatic releases 😅.

Byron:
@avoidscorn @edward-shen @joshtriplett It would be great if you could take a look and see if something important was missed in this first draft of the stability guide.  Thanks a lot.

STABILITY.md
```diff
@@ -0,0 +1,123 @@
...
+    └─────────────────────────────────────────────┘
+ ```
+
+### Tier 3: pre-release crates
+
+Pre-release crates is marked with major version number zero, for example `0.1.4` and lives in stability tier 3 _(->ST3)_.
+
+It's acceptable to let each breaking change be followed by a minor version release.
+Dependent crates are advised to use `=<version>` version requirements to avoid automatic updates of patch levels which
+may be breaking if they also depend on other `plumbing crates`.
```
edward-shen, line 71: Is this necessary? This is implied by semver, and fwiw is the convention in Rust's ecosystem for pre 1.0 crates anyways.

STABILITY.md
```diff
@@ -0,0 +1,123 @@
...
+    ╚═════════════════════════════════════════════╝
+                           │                       
+    Stability Tier 2 ─────────────────────────────┐
+    │                      ▼                      │
+    │    plumbing crates─────────────────────┐    │
+    │    │ ┌───────────┐       ┌───────────┐ │    │
+    │    │ │  git-odb  │       │  git-ref  │ │    │
+    │    │ └───────────┘       └───────────┘ │    │
+    │    │ ┌───────────┐       ┌───────────┐ │    │
+    │    │ │git-config │       │ git-pack  │ │    │
```
edward-shen, line 56: `git-config` is still pre-1.0, which would be a tier 3 stability.

STABILITY.md
```diff
@@ -0,0 +1,123 @@
...
+breaking changes or new features without forcing the use of `git` sources for dependencies in cargo manifests. Pre-release releases must pin all the pre-release
+crates they depend on to prevent automatic upgrades.
+
+Minor version updates for new features can be released when needed assuming there are no other breaking changes, updating the build identifiers for year and month accordingly.
+
+## The _Minimal Stable Rust Version_ (->MSRV)
+
+It is as low as it can be for the crate in questions. While there is no non-pre-release of a tier 1 crate, the MSRV is automatically assumed to be the latest stable version.
+
+Increasing the MSRV is not considered a breaking change and doesn't warrant a major version bump itself.
```
edward-shen, line 103: Is this convention? My intuition of semver would be that bumping MSRV would be considered backwards incompatible, but I suppose this isn't part of the API, so semver doesn't strictly apply here.

Should we make it at least a minor version bump? And/or suggest that we will attempt (but not guarantee) to only change msvp at minor or major versions?

STABILITY.md
Byron: How can it be made clearer that this is an example for the structure of crates and their maximum stability tier? 

Looking at it again I do realize that `git-repository` plans to provide direct struct field access to something from `git-config`, so that would make it ST1.

It's probably better to hide these types or to commit to more stability once it goes 1.0.

STABILITY.md
Byron: I think it's beneficial to spell out how this is handled even if it represents the status quo.

Also I find helpful to explicitly state that dependents should pin their dependencies for stability. For the latter to truly work `gitoxide` crates should also pin their dependencies, so that A-0.4 -> B 0.3 won't have a chance to break downstream if A-0.4.1 -> B -> 0.4 is released with dependent D using A (which auto-updates) and B 0.3, which won't auto-update.

STABILITY.md
Byron: > Is this convention? My intuition of semver would be that bumping MSRV would be considered backwards incompatible, but I suppose this isn't part of the API, so semver doesn't strictly apply here.

That was also my first intuition, strictly speaking the MSRV (as well as the edition) affect stability and can break existing users. However, from what I reckon in the MSRV implementation for cargo nightly, cargo will be able to select only crates that have a matching msrv. So it won't select a patch level that would potentially break you.

> Should we make it at least a minor version bump? 

Whether minor or patch, it's the same effect on downstream crates as without cargo support, they will automatically update.

> And/or suggest that we will attempt (but not guarantee) to only change msvp at minor or major versions?

Something like this I had written there before but removed it for the sake of simplicity and the preference for "let's wait for actual demand before complicating our life".

After re-reading this paragraph I realised it was a little 'split' with portions of the old text and the new one. It has now made clear that the MSRV is the latest stable unless there is demand for setting it. Hence this is likely to be revisited at some point.

Byron:
@edward-shen Something new just emerged: Foundation Crates. The motivation of this tiered system is to have the least amount of crates commit to a long time between potentially breaking major releases. `git-repository` acts as a boundary, providing useful and convenient functionality with high stability guarantees.

However it's unnecessarily hard to seal off all the details of the underlying crates which are usually exposed with types. For example, to keep a list of object ids most efficiently, you want to use `git_hash::ObjectId`. There are some useful types defined in the `git_ref` crate as well which are worth using even (or particularly) in ST1. The same, I believe, will hold true for `git-config` usage, which probably can provide access to some `Value` type. This why I have tentatively added a new crate, `git-config-base` (name is… probably not the best 😅, it only contains types, so maybe `git-config-types` is more specific), which could contain such a `Value` type.

However, it might well be possible that this isn't ever required as the API would be more like `get_bool`, `get_int` and so worth, but I think something more generic with conversions akin to `get("foo.bar").to_int()?` is neater.

What do you think?

Byron:
@edward-shen I probably should hold back in writing comments for a few minutes so that I can avoid back-paddling like now. When trying rip out `git-ref-base` I noticed that there is also ref logs to consider. And there is probably more and I realized that a whole lot of work would be caused by trying to not commit to stability tier 1 for `git-ref` or `git-config`. And I wonder why, just because there are still in pre-release and it will take them some time to stabilize. When that is done, it seems fair and standard business to grant them some stability and maybe keep things simpler.

Of course, as more and more of the `git-repository`  API materializes I might still change my mind for `git-ref` and summon a `git-ref-base`, but for now there shouldn't be a strong need.

STABILITY.md
Byron: > Also I find helpful to explicitly state that dependents should pin their dependencies for stability. For the latter to truly work gitoxide crates should also pin their dependencies, so that A-0.4 -> B 0.3 won't have a chance to break downstream if A-0.4.1 -> B -> 0.4 is released with dependent D using A (which auto-updates) and B 0.3, which won't auto-update.

Please ignore the latter as I think it's not necessary to protect downstream from that possibility with pre-release crates. This should only be done with 'alpha' releases of released software to not create issues like the ones encountered recently with `clap`.

Byron:
Thanks for everyones involvement. I am merging this branch because I plan some more breaking changes in ST3 crates to scratch them off the list of things that always bothered me, but am afraid of merge conflicts.

I invite additional discussions and changes in a discussion or PR. Thanks for your understanding.

PR #172, opened 2021-08-24
Allow creation of empty indices

mellowagain:


Byron:
Could you elaborate on the problem that the PR is solving? I think I am particularly interested understanding how empty indices can be useful and why they shouldn't be an error.

Thanks for your help.

PS: The advisory failure will be fixed in `main`

PR #185, opened 2021-08-29
cleanup and init a bare repository

Byron:
This time we can still get around using a full-blown git-config implementation as well.

* [x] refactor and some tests
* [x] repository open()
* [x] use thiserror in favor of quickerror (thanks to transparent errors and it being used already in dependency tree)
* [x] init bare
* [x] get git-config parsing to work on windows

### DIFF
--- a/git-repository/src/repository.rs
+++ b/git-repository/src/repository.rs
@@ -55,7 +55,7 @@ pub mod init {
     quick_error! {
         #[derive(Debug)]
         pub enum Error {
-            Init(err: crate::init::Error) {
+            Init(err: crate::path::create::Error) {
                 display("Failed to initialize a new repository")
                 from()
                 source(err)
@@ -78,7 +78,7 @@ pub mod init {
         /// Fails without action if there is already a `.git` repository inside of `directory`, but
         /// won't mind if the `directory` otherwise is non-empty.
         pub fn init(directory: impl AsRef<Path>) -> Result<Self, Error> {
-            let path = crate::init::into(directory.as_ref())?;
+            let path = crate::path::create::into(directory.as_ref())?;
             Ok(path.try_into()?)
         }
     }
--- a/git-repository/src/repository.rs
+++ b/git-repository/src/repository.rs
@@ -136,8 +136,8 @@ pub mod init {
         ///
         /// Fails without action if there is already a `.git` repository inside of `directory`, but
         /// won't mind if the `directory` otherwise is non-empty.
-        pub fn init(directory: impl AsRef<Path>) -> Result<Self, Error> {
-            let path = crate::path::create::into(directory.as_ref())?;
+        pub fn init(directory: impl AsRef<Path>, kind: crate::Kind) -> Result<Self, Error> {
+            let path = crate::path::create::into(directory.as_ref(), kind)?;
             Ok(path.try_into()?)
         }
     }
--- a/git-repository/src/repository.rs
+++ b/git-repository/src/repository.rs
@@ -24,27 +24,12 @@ pub mod from_path {
 
     use crate::Path;
 
-    pub type Error = git_odb::linked::init::Error;
-
     impl TryFrom<crate::Path> for crate::Repository {
-        type Error = Error;
+        type Error = crate::open::Error;
 
-        // TODO: Don't use this for cases where input paths are given to save on IOPs, there may be
-        //       duplicate file checks here.
         fn try_from(value: Path) -> Result<Self, Self::Error> {
-            let (git_dir, working_tree) = value.into_repository_and_work_tree_directories();
-            Ok(crate::Repository {
-                odb: git_odb::linked::Store::at(git_dir.join("objects"))?,
-                refs: git_ref::file::Store::at(
-                    git_dir,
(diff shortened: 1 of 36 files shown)


git-repository/src/repository.rs
```diff
@@ -81,7 +72,36 @@ pub mod open {
...
+
+        pub(in crate::repository) fn open_from_paths(
+            git_dir: PathBuf,
+            mut worktree_dir: Option<PathBuf>,
+        ) -> Result<Self, Error> {
+            if worktree_dir.is_none() {
+                let config = git_config::file::GitConfig::open(git_dir.join("config"))?;
+                let is_bare = config
+                    .value::<Boolean<'_>>("core", None, "bare")
+                    .map_or(false, |b| matches!(b, Boolean::True(_)));
```
Byron, line 88: @edward-shen I hope you are doing well!

Now that I am using `git-config` for the first time, I could imagine accessing values more comfortably. Unless I am missing something, are there plans to make this more ergonomic?
I could probably add what would help me with more convenient access, but love to align with you first. Thank you

Byron:
@edward-shen It looks like on windows [git-config can't parse a file](https://github.com/Byron/gitoxide/pull/185/checks?check_run_id=3459160323#step:5:877) that does parse fine on non-windows. It's based on [this template](https://github.com/Byron/gitoxide/blob/9e8a39e3cbd620bd48f379743df0d5783c33a86f/git-repository/src/assets/baseline-init/config#L3-L2) and all I do is to [substitute a value](https://github.com/Byron/gitoxide/blob/9e8a39e3cbd620bd48f379743df0d5783c33a86f/git-repository/src/path/create.rs#L172-L179). I would appreciate any hints at what could cause this, and will look into it if that starts blocking me. Thanks a bunch.

git-repository/src/repository.rs
edward-shen: That's the intent of ResolvedGitConfig. The current GitConfig was just a lightweight wrapper around the parser. I think I haven't written any accessor method on it, but that should be an easy add.

As for when I can get work done, I'm currently in the process of a cross continent move (NY -> CA), so unfortunately I don't know when I can work on it.

edward-shen:
> @edward-shen It looks like on windows [git-config can't parse a file](https://github.com/Byron/gitoxide/pull/185/checks?check_run_id=3459160323#step:5:877) that does parse fine on non-windows. It's based on [this template](https://github.com/Byron/gitoxide/blob/9e8a39e3cbd620bd48f379743df0d5783c33a86f/git-repository/src/assets/baseline-init/config#L3-L2) and all I do is to [substitute a value](https://github.com/Byron/gitoxide/blob/9e8a39e3cbd620bd48f379743df0d5783c33a86f/git-repository/src/path/create.rs#L172-L179). I would appreciate any hints at what could cause this, and will look into it if that starts blocking me. Thanks a bunch.

Maybe it's a newline issue? It's possible that the parser is always expecting a newline at that point, so it might not successfully parse it.

git-repository/src/repository.rs
Byron: > That's the intent of ResolvedGitConfig. The current GitConfig was just a lightweight wrapper around the parser. I think I haven't written any accessor method on it, but that should be an easy add.

With this I am happy to wait, and if you have ideas on how this API should look like, I'd be happy to add these methods as I go.

edward-shen:
I think most of the conversion between the two is already done, so all you
have to do is to implement the methods. I think something similar is fine,
e.g. a `.value()` method thats generic over some bytes form (TryFrom<[u8]>
was what I had before, but having a Deserialize bound might be nice), as
well as helper methods such as .bool() and .str() that make it easier to
access.

I don't have a strong vision on what the API should be like, and since
gitoxide is the primary user I see it perfectly reasonable to have one that
is best ergonomic for you.

Byron:
> Maybe it's a newline issue? It's possible that the parser is always expecting a newline at that point, so it might not successfully parse it.

I took a look and the [first version of the fix](https://github.com/Byron/gitoxide/pull/185/commits/2a2a89f68cc45e27a1cf0d33fc644ebabc762302) was quickly supplanted by [the second version](https://github.com/Byron/gitoxide/pull/185/commits/57203ce5d5e3c481b69c3ca173e4b00f11aaf7d7) which also counts newlines correctly again (I think).

>  I think something similar is fine,
e.g. a `.value()` method thats generic over some bytes form (TryFrom<[u8]>
was what I had before, but having a Deserialize bound might be nice), as
well as helper methods such as .bool() and .str() that make it easier to
access.

I will see how these utility methods could look like, the first stab at it would certainly be guided by my recent use of reading a boolean.

> I don't have a strong vision on what the API should be like, and since
gitoxide is the primary user I see it perfectly reasonable to have one that
is best ergonomic for you.

Perfect, thank you! I will be pinging you solely for the purpose of keeping you in the loop and to encourage discussion, but know well enough that you are very busy. Lastly, please feel free to guide this communication to work for you, I wouldn't want that to be a burden.

Long story short, I will be very happy to get into `git-config` more, and hope you will have a good time in the new location.

Issue #200, opened 2021-09-17
Typing upload-pack ERR responses


kim:
I have a test case which tries to fetch a ref the remote end doesn't know about:

```
15:49:04.445735 pkt-line.c:80           packet:  upload-pack< command=fetch
15:49:04.445784 pkt-line.c:80           packet:  upload-pack< agent=git/oxide-0.10.4
15:49:04.445787 pkt-line.c:80           packet:  upload-pack< 0001
15:49:04.445796 pkt-line.c:80           packet:  upload-pack< thin-pack
15:49:04.445800 pkt-line.c:80           packet:  upload-pack< include-tag
15:49:04.445803 pkt-line.c:80           packet:  upload-pack< ofs-delta
15:49:04.446266 pkt-line.c:80           packet:  upload-pack< want-ref refs/namespaces/hnrkez5kbhsrh97mimd8xgxf7w4cimcnc9wdo/refs/remotes/hyn68wpms9eognsszkrsikgbwkr9trotu8etrknwdw178z18bbgj3g/rad/id
15:49:04.446297 pkt-line.c:80           packet:  upload-pack> ERR unknown ref refs/namespaces/hnrkez5kbhsrh97mimd8xgxf7w4cimcnc9wdo/refs/remotes/hyn68wpms9eognsszkrsikgbwkr9trotu8etrknwdw178z18bbgj3g/rad/id
```

The `Display`ed error message from this is "The server response could not be parsed". `Debug`ging the error shows that it is two nested `io::Error`s.

I couldn't quite figure out where to look, but it seems to me like the pktline reader should recognise this as a protocol message, and return a dedicated error variant for it.

"unknown ref" is of course still an arbitrary string, but it would still be helpful to be able to pattern match on those kinds of errors.

Byron:
Thanks for reporting, [rant] error handling around these pseudo-streaming-iterator line readers wrapped in impls of std::io::Read is definitely not nice and begging for some sort of rewrite once GATs are there or… a version that just copies the line around to make the borrow check issues go away right now. Maybe a case of premature optimization[/rant].

Looking into this more, even with GATs or without these borrow-check workarounds error handling might be difficult to do on the top-level there is this implementation of `Read/AsyncRead` which needs an `std::io::Error`.  For code which knows what to expect, would dynamically casting the contained error even be possible?

### Understanding what's happening

Following the chain of errors bubbling up, it all starts with reading the packet line, where the server error is treated as string:

[...]

PR #201, opened 2021-09-19
transport err typing

Byron:
Fix for #200 

* [x] structured error in git-packetline
* [x] a flag to use ref-in-want in order to reproduce the issue on a high level - fails to reproduce as the current test suite operates with servers who don't support this feature at all. Unless there is a way to want invalid SHA1s, it's hard to trigger an error.
* [x] cause a packet-line based error
* [ ] obtain underlying packetline errors from io errors

### DIFF
--- a/etc/check-package-size.sh
+++ b/etc/check-package-size.sh
@@ -36,4 +36,4 @@ echo "in root: gitoxide CLI"
 (enter git-packetline && indent cargo diet -n --package-size-limit 15KB)
 (enter git-repository && indent cargo diet -n --package-size-limit 50KB)
 (enter git-transport && indent cargo diet -n --package-size-limit 30KB)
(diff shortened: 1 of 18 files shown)


kim:
Thanks!


> the current test suite operates with servers who don't support this feature at all

Does this not work?:

    git -C uploadpack.allowrefinwant daemon …

Byron:
> Does this not work?:

Thanks so much, I didn't think of the fact the the these tests use the daemon which is in fact configurable 🤦‍♂️. This works perfectly, and now there is a test to verify reasonable errors chains in case upload-pack bails out:

```
Error: The server response could not be parsed

Caused by:
    0: Upload pack reported an error
    1: unknown ref refs/heads/does-not-exist
```

Issue #198, opened 2021-09-14
[cargo smart-release] Changelog generation


Byron:
Now that it's easier than ever to publish a big amount of crates and do so safely for downstream users, it's absolutely crucial to unify and automate changelog generation.

`git-cliff` is quite inspirational for showing it is possible, and `git-conventional` is quite easy to work with and common enough to adopt (with adjustments) to generate valuable change logs automatically.

This should be done in a minimally configurable way only, while making 'debugging' feasible to understand why a piece of information might be missing.

Lastly, it should be possible to hand-edit changelogs and keep them save from future adjustments.

PR #209, opened 2021-09-26
Changelog support for smart-release (continuation 2)

### LINKED ISSUE #204: cargo-smart-release fails to install
I tried to install `cargo-smart-release` on nightly on ubuntu (WSL2), and got multiple compile errors:

image

image

Byron:
Fix for #198.

- [x] fix history traversal/change computation
- [x] reorder parsed tags as well (see git-packetline); the first headline indent should rule all of them
- [x] don't re-add statistical items unless there is none present (to allow deletion of them)
- [x] add a flag to add no statistical headline
- [ ] headlines for git-conventional kinds along with descriptions which can be hidden by keeping only their html tag around
- [x] add flag for making no git-conventional based insertions (or extend previous one, probably better to identify these with names)
- [ ] deduce version bump by looking at the changelog
- [ ] find a solution for Reference links, currently they get lost entirely
- [ ] add GitHub links for release headlines
- [ ] add headline for links to issues and PRs that were worked on
- [x] add change-log related flags for smart-release sub-command (allowing to handle changelogs as part of a release), rewriting 'Unreleased' headers into their respective tag name and date.
- [ ] create a tag object instead of tag reference if changelog is present, containing the changelog itself as text
- [ ] warn if the changelog needs updating, so we would have to compare the newly generated changelog with the one that's present 
- [ ] consider automating GitHub releases with `gh release create` - it supposedly pulls contents from tag objects, but it should be easy to pass them. [More information here](https://lib.rs/crates/parse-changelog).

After merge, release and fix #204, and setup install validation for smart-release.

----

* https://github.com/BurntSushi/bstr/issues/98 
* https://github.com/crate-ci/git-conventional/pull/16


Byron:
Rollup changes from a week to avoid long-running PRs.

PR #210, opened 2021-10-02
Fix immediate dereference warning

caspervonb:


Byron:
Thanks a lot for digging in and your first contribution!

_It's strange that clippy didn't pick this up before, maybe this was discovered by a recent nightly._

Issue #191, opened 2021-09-01
git-config towards 1.0


Byron:
A list of ideas for features that come up when using `git-config::file::GitConfig`

* `Boolean` can act as `bool` or be converted to one, maybe using deref
* `Integer` derefs to integer
* Something like `try_value() -> Result<Option<T>, E>` as currently I am ignoring all errors as it's too cumbersome to check if the conversion failed which probably should be an actual 
* could there be a more convenient API than [this](https://github.com/Byron/gitoxide/blob/d0a956fdb5a822dbd116792bfbe70d1532a95ec9/git-repository/src/repository.rs#L74-L76).

### Missing features

* [ ] cascading config files
* [ ] know which config files to cascade
* [ ] actual `PathBuf` instead of just a string
* [ ] comfortable API for accessing values
* [ ] include directive

### Related Discussions

- https://github.com/Byron/gitoxide/discussions/166
- https://github.com/Byron/gitoxide/discussions/147
- https://github.com/Byron/gitoxide/discussions/54

edward-shen:
What would you suggest for a concrete implementation then? A separate type `ParserValue` and a serde friendly enum? Or just enhancements to the existing `Value` (e.g. `as_*`) and `to_owed`?

Ironically enough, serde compatibility was always a goal from the beginning but never achieved (yet)

PR #225, opened 2021-10-19
feat(git-config): `add GitConfig::from_paths`

caspervonb:
This adds `GitConfig::from_paths` which constructs a config from all the sections of the given files.
Sections are additive so that multi-value semantics are preserved.

Towards #191

### DIFF
--- a/git-config/src/file/git_config.rs
+++ b/git-config/src/file/git_config.rs
@@ -136,6 +136,7 @@ impl<'event> GitConfig<'event> {
     }
 
     /// Constructs a `git-config` file from the provided paths in the order provided.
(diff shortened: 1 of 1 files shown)


edward-shen:
Thanks for the PR!

Re: multivar behavior: can you run/add a test that includes sections with the same key? I'd like to document that behavior in our tests.

Honestly that wording from the git-config doc is a little ambiguous. Can you also model a test after the behavior of `git config`?

caspervonb:
Digging into it little bit more, seems that git-config calls `git_config_from_file` repeatedly which just keeps pushing and parsing sections so this seems to be the way to go.

PR #229, opened 2021-10-20
ci: add workflow dispatch to stress workflow

caspervonb:
Adds a workflow_dispatch to the stress workflow for manual triggering


Byron:
That's great, I have never used that one. Let me try :)

PR #231, opened 2021-10-20
chore: rename workflows

caspervonb:
Nit and subjective, but feels cleaner.


Byron:
Nitty but not gritty 😅

PR #239, opened 2021-10-22
Include tests when running clippy

caspervonb:


Byron:
Good catch! Thank you.

PR #240, opened 2021-10-25
Add `GitConfig::from_env_paths` with git-like sequence resolution

caspervonb:
Towards #191


Byron:

I think it's OK to merge without tests given `git-config` is still unstable and that nobody is using this method yet.

PR #248, opened 2021-10-31
Simplification of `header_field_multi_line` and its invocation.

xmo-odoo:
Just a simple "cleanup" I saw while reading the code: 

* `EmptyValue` is already used by `header_field` and seems suitable for the same case here, the way it was previously called there was no way the value could be empty but since `header_field_multi_line` is already fallible it seems odd to make it panic.
* Swapped `lines` for `split_str`, avoids the terminal edge case which might be a tad slower (as it writes an empty string) but doesn't seem like it should be too critical.
* Somewhat similarly simplifies `write_to` by removing the condition entirely (should work because of the fix to (1), `header_field_multi_line` starts by writing a single header then writes the followup so it should work fine for both cases).

### DIFF
--- a/git-object/src/commit/write.rs
+++ b/git-object/src/commit/write.rs
@@ -1,7 +1,5 @@
 use std::io;
 
-use bstr::ByteSlice;
-
 use crate::{encode, encode::NL, Commit, CommitRef, Kind};
 
 impl crate::WriteTo for Commit {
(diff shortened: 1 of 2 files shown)


Byron:
Thanks a lot for this contribution, I love the simplification it brings and, even after running the stress test (which I believed was the cause for some of the complexity thanks to round-tripping) everything still works perfectly.

One would have hoped for a measurable speed boost, but that's not apparently the case. It's not truly surprising either as the cost of git-object interactions is dominated by accessing packs.

PR #259, opened 2021-11-19
Discovery: data consistency and resource usage

Byron:
Concurrent writes to packs observable by applications pose a challenge to the current implementation and we need to find ways around that.

#### Motivation

Use this document to shed light on the entire problem space surrounding data consistency and resource usage of packed objects to aid in finding solutions
that are best for various use cases without committing to high costs in one case or another.


Byron:
Even though it's early in the process of this discovery I'd appreciate any (and early) feedback. Thank you :)

CC @kim @joshtriplett

kim:
You still seem to be missing the most obvious optimisation: gitoxide currently
maps both the index and the data file eagerly. There is no good reason to do
this, except for very specialised use cases (eg. implementing `git repack`). I
think it is fair to say that the majority of applications will **not** know
upfront which objects (and therefore which packs) they will access. Those
applications will most likely benefit from mapping data files on-demand, even if
you don't provide a threadsafe implementation and don't account for concurrent
modifications of the odb.

Also your definition of scalability does not consider that resource accounting
is independent of the number of available cores, so I would find it appropriate
to not make bold claims to that end.

Byron:
Thanks for your help, it's appreciated :).

> You still seem to be missing the most obvious optimisation: gitoxide currently
maps both the index and the data file eagerly. There is no good reason to do
this, except for very specialised use cases. […]

To me, avoiding _unconditional_ interior mutability to support lazy loading of packs allows for thread-safe sharing of memory maps to scale object access with cores. Maybe this is a special case or an optimization many will skip. I highlight the word 'unconditional' because also I believe lazy loading should be supported, probably even as the default, but not be the only option. The use-cases I mention in the discovery will try to make a case for different options to handle this, and I hope to have a technical sketch for a solution that serves all of them soon.

> Also your definition of scalability does not consider that resource accounting
is independent of the number of available cores, so I would find it appropriate
to not make bold claims to that end.

I'd be glad if you could point me to this statement so it can be adjusted as I agree that it shouldn't even be implied that resource accounting is somehow dependent on the number of cores.

Issue #270, opened 2021-12-05
Commit ancestor traversal sorted by commit time


Byron:
The `Ancestors` iterator `git-traverse` should learn to dispatch Commits sorted by their date.

This capability is needed in many applications, which currently have to sort after storing all commits in a Vec. Ideally this can be done on the fly, to the extend possible.

Byron:
@oknozor Maybe this issue would be something for you. It's involving standard tests and Rust, as opposed to issue 269 which I am getting done myself why the memory is fresh.

It's mainly cleanup which certainly is a thankless first issue with lots of implicit complexity, most of which isn't mentioned in the issue itself.

oknozor:
Just to be sure I understand whant you want here, let's take a simple example  : 
```
# Commit in year 2000
git checkout -q -b main
GIT_COMMITTER_DATE="2000-01-02 00:00:00 +0000" git commit -q --allow-empty -m c1

# Commit in year 2001
git checkout -q -b branch1
GIT_COMMITTER_DATE="2001-01-02 00:00:00 +0000" git commit -q --allow-empty -m b1c1

# Commit in year 2000
git checkout -q main
GIT_COMMITTER_DATE="2000-01-02 00:00:00 +0000" git commit -q --allow-empty -m c2

# Commit from branch1 made in 2001 merged in 2002
GIT_COMMITTER_DATE="2002-01-02 00:00:00 +0000" git merge branch1 -m m1b1
```

This would produce the following git log : 
```
❯ git log --date=format:'%Y' --pretty=format:'%cd %H% -%d %s' --abbrev-commit --graph  
*   2002 288e509293165cb5630d08f4185bdf2445bf6170- (HEAD -> main) m1b1
|\  
| * 2001 bcb05040a6925f2ff5e10d3ae1f9264f2e8c43ac- (branch1) b1c1
* | 2000 9902e3c3e8f0c569b4ab295ddf473e6de763e1e7- c2
|/  
* 2000 134385f6d781b7e97062102c6a483440bfda2a03- c1
```

Currently they are not produced in the correct  order using the `Ancestors` iterator implementation  : 

```
# c2 and b1c1 are inverted
288e509293165cb5630d08f4185bdf2445bf6170 (m1b1)
9902e3c3e8f0c569b4ab295ddf473e6de763e1e7 (c2) 
bcb05040a6925f2ff5e10d3ae1f9264f2e8c43ac (b1c1) 
[...]

Issue #251, opened 2021-11-04
Reference Name Joining


FintanH:
Hey :wave: 

I've been perusing the `git-ref` family of names, i.e. `FullName`, `FullNameRef`, and `PartialNameRef`. I want to see if I'm using it right and also suggest a couple of additions that might be helpful for consumers of the API.

My use case is the following. I want to end up using `try_find_reference`, which takes one parameter that is `impl TryInto<PartialNameRef>`. I have a structural representation of the reference which looks like:
```rust
pub enum Remote {
    Default,
    Peer(PeerId),
}

pub struct Reference {
    pub remote: Remote,
    pub urn: Urn,
}
``` 

Now, I can't directly write a `TryFrom` for `PartialNameRef` as far as I can tell because of the associated lifetimes. But I can write a `TryFrom` for `FullName` and use `to_partial`. I'm wondering if that's the correct and intended way of using the API?

Regarding my suggestions, when I wrote a `TryFrom` for `Remote`, I wasn't able to reuse this when writing `TryFrom` for `Reference`, since there's no way to `join` two `FullName`s. I ended up having to use `format!` and use the `TryFrom` instance on `&String`. It would be useful to be able to join two names to get another valid name.

The other (possibly) useful suggestion would be to have a macro for writing static names, that are known to be valid at compile time. We used this technique in radicle-link[[0]] and it was extremely nice compare to using `try_from` everywhere :)

Let me know what you think, and perhaps I could help contribute with these additions :v: 

[0]: https://github.com/radicle-dev/radicle-link/blob/master/macros/src/lib.rs#L16-L46

Byron:
Thanks for posting, I remember that in order to make it convenient to use, I kept adding TryFrom implementations to cover the inputs encountered during testing. It's likely that many more are missing and I hope it's possible to add them when discovered.

> Now, I can't directly write a TryFrom for PartialNameRef as far as I can tell because of the associated lifetimes. But I can write a TryFrom for FullName and use to_partial. I'm wondering if that's the correct and intended way of using the API?
> Regarding my suggestions, when I wrote a TryFrom for Remote, I wasn't able to reuse this when writing TryFrom for Reference, since there's no way to join two FullNames. I ended up having to use format! and use the TryFrom instance on &String. It would be useful to be able to join two names to get another valid name.

I have added a test to validate currently possible inputs and to reproduce something akin to what you are describing or so I hope.

[...]

Issue #11, opened 2020-08-14
Benchmark alternate malloc implementations (jemalloc, scudo)


joshtriplett:
Given that gitoxide is allocating memory and working across many threads, it would be worthwhile to benchmark using alternate memory allocators.

It'd be *trivial* to test `jemalloc`: just use the [`jemallocator`](https://github.com/gnzlbg/jemallocator) crate from the top-level gitoxide bin crate, following the instructions in its README.

The `scudo` allocator has even more performance, as well as some resilience against some kinds of security-related bugs. See https://expertmiami.blogspot.com/2019/05/what-is-scudo-hardened-allocator_10.html for performance information. However, it isn't packaged for Rust yet, so it won't be quite as easy to try.

Byron:
Agreed, it's absolutely worth running some experiments.

My goal would be to offer feature toggles to make switching between implementations easier. By the looks of it, it doesn't take much indeed and will play nice with feature toggles as well.

joshtriplett:
I ran some experiments, and the scudo allocator provided much less performance than that blog post would suggest. jemalloc and mimalloc seem like the primary reasonable options. I'd suggest trying jemallocator and mimallocator, and benchmarking.

Byron:
Thanks for the update. Till now the only place where a lot of allocations can happen is when owned Objects are created - each field of the decoded object usually results in an allocation.

Whether or not an application runs into issues with that probably depends on the application making the gitoxide top level crate the one where the allocator could make a difference. The only command to date which can allocate a lot is `gixp pack-verify --re-encode …`, which seems to slow about 30% when using the default allocator.

rickprice:
Hi Sebastian, would you still like some help with this?
Do we have any tests currently for doing benchmarks?
My work (ActiveState) has time  (currently 2 days) every quarter where we can work on projects, and I would like to help move gitoxide forward.

Byron:
Hi @rickprice, I certainly would love some help and benchmarks have been in my mind for a while now. Thus far, there are _experiments_ which effectively serve as comparisons to `libgit2`, but there is no benchmark suite yet.

Benchmarks should be done per crate and should probably use `criterion` to drive them, even though `glassbench` might also be usable but doesn't yet support [throughput](https://github.com/Canop/glassbench/issues/1), a feature I really like.

[...]

PR #272, opened 2021-12-06
Feat/traversal sort by committer date

oknozor:
This is a work in progress for #270

### DIFF
--- a/git-traverse/src/commit.rs
+++ b/git-traverse/src/commit.rs
@@ -4,9 +4,11 @@ pub struct Ancestors<Find, Predicate, StateMut> {
     predicate: Predicate,
     state: StateMut,
     mode: Parents,
+    sorting: Sorting,
 }
 
 /// Specify how to handle commit parents during traversal.
+#[derive(Copy, Clone)]
 pub enum Parents {
     /// Traverse all parents, useful for traversing the entire ancestry.
     All,
@@ -17,6 +19,22 @@ pub enum Parents {
 impl Default for Parents {
     fn default() -> Self {
         Parents::All
+    }
+}
+
+/// Specify how to sort commits during traversal.
+#[derive(Copy, Clone)]
+pub enum Sorting {
+    /// Default order, sort commit looking up the first reachable parent
+    GraphOrder,
+    /// Order commit looking up the most recent parent, since only parents are looked up
+    /// this ordering is partial
+    ByCommitterDate,
+}
+
+impl Default for Sorting {
+    fn default() -> Self {
+        Sorting::GraphOrder
     }
 }
 
@@ -31,7 +49,7 @@ pub mod ancestors {
     use git_object::CommitRefIter;
     use quick_error::quick_error;
 
-    use crate::commit::{Ancestors, Parents};
+    use crate::commit::{Ancestors, Parents, Sorting};
 
     quick_error! {
         /// The error is part of the item returned by the [Ancestors] iterator.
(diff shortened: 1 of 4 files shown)


git-traverse/src/commit.rs
```diff
@@ -151,6 +176,80 @@ pub mod ancestors {
...
+                        }
+
+                        for token in commit_iter {
+                            match token {
+                                Ok(git_object::commit::ref_iter::Token::Parent { id }) => {
+                                    let mut vec = vec![];
+                                    let parent = (self.find)(id.as_ref(), &mut vec);
+
+                                    // Get the parent committer date
+                                    let parent_committer_date = parent
```
Byron, line 211: This could be simplified to:

```Rust
 let parent_committer_date = parent.and_then(|mut parent| parent.committer().map(|committer| committer.time));
```

git-traverse/src/commit.rs
```diff
@@ -151,6 +176,80 @@ pub mod ancestors {
...
+    impl<Find, Predicate, StateMut> Ancestors<Find, Predicate, StateMut>
+    where
+        Find: for<'a> FnMut(&oid, &'a mut Vec<u8>) -> Option<CommitRefIter<'a>>,
+        Predicate: FnMut(&oid) -> bool,
+        StateMut: BorrowMut<State>,
+    {
+        fn next_by_commit_date(&mut self) -> Option<Result<ObjectId, Error>> {
+            let state = self.state.borrow_mut();
+            let res = state.next.pop_front();
+            let mut parents_with_date = vec![];
```
Byron, line 195: This array should be part of the state to avoid re-allocation each time one calls `next()`.

With that in place, one would probably clear it here.

Byron (commented):
Despite it being a WIP, it looks pretty complete to me, great work!

oknozor:
I have updated according to your review, let me know if I still need to change anything.

git-traverse/src/commit.rs
```diff
@@ -151,6 +177,80 @@ pub mod ancestors {
...
+                match (self.find)(&oid, &mut state.buf) {
+                    Some(mut commit_iter) => {
+                        if let Some(Err(decode_tree_err)) = commit_iter.next() {
+                            return Some(Err(decode_tree_err.into()));
+                        }
+
+                        for token in commit_iter {
+                            match token {
+                                Ok(git_object::commit::ref_iter::Token::Parent { id }) => {
+                                    let mut vec = vec![];
```
Byron, line 208: Sorry I missed this the first time around. Since another buffer is needed, it should probably be part of the state as well.

git-traverse/src/commit.rs
```diff
@@ -151,6 +177,80 @@ pub mod ancestors {
...
+                            }
+                        }
+                    }
+                    None => return Some(Err(Error::NotFound { oid })),
+                }
+            }
+
+            state
+                .parents_with_date
+                .sort_by(|(_, time), (_, other_time)| other_time.cmp(&time));
```
Byron, line 233: Here I would find `time.cmp(&other_time).reverse()` clearer, the `reverse()` makes a bit more obvious that this is going to be descending, as opposed to comparing the arguments in another order.

Byron (changes requested):
This time around I took a closer look at the tests and thank you very much for cleaning them up before adding more.

oknozor:
Just updated according to the latest review

Byron:
Thanks a lot for contributing! I hope thus far the experience was enjoyable, and if so, I would be happy to help you find another task. _(And if not, I'd be eager to learn so the experience can be improved, of course)_.

oknozor:
Definitely an enjoyable experience :). I'd be eager to work on something else !

Byron:
Awesome, thank you!

Candidates for a similar amount of work, give or take, would be #251 or #11. If these aren't for you, you could give me a hint of what interests you, or peruse the [collaboration board](https://github.com/Byron/gitoxide/projects/1) yourself for clues.

PR #286, opened 2021-12-29
fix(core/organize): keep non "git" repository name extensions

cmoog:
When converting a remote URL to a destination directory name,
keep extension strings if they are not equal to "git". For example,
https://github.com/jonhoo/rust-for-rustaceans.com should _not_ have the
extension stripped from its directory name.

### DIFF
--- a/gitoxide-core/src/organize.rs
+++ b/gitoxide-core/src/organize.rs
@@ -177,7 +177,11 @@ fn handle(
             match kind {
                 RepoKind::Bare => path,
                 RepoKind::WorkingTree => {
-                    path.set_extension("");
+                    if let Some(ext) = path.extension() {
+                        if ext == "git" {
+                            path.set_extension("");
+                        }
+                    }
                     path
                 }
             }


gitoxide-core/src/organize.rs
```diff
@@ -177,7 +177,11 @@ fn handle(
             match kind {
                 RepoKind::Bare => path,
                 RepoKind::WorkingTree => {
-                    path.set_extension("");
+                    if let Some(ext) = path.extension() {
+                        if ext == "git" {
```
cmoog, line 181: This would also work, but I prefer the existing semantics.
```rust
if path.extension().map_or_else(|| false, |e| e == "git") {
    path.set_extension("");
}
```

Byron:

gitoxide-core/src/organize.rs
Byron: Sometimes plain and simple is the easiest to read, indeed.

Issue #301, opened 2022-01-20
git-worktree


Byron:
### Tasks for cloning

* [ ] checkout entire index OR checkout from tree (in case the index doesn't exist yet).
    - This would allow the index to be created while we are checking out, maybe there is a benefit.

Byron:
Moved to [this Feature Proposal Discussion](https://github.com/GitoxideLabs/gitoxide/discussions/2779). This issue is being closed while the proposal is reviewed; it will become an issue again once review is complete.

_AI disclosure: This comment was written and posted by OpenAI Codex on behalf of Byron._

PR #315, opened 2022-01-23
Checkout of index entries into a directory

AP2008:


Byron (commented):
It might be worth taking a look at the reference in #301 to see what it means for git to handle multi-threading correctly. For one, these probably require different code paths, furthermore special care has to be taken to handle file systems with case folding correctly.

Besides that, it's neat to see what rayon could do, even though ultimately any parallel code has to make due with `git_features::parallel` as these allow to turn parallelism off entirely for completely single-threaded builds.

Furthermore I would have expected object ids from entries to be used to load contents to store in destination files, along with a more test-driven approach. For instance, it's unclear how this code gets executed and validated.

AP2008:
What function do I use to extract the contents using the `ObjectId` ? I have started writing tests. Should I use the `.git` repos used for testing in `git-index` to test `git-worktree` ?

Byron:
As an addition, I think it would be best to reduce the abstraction level to the bare minimum without forcing to create a `Worktree` abstraction just yet. What's happening here can happen in a function, without the need for any intermediate state. 

Furthermore, I think it's beneficial to implement this without threads at first and strive for correctness. From there it should be possible to derive a multi-threaded checkout to reach parity with what `git` can already do.

For examples on how to avoid dependencies and provide facilities to obtain objects, the `git-traverse` and `git-diff` crates are probably helpful examples.

Thanks

AP2008:
Should I retrieve the contents of an `ObjectId` using `git_odb::Handle` and `git_odb::FindExt` where the `git_odb::Handle` will be passed by the caller ?
What should be the behaviour when the `mode` of an entry is a `DIR`, `SYMLINK` or `COMMIT` ?
The data retrieved using `git_odb::Handle` can be copied to the new files when the `mode` is a `FILE` or `FILE_EXECUTABLE`.

Byron:
> Should I retrieve the contents of an ObjectId using git_odb::Handle and git_odb::FindExt where the git_odb::Handle will be passed by the caller ?
As mentioned earlier, [`git-traverse` leads the way](https://github.com/Byron/gitoxide/blob/2f2d856efe733d3cf81110c0e0607d2e7c40d968/git-traverse/src/tree/breadthfirst.rs#L67). In tests, it's ok to pass a closure that uses an object database handle of sorts as you see fit.

> What should be the behaviour when the mode of an entry is a DIR or COMMIT ?

For `DIR` I think a `todo!()` is appropriate as it requires sparse index handling (_I think_), and for `COMMIT` it could be whatever `git` does. Does it create this as empty directory maybe?

> In the case of SYMLINK, I am creating the symlink using os-specific functions (#cfg[target_family = "unix" or "windows")

For handling symlinks and executables I think it's important provide options that mirror options of git itself, which allows to completely ignore the file mode for instance which is typically the case on windows. There is certainly special handling for symlinks on windows as well, along with options that could be passed by the caller. Of course, ultimately there is no way around gating certain functionality depending on the platform or OS, but on top of that git provides options.

AP2008:
> As mentioned earlier, [`git-traverse` leads the way](https://github.com/Byron/gitoxide/blob/2f2d856efe733d3cf81110c0e0607d2e7c40d968/git-traverse/src/tree/breadthfirst.rs#L67). In tests, it's ok to pass a closure that uses an object database handle of sorts as you see fit.

The [tests](https://github.com/Byron/gitoxide/blob/498072ea42dca7b9d00bedba42829bdac92195b9/git-traverse/tests/tree/mod.rs#L6) in `git-traverse` use `git-odb` as well. I was planning to do something like this:
```rust
use git_index::{entry::Mode, State};
use git_odb::{FindExt, Handle};

pub fn copy_index<P: AsRef<Path>>(state: &State, object_handle: &Handle, path: P) -> Result<()> {
    let entries = state.entries();
    let mut buf = Vec::new();
    for entry in entries {
        let obj = object_handle.find(entry.id, &mut buf)?;
        ...
    }
   ...
}
```

Byron:
And that kind of code is ok in tests, but there can't be a dependency of `git-worktree` to `git-odb`. Instead, this is abstracted into a closure [like this](https://github.com/Byron/gitoxide/blob/e3977fe033550bfd3297cdd674934e40476aa38b/git-pack/tests/pack/data/output/count_and_entries.rs#L246).

AP2008:
> And that kind of code is ok in tests, but there can't be a dependency of `git-worktree` to `git-odb`. Instead, this is abstracted into a closure [like this](https://github.com/Byron/gitoxide/blob/e3977fe033550bfd3297cdd674934e40476aa38b/git-pack/tests/pack/data/output/count_and_entries.rs#L246).

My function should take a closure ?

git-worktree/src/lib.rs
```diff
@@ -1 +1,62 @@
...
+                    if linked_to_path.exists() {
+                        std::fs::copy(linked_to_path, dest)?;
+                    } else {
+                        std::fs::write(dest, linked_to)?;
+                    }
+                }
+            }
+            Mode::DIR => todo!(),
+            Mode::COMMIT => todo!(),
+            _ => {}
```
Byron, line 58: Probably it's best to put a `todo()` or `unreachable!()` here as all cases should be clear in code (eventually).

git-worktree/src/lib.rs
```diff
@@ -1 +1,62 @@
...
+        let dest = path.join(entry.path(state).to_string());
+        create_dir_all(dest.parent().expect("Path has no parent"))?;
+        match entry.mode {
+            Mode::FILE | Mode::FILE_EXECUTABLE => {
+                let obj = find(&entry.id, &mut buf).unwrap();
+                std::fs::write(dest, obj.data)?;
+            }
+            Mode::SYMLINK => {
+                let obj = find(&entry.id, &mut buf).unwrap();
+                let linked_to = str::from_utf8(obj.data)?;
```
Byron, line 35: There is no way UTF-8 can be assumed here, this can't be more than a `Path` or `OsString`.

git-worktree/src/lib.rs
```diff
@@ -1 +1,62 @@
...
+where
+    P: AsRef<Path>,
+    Find: for<'a> FnMut(&oid, &'a mut Vec<u8>) -> Option<Data<'a>>,
+{
+    let path = path.as_ref();
+    let entries = state.entries();
+    let mut buf = Vec::new();
+    for entry in entries {
+        let dest = path.join(entry.path(state).to_string());
+        create_dir_all(dest.parent().expect("Path has no parent"))?;
```
Byron, line 27: I think most of the code already written optimizes for reading the `expect()` statement, so here it would be something like `file path has a directory`.

git-worktree/src/lib.rs
```diff
@@ -1 +1,62 @@
...
+pub fn copy_index<P, Find>(state: &State, path: P, mut find: Find, opts: Options) -> Result<()>
+where
+    P: AsRef<Path>,
+    Find: for<'a> FnMut(&oid, &'a mut Vec<u8>) -> Option<Data<'a>>,
+{
+    let path = path.as_ref();
+    let entries = state.entries();
+    let mut buf = Vec::new();
+    for entry in entries {
+        let dest = path.join(entry.path(state).to_string());
```
Byron, line 26: One cannot convert to a string as it assumed UTF-8 encoding, which is happening lossily here. Converting from bytes back to an OsStr isn't easy, but I think there is a crate for that `os_str_bytes` might be helpful.

Byron (commented):
I did a quick review to provide some guidance, while also pointing out the complexity that is in something that seems deceptively simple at first.

Thus it's probably best to take note of everything that is entailed in checking out a git file, and decide what to implement in this PR and maybe what to push into the next one to avoid long-lived PRs.

I strongly recommend to ditch multi-threading in the first iteration, and push it back once everything else works with a focus on correctness, ideally with an encompassing test for certain special behavior. As this can become complex quickly, putting less in each PR is probably more.

Even though it doesn't help to reduce complexity, it should help scoping this PR by looking at the requirements to checking out an entry coming [from the respective code in git](https://github.com/git/git/blob/main/entry.c#L459:L459). There is a lot of cases to handle and to test for.

git-worktree/Cargo.toml
```diff
@@ -13,3 +13,14 @@ doctest = false
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+git-index = { version = "^0.1.0", path = "../git-index" }
+rayon = "1.5.1"
```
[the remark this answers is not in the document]
AP2008: Yes. I missed it while cleaning up.

AP2008:
The `FILE` and `SYMLINK` implementations are pretty much the same as they are in git.
I didn't find much about `FILE_EXECUTABLE` in git's implementation. Should it just be made executable ?
The other implementations may require a submodule implementation(https://github.com/git/git/blob/main/entry.c#L377:L377) (Does gitoxide have one ?).
Or is there some other way ?

Byron:
> I didn't find much about FILE_EXECUTABLE in git's implementation. Should it just be made executable ?

I think [here is where the executable bit is set](https://github.com/git/git/blob/main/entry.c#L79:L79) - it's done while opening/creating the file. Maybe something similar is possible with Rust as well.

> The other implementations may require a submodule implementation(https://github.com/git/git/blob/main/entry.c#L377:L377) (Does gitoxide have one ?).

This should probably be outsourced into a function that is passed as parameter, that task cannot be performed by the `git-worktree` crate.

Generally I recommend looking and noting all features that are currently missing here (but available in git) maybe along with references to the git source code. That way it's possible to determine how far this PR should go with its implementation, and what could be next to get to an MVP level of correctness and capability.

AP2008:
> I think [here is where the executable bit is set](https://github.com/git/git/blob/main/entry.c#L79:L79) - it's done while opening/creating the file. Maybe something similar is possible with Rust as well.

Done! Had to modify the tests a bit. I set the executable bit only on unix-based systems. On windows, a regular file is created.

Byron:
In order to get this PR on track I think it's best to have an in-person chat on how to proceed, so I reached out to you by email and hope you can make time.

Thanks a lot.

AP2008:
- Should files marked with the flag `SKIP_WORKTREE` be skipped while copying.
- Should the copied files have the timestamps present in the entries ?

Byron:
> Should files marked with the flag SKIP_WORKTREE be skipped while copying.

I think so, but can't look up the git source right now.

> Should the copied files have the timestamps present in the entries ?

I think it's the other way around, the index entries are adjusted to have the new files stat information.

AP2008:
What do the `secs` and `nsecs` fields in `entry::Time` represent ? (Unix time ?)

Byron:
Yes, it's something that can be converted into `std::time::SystemTime`.

AP2008:
How do I change the timestamps of the `entries` as the `entries` functions returns an immutable reference ?

Byron:
Feel free to add a method that allows mutable access.

AP2008:
Why is the time stored in seconds and nanoseconds as u32 ? The nanoseconds will overflow a u32. The std lib stores and returns them as u64 and u128 respectively.

Byron:
> Does the nanos field represent the fractional part of the seconds or does it represent the actual nanoseconds since epoch ?

That's a great question, and I think it's save to assume that it's the fractional part of each second with nanosecond precision.  I assume the question arises from having `SystemTime` that needs to brought down to `u32` to match the in-memory representation git uses. In theory, `nsec` is never bigger than 1000000000, one billion.

AP2008:
Implemented the skip-worktree and timestamp behaviours. Please take a look.

git-worktree/Cargo.toml
```diff
@@ -13,3 +13,13 @@ doctest = false
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+git-index = { version = "^0.1.0", path = "../git-index" }
+anyhow = "1.0.42"
```
Byron, line 17: `anyhow` cannot be used in plumbing crates, `quickerror` is in its place here.

git-worktree/src/lib.rs
```diff
@@ -1 +1,95 @@
...
+use std::os::unix::fs::PermissionsExt;
+
+/// Copy index to `path`
+pub fn copy_index<P, Find>(state: &mut State, path: P, mut find: Find, opts: Options) -> Result<()>
+where
+    P: AsRef<Path>,
+    Find: for<'a> FnMut(&oid, &'a mut Vec<u8>) -> Option<Data<'a>>,
+{
+    let path = path.as_ref();
+    let paths = state
```
Byron, line 28: I think it would be better to traverse `entries()` at first, and keep track of paths which later need their `stat` information be update by keeping only that information around. It would then be used to update the respective index entry with.

git-worktree/src/lib.rs
```diff
@@ -1 +1,95 @@
...
+        .entries()
+        .iter()
+        .map(|e| e.path(state).to_path().map(|p| p.to_path_buf()))
+        .collect::<Result<Vec<_>, _>>()?;
+    let mut buf = Vec::new();
+    for (i, entry) in state.entries_mut().iter_mut().enumerate() {
```
Byron, line 34: Better use `entries()` here, related to comment above.

git-worktree/src/lib.rs
```diff
@@ -1 +1,95 @@
...
+                        if dest.is_file() {
+                            std::os::windows::fs::symlink_file(linked_to, dest)?;
+                        } else {
+                            std::os::windows::fs::symlink_dir(linked_to, dest)?;
+                        }
+                    }
+                } else {
+                    let linked_to_path = path.join(linked_to);
+                    if linked_to_path.exists() {
+                        std::fs::copy(linked_to_path, dest)?;
```
Byron, line 71: I don't think that's how git handles symlinks, as for this would make it look like the content of the file changed. Instead, I think it will write the `linked_to` portion directly into a file - after all, that's all symlinks are on linux, which makes sure the hash matches.

git-worktree/src/lib.rs
```diff
@@ -1 +1,95 @@
...
+        .entries()
+        .iter()
+        .map(|e| e.path(state).to_path().map(|p| p.to_path_buf()))
+        .collect::<Result<Vec<_>, _>>()?;
+    let mut buf = Vec::new();
+    for (i, entry) in state.entries_mut().iter_mut().enumerate() {
+        if entry.flags.contains(Flags::SKIP_WORKTREE) {
+            continue;
+        }
+        let time = SystemTime::now().duration_since(SystemTime::UNIX_EPOCH)?;
```
Byron, line 38: This must be the time returned by a `metadata()` call, as it must match what's actually on disk. Otherwise git will think the file changed.

AP2008:
The review changes have been fixed and implemented. Please take a look.

git-worktree/src/lib.rs
```diff
@@ -1 +1,111 @@
...
+                let obj = find(&entry.id, &mut buf).unwrap();
+                std::fs::write(&dest, obj.data)?;
+                if entry.mode == Mode::FILE_EXECUTABLE {
+                    #[cfg(unix)]
+                    fs::set_permissions(&dest, fs::Permissions::from_mode(0o777)).unwrap();
+                }
+                let met = std::fs::metadata(&dest)?;
+                // Set both fields to mtime if ctime is not present and vice-versa
+                // If both fields are not present, will raise an Err
+                let ctime = met.created().or_else(|_| met.modified())?;
```
Byron, line 56: This is well-meant and it would make sense to do that, too. Git unfortunately [doesn't go this far](https://github.com/git/git/blob/main/read-cache.c#L154:L154) and instead just sets these values to zero if the fields do not exist.

That said, if the `metadata()` call succeeds, there will be data for `created()` and `modified()` if the fields exist, and if not, these should just be set to zero. Zero is the value the struct fields would have after a memset.

The reason we do this is to write exactly the same index that git would write, to not alter its behavior if gitoxide and git are used on the same repo.

git-worktree/src/lib.rs
```diff
@@ -1 +1,111 @@
...
+        create_dir_all(dest.parent().expect("path is empty"))?;
+        match entry.mode {
+            Mode::FILE | Mode::FILE_EXECUTABLE => {
+                let obj = find(&entry.id, &mut buf).unwrap();
+                std::fs::write(&dest, obj.data)?;
+                if entry.mode == Mode::FILE_EXECUTABLE {
+                    #[cfg(unix)]
+                    fs::set_permissions(&dest, fs::Permissions::from_mode(0o777)).unwrap();
+                }
+                let met = std::fs::metadata(&dest)?;
```
Byron, line 53: For good measure, let's use `symlink_metadata()` everywhere.

git-worktree/src/lib.rs
```diff
@@ -1 +1,111 @@
...
+                        if dest.is_file() {
+                            std::os::windows::fs::symlink_file(linked_to, &dest)?;
+                        } else {
+                            std::os::windows::fs::symlink_dir(linked_to, &dest)?;
+                        }
+                    }
+                } else {
+                    std::fs::write(&dest, obj.data)?;
+                }
+                let met = std::fs::metadata(&dest)?;
```
Byron, line 77: For good measure, let's use `symlink_metadata()` everywhere. 
This is particularly important here as we are never interested in following the symlink we potentially just created to a file that definitely didn't change.

git-worktree/src/lib.rs
```diff
@@ -1 +1,111 @@
...
+        ToU32Error(err: std::num::TryFromIntError) { from() }
+        IoError(err: std::io::Error) { from() }
+    }
+}
+
+/// Copy index to `path`
+pub fn copy_index<P, Find>(state: &mut State, path: P, mut find: Find, opts: Options) -> Result<(), Error>
+where
+    P: AsRef<Path>,
+    Find: for<'a> FnMut(&oid, &'a mut Vec<u8>) -> Option<Data<'a>>,
```
Byron, line 34: Let's let the return object be a `git_object::BlobRef` (_fully qualified_) instead of bare data. This forces an additional type check which should never fail.

git-worktree/src/lib.rs
```diff
@@ -1 +1,111 @@
 #![forbid(unsafe_code, rust_2018_idioms)]
+//! Git Worktree
+
+use git_hash::oid;
+use git_index::{
```
Byron, line 5: Let's fully qualify the path to `git_index` items, instead of importing them to somewhat generic type names.

git-worktree/src/lib.rs
```diff
@@ -1 +1,111 @@
...
+            }
+            Mode::DIR => todo!(),
+            Mode::COMMIT => todo!(),
+            _ => unreachable!(),
+        }
+    }
+    let entries = state.entries_mut();
+    for (ctime, mtime, i) in entry_time {
+        let mtime = mtime.duration_since(std::time::UNIX_EPOCH)?;
+        entries[i].stat.mtime.secs = u32::try_from(mtime.as_secs())?;
```
Byron, line 92: For good measure, this should be `let stat = &mut entries[i].stat`

Byron (commented):
It's definitely getting there!
I hope I get to doing the full fledged analysis soon to learn how to tame this beast.

git-index/src/lib.rs
```diff
@@ -22,6 +22,10 @@ mod access {
         pub fn entries(&self) -> &[Entry] {
```
Byron, line 22: The lib should start out with: `#![deny(unsafe_code, rust_2018_idioms)]`.

git-worktree/src/lib.rs
AP2008: Even if the `metadata` call succeeds, `mtime` or `ctime` may fail. My initial assumption was that they would succeed but `mtime` succeeds on unix and `ctime` fails as unix does not store creation time.

To quote https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_09, 
> Each file has three distinct associated timestamps: the time of last data access, the time of last data modification, and the time the file status last changed

Byron: It's undisputed that they may fail, it's the fallback which isn't correct as git keeps these values at 0. There is code linked for reference which means that it will just copy whatever is present in a previously 0-initialized struct.

git-index/src/lib.rs
AP2008: Doesn't it have one here: https://github.com/Byron/gitoxide/blob/91c064e48692f25f04a3d1ae8dacb7056d307bda/git-index/src/lib.rs#L1 ?

AP2008:
Fixed. Please take a look.

git-worktree/src/lib.rs
```diff
@@ -1 +1,127 @@
...
+
+#[cfg(unix)]
+use std::os::unix::fs::PermissionsExt;
+
+quick_error! {
+    #[derive(Debug)]
+    pub enum Error {
+        Utf8Error(err: git_object::bstr::Utf8Error) {
+            from()
+            display("Utf8 Error: {}", err)
```
Byron, line 22: This `Display()` implementation is probably worse than having none. The idea is to provide context about what this part of the program tries to do when it happened. The same is true for all following `display(…)` specifications.

Byron (changes requested):
Maybe let me resolve conversation items next time so I get to decide if it's resolved or not.

Something you could do to mark them as done is to give it some sort of reaction - that's how I do it in one of my PRs, facilitating the reviewers work who otherwise would have to find and go through resolved comments otherwise before reviewing everything again.

git-worktree/src/lib.rs
AP2008: Should it be something like `"Couldn't convert path to UTF8"` ?

AP2008:
Committed. Please take a look.

git-index/src/lib.rs
Byron: Sorry, I was looking at `git-index` all the time 🤦‍♂️.

git-worktree/src/lib.rs
```diff
@@ -1 +1,126 @@
...
+{
+    let path = path.as_ref();
+    let mut buf = Vec::new();
+    let mut entry_time = Vec::new(); // Entries whose timestamps have to be updated
+    for (i, entry) in state.entries().iter().enumerate() {
+        if entry.flags.contains(git_index::entry::Flags::SKIP_WORKTREE) {
+            continue;
+        }
+        let dest = path.join(entry.path(state).to_path()?);
+        create_dir_all(dest.parent().expect("path is empty"))?;
```
Byron, line 52: All over the codebase, one will write what's expected to be the case, instead of what went wrong. Here it would be more like `"entry paths are never empty"`.

git-worktree/src/lib.rs
```diff
@@ -1 +1,126 @@
...
+            from()
+            display("Could not read file time in proper format: {}", err)
+        }
+        ToU32Error(err: std::num::TryFromIntError) {
+            from()
+            display("Could not convert seconds to u32: {}", err)
+        }
+        IoError(err: std::io::Error) {
+            from()
+            display("IO Error: {}", err)
```
Byron, line 33: "IO error while writing blob or reading file metadata" seems to be the reason for IO errors to occur. That should be reflected in the message.

git-worktree/src/lib.rs
```diff
@@ -1 +1,126 @@
...
+        }
+        let dest = path.join(entry.path(state).to_path()?);
+        create_dir_all(dest.parent().expect("path is empty"))?;
+        match entry.mode {
+            git_index::entry::Mode::FILE | git_index::entry::Mode::FILE_EXECUTABLE => {
+                let obj = find(&entry.id, &mut buf).unwrap();
+                std::fs::write(&dest, obj.data)?;
+                if entry.mode == git_index::entry::Mode::FILE_EXECUTABLE {
+                    #[cfg(unix)]
+                    fs::set_permissions(&dest, fs::Permissions::from_mode(0o777)).unwrap();
```
Byron, line 59: There should never be a panic if something fails, but instead of avoiding `unwrap` here, I think file creation/opening should take care of the execute bit like [git does](https://github.com/git/git/blob/main/entry.c#L81:L82). This can be done with [`custom_flags()`](https://doc.rust-lang.org/std/os/unix/fs/trait.OpenOptionsExt.html#tymethod.custom_flags).

git-worktree/src/lib.rs
```diff
@@ -1 +1,126 @@
...
+    let mut entry_time = Vec::new(); // Entries whose timestamps have to be updated
+    for (i, entry) in state.entries().iter().enumerate() {
+        if entry.flags.contains(git_index::entry::Flags::SKIP_WORKTREE) {
+            continue;
+        }
+        let dest = path.join(entry.path(state).to_path()?);
+        create_dir_all(dest.parent().expect("path is empty"))?;
+        match entry.mode {
+            git_index::entry::Mode::FILE | git_index::entry::Mode::FILE_EXECUTABLE => {
+                let obj = find(&entry.id, &mut buf).unwrap();
```
Byron, line 55: There can be no panics, and not finding an object should just be an error as descriptive [as this one](https://github.com/git/git/blob/main/entry.c#L297:L299).

git-worktree/tests/copy_index/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+    let copy_files = dir_structure(output);
+
+    let srepo_files: Vec<_> = repo_files.iter().flat_map(|p| p.strip_prefix(&path)).collect();
+    let scopy_files: Vec<_> = copy_files.iter().flat_map(|p| p.strip_prefix(output)).collect();
+    assert!(srepo_files == scopy_files);
+
+    for (file1, file2) in repo_files.iter().zip(copy_files.iter()) {
+        assert!(fs::read(file1)? == fs::read(file2)?);
+        #[cfg(unix)]
+        assert!(fs::symlink_metadata(file1)?.mode() & 0b111 << 6 == fs::symlink_metadata(file2)?.mode() & 0b111 << 6);
```
Byron, line 43: If there is equality, use `assert_eq!`. It's odd that clippy doesn't warn about this.
Especially with assertions that aren't entirely obvious to the untrained eye, can you use the third argument to say what it is testing?

git-worktree/tests/copy_index/mod.rs
```diff
@@ -0,0 +1,91 @@
+use crate::{dir_structure, fixture_path};
+use git_object::bstr::ByteSlice;
+use git_odb::FindExt;
+use git_worktree::{copy_index, Options};
+use std::fs;
+
+#[cfg(unix)]
+use std::os::unix::prelude::MetadataExt;
+
+type Result<T = ()> = std::result::Result<T, Box<dyn std::error::Error>>;
```
Byron, line 10: This should be on `crate` level, this is a module. Usages should be `crate::Result`;

git-worktree/tests/copy_index/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+#[test]
+fn test_copy_index_without_symlinks() -> Result<()> {
+    let path = fixture_path("make_repo");
+    let path_git = path.join(".git");
+    let mut file = git_index::File::at(path_git.join("index"), git_index::decode::Options::default())?;
+    let output_dir = tempfile::tempdir()?;
+    let output = output_dir.path();
+    let odb_handle = git_odb::at(path_git.join("objects")).unwrap();
```
Byron, line 56: If a test returns `crate::Result`, there should be no `unwrap()` anymore. Using `crate::Result` is only recommended at the very end of the development cycle when it's all working and stable, as it's generally harder to see where it breaks then, compared to a panic using `unwrap()`. Probably `anyhow` is better in that regard.

It's generally OK to use `unwrap()` in tests while having it return `()`, as in no explicit return value, if that makes debugging easier.

git-worktree/tests/copy_index/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    copy_index(
+        &mut file,
+        &output,
+        move |oid, buf| {
+            odb_handle
+                .find(oid, buf)
```
Byron, line 63: There is `find_blob` that you could use instead.

Byron (commented):
It's getting there. As a general note, production code should never use `unwrap()`.

AP2008:
Committed. Please take a look.

git-worktree/src/lib.rs
```diff
@@ -1 +1,126 @@
...
+        }
+        ToU32Error(err: std::num::TryFromIntError) {
+            from()
+            display("Could not convert seconds to u32: {}", err)
+        }
+        IoError(err: std::io::Error) {
+            from()
+            display("IO error while writing blob or reading file metadata or changing filetype: {}", err)
+        }
+        FindOidError(oid: git_hash::ObjectId, path: std::path::PathBuf) {
```
Byron, line 35: For consistency, this should be called `NotFound { oid: git_hash::ObjectId, path: PathBuf}`.

git-worktree/src/lib.rs
```diff
@@ -1 +1,126 @@
...
+            display("unable to read sha1 file of {} ({})", path.display(), oid.to_hex())
```
Byron, line 36: There is no such thing as a `SHA1 file`, for all you know an object couldn't be found. This text should be reworded to reflect that.
I like that the path at which the object would be checked out at is mentioned as well.

git-worktree/src/lib.rs
Byron: Is this something you tried?

git-worktree/src/lib.rs
```diff
@@ -1 +1,126 @@
...
+                let obj = find(&entry.id, &mut buf).ok_or_else(|| Error::FindOidError(entry.id, path.to_path_buf()))?;
+                std::fs::write(&dest, obj.data)?;
+                if entry.mode == git_index::entry::Mode::FILE_EXECUTABLE {
+                    #[cfg(unix)]
+                    fs::set_permissions(&dest, fs::Permissions::from_mode(0o777))?;
+                }
+                let met = std::fs::symlink_metadata(&dest)?;
+                let ctime = met
+                    .created()
+                    .map_or(Ok(Duration::from_secs(0)), |x| x.duration_since(std::time::UNIX_EPOCH));
```
Byron, line 68: This should be `let ctime = met.created()?.duration_since(std::time::UNIX_EPOCH)?;
`

git-worktree/tests/copy_index/mod.rs
Byron: > Especially with assertions that aren't entirely obvious to the untrained eye, can you use the third argument to say what it is testing?

What about the third argument, a message describing what it is supposed to test?

git-worktree/tests/copy_index/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+use git_object::bstr::ByteSlice;
+use git_odb::FindExt;
+use git_worktree::{copy_index, Options};
+use std::fs;
+
+#[cfg(unix)]
+use std::os::unix::prelude::MetadataExt;
+
+#[test]
+fn test_copy_index() -> Result<()> {
```
Byron, line 11: Please use `crate::Result<()>`.

git-worktree/src/lib.rs
```diff
@@ -1 +1,126 @@
...
+                let mtime = met
+                    .modified()
+                    .map_or(Ok(Duration::from_secs(0)), |x| x.duration_since(std::time::UNIX_EPOCH));
+                entry_time.push((ctime?, mtime?, i));
+            }
+            git_index::entry::Mode::SYMLINK => {
+                let obj = find(&entry.id, &mut buf).unwrap();
```
Byron, line 75: Why is the `unwrap()` fixed above but not here?

Byron (changes requested):
It's important to stop amending to the same commit, instead, create as many commits on top of this one and push these instead, ideally with more descriptive messages than `implemented git-worktree`

Every time there is a change I have to look through everything and match all my comments, the time this takes is considerable. This gets frustrating as it seems no great effort is made to fully address all change requests either by applying a fix or by proposing alternatives.

Lastly, it would be great to see this rebased onto the latest `main` branch to resolve the conflict.

Thank you

git-worktree/src/lib.rs
AP2008: I tried implementing you suggestion and was able to but it made that part of the code more complicated.
So, I decided to set the permission bit after writing. Is this fine ?

Byron: Yes, please, use `custom_flags()` - let's not use more IO operations than git for critical functionality like this.

git-worktree/src/lib.rs
AP2008: The same error is used in the git source code you linked: https://github.com/git/git/blob/main/entry.c#L297:L299

git-worktree/src/lib.rs
AP2008: Fixed: https://github.com/Byron/gitoxide/pull/315/commits/eaee85595dc658549e62e3292b025ec016e70abd

git-worktree/src/lib.rs
AP2008: Fixed: https://github.com/Byron/gitoxide/pull/315/commits/eaee85595dc658549e62e3292b025ec016e70abd

git-worktree/src/lib.rs
```diff
@@ -1 +1,126 @@
...
+use std::path::Path;
+use std::time::Duration;
+
+#[cfg(unix)]
+use std::os::unix::fs::PermissionsExt;
+
+quick_error! {
+    #[derive(Debug)]
+    pub enum Error {
+        Utf8Error(err: git_object::bstr::Utf8Error) {
```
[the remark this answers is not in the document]
AP2008: Fixed: https://github.com/Byron/gitoxide/pull/315/commits/eaee85595dc658549e62e3292b025ec016e70abd

git-worktree/src/lib.rs
AP2008: This will fail on Unix as unix does not store `ctime`. My implementation just sets the `ctime` to `0` if it does not exist as you had previously suggested.

git-worktree/src/lib.rs
AP2008: Fixed: https://github.com/Byron/gitoxide/commit/e838eaa5721d8b1b13155aa81234c9c44d9b15fe
Reduces 2 io calls (metadata and permissions)

git-worktree/tests/copy_index/mod.rs
AP2008: Done: https://github.com/Byron/gitoxide/commit/25a9dc16dbb26e9aa0f3379b2af53cc0baa96663

git-worktree/tests/copy_index/mod.rs
AP2008: Done: https://github.com/Byron/gitoxide/commit/25a9dc16dbb26e9aa0f3379b2af53cc0baa96663

Byron (commented):
I think it's time to merge `main` back and resolve conflicts in the process.

Can't be long now till it's ready for a merge.

git-worktree/tests/copy_index/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+        &mut file,
+        &output,
+        move |oid, buf| odb_handle.find_blob(oid, buf).ok(),
+        Options { symlinks: false },
+    )?;
+
+    let repo_files = dir_structure(&path);
+    let copy_files = dir_structure(output);
+
+    let srepo_files: Vec<_> = repo_files.iter().flat_map(|p| p.strip_prefix(&path)).collect();
```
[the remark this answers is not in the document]
AP2008: Inlined: https://github.com/Byron/gitoxide/pull/315/commits/a4b880cf17665b61e3f7f193de57704b1db5318f

git-worktree/src/lib.rs
AP2008: Done: https://github.com/Byron/gitoxide/pull/315/commits/a4b880cf17665b61e3f7f193de57704b1db5318f

git-worktree/src/lib.rs
AP2008: Done: https://github.com/Byron/gitoxide/pull/315/commits/a4b880cf17665b61e3f7f193de57704b1db5318f

AP2008:
Fixed and resolved conflicts!

Byron:
Thanks a lot for your first contribution! There is a lot of potential, and I think it's best channeled after I did my due diligence with analyzing and understanding git-worktree myself. This should lead to smaller tasks that are by themselves better understood - please feel free to subscribe to [the tracking issue](https://github.com/Byron/gitoxide/issues/301) for updates, and I will ping you once I think it's ready to start.

I also recommend looking at the crate as it is now.

Byron:
I am closing the PR as github seems to not recognize it's already merged.

ErichDonGubler:
@Byron (CC @AP2008): RE: https://github.com/Byron/gitoxide/pull/315#pullrequestreview-871543049: I would strongly recommend `git`'s `git rebase --autosquash` functionality for alleviating this pain. It makes it a lot easier to track the completion of feedback items, often letting the reviewee point to Git commit hashes instead of just stating, "I promise I fixed this up!" and leaving you to manually diff code yourself.

If you'd like me to give some more concrete examples, it might be the motivation I need to finally commit to writing a blog post about it. :) I've been dealing with similar review pain points in my professional work at @NZXTCorp.

Byron:
Thanks @ErichDonGubler , that's a great suggestion and a feature I wasn't aware off. A quick search brought me to [this article](https://thoughtbot.com/blog/autosquashing-git-commits) which revealed everything one would want to know on how to use it effectively. 
Maybe that's the blog post you'd write, or even the one you have written in the mean while :).

As a summary, I have a learned a lot when trying to create PRs myself and when reviewing them, and it's not exactly working perfectly out of the box as it requires some experience for both involved parties.  `--autosquash` along with `commit --fixup` can really make the difference here, so I am looking forward to apply it in practice.

Lastly, I think it would be great if GitHub would automatically detect `--fixup` commits and maybe if provided with the URL of the comment they are fixing, adds this information to the discussion automatically. Is it that sophisticated? I will definitely try to find out :).

ErichDonGubler:
> Lastly, I think it would be great if GitHub would automatically detect `--fixup` commits and maybe if provided with the URL of the comment they are fixing, adds this information to the discussion automatically. Is it that sophisticated? I will definitely try to find out :).

I wish that this was the case, but AFAIK GitHub has no functionality built around a `--fixup` flow -- just happiness from keeping history append-only during critical points of a review. 😅

PR #321, opened 2022-01-24
fix(git-config): fix usage example in README.md

sassman:
this PR just fixes the git-config usage example in the readme.

### DIFF
--- a/git-config/README.md
+++ b/git-config/README.md
@@ -31,9 +31,9 @@ let input = r#"
 [other "internal"]
   hello = world
 "#;
-let mut config = GitConfig::from(input)?;
-let boolean = config.get_config::<Boolean>("core", None, "some-bool");
-config.set_value("other", Some("internal"), "hello", "clippy!");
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot, much appreciated.

I think there is a crate that is able to either run Rust code in the README or put Rust code from where it is tested in the README. Maybe this would be an option at some point to protect from this kind of regression.

sassman:
yes, there is https://github.com/rust-lang/rfcs/blob/master/text/1990-external-doc-attribute.md

that has this syntax sugar:

```rust
#[doc(include = "../docs/example.md")]
fn my_func() {
  // Hidden implementation
}
```

- that solves the problem of, docs contain test code that should better run on `cargo test`
- and that documentation is not anymore duplicate (doc in code and external readme/doc files)

Byron:
That's nice, even built-in. I checked the RFC to see if there was a way to take a portion of a file only, but that doesn't seem to be available just yet.

PR #323, opened 2022-01-24
feat(git-config:subsections): implement a draft for subsection fetching

### LINKED ISSUE #319: is there a way to retrieve a subsection name
given a gitconfig with the `[url xxxx]` sections as below.

I need to iterate over all `url` sections and be able to retrieve the `insteadOf` config key (which works already) and the `xxxx` that is the subsection name (see the todo note)

## example

```rust
    use git_config::file::GitConfig;
    use git_config::parser::{Key, Parser};
    use git_config::values::Value;
    use std::fs::read_to_string;

    let input = r#"
[url "ssh://git@github.com/"]
    insteadOf = https://github.com/
[url "ssh://git@bitbucket.org"]
    insteadOf = https://bitbucket.org/
"#;
    let config = GitConfig::try_from(input).unwrap();
    let url = config.se.sections_by_name("url");
    assert_eq!(url.len(), 2);

    for (i, u) in url.iter().enumerate() {
        let instead_of = u.value(&Key::from("insteadOf"));

        // todo(sub-section-name)
        // let xxxxx = u.retrieve_parent_sub_section_name();
        // assert_eq!(xxxxx, "ssh://git@github.com/")

        if i == 1 {
            assert_eq!(
                instead_of.unwrap().deref(),
                "https://github.com/".as_bytes()
            )
        } else {
            assert_eq!(
                instead_of.unwrap().as_ref(),
                "https://bitbucket.org/".as_bytes()
            )
        }
    }
```

Maybe there is also a better way of doing this, then I would be very pleased if you can give me a hint.

sassman:
- introduce method `sections_by_name_with_header` to allow iterating over tuples of
  section header and section body
- remove the integration test in favor of a doctest
- doctest is also a comprehensive usage example
- leave a todo note to illustrate the inconsitent order issue, so sections
  do not always arrive in the same order
- fixes #319

### DIFF
--- a/git-config/src/file/git_config.rs
+++ b/git-config/src/file/git_config.rs
@@ -463,6 +463,70 @@ impl<'event> GitConfig<'event> {
             .collect()
     }
 
+    /// Get all sections that match the provided name. Returning the section header and body.
+    ///
+    /// # Example
+    ///
+    /// Provided the following config:
+    /// ```plain
+    /// [url "ssh://git@github.com/"]
+    ///     insteadOf = https://github.com/
+    /// [url "ssh://git@bitbucket.org"]
+    ///     insteadOf = https://bitbucket.org/
+    /// ```
+    /// Calling this method will yield all section bodies and their header:
+    ///
+    /// ```rust
+    /// use git_config::file::{GitConfig, GitConfigError};
+    /// use git_config::parser::Key;
+    /// use std::borrow::Cow;
+    /// use std::convert::TryFrom;
+    /// use nom::AsBytes;
+    ///
+    /// let input = r#"
+    /// [url "ssh://git@github.com/"]
+    ///    insteadOf = https://github.com/
+    /// [url "ssh://git@bitbucket.org"]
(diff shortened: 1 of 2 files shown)


Byron:
That's great work, thanks so much!
I particularly like the idea of following this crates precedence and use doc-tests, it's something that I definitely under-use at the moment.

The `// TODO` can certainly remain for another day, and after taking a look I would actually have expected these to be stable.
Lastly, looking at the example, it shows once more why I avoid using `git-config` in its current state as it's quite cumbersome to obtain values at the moment. It's definitely something that will be improved in future iterations. However, since you are using it more intensely [than I do](https://github.com/Byron/gitoxide/blob/8b9da35b3e0d3458efcac150f7062c9d7382a6c4/git-repository/src/repository.rs#L109), please be welcome to make suggestions or PRs that improve the usability, breaking or not :).

sassman:
Thank you too. 
If I do see specific actions on how to iterate on the API for git-config, I will take them and come up with suggestions. But I know that this is not a trivial thing to do.

Issue #326, opened 2022-02-02
Add Path value to git-config


svetli-n:
See also https://github.com/Byron/gitoxide/blob/main/crate-status.md#git-config

Byron:
I have created [a new tracking](https://github.com/Byron/gitoxide/issues/331) issue which can be an umbrella for all PRs related to git-config, avoiding to have an issue per PR.

PR #327, opened 2022-02-02
Path value

svetli-n:
Closes https://github.com/Byron/gitoxide/issues/326

### DIFF
--- a/git-config/src/values.rs
+++ b/git-config/src/values.rs
@@ -382,13 +382,13 @@ pub mod path {
     }
 
     #[cfg(test)]
-    mod tests {
+    mod interpolate_tests {
         use std::borrow::Cow;
 
         use crate::values::{path::interpolate::Error, Path};
 
         #[test]
-        fn not_interpolated() {
+        fn no_interpolation_for_paths_without_tilde_or_prefix() {
             let path = &b"/foo/bar"[..];
             let actual = Path::from(Cow::Borrowed(path));
             assert_eq!(&*actual, path);
@@ -399,7 +399,7 @@ pub mod path {
         }
 
         #[test]
-        fn empty_is_error() {
+        fn empty_path_is_error() {
             assert!(matches!(
                 Path::from(Cow::Borrowed("".as_bytes())).interpolate(None),
                 Err(Error::Missing { what: "path" })
@@ -407,7 +407,7 @@ pub mod path {
         }
 
         #[test]
-        fn prefix_interpolated() {
+        fn prefix_substitutes_git_install_dir() {
             for git_install_dir in &["/tmp/git", "C:\\git"] {
                 for (val, expected) in &[
                     (&b"%(prefix)/foo/bar"[..], "foo/bar"),
@@ -431,7 +431,7 @@ pub mod path {
         }
 
         #[test]
-        fn disabled_prefix_interpolation() {
+        fn prefix_substitution_skipped_with_dot_slash() {
             let path = "./%(prefix)/foo/bar";
             let git_install_dir = "/tmp/git";
             assert_eq!(
@@ -443,7 +443,7 @@ pub mod path {
         }
 
         #[test]
-        fn tilde_interpolated() {
+        fn tilde_substitutes_current_user() {
             let path = &b"~/foo/bar"[..];
             let expected = format!(
                 "{}{}foo/bar",
@@ -459,16 +459,16 @@ pub mod path {
 
         #[cfg(target_os = "windows")]
         #[test]
-        fn user_interpolated() {
+        fn tilde_with_given_user_is_unsupported_on_windows() {
             assert!(matches!(
                 Path::from(Cow::Borrowed(&b"~baz/foo/bar"[..])).interpolate(None),
-                Err(path::interpolate::Error::UserInterpolationUnsupported)
+                Err(Error::UserInterpolationUnsupported)
             ));
         }
 
         #[cfg(not(target_os = "windows"))]
         #[test]
-        fn user_interpolated() {
+        fn tilde_with_given_user() {
             let user = std::env::var("USER").unwrap();
             let home = std::env::var("HOME").unwrap();
             let specific_user_home = format!("~{}", user);
--- a/git-config/src/values.rs
+++ b/git-config/src/values.rs
@@ -298,8 +298,8 @@ quick_error! {
         PwdFileQuery {
             display("User home info missing")
         }
-        Unsupported {
-            display("Not available on this platform")
+        UserInterpolationUnsupported {
+            display("User interpolation is not available on this platform")
         }
     }
 }
@@ -308,9 +308,23 @@ quick_error! {
 ///
 /// Git represents file paths as byte arrays, modeled here as owned or borrowed byte sequences.
 #[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
-#[allow(missing_docs)]
 pub struct Path<'a> {
+    /// The interpolated path
     pub value: Cow<'a, [u8]>,
+}
+
+impl<'a> std::ops::Deref for Path<'a> {
+    type Target = [u8];
+
+    fn deref(&self) -> &Self::Target {
+        self.value.as_ref()
+    }
+}
+
+impl<'a> AsRef<[u8]> for Path<'a> {
+    fn as_ref(&self) -> &[u8] {
+        self.value.as_ref()
+    }
 }
 
 impl<'a> TryFrom<Cow<'a, [u8]>> for Path<'a> {
@@ -373,7 +387,7 @@ impl Path<'_> {
 
     #[cfg(target_os = "windows")]
     fn interpolate_user(_val: Cow<[u8]>, _slash: u8) -> Result<Path, PathError> {
-        Err(PathError::Unsupported)
+        Err(PathError::UserInterpolationUnsupported)
     }
 
     #[cfg(not(target_os = "windows"))]
@@ -1524,12 +1538,11 @@ mod path {
     #[test]
     fn not_interpolated() {
         let path = &b"/foo/bar"[..];
-        let borrowed_path = Cow::Borrowed(path);
-        assert_eq!(
-            Path::try_from(borrowed_path).unwrap(),
-            Path {
-                value: Cow::Borrowed(path)
-            }
+        let actual = Path::try_from(Cow::Borrowed(path)).unwrap();
+        assert_eq!(&*actual, path);
+        assert!(
+            matches!(&actual.value, Cow::Borrowed(_)),
+            "it does not unnecessarily copy values"
         );
     }
 
@@ -1547,17 +1560,24 @@ mod path {
         let git_install_dir = "/tmp/git";
         let expected = format!("{}/foo/bar", git_install_dir);
         assert_eq!(
-            Path::interpolate(val, Some(std::path::Path::new(git_install_dir))).unwrap(),
-            Path {
-                value: Cow::Borrowed(expected.as_bytes())
-            }
+            &*Path::interpolate(val, Some(std::path::Path::new(git_install_dir))).unwrap(),
+            expected.as_bytes()
+        );
+    }
+
+    #[test]
+    fn disabled_prefix_interpoldation() {
+        let path = &b"./%(prefix)/foo/bar"[..];
+        let git_install_dir = "/tmp/git";
+        assert_eq!(
+            &*Path::interpolate(Cow::Borrowed(path), Some(std::path::Path::new(git_install_dir))).unwrap(),
+            path
         );
     }
 
     #[test]
     fn tilde_interpolated() {
         let path = &b"~/foo/bar"[..];
-        let borrowed_path = Cow::Borrowed(path);
         let home = dirs::home_dir()
             .expect("empty home")
             .to_str()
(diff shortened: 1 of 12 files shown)


svetli-n:
@Byron Related to the encoding discussion earlier, I return Err when encoding error, like the existing code, e.g. https://github.com/Byron/gitoxide/blob/main/git-config/src/values.rs#L332

WDYT?

git-config/src/values.rs
```diff
@@ -4,6 +4,7 @@ use std::{borrow::Cow, convert::TryFrom, fmt::Display, str::FromStr};
 
 #[cfg(feature = "serde")]
 use serde::{Serialize, Serializer};
+use std::path::{Path as StdPath, PathBuf};
```
Byron, line 7: I would go with fully qualified paths if there is a conflict, like `std::path::Path`.

git-config/src/values.rs
```diff
@@ -237,6 +243,7 @@ impl From<&Value<'_>> for Vec<u8> {
             Value::Boolean(b) => b.into(),
             Value::Integer(i) => i.into(),
             Value::Color(c) => c.into(),
+            Value::Path(p) => p.value.to_str().expect("path is valid utf-8").as_bytes().into(),
```
Byron, line 246: No encoding can be assumed, definitely not UTF-8. I didn't look into git's source code yet to see what kind of requirements apply to a path stored in a git-config file. The assure round-tripping, what's stored in `p` can't be the decoded value. This crate is already provide means of 'normalizing' values, which deals with escaping values and the likes, probably using this machinery be valuable here.

git-config/src/values.rs
```diff
@@ -273,6 +281,68 @@ impl Serialize for Value<'_> {
...
+pub struct Path<'a> {
+    pub value: Cow<'a, StdPath>,
+}
+
+impl<'a> TryFrom<&'a str> for Path<'a> {
+    type Error = String;
+
+    fn try_from(value: &'a str) -> Result<Self, Self::Error> {
+        let path = StdPath::new(value);
+        if !path.exists() {
```
Byron, line 295: Where does the requirement for paths needing to exist come from?

git-config/src/values.rs
```diff
@@ -250,6 +257,7 @@ impl Display for Value<'_> {
             Value::Boolean(b) => b.fmt(f),
             Value::Integer(i) => i.fmt(f),
             Value::Color(c) => c.fmt(f),
+            Value::Path(p) => p.value.to_str().expect("path is valid utf-8").fmt(f),
```
Byron, line 260: Depending on how the path's encoding is handled, I think it would be very similar if not the same as `Value::Other()`.

git-config/src/values.rs
```diff
@@ -1390,3 +1460,30 @@ mod color_attribute {
...
+    use std::borrow::Cow;
+    use std::path::Path as StdPath;
+    use tempfile::tempdir;
+    use std::fs::File;
+
+    #[test]
+    fn ok() {
+        let dir = tempdir().unwrap();
+        let config_path = dir.path().join("foo");
+        let config_path = config_path.as_path();
```
Byron, line 1476: It's certainly a good way to warm up with these tests. What I'd be interested in seeing is how paths are handled within a git config file.

To my mind, they are indistinguishable from strings, the caller knows if something is supposed to be a path or not.

Thus it's the way a string-like value (I think it's `Other`) here is accessed, it's the callers choice to interpret it as `Path` and the API should help to resolve it as needed.

If this is correct, there would be no `Path` variant in `Value`, as it can't be known in advance.  (I see now why it checks for the existence of a path - it's really to find out if a string value is a path or not. This, however, can't be determined for sure without resolution being available at parse time, not to talk about the IO costs of parsing a config file that way.)

Byron (commented):
tl;dr: It's probably a good idea to read all comments in order, they 'add up' as my understanding improves as I write them. Here is a summary of what I know, hopefully that leads to a course of action.

----

I took a look and couldn't find any code that would actually test serialization. It appears that `serde` was originally intended to be used for serialization, but without tests I am not sure.

Here is what I know:

- this crate has a complete parser and machinery for producing human-friendly errors
- it can do zero-copy
- it maintains enough information to theoretically write values back as it parsed them (leaving comments). However, we can't find tests for that.
- everything is well documented, there are many doc-tests, which should help a lot in figuring out what's needed to get this crate over the finishing line.
- There is a lot of complexity to support round-tripping, but it's seemingly not tested.
   - In that vein, I didn't see normalized values being denormalized when serialized, so probably serde isn't meant to be used for writing `git-config` files (or I am missing something, or it's not yet implemented)

For now it seems like denormalizing values eagerly is how it's done here (I would have preferred to do that lazily), so eager normalization is probably what would happen to paths as well.

I think it's worth focussing on tests that interpret strings as paths and experiment with an API that makes this somewhat comfortable. Maybe from there one gains enough knowledge to understand how to approach this crate and which steps to take next. Right now, my main issue is that there is a lot of complexity without the necessary tests/features to justify it. It's still possible I am missing something obvious here, and a way to deal with this would be a complete study/read of all 7k lines here.

git-config/src/values.rs
svetli-n: Yep, I also feel like looking into git's source code to see how they handle path. And will also look closer at 'normalizing' values.

But see also how existing code "assumes" UTF-8. For example, 
       ```
 T::try_from(self.get_raw_value(section_name, subsection_name, key)?)
            .map_err(|_| GitConfigError::FailedConversion)
```
would go to 
`impl TryFrom<Cow<'_, [u8]>> for Integer {`
then to e.g
`Self::from_str(std::str::from_utf8(s).map_err(|_| ())?).map_err(|_| ())`

So we can get `GitConfigError::FailedConversion`

That should be ok either?

File: git-config/src/values.rs (hunk 3 above)
Comment (svetli-n) [reply git-config/src/values.rs:295]: like you said to be able to distinguish from str but yes its IO

File: git-config/src/values.rs (hunk 4 above)
Comment (svetli-n) [reply git-config/src/values.rs:260]: May be, but then there are not std::path::Path related methods, not sure of these are needed. So good reason to look at git

File: git-config/src/values.rs (hunk 5 above)
Comment (svetli-n) [reply git-config/src/values.rs:1476]: Yep, if we can't deal with Path here, it can be bubbled up the call stack. But then the value abstraction is may be abit broken. 

Or we can do like Integer, described above and get `FailedConversion`. The caller then would know its encoding error and can ask for `Other`?

**Comment by svetli-n** [comment]
Looked at git and it seems to me that its treating Path as possibly interpolated string: https://github.com/git/git/blob/master/config.c#L1253

hence these changes, WDYT?

File: git-config/src/values.rs (hunk 2 above)
Comment (Byron) [reply git-config/src/values.rs]: If it was enforced, it indeed wouldn't. However, here it seems the caller decides which type to use and if that is `String` or `str`, they probably know the value is not expected to be encoded differently.

The existing usages of `str::from_utf8(…)` also only appear in places where this is expected to succeed, i.e. for decimals or booleans, which are known to be ascii only.

File: git-config/src/values.rs (hunk 5 above)
Comment (Byron) [reply git-config/src/values.rs]: > But then the value abstraction is may be abit broken?

I don't know if it's broken, maybe it's something that most users wouldn't need as they usually know what type to expect when querying a value, so no need to 'try' interpreting it as `Value` seems to be doing.

> Or we can do like Integer, described above and get FailedConversion. The caller then would know its encoding error and can ask for Other?

Maybe, to me it all depends on how it's going to be used and for all I know, I'd ask for `path_value(…<context for resolution>)` and it can handle things correctly from there.

File: git-config/src/values.rs [hunk 6]
```diff
@@ -273,6 +280,47 @@ impl Serialize for Value<'_> {
...
+
+impl Path<'_> {
+    fn interpolate(val: Cow<'_, [u8]>) -> Result<Path<'_>, ()> {
+        if val.is_empty() {
+            return Err(());
+        }
+
+        if val.starts_with(b"~/") {
+            let mut home = env::var("HOME").map_err(|_| ())?;
+            if env::consts::OS == "windows" {
```
Byron, line 305: Maybe use `cfg!(target_os = "windows")` here to get it at compile time.

git-config/src/values.rs
```diff
@@ -273,6 +280,47 @@ impl Serialize for Value<'_> {
...
+                home = home.replace("\\", "/");
+            }
+            let mut expanded = vec![];
```
Byron, line 308: Could this be `let mut expanded = home.to_owned()`?;

git-config/src/values.rs
```diff
@@ -273,6 +280,47 @@ impl Serialize for Value<'_> {
...
+            expanded.extend(home.as_bytes().into_iter());
+            if !home.ends_with("/") {
+                expanded.push(b'/');
+            }
+            expanded.extend(&val[2..]);
+
+            return Ok(Path {
+                value: Cow::Owned(expanded),
+            });
+        }
```
Byron, line 318: What about `{ Ok(…) } else { Ok(…)  }`, to avoid having to name `return`.

git-config/src/values.rs
```diff
@@ -273,6 +280,47 @@ impl Serialize for Value<'_> {
...
+    }
+}
+
+impl Path<'_> {
+    fn interpolate(val: Cow<'_, [u8]>) -> Result<Path<'_>, ()> {
+        if val.is_empty() {
+            return Err(());
+        }
+
+        if val.starts_with(b"~/") {
```
Byron, line 303: What about adding `const TILDE_PREFIX: &[u8] = b"~/"; `, to allow using its lengths later like so.

```Rust
let (_prefix, val) = val.split_at(TILDE_PREFIX.len());
```

git-config/src/values.rs
```diff
@@ -300,24 +303,38 @@ impl Path<'_> {
...
-            if !home.ends_with("/") {
-                expanded.push(b'/');
-            }
-            expanded.extend(&val[2..]);
+        // TODO handle %(prefix) case
 
-            return Ok(Path {
+        const SLASH: u8 = b'/';
+
+        if val.starts_with(b"~/") {
```
Byron, line 310: There is a lot of error cases and I think it's time for upgrading to `quickerror`. This will also do away with the manual error mapping in many if not all cases.

git-config/src/values.rs
```diff
@@ -1471,6 +1488,23 @@ mod path {
         );
     }
 
+    #[test]
+    fn user_interpolated() {
+        let user = std::env::var("USER").unwrap();
+        let path = format!("~{}/foo/bar", user);
+        let path = path.as_bytes();
```
Byron, line 1495: I wouldn't re-assign for `as_bytes()` here and further down.

git-config/tests/integration_tests/file_integeration_test.rs
```diff
@@ -76,7 +76,17 @@ fn get_value_for_all_provided_values() -> crate::Result {
     assert_eq!(
         file.value::<git_config::values::Path>("core", None, "location")?,
         git_config::values::Path {
-            value: Cow::Borrowed(b"/tmp")
+            value: Cow::Borrowed(
+                format!(
+                    "{}/tmp",
+                    dirs::home_dir()
+                        .ok_or("empty home dir")?
```
Byron, line 83: It's better to `expect("msg")` here instead of substituting values which are then seemingly the expected value.

Byron (commented):
Currently `gitoxide` generally has the problem of having to convert BString/Bytes back to Paths, which goes through a UTF-8 decode step which isn't what git does, thus making it less compatible.

I will take a look and see how this can be solved more generally.

Byron:
Despite of what I said earlier, it's probably in our interest to leave `git-url` out of it for now. It doesn't support (and shouldn't support) `%(prefix)` expansion. Maybe it's possible to reuse some code later, so definitely something to keep an eye out, for later :).

svetli-n:
> Currently `gitoxide` generally has the problem of having to convert BString/Bytes back to Paths, which goes through a UTF-8 decode step which isn't what git does, thus making it less compatible.
> 
> I will take a look and see how this can be solved more generally.

Cool, I see https://github.com/Byron/gitoxide/pull/334

Byron:
> Cool, I see https://github.com/Byron/gitoxide/pull/334

Yes, and the short version is that functions in `git_features::path::` should be used to do transformation between `Path/OsStr` and bytes and vice-versa. It's a fallible operation, and depending on the case one should either bubble it up, or call the corresponding `*_or_panic_on_windows()` to panic instead.

svetli-n:
> > Cool, I see #334
> 
> Yes, and the short version is that functions in `git_features::path::` should be used to do transformation between `Path/OsStr` and bytes and vice-versa. It's a fallible operation, and depending on the case one should either bubble it up, or call the corresponding `*_or_panic_on_windows()` to panic instead.

Ok, I'll refactor to using `git_features::path::` and `quick_error`

How about having something like 
`fn replace<'a>(path: impl Into<Cow<'a, [u8]>>, find: &[u8], replace: &[u8]) -> Cow<'a, [u8]>` in `pub mod convert` to simplify `interpolate`?

Byron:
> How about having something like
fn replace<'a>(path: impl Into<Cow<'a, [u8]>>, find: &[u8], replace: &[u8]) -> Cow<'a, [u8]> in pub mod convert to simplify interpolate?

That's definitely something to do once there is another crate needing that functionality, that is, replacing a substring globally without another string. Right now that doesn't seem to be the case, so I'd hold off and keep that local and specific to `git-config`.

Byron:
Hi @svetli-n, thanks a lot for all the work! I took some time myself to dig in and came up with a few changes that are explained in detail in the respective commits.

Please let me know what you think (except that CI seems to be failing 😅), I will fix that later today. Thank you

Byron:
I had a little bit of an epiphany while continuing to refactor which reduced the complexity of the implementation quite a bit while containing it in its own module. The main realization was that git does what it does with slashes because it assumes slashes internally. Once there is an `std::path::Path` though, there is no need for doing this as it's handled for you - a type that `Path::interpolate(…)` now conveniently returns.

The learning here could be that these separator conversions only need to be used to assure paths are using slashes at rest to not break compatibility with git, while transforming them to std::path::Path right before usage.

Byron:
Whohooo, great catch, thanks a lot!
Merged :).

Issue #331, opened 2022-02-08
git-config towards 1.0


Byron:
An issue to see what's missing to make `git-config` comfortably usable by `git-repository`.

###  Tasks

* [x] ❗️a way to know about the source of a value, i.e. it's possible to know which config file is providing an obtained value.
   - this is important for reading values with security implications, so one would avoid accepting values from configuration files not owned by the current user.
   - [x] a way to restrict queries to a sub-set of the available sources
   - [x] #460
anymore as query entails git-sec.
* [x] #459 
* [x] cascading config files (review/validate)
    * [x] `global` config instantiation (the one without a local repository)
    * [x] know which config files to cascade
* [x] handle implicit keys like `k` and `k=` losslessly during serialization
* [x] #456
    * [x] #457
* [x] correct parsing of `Colors` (e.g. hex input, attributes, inverted attributes, order independence) - _validate_
* [x] use `bstr` internally instead of `[u8]`
* ~~[`flock` config files](https://github.com/git/git/blob/master/config.c#L1954:L1954) before reading them~~
    - a linux-specific thread-safety mechanism, with nothing to do with `flock`.
* advanced features
    * [x] #327
    * [x] parse quantities when dealing with integer [as described here](https://git-scm.com/docs/git-config#Documentation/git-config.txt---typelttypegt)
    * [x] include (might require [real-path support](https://github.com/git/git/blob/main/path.c#L731:L731) in path interpolation) (#345)
    * [x] includeIf (don't forget `from_env()` should pick it up too).
        * [x] Assure `incldudeIf` dotted-key notation works for `GIT_CONFIG_KEY` as well, [see TODO and related comment](https://github.com/Byron/gitoxide/pull/432#discussion_r908831861).
    * [x] assure our multi-value implementation allows for [last-one wins semantics](https://github.com/git/git/blob/30cc8d0f147546d4dd77bf497f4dec51e7265bd8/config.c#L2394). They seem to change the order of values.
    * [x] skip [BOM](https://github.com/git/git/blob/master/config.c#L1005:L1005), `unicode-bom` will do the job well.
* [x] comfortable API for accessing values
* [x] validate sub-sections are [case-sensitive](https://github.com/git/git/blob/main/config.c#L44:L44) when reading (and writing)
     - This applies only to legacy sub-section syntax and we don't implement that, comparing case-insensitive instead. Listed in known deviations.
* [x] remove ResolvedConfig entirely - `GitConfig` can do it all
[...]

PR #345, opened 2022-03-05
Add git-config include.

svetli-n:
#331


svetli-n:
@Byron plz take a look when taking a break from multithreading :-)

Byron:
Thanks for taking a shot at this.

There seems to be an issue with the way commits/pushes are done - this is how it looks to me:

Screen Shot 2022-03-06 at 8 16 05 AM

git-config/src/file/git_config.rs
```diff
@@ -1673,6 +1704,125 @@ mod from_paths {
...
+
+        let a_path = dir.path().join("a");
+        fs::write(
+            a_path.as_path(),
+            r"
+            [core]
+              a = false
+            ",
+        )
+        .expect("Unable to write config file");
```
Byron, line 1719: In tests, just unwrapping is fine in 'step one: getting it to work'. In 'step two: making it pretty' one can replace them with `?`.

git-config/src/file/git_config.rs
```diff
@@ -1673,6 +1704,125 @@ mod from_paths {
         assert_eq!(config.len(), 1);
     }
 
+    #[test]
```
Byron, line 1707: While tests are failing and that is expected, when pushing you can `#[ignore]` them.

That way CI stays green while you are working on it.

git-config/src/file/git_config.rs
```diff
@@ -1673,6 +1704,125 @@ mod from_paths {
...
+        fs::write(
+            c_path.as_path(),
+            format!(
+                r"
+            [core]
+              c = 12
+            [include]
+              path = {}
+            ",
+                b_path.as_path().to_str().unwrap()
```
Byron, line 1806: While absolute paths are definitely great to have, I would also use relative paths from the start to be sure that isn't missed.

Please ignore if that's going to be in a separate test, but to me it seems these can be mixed quite easily for a more realistic (but more complex) test.

Byron (commented):
I took a look mostly at the tests this time, with a few general comments.

git-config/src/file/git_config.rs
```diff
@@ -147,18 +148,57 @@ impl<'event> GitConfig<'event> {
...
+            if depth >= MAX_DEPTH {
+                return Ok(());
+            }
+            for path in paths.iter() {
+                let other = GitConfig::open(path)?;
+                let mut section_ids: Vec<_> = other.section_headers.keys().collect();
+                section_ids.sort();
+                for section_id in section_ids {
+                    let section_header = &other.section_headers[&section_id];
+                    if section_header.name.0 == "include" {
```
svetli-n, line 173: could not use `multi_value` since it breaks section ordering

git-config/src/file/git_config.rs
```diff
@@ -147,18 +148,57 @@ impl<'event> GitConfig<'event> {
...
+                        let path_values = other.sections[&section_id].values(&Key::from("path"));
+                        let mut paths = vec![];
+                        for path_value in path_values {
+                            let interpolated_path = crate::values::Path::from(path_value).interpolate(None).unwrap();
+                            let path = if interpolated_path.is_relative() {
```
svetli-n, line 178: could nto use `canonicalize()` since the path needs to be relative to the current `path`

git-config/Cargo.toml
```diff
@@ -25,7 +25,7 @@ quick-error = "2.0.0"
 [dev-dependencies]
 serial_test = "0.5.1"
 serde_derive = "1.0"
-criterion = "0.3"
+#criterion = "0.3.5"
```
Byron, line 28: Was this disabled intentionally? I think it's used by the benchmark in the `benches` directory.

git-config/src/file/git_config.rs
Byron: That's good to know, thanks for the note.

Maybe we keep that in mind and ask ourselves if maintaining section ordering is something we would want or need, especially if this affects the [last-one wins](https://github.com/git/git/blob/main/config.c#L2267:L2276) semantics.

git-config/src/file/git_config.rs
Byron: I still think that canonicalized paths should be possible for use exclusively in the `seen` hashset.

git-config/src/file/git_config.rs
```diff
@@ -147,18 +148,54 @@ impl<'event> GitConfig<'event> {
...
+                let other = GitConfig::open(path)?;
+                let mut section_ids: Vec<_> = other.section_headers.keys().collect();
+                section_ids.sort();
+                for section_id in section_ids {
+                    let section_header = &other.section_headers[&section_id];
+                    if section_header.name.0 == "include" {
+                        let path_values = other.sections[&section_id].values(&Key::from("path"));
+                        let mut paths = vec![];
+                        for path_value in path_values {
+                            let interpolated_path = crate::values::Path::from(path_value).interpolate(None).unwrap();
```
Byron, line 174: We'd definitely need a value for `None`, context should be passed in to the method we are in.

git-config/src/file/git_config.rs
```diff
@@ -147,18 +148,54 @@ impl<'event> GitConfig<'event> {
...
+                            let path = if interpolated_path.is_relative() {
```
Byron, line 175: I think making this explicit is absolutely OK, just wanted to mention that `path.join(absolute_path)` does the right thing and produces `absolute_path` automatically. Thus `join()`ing paths is always safe and can be done unconditionally.

Byron (commented):
Thanks a lot! 

I think the biggest change still needed is to pass in context so `interpolate(…)` has something to work with.

git-config/Cargo.toml
Byron: Will fix itself when rebasing against master.

git-config/src/file/git_config.rs
Byron: Will be done.

git-config/src/file/git_config.rs
Byron: Got that wrong, this cannot happen here. Please ignore.

git-config/src/file/git_config.rs
```diff
@@ -1673,6 +1747,191 @@ mod from_paths {
...
+            let prev_path = dir.path().join(&(i - 1).to_string());
+            fs::write(
+                path.as_path(),
+                format!(
+                    "
+                    [core]
+                      i = false 
+                    [include]
+                      path = {}",
+                    prev_path.as_path().to_str().unwrap().replace("\\", "\\\\")
```
Byron, line 1859: Since this conversion is done a lot, I think it would be better to have a little function for it that is used to describe why we do that. I'd think it's just important for windows and it could say so either with a descriptive function name or a short function name and some description for the function itself.

git-config/src/file/git_config.rs
```diff
@@ -1673,6 +1747,191 @@ mod from_paths {
...
+        );
+
+        assert_eq!(
+            config.get_raw_value("core", None, "a"),
+            Ok(Cow::<[u8]>::Borrowed(b"false"))
+        );
+    }
+
+    #[test]
+    fn nested_include_has_max_depth_10() {
```
Byron, line 1832: I love this test exists. It's great :).

git-config/src/file/git_config.rs
```diff
@@ -131,8 +157,10 @@ impl<'event> GitConfig<'event> {
     /// Returns an error if there was an IO error or if the file wasn't a valid
     /// git-config file.
     #[inline]
-    pub fn open<P: AsRef<Path>>(path: P) -> Result<Self, ParserOrIoError<'static>> {
-        parse_from_path(path).map(Self::from)
+    pub fn open<P: AsRef<Path>>(path: P) -> Result<Self, from_paths::Error> {
```
Byron, line 160: I don't thinks `open(…)` should use `from_paths::Error` - it doesn't do any include resolution and to my mind seems unrelated.
Maybe I am missing something.

Byron (commented):
I really like seeing how this PR is developing, and `includeIf` is probably in reach now, too.

git-config/src/file/git_config.rs
svetli-n: that's right, I went too far with it, rolled back!

git-config/src/file/git_config.rs
```diff
@@ -145,28 +172,70 @@ impl<'event> GitConfig<'event> {
...
-            let other = Self::open(path)?;
-            for (section_id, section_header) in other.section_headers {
-                config.push_section(
-                    section_header.name.0.to_owned(),
-                    section_header.subsection_name.to_owned(),
-                    other.sections[&section_id].clone(),
-                );
+        let mut seen = HashSet::new();
+        let paths: Vec<_> = paths.iter().map(|p| p.into()).collect();
+        const MAX_DEPTH: u8 = 10;
```
Byron, line 179: Since this is a plumbing crate, I think it should be the application that decides where this limit is. Could this be a parameter?

If so, this might pose the question of these parameters, all related to resolving includes, should be part of an `Options` struct of sorts.

git-config/src/file/git_config.rs
```diff
@@ -145,28 +172,70 @@ impl<'event> GitConfig<'event> {
     ///
     /// [`git-config`'s documentation]: https://git-scm.com/docs/git-config#Documentation/git-config.txt-FILES
     #[inline]
-    pub fn from_paths(paths: &[&Path]) -> Result<Self, ParserOrIoError<'static>> {
+    pub fn from_paths(paths: &[&Path], git_install_dir: Option<&std::path::Path>) -> Result<Self, from_paths::Error> {
```
Byron, line 175: I think it's always good to 'not lie' about the inputs, so if a `Vec<PathBuf>` is needed, it's better to have people pass exactly that. That way, if the caller happens to have that, it can just pass it on without extra copies being created unnecessarily.
It's just an idea, and maybe it's not the right thing to do here.

git-config/src/file/git_config.rs
```diff
@@ -202,7 +272,7 @@ impl<'event> GitConfig<'event> {
         }
 
         let paths = paths.iter().map(PathBuf::as_path).collect::<Vec<&Path>>();
```
Byron, line 274: Yes, when seeing this I think the input to `from_paths()` should be `Vec<PathBuf>`.

Byron (commented):
Thanks for the fixes!
Some more ideas came up, I think they may help prepare the code for `includeIf` as well.
My plan is to merge this one next time.

git-config/src/file/git_config.rs
svetli-n: indeed

git-config/src/file/git_config.rs
svetli-n: like  `pub struct Options<'a> {` ?

Byron: That looks good to me!

git-config/src/file/git_config.rs
```diff
@@ -1673,6 +1770,190 @@ mod from_paths {
...
+            .unwrap();
+        }
+
+        let config = GitConfig::from_paths(vec![dir.path().join((max_depth - 1).to_string())], &options).unwrap();
+        assert_eq!(
+            config.get_raw_value("core", None, "i"),
+            Ok(Cow::<[u8]>::Borrowed(b"true"))
+        );
+
+        let config = GitConfig::from_paths(vec![dir.path().join(max_depth.to_string())], &options).unwrap();
```
Byron, line 1893: I know you tested it, but when looking at [the code in git](https://github.com/git/git/blob/master/config.c#L183:L183) I noticed that it can indeed abort if the max-depth is exceeded.

Hence I think there should be an error instead of a potentially different value. This seems fair now especially since `max_depth` can be configured.

Byron (commented):
I was about to prepare this PR for merging when I noticed one last thing that I couldn't fix in the little time that remained.

Once it is addressed I'd love to merge. Thanks.

git-config/src/file/git_config.rs
```diff
@@ -142,13 +143,23 @@ pub mod from_paths {
...
-        /// the location where gitoxide is installed
-        pub git_install_dir: Option<&'a std::path::Path>,
-        /// max_depth of nested includes to follow
+        /// the location where gitoxide or git is installed
+        pub git_install_dir: Option<Cow<'a, std::path::Path>>,
+        /// The maximum length of the file include chain built by following nested includes.
         pub max_depth: u8,
     }
+
+    impl<'a> Default for Options<'a> {
```
svetli-n, line 155: Great with Default!

git-config/src/file/git_config.rs
svetli-n: Great you saw it, I never tested this. What I manually tested with git was including non-existing path which is ignored and we handle it. 

I think we had discussion to not have max depth at all, then we decided to add it and I didn't really look at how git is handling it as I should have. 

I'll add MaxDepth quick error and return that instead.

git-config/src/file/git_config.rs
```diff
@@ -187,6 +274,7 @@ impl<'event> GitConfig<'event> {
             // These two are supposed to share the same scope and override
             // rather than append according to git-config(1) documentation.
             if let Ok(xdg_config_home) = env::var("XDG_CONFIG_HOME") {
+                // TODO make sure the path exists, as git does
```
Byron, line 277: Is this something to address in this PR or is it for later? Both is fine, your call.

git-config/src/file/git_config.rs
Byron: Awesome, thanks!

Good it was caught, and thanks for the refresher on non-existing files, already forgot about it.

git-config/src/file/git_config.rs
svetli-n: actually nothing to address, can be NULL: https://github.com/git/git/blob/master/config.c#L1953

Byron: Thanks, sounds reasonable, especially since we also ignore non-existing paths.

Byron:
So close :)! 

When refactoring and looking at the code I noticed that it would by its very nature loose the include sections themselves. Maybe I am wrong about this, but could you add a test to assure that doesn't happen?

I double-checked with `git config --get-all --includes  include.path` and could indeed query all values underneath the include sections themselves.

Thank you.

Byron:
As discussed, two more small items on the list of the ever expanding scope.

- allow to emulate something like '--no-include'
- check if `GIT_CONFIG_VALUES` may contain `include.path` and have them followed, and if so, we'd need that too.

svetli-n:
Ready may be? :-)

Byron:
Thanks a lot for your contribution, it's merged!

Please note also about the comment in 3cbe072b024848c0133b6800dc84e68f58eee621, I think it would be valuable to split tests into integration tests without affecting the API, and factor the unit tests out into their own module files (without being integration tests). Even though this refactoring could be done at any time, I recommend doing it before proceeding with `includeIf` support.

svetli-n:
Awesome & thanks for the valuable guidance while working on it!

Yes, I was unsure about the tests and will act on the comment in 3cbe072b024848c0133b6800dc84e68f58eee621

PR #352, opened 2022-03-12
a way to parallelize with less contention

Byron:


Byron:
Failure - tried shared-memory with chunks and work-stealing, chunks without work-stealing, and individual items without work-stealing.
It was always slower than channels, fastest was chunks without work-stealing, but still about 10% slower for some reason. Chunk-size is typically 1000.
I think it's mainly the file system that is sensitive to the order in which files are written and what not.

PR #372, opened 2022-03-31
derive PartialEq on ResolvedGitConfig

mzr:
I have two separate solutions that generate git config. I would like to compare that config file generated by those two is logically identical.

### DIFF
--- a/git-config/src/file/resolved.rs
+++ b/git-config/src/file/resolved.rs
@@ -19,6 +19,7 @@ enum ResolvedTreeNode<'event> {
 /// not remember comments nor whitespace. Additionally, values are normalized
 /// upon creation, so it's not possible to retrieve the original value.
 #[allow(clippy::module_name_repetitions)]
+#[derive(PartialEq, Eq, Debug)]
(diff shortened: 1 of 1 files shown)


git-config/src/file/resolved.rs
```diff
@@ -19,6 +19,7 @@ enum ResolvedTreeNode<'event> {
 /// not remember comments nor whitespace. Additionally, values are normalized
 /// upon creation, so it's not possible to retrieve the original value.
 #[allow(clippy::module_name_repetitions)]
+#[derive(PartialEq)]
```
Byron, line 22: Is it possible to also derive `Eq`?

Byron (commented):
Thanks a lot!

Please be warned that this crate will change a lot, even though it's save to use the current minor version as breakage is always indicated by semver.

git-config/src/file/resolved.rs
mzr: yup

mzr: I also added Debug

Byron:
Thanks a lot!

PR #374, opened 2022-04-01
git-config: includeIf support

svetli-n:
git includeIf

TODO
- [ ] Both the symlink & realpath versions of paths will be matched outside of $GIT_DIR. E.g. if ~/git is a symlink to /mnt/storage/git, both gitdir:~/git and gitdir:/mnt/storage/git will match.


Byron:
As per our session, this could be done:

- `sections_by_name_with_header()` can be used to get includeIf information
- gitdir and branch (optional) can be passed in options
- there can be a utility method to determine whether to include the path or not, initially without globbing
- lastly, the `glob` crate can help to do the pattern matching

git-config/src/file/git_config.rs
```diff
@@ -295,10 +295,11 @@ impl<'event> GitConfig<'event> {
                 }
             }
             if let Some(branch_name) = options.branch_name {
-                if branch_name.category() == Some(git_ref::Category::LocalBranch) {
-                    let branch_name = branch_name.shorten();
-                    if let Some(condition) = condition.strip_prefix("onbranch:") {
-                        return condition == branch_name;
+                if let Some((category, branch_name)) = branch_name.category_and_short_name() {
```
Byron, line 298: This should rather be: 
```Rust
if let Some((git_ref::Category::LocalBranch, branch_name)) = branch_name.category_and_short_name() {…}
```

That way the new API reduces the nesting level.

git-config/src/file/git_config.rs
```diff
@@ -272,14 +277,37 @@ impl<'event> GitConfig<'event> {
...
+                    }
+                }
+            }
             false
         }
 
         fn resolve_includes_recursive(
             target_config: &mut GitConfig,
-            target_config_path: Option<&Path>,
+            target_config_path: &Option<&Path>,
```
Byron, line 311: Options should never be passed as reference, but should always be owned. That way they are owned where they are used, which doesn't affect their content at all.

git-config/src/file/git_config.rs
```diff
@@ -289,10 +294,25 @@ impl<'event> GitConfig<'event> {
...
+                .unwrap_or_default();
+
+            for (header, body) in target_config.sections_by_name_with_header("includeIf") {
+                if let Some(condition) = &header.subsection_name {
+                    if include_condition_is_true(condition, options) {
+                        let paths: Vec<values::Path> = body
+                            .values(&Key::from("path"))
+                            .iter()
+                            .map(|path| values::Path::from(path.clone()))
+                            .collect();
```
Byron, line 338: No need to collect here - `extend()` below takes an iterator.

git-config/tests/integration_tests/git_config_integration_tests.rs
```diff
@@ -674,6 +674,87 @@ mod from_paths_tests {
...
+        let d_path = dir.path().join("d");
+        let inside_a_x_path = dir.path().join("a").join("x").join(".git");
+
+        fs::write(
+            a_path.as_path(),
+            format!(
+                r#"
+            [core]
+              b = 1
+            [includeIf "gitdir:~/.git"]
```
Byron, line 693: What does `git` do if the `gitdir` is `~/repo` but the passed in `git_dir` for comparison is `~/repo` or `~/repo/` (trailing slash). Does it normalize these paths to some extend? I would expect it, but never checked.

I am asking because this is the kind of 'hidden smartness' that probably can be implemented before globbing.

git-config/tests/integration_tests/git_config_integration_tests.rs
```diff
@@ -674,6 +674,87 @@ mod from_paths_tests {
...
+
+        // absolute path
+        let options = from_paths::Options {
+            git_dir: Some(inside_a_x_path.as_path()),
+            ..Default::default()
+        };
+        let config = GitConfig::from_paths(vec![a_path.clone()], &options).unwrap();
+        assert_eq!(config.get_raw_value("core", None, "b"), Ok(Cow::<[u8]>::Borrowed(b"2")));
+
+        // tilde ~ path
```
Byron, line 739: Instead of using comments, I recommend using 'doc-strings' in assertions themselves, like this:

```Rust
assert_eq!(config.get_raw_value("core", None, "b"), Ok(Cow::<[u8]>::Borrowed(b"3")), "tilde ~ is resolved to the home directory");
```

Byron (commented):
It's feels like this is nearly done already :) - great work!

It also seems that there is a _groove_ to the tests, using include files with marker values that are clearly enumerated.

git-config/src/file/git_config.rs
svetli-n: > Options should never be passed as reference, but should always be owned

Is this a general rule / Do you mean that since its Option of reference there is a cheap copy of it so no point to pass by reference similar to not passing reference to e.g. i32?

git-config/src/file/git_config.rs
svetli-n: ah, ok!

git-config/tests/integration_tests/git_config_integration_tests.rs
svetli-n: GIT_DIR has to be valid or there's error

```
➜  /Users/svetlin  GIT_DIR=~/IdeaProjects/gitoxide/.git git config --get core.filemode

false
➜  /Users/svetlin  GIT_DIR=~/IdeaProjects/gitoxide git config --get core.filemode

➜  /Users/svetlin  GIT_DIR=~/IdeaProjects/gitoxide/ git config --get core.filemode

➜  /Users/svetlin
```

svetli-n: No errors in the opposite case:

with
```
[includeIf "gitdir:~/IdeaProjects/gitoxide"]
        path = ~/tmp/custom-git.cfg
```
or 
```
[includeIf "gitdir:/non-existing"]
        path = ~/tmp/custom-git.cfg
```
filemode is not changed
```
➜  /Users/svetlin  GIT_DIR=~/IdeaProjects/gitoxide/.git git config --get core.filemode

true
```

with
```
[includeIf "gitdir:~/IdeaProjects/gitoxide/"]
        path = ~/tmp/custom-git.cfg
```
filemode is changed
```
➜  /Users/svetlin  GIT_DIR=~/IdeaProjects/gitoxide/.git git config --get core.filemode

false
```

git-config/tests/integration_tests/git_config_integration_tests.rs
svetli-n: great, thanks

svetli-n:
@Byron feels like we need to implement https://github.com/git/git/blob/142430338477d9d1bb25be66267225fb58498d92/wildmatch.c#L55 or other ideas?

git-config/src/file/git_config.rs
Byron: Yes, everything that that is `Copy` should be copied in. For me it's only a rule because the callee can do more if its values are owned. This is particularly true for `Option` which is `Copy` if its contents is copy. And `Some(&_)` are always copy as shared borrows are copy.

git-config/tests/integration_tests/git_config_integration_tests.rs
Byron: Thanks for sharing your results! Changing the `GIT_DIR` environment variable wasn't my intention, I rather wanted to know what happens if the paths aren't normalized like this:

```git-config
# trailing slashes seem to work, right?
[includeIf "gitdir:~/IdeaProjects/gitoxide/"]
        path = ~/tmp/custom-git.cfg

# what about using the .git suffix?
[includeIf "gitdir:~/IdeaProjects/gitoxide/.git"]
        path = ~/tmp/custom-git.cfg

# what happens in bare repositories? Those without a work tree? Then there is only the full `GIT_DIR` as valid directory.
[includeIf "gitdir:~/IdeaProjects/gitoxide.git"]
        path = ~/tmp/custom-git.cfg

# And again, bare with trailing slash.
[includeIf "gitdir:~/IdeaProjects/gitoxide.git/"]
        path = ~/tmp/custom-git.cfg

# Since `gitoxide.git` is bare, the following shouldn't match
[includeIf "gitdir:~/IdeaProjects"]
        path = ~/tmp/custom-git.cfg
```

Since the work tree directory works as well as the git dir, I think the context must know if the repository is bare or not, or better, pass the working tree directory separately to assure we don't do anything forbidden. Just some thoughts born out of ignorance, and I hope it makes some sense, too 😅.

Byron:
I think the `ignore` crate manages to use a regex engine to do all matching, and it also manages to handle some special cases. [Here is the regex engine syntax](https://docs.rs/regex/1.5.5/regex/#syntax) which knows things like `[[:digit:]]` and [here is a way to process a wildcard pattern](https://github.com/BurntSushi/ripgrep/blob/master/crates/ignore/src/gitignore.rs#L499) to match what git seems to need.

For now it's OK to have only a couple of tests for that and lean strongly on the `ignore` crate as I will have to go 'all in' to it anyway for my work on the worktree's git-ignore and git-attribute file handling. Then I will make sure to provide this kind of functionality in a crate.

However, if you feel more like implementing what's in `wildcard.c` as a fallback, that's fine with me too. I'd see it for cases where for some reason the regex based version doesn't work, or one wants a build without regex, as something like `wildcard.rs` would probably be perfectly correct assuming a perfect port.

Byron:
After thinking about this for a moment, I think it's best to not get trapped into wildcard support right now and rather focus on correctness and the big picture. For example, what's more important is to eventually get to resolving the entire config file hieararchy correctly (as much as possible, if there is no glob support) to allow early adopters like `onefetch` use gitoxide's configuration support.

Since I am working on worktree right now which needs glob support as well, I will prioritize globbing and do that next. That way, it should be available by the time `git-config` approaches its 1.0 milestone. 

I hope that's OK, it's something we can definitely brainstorm about during our session.

svetli-n:
> 



> I hope that's OK, it's something we can definitely brainstorm about during our session.

Sounds good.

> Edit: I have factored pattern parsing (and soon matching) out into the `git-glob` crate. At some point this can be used to easily parse patterns and use them for matching.

Yes, I saw the PR, looking forward :-)

git-config/tests/integration_tests/git_config_integration_tests.rs
Byron: Thanks so much! Also interesting that it adds `**` automatically on trailing slash, but it's probably a brute-force way to support `/.git` suffixes. It's most certainly intended so people can say: include this file if it is in this parent directory, leading to every repository in it being able to include the file.

By the way, this could be implemented as a prefix-search as well, so no glob needed in that particular case.

Byron: Oh, and I think we should definitely have tests for that, too. The existing test can probably be extended for this.

git-config/tests/git_config/mod.rs
```diff
@@ -675,6 +676,13 @@ mod from_paths_tests {
         );
     }
 
+    #[ignore]
```
svetli-n, line 679: @Byron fails when `matches`'s  `value: impl Into<&'a BStr>` (so `from_paths::Options.git_dir`) is an absolute path. What am I missing?

svetli-n:
Not sure why cargo check MSRV fails here and not on other PRs

Byron:
I took a look at the CI failure and realized that it never ran on `cargo msrv` due to my mistake. cargo msrv just builds `git-repository` without the `unstable` features, which would pull in additional crates. Will fix this right away.

The other CI failure was related to a windows path not matching, and I'd need advice if special support is required. The glob code didn't see backslashes as path separator, so I assumed this has to happen before matching. Your advice is definitely welcome.

git-config/tests/git_config/mod.rs
Byron: You have found a shortcoming that I will fix one way or another. The issue is that thus far all matches are considered relative to a repository, and there is an optimization from `ignore` I think that treats absolute patterns (starting with a 'slash' as special and removes the leading slash. When matching against absolute values, even if it's the same as the pattern, they won't match anymore. `git` doesn't do that kind of processing and I will see if I can remove it as well.

Byron:
If you merge `main` back you will find the MSRV issue as well as the glob issue fixed. This shows that the test suite is biased towards matching repo-relative paths, so it lost its ability to match patterns patterns normally. Now `matches(…)` behaves like one would expect, and hope it doesn't have anything left to surprise you with. One thing you should know about though: patterns still have their trailing slashes removed, so this is something you'd have to either compensate for, or use `wildmatch()` directly.

svetli-n:
Awesome, thanks!

svetli-n:
@Byron looking forward review of git-path changes / `real_path()` & tests

Byron:
This review was great fun!

First of all, I like how the tests are done, they were easy to follow and do indeed test the most important bits and pieces. I left a few notes in the form of TODOs for tests I'd love to see, if possible.

The biggest changes were structural in nature, as well as turning the recursive implementation which cloned and copied _a lot_ into something that does so only when needed. The thorough tests were paramount in enabling these refactorings. Let me also highlight the importance of test isolation even within one catch all test case.

Besides that, please take a look at the steps I took to get there, even commits with `refactor` subject have an explanatory body.

I hope that helps moving forward!

git-path/src/realpath.rs
```diff
@@ -0,0 +1,84 @@
...
+    MaxSymlinksExceeded { max_symlinks: u8 },
+    #[error(transparent)]
+    ReadLink(#[from] std::io::Error),
+    #[error("Empty is not a valid path")]
+    EmptyPath,
+    #[error("Parent component of {} does not exist", .path)]
+    MissingParent { path: String },
+}
+
+pub(crate) mod function {
```
svetli-n, line 17: Now that it's own module, what is the benefit of nesting it in another module? Or why is it not just `pub fn realpath()`

git-path/src/realpath.rs
```diff
@@ -18,15 +20,13 @@ pub(crate) mod function {
     use std::path::{Path, PathBuf};
 
     /// TODO
-    pub fn realpath(path: &Path, cwd: &Path, max_symlinks: u8) -> Result<PathBuf, Error> {
-        // TODO add test
-
+    pub fn realpath(path: impl AsRef<Path>, cwd: impl AsRef<Path>, max_symlinks: u8) -> Result<PathBuf, Error> {
```
svetli-n, line 23: by doing so we accept more input types, which seems always better or is there a case where `&T` arg is preferable over` impl AsRef<T>` arg ?

git-path/src/realpath.rs
```diff
@@ -46,8 +46,8 @@ pub(crate) mod function {
                     }
                     CurDir => traverse(input_path, num_symlinks, max_symlinks, real_path),
                     ParentDir => {
-                        let parent_path = PathBuf::from(real_path.parent().ok_or(Error::MissingParent {
-                            path: real_path.to_string_lossy().into(),
+                        let parent_path = PathBuf::from(real_path.parent().ok_or_else(|| Error::MissingParent {
```
svetli-n: why is this better?

git-path/tests/realpath/mod.rs
```diff
@@ -32,100 +32,114 @@ fn create_symlink(link: &Path, link_dest: &Path) {
 }
 
 #[test]
-fn real_path_tests() {
+fn assorted() {
     let cwd = tempdir().unwrap();
     let cwd = cwd.path();
+    let symlinks_disabled = 0;
```
svetli-n, line 41: great name 💯

svetli-n:
> I hope that helps moving forward!

Thanks for the nice step-wise refact!

git-path/src/realpath.rs
Byron: Maybe I am not fully understanding the question as it's made on code that I don't think exists like that anymore. 
But just looking at the diff in the main page then the answer is:

- `to_string_lossy()` should always be avoided in favor of the original data that isn't degenerated. The path isn't only for printing, but callers may also use it themselves.
- `ok_or_else()` is something that clippy wants due to the function call.

git-path/src/realpath.rs
Byron: `&T` is preferable when used in `Option<&T>` as `Option<impl AsRef<T>>` makes it hard to just call the method with `None` - the compiler will argue that it doesn't know the type of the option that way, which makes sense. In that case, do `impl Into<Option<T>>` which is powerful as one can then call `f(value)` and `f(None)`.

git-path/src/realpath.rs
Byron: Having `fn realpath()`  in `pub mod realpath` would make its path (also in the docs)` realpath::realpath` even though one might import it later to `crate::realpath()`, which leaves it at two places.

The package private module `function` is just there to keep `fn realpath()` in `mod realpath` but hide it from being exported there. Instead it's being pulled up later to be exported in `crate::realpath()` while it's actually located in `mod realpath`. There, however, it's not available so to the API user it exists only in one place, not in two. Ambiguous paths I avoid and they can only happen with `prelude` modules which are designed to contain typical imports. The latter you find only in `git-repository` I think.

svetli-n:
@Byron its first working version, please review. I will change small things but it's very close to done IMO.

Byron:
I took the liberty of finally adjusting the structure to be closer to how I always envisioned it, and tried to think of tests by looking at the documentation.

Maybe they exist already but I couldn't identify them. Also I didn't look at the actual code yet, but I will once the tests are closer to the documentation to aid my understanding what to look out for.

We can also consider merging this one and improve the tests in a new PR, and it's something I'd vote for if you are OK with it as I'd like to have the `git-path::realpath` functionality.

### Missing tests 

_…or they could be more clear as I compared to the documentation_. Please try to to decouple the tests so it's not just a big one, but rather have multiple potentially with shared code for setup.

- **onbranch**
   - trailing slash automatically adds `/**` to match sub-directories essentially (using path matcher), so `foo/` matches `foo/bar`
   - simple globs like `foo*` matching `foobar` but not `foo/bar` (because of the path matching, possibly check git)
- **gitdir**
   - a single `~` matches the home directory
   - a test validating user wildcards are respected and that path matching is used

Byron:
Actually, scratch that :D.

Let's merge this one as there is breakage that is already fixed in `main` - so it's a sign telling me to merge now and we take over remaining tasks into the next PR for `includIf` fixups :D.

Issue #298, opened 2022-01-20
**vergen**


Byron:
* [x]  information about branch and commit
* [ ] commit name (tag + shorthash something)

https://lib.rs/crates/vergen

nstinus:
Just hit [this](https://github.com/rustyhorde/vergen/issues/363#issue-2459647026). Is there a way to get a describe with match arguments like the git command line?

Byron:
Currently it's not possible to do that. To enable this, one would either have to…

* …add a new variant to [`SelectRef`](https://github.com/GitoxideLabs/gitoxide/blob/af0383254422b70d53f27572c415eea2e4154447/gix/src/commit.rs#L104-L170) which is responsible for selecting the available name/commit-id mappings.
    - This could even be implemented with a filter function so that users can implement whatever filter logic they want, or alternatively, a way to pass the name mappings directly.
* …use [the plumbing](https://github.com/GitoxideLabs/gitoxide/blob/af0383254422b70d53f27572c415eea2e4154447/gix/src/commit.rs#L228-L237) directly.
    - This would allow the client side to implement this today, without having to wait for (or contribute to) upstream.

I hope that helps.

Byron:
Moved to [this Oxidize Discussion](https://github.com/GitoxideLabs/gitoxide/discussions/2765). This issue is being closed because integration tracking now lives in Discussions.

_AI disclosure: This comment was written and posted by OpenAI Codex on behalf of Byron._

PR #375, opened 2022-04-06
run both windows toolchains (#298)

Byron:
Try to reproduce the [vergen issue](https://github.com/Byron/vergen/runs/5845150345?check_suite_focus=true#step:10:506).

```
Err` value: The ref file could not be read in full
Caused by:
    The filename, directory name, or volume label syntax is incorrect. (os error 123)', src\gen.rs:673:76
```


Byron:
Doesn't reproduce, need to add information to the error to see which path it can't handle. Fix it from there.

Issue #376, opened 2022-04-07
git-config parsing bug with empty configs


mzr:
File like
```
[section]
    foo =  
    bar = a
```

parses as 
```
ResolvedGitConfig(
    {
        (
            SectionHeaderName(
                "section",
            ),
            None,
        ): {
            Key(
                "foo",
            ): [
                10,
                32,
                32,
                32,
                32,
                98,
                97,
                114,
                32,
                61,
                32,
                97,
            ],
        },
    },
)
```

This doesn't make sense to me that it's being parsed like this, because git seems to be handling it just fine:
```
➜  git config --get section.bar
a
➜  git config --get section.foo

```

No matter what goes after `foo =` it is interpreted as a value for `foo`. I think can fix that given some code pointers.

mzr:
I took a brief look at the code. I think it happens because continuations don't look whether newline has been escaped or not.

mzr:
O took a glance at the code and I think the logic that handles continuations (multi-line values) doesn't account for empty values.

PR #377, opened 2022-04-07
[bugfix] Do not treat empty values in sections like multi-line values

mzr:
Fixing #376. Added a test for it. Ran `cargo test` - all passed.


Byron:
Thanks again for this contribution!

This makes me think that there is probably even more work to be done on `git-config` than I thought as previously I was trusting the parser.

PR #378, opened 2022-04-09
 Parse quantities when dealing with integer

svetli-n:
Related to https://github.com/Byron/gitoxide/issues/331


Byron:
Thanks a lot, very useful :)!

Note that I made a few modifications before merging:

- bring integration test layout closer to how it's done in all the other gitoxide crates (may cause conflicts with remaining PR)
- make suffix parsing case-insensitive which is how git does it
- follow up on clippy lints

Byron:
Closing as it looks like GitHub didn't notice it's now part of main using a fast-forward merge after a rebase.

PR #381, opened 2022-04-13
feat(git-config): add suppport for android

MrAdityaAlok:
Do not interpolate `~user/` on Android ([Termux](https://termux.org)).
There is no meaning of it. It is single user system.

Signed-off-by: Aditya Alok <dev.aditya.alok@gmail.com>


Byron:
Thanks a lot, looks good to me!

PR #384, opened 2022-04-14
git-testtools - transparent freeze support

Byron:
On windows, there is one test in particular which takes more than 60s to prepare the fixture for. It would be great to speed this up by simply untarring the fixture script's output.

To achieve this, one should be able to easily tar up the result of fixture scripts and relate them to the script version that created them (by MD5). These are then added via git-lfs and picked-up automatically if there is an archive with the correct version.

Note that this has to be done selectively or maybe needs some 'smart' fixups after unpacking due to different default config options on different operating systems.

See https://github.com/Byron/gitoxide/pull/382#issuecomment-1098585307 for where the idea comes from with some code for untarring archives.

### Tasks

* [x] assure windows actually uses caches to cut down on 60s cache time case
* [x] place target triple in ID to allow excluding it
* ~~allow options to explicitly exclude cache creation for some caches that don't work well for some reason (and remove existing ones)~~
* [x] use caches on MacOS as well

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-**/generated-archives/*/*.tar.xz filter=lfs diff=lfs merge=lfs -text
+**/generated-archives/*.tar.xz filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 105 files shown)


Byron:
Outcome: Windows tests run 3 minutes faster, and we have archives of all the states generated by git different git versions (but only if we regenerate regularly). Since linux will always run git we will see if something changes that breaks us.

PR #387, opened 2022-04-15
Parse expiry date

svetli-n:


Byron:
Tasks we came up with during our session:

- move `git-actor::Time` to `git-date` crate
- implement `parse(input: &BStr) -> Time` to the extend necessary, but be sure to have `#[ignore]` tests for all cases that aren't implemented yet

PR #389, opened 2022-04-18
git-repository: `describe` does not return more recent tag, in case of duplicate tags

Byron:
If a commit has multiple tags, `describe` should return the most recent one.

Commands to reproduce:
```
❯ git tag v2 -m "Testing tags v2"
❯ git tag v0 -m "Testing tags v0" "HEAD~1"
❯ git tag v1 -m "Testing tags v1"
❯ git describe
v1
❯ gix repository commit describe
v2
```

### DIFF
--- a/git-object/src/commit/ref_iter.rs
+++ b/git-object/src/commit/ref_iter.rs
@@ -77,7 +77,6 @@ impl<'a> CommitRefIter<'a> {
     }
 
     /// Returns the committer signature if there is no decoding error.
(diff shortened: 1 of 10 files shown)


Byron:
Sorry, @davidkna, I wasn't aware that changing an issue into a PR via the Github API changes ownership as undesired side-effect. I won't be doing this again and instead just link a new PR to the original issue.

davidkna:
@Byron I don't mind. Thank you for fixing this!

PR #393, opened 2022-04-18
More test coverage for Repository::state::InProgress

inferiorhumanorgans:
This also caught a flaw in the `CherryPickSequence` and `RevertSequence` logic.


git-repository/tests/fixtures/make_am_repo.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+set -eu -o pipefail
+
+git init -q
+
+echo 1.main > 1
```
Byron, line 6: The issue I am having with naming files with numbers while redirecting into them is that `… >1` has a different meaning than `> 1`, and it's just a little too close for me to feel comfortable while reading the script.

I hope you can see the point and avoid such names in the future, and I am happy to rename it for you for now as well.

Byron (changes requested):
Thanks a million, it's great to see this fully tested (and of course to see that testing is truly worth it to catch actual issues).

I can do some refactoring on top (see my comment) to make it mergable, but would ask you to try to use `git-lfs` to not checkin actual binary files of fixture archives like is the state now.

It's surprising that git allows to add these in the first place as I would expect it to fail if a configured filter isn't found (it's in `.gitattributes` in the root).

Please let me know in case git-lfs isn't working for you for some reason and I will cleanup the history accordingly. I also realized that the test system definitely needs a switch to ignore archives which is something I'd use to validate the scripts actually work on MacOS and maybe even on Windows.

Byron:
Thanks a lot for your continued contribution, it's much appreciated!

I took the liberty to do the refactors myself as it's faster to do that than to review again, thanks for your understanding.
Since I had to squash the commits to make removing the binary archives from the history easier, I have to close this commit as its content is contained in #395 .

git-repository/tests/fixtures/make_am_repo.sh
inferiorhumanorgans: Ah, the joys of sh.  The only thing I'd add is that for this script I saved the path to the patch in [`${patch_path}`](https://github.com/Byron/gitoxide/blob/e7b2d8f446b41b26b518abf7d1b048605ef2bbe8/git-repository/tests/fixtures/make_am_repo.sh#L14) but the call to [`git am`](https://github.com/Byron/gitoxide/blob/e7b2d8f446b41b26b518abf7d1b048605ef2bbe8/git-repository/tests/fixtures/make_am_repo.sh#L22) references the file directly.

The latter should be `git am "${patch_path}"`.

Byron: Thanks for the hint. It's adjusted in cfaf31fc654472acf1aacacb516b58a3295cffcd.

PR #405, opened 2022-05-05
[git-config] allow underscores in section names and config entry keys

### LINKED ISSUE #404: [git-config][bug] parser fails on underscores
The parser doesn't handle undersores (`_`) well:

```
use git_config::file::GitConfig;
use anyhow::Result;

fn main() -> Result<()> {

        let conf = r#"
            [a_b]
                ab = b
        "#;

        let _: GitConfig = conf.try_into()?;

    Ok(())
}
```

gives:

```
➜  gitconfig git:(master) ✗ cargo run
   Compiling gitconfig v0.1.0 (/home/mzr/sources/gitconfig)
    Finished dev [unoptimized + debuginfo] target(s) in 0.58s
     Running `target/debug/gitconfig`
Error: Got an unexpected token on line 2 while trying to parse a section header: '[a_b]
    ' ... (27 characters omitted)
```

mzr:
Fixes #404

### DIFF
--- a/git-config/src/parser.rs
+++ b/git-config/src/parser.rs
@@ -1101,7 +1101,7 @@ fn section<'a, 'b>(i: &'a [u8], node: &'b mut ParserNode) -> IResult<&'a [u8], (
 fn section_header(i: &[u8]) -> IResult<&[u8], ParsedSectionHeader> {
     let (i, _) = char('[')(i)?;
     // No spaces must be between section name and section start
-    let (i, name) = take_while(|c: u8| c.is_ascii_alphanumeric() || c == b'-' || c == b'.')(i)?;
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot for the initiative!

Trying the same with git yields the same results though, underscores are rejected…

Screen Shot 2022-05-06 at 07 16 45

…with something like this: `fatal: bad config line 35 in file .git/config`.

Thus the parser is handling it exactly according to spec, and there is no incentive to make it more flexible than the parser git uses as `gitoxide` cannot cause incompatibility to `git` in any way.

mzr:
Oh, you're right. I forgot to check that. 😅

Byron:
No worries, I am thankful as the PR also uncovered a missing test which has since been added.
Truth to be told, I was thinking to merge as it seemed strange underscores aren't supported as well.

Issue #413, opened 2022-05-12
RevSpec parsing


Byron:
### Basic Parsing

It would be great to be part of the development in case you want to open a draft PR sooner than later. Parsing would go into the `git-revision` crate and have the following requirements.

- parse plain bytes (or `&BStr`), not `&str`
- call functions to signal which item was parsed, that is it doesn't parse into a data structure

Also see [this discussion](https://github.com/Byron/gitoxide/discussions/412) for more insights.

* [ ] a callback based parser

### Revision Ranges

**Prerequisistes**
* [ ] Git Revision Parsing - See #408 
* [ ] Git Merge Base - See #409 
* [ ] `..` (two-dot) Range Notation
* [ ] `...` (three-dot) Symmetric Difference Notation

Byron:
Moved to [this Feature Proposal Discussion](https://github.com/GitoxideLabs/gitoxide/discussions/2774). This issue is being closed while the proposal is reviewed; it will become an issue again once review is complete.

_AI disclosure: This comment was written and posted by OpenAI Codex on behalf of Byron._

PR #414, opened 2022-05-12
feat: Implement `TryFrom<&str>` for `Prefix`

kalkin:
Currently there is no easy way to create a `struct Prefix` from a hex string. The method `Parser::from_hex()` is NIY.

`Parser::from_hex()` has the wrong API. It currently is defined to return `Self`, while the proper way would be to return `Result<Self, Error>`. I think the cleaner way would be to implement `std:;convert::TryFrom` for `Prefix`.


kalkin:
Handling prefixes is also important for #413

Byron:
Thanks a lot, much appreciated! I particularly like the exhaustive tests.

Please have a look at the commits I put on top as they reveal some patterns that are used throughout all `gitoxide` crates.

PR #415, opened 2022-05-13
Pathspecs

SidneyDouw:
First tiny commit to coordinate and start tracking progress in here.

Some references on how pathspecs work [here](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefpathspecapathspec), [here](https://css-tricks.com/git-pathspecs-and-how-to-use-them/) and [here](https://github.com/git/git/blob/master/pathspec.c).

### Tasks

* [x] parsing
  * [x] error handling
  * [x] attribute parsing
  * [x] handle escaped attribute values correctly
* [x] error handling is not more strict than git (and we have proof)
  * [x] more test cases
  * [x] establish baseline (with git-testtools)

### Out of scope

* pattern matching


git-pathspec/tests/pathspec.rs
```diff
@@ -0,0 +1,11 @@
+#[test]
```
Byron, line 1: It's good to put down some tests you think will be implemented some time!

Ideally these are ignored while they serve as markers to not forget them, without breaking CI (_we never want to break it and fix it swiftly if it happens anyway_).

Byron:
### From previous session

- Could the attribute name use the same 'idea' as used for `FullName` in `git-ref`? This assumes that `git-attributes` parsing validates attribute names.
- what about whitespace in paths (trailing maybe?)

### Session Notes

- test to assure valid portions aren't repeatable (motivate a `break`)
- fix complex clippy signature warning, probably by returning a pattern right away
- fix ignored test
- could the `git-attributes` parsing error be more distinct (if that was fixed in the `git-attributes` crate?)
- repetition tests for attributes (doesn't work) and for things like 'top' (didn't check)

Byron:
### From previous session

- Could the attribute name use the same 'idea' as used for `FullName` in `git-ref`? This assumes that `git-attributes` parsing validates attribute names.
- could the `git-attributes` parsing error be more distinct (if that was fixed in the `git-attributes` crate?)

### Session Notes

- Maybe [this helps](https://github.com/Byron/gitoxide/blob/d80b321cf1863ad4436e69c0ee436f628f72531a/git-attributes/src/parse/attribute.rs#L87) for figuring out how which values to escape.
- consider running a baseline script and parsing the result into a global lookup table, see `rev-parse-delegate` as example
- try to pass `&mut p` to `parse_long_…` for much less boiler plate. Destructure in function signature.

Byron:
### From previous session

- Could the attribute name use the same 'idea' as used for `FullName` in `git-ref`? This assumes that `git-attributes` parsing validates attribute names.
- could the `git-attributes` parsing error be more distinct (if that was fixed in the `git-attributes` crate?)

### Session Notes

- Maybe [this helps](https://github.com/Byron/gitoxide/blob/d80b321cf1863ad4436e69c0ee436f628f72531a/git-attributes/src/parse/attribute.rs#L87) for figuring out how which values to escape.

Byron:
### Session Notes

- rewrite escaping once more with feeling
- Avoid tuple conversions entirely, use the new `Attribute*` and change things where needed.
- it should work if `Iter::new(…, ~~line-number~~)` is working without line number.

Byron:
### Session Notes

- to_owned() for Name, avoid into()
- create fuzz target and run it once

SidneyDouw:
image

fuzzer iterations

Byron:
### Session Notes

- Use `Assignment` instead, and consolidate `Name` , `NameRef` and `Assignment` types. `Name` is a validated name, `NameRef` probably as well for type safety.

```patch
diff --git a/git-attributes/src/lib.rs b/git-attributes/src/lib.rs
index 3b5beb797..9f3812456 100644
--- a/git-attributes/src/lib.rs
+++ b/git-attributes/src/lib.rs
@@ -53,20 +53,20 @@ pub enum State {
     Unspecified,
 }
 
-/// Holds and owns data that represent one validated attribute
+/// Represents a validated attribute name
 #[derive(PartialEq, Eq, Debug, Hash, Ord, PartialOrd, Clone)]
-pub struct Name(BString, State);
+pub struct Name(pub(crate) CompactString);
 
 /// Holds validated attribute data as a reference
 #[derive(PartialEq, Eq, Debug, Hash, Ord, PartialOrd)]
-pub struct NameRef<'a>(&'a BStr, StateRef<'a>);
+pub struct AssignmentRef<'a>(&'a str, StateRef<'a>);
 
 /// Name an attribute and describe it's assigned state.
 #[derive(PartialEq, Eq, Debug, Hash, Ord, PartialOrd, Clone)]
 #[cfg_attr(feature = "serde1", derive(serde::Serialize, serde::Deserialize))]
 pub struct Assignment {
     /// The name of the attribute.
-    pub name: CompactString,
+    pub name: Name,
     /// The state of the attribute.
     pub state: State,
 }
diff --git a/git-attributes/src/match_group.rs b/git-attributes/src/match_group.rs
index 539989b57..59188c1b2 100644
--- a/git-attributes/src/match_group.rs
+++ b/git-attributes/src/match_group.rs
@@ -7,7 +7,7 @@ use std::{
 };
 
 fn attrs_to_assignments<'a>(
-    attrs: impl Iterator<Item = Result<crate::NameRef<'a>, crate::name::Error>>,
+    attrs: impl Iterator<Item = Result<crate::AssignmentRef<'a>, crate::name::Error>>,
 ) -> Result<Vec<Assignment>, crate::name::Error> {
     attrs
         .map(|res| {
diff --git a/git-attributes/src/name.rs b/git-attributes/src/name.rs
index 78d3ad67c..b54d505e5 100644
--- a/git-attributes/src/name.rs
+++ b/git-attributes/src/name.rs
@@ -1,7 +1,7 @@
-use crate::{Name, NameRef, StateRef};
+use crate::{AssignmentRef, Name, StateRef};
 use bstr::{BStr, BString, ByteSlice};
 
-impl<'a> NameRef<'a> {
+impl<'a> AssignmentRef<'a> {
     pub fn name(&self) -> &'a BStr {
         self.0
     }
@@ -15,8 +15,8 @@ impl<'a> NameRef<'a> {
     }
 }
 
-impl<'a> From<NameRef<'a>> for Name {
-    fn from(v: NameRef<'a>) -> Self {
+impl<'a> From<AssignmentRef<'a>> for Name {
+    fn from(v: AssignmentRef<'a>) -> Self {
         Name(v.0.to_owned(), v.1.into())
     }
 }
diff --git a/git-attributes/src/parse/attribute.rs b/git-attributes/src/parse/attribute.rs
index 91a37231d..bcc0f7dfe 100644
--- a/git-attributes/src/parse/attribute.rs
+++ b/git-attributes/src/parse/attribute.rs
@@ -1,4 +1,4 @@
-use crate::{name, NameRef, StateRef};
+use crate::{name, AssignmentRef, StateRef};
 use bstr::{BStr, BString, ByteSlice};
 use std::borrow::Cow;
 
@@ -42,7 +42,7 @@ impl<'a> Iter<'a> {
         Iter { attrs: attrs.fields() }
     }
 
-    fn parse_attr(&self, attr: &'a [u8]) -> Result<NameRef<'a>, name::Error> {
+    fn parse_attr(&self, attr: &'a [u8]) -> Result<AssignmentRef<'a>, name::Error> {
         let mut tokens = attr.splitn(2, |b| *b == b'=');
         let attr = tokens.next().expect("attr itself").as_bstr();
         let possibly_value = tokens.next();
@@ -58,7 +58,7 @@ impl<'a> Iter<'a> {
                     .unwrap_or(StateRef::Set),
             )
         };
-        Ok(NameRef(check_attr(attr)?, state))
+        Ok(AssignmentRef(check_attr(attr)?, state))
     }
 }
 
@@ -78,7 +78,7 @@ fn check_attr(attr: &BStr) -> Result<&BStr, name::Error> {
 }
 
 impl<'a> Iterator for Iter<'a> {
-    type Item = Result<NameRef<'a>, name::Error>;
+    type Item = Result<AssignmentRef<'a>, name::Error>;
 
     fn next(&mut self) -> Option<Self::Item> {
         let attr = self.attrs.next().filter(|a| !a.is_empty())?;
```

Byron (changes requested):
Maybe this is something that could go, as a link to the commit on GitHub, into a README file.

Generally we don't want to reproduce text verbatim as it may cause issues related to license incompatibility.

Thus I'd ask you to replace this commit with another one, potentially adding a stub README.md with a link to useful information.

That said, I think over time `git` migrated away from using `fnmatch` to their own `wildmatch` which has a special `path` mode to not cause `*` to match `/`, thus not matching into sub-directories anymore as described here. That's for the better, I think, as it does away with this unintuitive behaviour.
Also: Amazing how old `git` already is :D.

Lastly, please let me know if I should review it in order to merge, it seems like all outstanding issues have been addressed. Thank you.

Byron:
Fantastic, thanks so much!

I made a few modifications which you might find interesting. Particularly the ones that avoid allocations - now the implementation is entirely allocation free if there is nothing to unescape.

Now I am very much looking forward to actually matching with them, let's see when that can happen.

PR #418, opened 2022-05-14
git-repository: allow reading information about remote branch

davidkna:
I added two helper functions. For a given branch name, they return the remote target reference (this allows getting to the remote branch name) and the name of the associated remote.

`remote_ref` isn't zero-copy, because I couldn't make it return a `Cow<FullnameRef>`.

### DIFF
--- a/cargo-smart-release/src/git/mod.rs
+++ b/cargo-smart-release/src/git/mod.rs
@@ -117,10 +117,10 @@ pub fn author() -> anyhow::Result<git_repository::actor::Signature> {
     .to_owned())
 }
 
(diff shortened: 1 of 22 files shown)


Byron:

It was the perfect reason to finally fix the clutch that was `FullNameRef<'_>` and turn it into a proper `&FullNameRef` to make `Cow` possible.

This should also pave the way for more improvements to the `git-ref` API around the usage of `PartialNameRef` (or _`PartialNameCow`_ how it's called now. Maybe soon it can become a real `Cow<'_, PartialNameRef>`.

As a side-note: I didn't think much about remotes yet so can't promise their handling doesn't change drastically, but we will deal with it once we get there. What's here looks like a good start.

PR #420, opened 2022-05-18
Fix markdown rendering issue

spenserblack:
Just a quick fix, because the rendered code span looked strange to me :slightly_smiling_face:


Byron:
Thanks a lot! And thanks for your first cross-contribution :)!

PR #421, opened 2022-05-19
git-discover: add ceiling_dirs option to upwards discovery

davidkna:
This is useful for instance to avoid going into expensive network paths. I made the type `Cow` to make it easier to add something like `Options::from_env` in the future. I also intend to implement something to keep discovery on the same device in another PR, though if desired I could also add this here.

### DIFF
--- a/git-discover/src/upwards.rs
+++ b/git-discover/src/upwards.rs
@@ -1,4 +1,5 @@
-use std::path::PathBuf;
+use std::borrow::Cow;
+use std::path::{Path, PathBuf};
 
 /// The error returned by [git_discover::discover()][function::discover()].
 #[derive(Debug, thiserror::Error)]
@@ -24,24 +25,30 @@ pub enum Error {
 
 /// Options to help guide the [discovery][function::discover()] of repositories, along with their options
 /// when instantiated.
-pub struct Options {
+pub struct Options<'a> {
     /// When discovering a repository, assure it has at least this trust level or ignore it otherwise.
     ///
     /// This defaults to [`Reduced`][git_sec::Trust::Reduced] as our default settings are geared towards avoiding abuse.
     /// Set it to `Full` to only see repositories that [are owned by the current user][git_sec::Trust::from_path_ownership()].
     pub required_trust: git_sec::Trust,
+    /// When discovering a repository, ignore any repositories that are located in these directories or any of their parents.
+    pub ceiling_dirs: Cow<'a, [Cow<'a, Path>]>,
 }
 
-impl Default for Options {
+impl Default for Options<'_> {
     fn default() -> Self {
         Options {
             required_trust: git_sec::Trust::Reduced,
+            ceiling_dirs: Cow::Borrowed(&[]),
         }
     }
 }
 
 pub(crate) mod function {
-    use std::path::{Path, PathBuf};
+    use std::{
+        borrow::Cow,
+        path::{Path, PathBuf},
+    };
 
     use git_sec::Trust;
 
@@ -55,7 +62,10 @@ pub(crate) mod function {
     // TODO: tests for trust-based discovery
     pub fn discover_opts(
         directory: impl AsRef<Path>,
-        Options { required_trust }: Options,
+        Options {
+            required_trust,
+            ceiling_dirs,
+        }: Options<'_>,
     ) -> Result<(crate::repository::Path, git_sec::Trust), Error> {
         // Canonicalize the path so that `git_discover::parent` _actually_ gives
         // us the parent directory. (`git_discover::parent` just strips off the last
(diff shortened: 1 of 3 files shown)


Byron:
Thanks a lot for tackling this, it's definitely much appreciated!

I simplified the `Cow` away to achieve better ergonomics, it's OK to force those who have a borrowed Path to allocate, and found bunch of code that doesn't seem to be necessary. Maybe a test is missing to justify it, could you take a look?

Regarding the 'do not cross filesystems' option, that would be great to have as well. A separate PR would be preferred so this one doesn't have to stay open for too long.

git-discover/src/upwards.rs
```diff
@@ -147,15 +188,19 @@ pub(crate) mod function {
...
-        match c {
-            Prefix(p) => p.as_os_str().len(),
-            CurDir => 1,
-            ParentDir => 2,
-            Normal(p) => p.len(),
-            RootDir => 1,
-        }
+    /// Find the number of components parenting the `base_path` before the first directory in `ceiling_dirs`.
+    /// `base_path` needs to be an absolute path. Non-absolute `ceiling_dirs` are discarded if `base_path` is absolute.
+    // TODO: Handle this in a verbatim-path-prefix-neutral way on Windows (introduced by `path::canonicalize`).
```
Byron, line 193: Before I forget: could there be a test that triggers this by failure? It could be marked with `should_panic` or it could be ignored to indicate some work needs to be done to make it pass.

git-discover/src/upwards.rs
```diff
@@ -24,18 +26,21 @@ pub enum Error {
...
 /// when instantiated.
-pub struct Options {
+pub struct Options<'a> {
     /// When discovering a repository, assure it has at least this trust level or ignore it otherwise.
     ///
     /// This defaults to [`Reduced`][git_sec::Trust::Reduced] as our default settings are geared towards avoiding abuse.
     /// Set it to `Full` to only see repositories that [are owned by the current user][git_sec::Trust::from_path_ownership()].
     pub required_trust: git_sec::Trust,
+    /// When discovering a repository, ignore any repositories that are located in these directories or any of their parents.
+    pub ceiling_dirs: &'a [PathBuf],
```
davidkna, line 36: The reason I also went with `Cow` for the list is that, a `Options::from_env` would need to use a vector to parse this.

git-discover/src/upwards.rs
```diff
@@ -75,7 +83,29 @@ pub(crate) mod function {
...
 
         let mut cursor = dir.clone().into_owned();
+
+        let max_height = if !ceiling_dirs.is_empty() {
+            // Ceiling directory discovery requires us to canonicalize the path
+            // TODO: this works nonetheless. Is a test missing? Or does 'absolutize()' do the job already?
+            // is_canonicalized = true;
+            // cursor = cursor
+            //     .canonicalize()
+            //     .map_err(|_| Error::InaccessibleDirectory { path: cursor })?;
```
davidkna, line 93: Without this, this was breaking the `upwards::from_dir_with_dot_dot` test (the `is_empty` causes it not to fail now).

Byron: Thinking about it, the prefix check would only be sensible on a canonicalized directory _or_ if all paths are relative to the same root. The latter is something that could also be documented as requirement. Maybe there could be an assertion for this as well.

The reason I am not keen to canonicalize by default is that it alters paths that the user as provided, which might elongate them which makes it uncomfortable to read.

It's a bit hard for me to reason about the roots of paths in this portion of the code unfortunately, but it should always be possible to pin requirements with tests.

git-discover/src/upwards.rs
Byron: To me that seems alright, probably I am missing something. If you plan to submit a PR to integrate ceilings with `git-repository` we can tackle issues then.

git-discover/src/upwards.rs
davidkna: > Thinking about it, the prefix check would only be sensible on a canonicalized directory or if all paths are relative to the same root. The latter is something that could also be documented as requirement. Maybe there could be an assertion for this as well.

The `GIT_CEILING_DIRECTORIES` env variable for this requires absolute paths, and I think non-absolute paths would be too brittle to be useful in practice. Or do you mean same root as in something like drives on Windows?

> The reason I am not keen to canonicalize by default is that it alters paths that the user as provided, which might elongate them which makes it uncomfortable to read.

Would not setting the cursor be acceptable? Though that would mean calling `canonicalize` twice in some circumstances.

git-discover/src/upwards.rs
davidkna: Done.

Byron:
Thanks a lot! I thought about this a little more and believe have found a way to deal with possibly non-matching ceiling dirs. It's now an error by default to pass ceilings of which none matches the directory to search. This makes sure the user knows the relative root of the paths it passes. It will, however, absolutize the ceilings as this is the one transformation we apply unconditionally to the search directory as well.

All tests have been renamed and the special windows test is now unnecessary.

git-discover/src/upwards.rs
Byron: > The GIT_CEILING_DIRECTORIES env variable for this requires absolute paths, and I think non-absolute paths would be too brittle to be useful in practice. Or do you mean same root as in something like drives on Windows?

Now that I was about to write about this I realize that what's merged by me still has shortcomings, so I am going to fix this on main.
The current implementation kicks the can down the street and tells the caller to assure paths are normalized according to their requirements.

Issue #424, opened 2022-05-21
[git-sec] adopt git-for-windows quirks


davidkna:
### Summary 💡

Git-for-Windows has some exceptions for the ownership checks. The home directory is accepted unconditionally (which is owned by admin instead of the user), and admin-group-owned folders are accepted if the user is an admin (helpful in environments like Github-Actions).

Git-for-Windows references:
https://github.com/git-for-windows/git/blob/0cd1496fc3dd78dadbdf3644ec4d384188573e6f/compat/mingw.c#L3503-L3515

https://github.com/git-for-windows/git/blob/0cd1496fc3dd78dadbdf3644ec4d384188573e6f/compat/mingw.c#L3535-L3543

If this is something you would be happy to adopt, I would be happy to make a contribution

### Motivation 🔦

* Allow opening repos in home directory on Windows
* Allows the check to work in GHA without extra workarounds

Byron:
> If this is something you would be happy to adopt, I would be happy to make a contribution

Yes, `gitoxide` would love to act similarly on Windows.

#### _For context_

When I implemented this, I intentionally left out the special case related to the admin home as `gitoxide` implements much finer grained control over what's permitted in a repository. By default, even with reduced Trust, typical repositories are readable and even writable without hickups. `gitoxide` will implement additional safe-guards by associating trust with configuration values, which is how it can say that 'we want executable paths only if we own the file containing it (or we deem it OK for other reasons)'. That way attacker-controlled configuration or files will simply be ignored while falling back to configuration or files which are trusted.

However, as it also implements the strictly optional [full-bailout](https://github.com/Byron/gitoxide/blob/a7ac64cd801b985790b5717be1a5dc722b2ae3a9/git-repository/src/open.rs#L231-L234) like git, it would definitely be beneficial to implement the same logic to determine trust on windows as git does.

davidkna:
closed via https://github.com/Byron/gitoxide/pull/426

PR #426, opened 2022-05-22
git-sec: adopt git-for-windows exception rules

davidkna:
This PR implements the changes proposed in #424:

> Git-for-Windows has some exceptions for the ownership checks. The home directory is accepted unconditionally (which is owned by admin instead of the user), and admin-group-owned folders are accepted if the user is an admin (helpful in environments like Github-Actions).
>
> Git-for-Windows references:
> https://github.com/git-for-windows/git/blob/0cd1496fc3dd78dadbdf3644ec4d384188573e6f/compat/mingw.c#L3503-L3515
> 
> https://github.com/git-for-windows/git/blob/0cd1496fc3dd78dadbdf3644ec4d384188573e6f/compat/mingw.c#L3535-L3543

I couldn't make `CheckTokenMembership` work with the token/handle already being opened, so I instead I made it forward a null value to have the Windows API handle the details. As part of that attempt, I also made the code prefer using a thread token instead of a process token, which seems to be preferred, and added a `CloseHandle` call for the old handle.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1536,6 +1536,7 @@ name = "git-sec"
 version = "0.1.1"
 dependencies = [
  "bitflags",
+ "dirs",
  "libc",
  "serde",
  "tempfile",
--- a/git-discover/tests/upwards/mod.rs
+++ b/git-discover/tests/upwards/mod.rs
@@ -3,18 +3,7 @@ use std::path::PathBuf;
 use git_discover::repository::Kind;
 
 fn expected_trust() -> git_sec::Trust {
-    #[cfg(not(windows))]
-    {
-        git_sec::Trust::Full
-    }
-    #[cfg(windows)]
-    {
-        if is_ci::cached() {
-            git_sec::Trust::Reduced
-        } else {
-            git_sec::Trust::Full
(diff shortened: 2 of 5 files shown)


Byron:
Thanks a million! This looks good to me, but then again, I am not an authority on how to do these things. The group membership check was already implemented at some point _and made by a Microsoft employee_, hence it probably is the authoritative version: https://github.com/Byron/gitoxide/commit/9a3f0ba8277d92eb75129931993bddbd9961ccdd#diff-71eaf9b8d0991458e6650383da1d69a99dff38608f29fa716c03e67b9388c0ffR70 . If you find anything there to remove the TODO, you could do so in another PR.

Thanks for all your help with this, you are definitely pushing `gitoxide` along considerably!

Byron:
Thanks for the great suggestion! I tried it (actually the mount was already there as drive Z, I just didn't see it), but the same issue ensues. I am actually happy about it as it would be strange to have different permissions depending on the way one reaches the very same resource. Maybe there is more to mapping a drive though, after all this one was pre-created. For now it seems OK to keep this test-only environment variable switch.

PR #428, opened 2022-05-24
parse revision specifications

Byron:
'Pure' parsing of rev-specs.

#### Revisions

- [x] long hex and short hex sha
- [x] `git describe` output
- [x] refnames
- [x] @ for `HEAD`

#### `@` with ref-name or branch-name anchor
- [x] `[refname]`@{`date`} syntax (with just initial parsing of dates, see `git-date`
- [x] `refname`@{`n`}`
- [x] `@{`n`}`
- [x] `@{`-n`}`
- [x] `[branchname]`@{upstream|u} and `[branchname]`@{push}

#### any anchor

- [x] `rev`^`[n]` (`<rev>^0` == commit)
- [x] `rev`^{`type`} (type = [commit, tree, tag, blob] with `object` needing it to exist
- [x]  `rev`^{} (deref tag)
- [x] `rev`^{/`text or regex`} (skip actual regex handling which has additional rules)
- [x] `rev`~`[n]`
- [x] `rev`:`path` 
- [x] :/`text or regex`
- [x] :[`n`:]`path` (stage entry lookup)

#### Ranges

- [x] ^`rev` - commit exclusion
- [x] `a`..`b` - two dot range
- [x] `a`...`b` - three dot symmetric difference

### For git-repository

- [disambiguation tests](https://github.com/git/git/blob/master/t/t1512-rev-parse-disambiguation.sh#L1:L1)

### Out of scope

- `gix repo revspec explain` - just say something each time the delegate is called.
- integration with `git-repository`
- ~~[revision ranges](https://git-scm.com/docs/git-rev-parse#_revision_range_summary)~~ - need to parse it as otherwise valid specs can't be parsed.


kalkin:
I see you started an own implementation. That's good, because I'm currently very busy with moving and starting a new job. Sorry for letting you down, hope I can help out at some other point, once everything settled.

Byron:
No worries, you are welcome any time, all the best!

PR #430, opened 2022-05-27
Add discovery opt env-overrides & env discovery helpers

davidkna:
This PR adds a `upwards::Options::load_env_overrides`-function in git-discover to load environment overrides for the `ceiling_dirs` and `cross_fs` options. This required changing the `ceiling_dirs`-type to `Vec<_>`.

Furthermore, in `git-repository` add simple `discover_with_environment_overrides` and `discover_with_environment_overrides_opt`-helpers that both handle `GIT_DIR` and fallbacks to upwards discovery with `load_env_overrides`.

Only `parse_ceiling_dirs` in `load_env_overrides` are tested, because anything more would require either environment mocking or integrations tests, but I've tested the functionality.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1207,6 +1207,7 @@ version = "0.1.3"
 dependencies = [
  "bstr",
  "defer",
+ "git-config",
  "git-hash",
  "git-path",
  "git-ref",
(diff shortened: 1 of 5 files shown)


Byron:
Thanks a lot for these improvements! 

The only major change I had to made was the removal of the `git-config` dependency, which means that the code handling this environment variable had to move into `git-repository`. Since you also added very useful discovery helpers there I presume this is what you would be using so the result is very much the same. If not please let me know and I am sure we can figure out alternatives to let `git-discover` handle this particular option.

Byron:
Here is the review-video: https://youtu.be/g6o-nHyrxsI

PR #431, opened 2022-05-27
git-discover: avoid shortening absolute paths

davidkna:
Avoid calling `shorten_path_with_cwd` if original path was absolute. I think it's reasonable to expect that absolute paths are left untouched.


Byron:
Thanks a lot! I totally agree that it's in the users interest to maintain absolute paths, and I definitely was biased towards short paths when I implemented this thinking of not much else.

Byron:
Thanks a lot! I totally agree that it's in the users interest to maintain absolute paths, and I definitely was biased towards short paths when I implemented this thinking of not much else.

PR #432, opened 2022-05-27
IncludeIf cleanup and fixes

svetli-n:
Continuation of https://github.com/Byron/gitoxide/pull/374


Byron:
### Session Notes

- I will try to refactor the on-branch tests to be closer to the spec as basis for (potentially) further refactoring of the gitdir tests to be closer to spec also.
- In any case, we try to get this ready for merge so work can continue on `git-date`, which will be fundamental to supporting `revspec` parsing.

Byron:
I mainly ended up refactoring the `gitdir_matches` function.

### Things I noticed

- `gitdir/i` is probably not correctly implemented as it doesn't pass the case-insensitive option to the wildcard function

### Tests I am interested in

* **gitdir**
  - `if condition.contains('\\')` doesn't seem to have a test
  -  A test showing what happens with `..`. Maybe it's superfluous though as it's relative either way.

svetli-n:
@Byron thanks! Please take a look again.

Byron:
Thanks a lot, I will! Please note that I forgot to [share the video link](https://youtu.be/BHcmdGnhiLE) yesterday, and should also have mentioned that I didn't get to refactor some of the tests as originally planned.

I will try again today 😁.

Byron:
This time I made it, and I hope you can refactor the existing `gitdir` and `gitdir/i` tests to test one thing at time only, while following the `git-config` specification for these conditions closely. The test-setup I am using can probably be adapted to work for `gitdir`.

I think it's a nice exercise, but I'd understand if you wanted to put it to the test with something new like `git-date`. Just let me know and I can finish refactoring the tests.

Byron:
@svetli-n Since I am making some changes to `git-date` which should be the basis for your work, it would be best if you could let me know if/when you plan to start on `git-date` so I can get it merged back into `main` in time. Please ignore if there are no such plans yet :).

svetli-n:
@Byron I'm planning to start on it later this week, after refactoring the includeIf tests

Byron:
@svetli-n Sounds good! Once I get note here I will be sure the latest version of `git-date` is available to you on `main`.

BlackHoleFox:
Pointed here by @Byron (thanks) and gave it a try with my changes from `#437` applied locally. I know its not complete yet, but hopefully it doesn't hurt to report that my setup isn't yet working.

Setup:
-------
`~/.gitconfig`:
```git-config
[user]
    email = "user1@example.com"
    name = "User1"

[includeIf "gitdir:%(prefix)/V:/other/"]
    path = V:/other/.gitconfig
```

`V:/other/.gitconfig`:
```git-config
# <snip some other stuff...>

[includeIf "gitdir:%(prefix)/V:/other/nested/"]
    path = V:/other/nested/.gitconfig
```

`V:/other/nested/.gitconfig`:
```git-config
[user]
    email = "user2@example.com"
    name = "User2"

[commit]
    gpgsign = false
```

`git config user.name` in home folder:
```shell
git config user.name
User1
```

`git config user.name` somewhere in `V:/other/nested`:
```shell
git config user.name
User2
```

`git-config` usage:
```rust
let config = File::from_env_paths(Default::default()).unwrap();
let current_git_user = config.string("user", None, "name").expect("no user configured");

println!("{current_git_user}");
```

but in both the directories, it returns `User1`. `config.strings()` also only has `User1` in it.

Byron:
> `let config = File::from_env_paths(Default::default()).unwrap();`

The `Default::default()` is the issue, it should be options setting the context of the git repository, that is its git directory and its branch.

The `git-config` crate is low-level and for you will be more pleasant to use as part of `git-repository`, which still lacks proper integration. I am saying this to warn about mismatched expectations, as `git-config` might not be what you would like to use it for, so shelling work out to `git` might be the way to go for now.

BlackHoleFox:
> The Default::default() is the issue, it should be options setting the context of the git repository, that is its git directory and its branch.

Oops, my bad. The docs didn't say if it made a difference resolving them. My initial expectation was it wouldn't since the configuration files are kinda independent from the actual repo. I tried setting it and I got the same results though fwiw.

```
 DEBUG: V:\other\nested\repo\.git\
 DEBUG: Some(["User1"])
 DEBUG: git user: "User1"
 ```

> The git-config crate is low-level and for you will be more pleasant to use as part of git-repository, which still lacks proper integration.

Not sure, tbh. My use case doesn't really care about the repositories it operates in beyond it just being "a repository." All the use case is for is to obtain which Git user is currently selected based off the total config hierarchy. When I saw `git_config` on crates.io, the description seemed to fit the bill. Is that specific expectation wrong?

**Comment by Byron** [comment]
Thanks for sharing your experience!

> EDIT: After fiddling with this more, it seems to just be a discrepancy between Git (for Windows, maybe?) and gitoxide. The part throwing git-config off was %(prefix)

`gitoxide` can resolve and handle `%(prefix)`, for that the `git_install_dir` must be set in the options. If it is not, you should actually see an error. And if it did something that didn't work and we can be sure it's not an invalid/unexpected value, then `gitoxide` should implement those 'special behaviours' as well. Let's not go easy on `gitoxide`, it has to be correct 😁.

Hence it would be nice if we could dig in to derive another test-case from it. We do test path resolution with `%(prefix)`, but apparently not includes.

> My use case doesn't really care about the repositories it operates in beyond it just being "a repository."

I think that use case is perfectly valid and is a good motivation to make the `git-config` initialization easy enough even on plumbing level. This means users of the API will still have to set the `Options` correctly, but from there it should be doing what git does and give the same answers when the config is queried.

**Comment by Byron** [comment]
### Session Notes

* find the reason for requiring `[serial]` in the first place
* find a way to not have to create temp-directories  in the current users home
* `create_symlink()` should be tested and should work. but maybe not?

**Comment by Byron** [comment]
> @Byron setting $HOME to tmp dir with Command.env() can work for git but not for gitoxide which we do not run as command so need to set $HOME with std::evn::set_var(), which definitely makes tests with ~ serial.

Right, now I understand. This is why the tempdirs are in home in the first place.

> Another alternative is to pass it to gitoxide config in Options, similar to git_dir (so to not read it from env when interpolating)

That sounds like a safe way to so this, and also moves responsibility of trusting/vetting the value to the caller. I like that - `git-repository` is the one to deal with this then.

**Comment by Byron** [comment]
It feels like I spent the whole day to get the tests to the point where they wouldn't write into my own `~/.gitconfig` file and allow them to be fully self-contained. A highlight was that git does do a `realpath(home)` for handling `includes` for conditions only, which is what forced you to bypass the symlink in `/var/private` on MacOS using `tempdir_in(home)`.

There is still another day of work in it at least to look at everything else, I will try to do that tomorrow with a clearer head.

**Comment by Byron** [comment]
Lesson learned for windows: Use `git_path::realpath()` instead of `canonicalize()` as the latter has the famous `questionmark` problem on windows, creating paths that git doesn't handle well anymore. It seems you ran into this before as well, and I think there are some tests left that fix it in some way or another.

File: git-config/tests/file/from_paths/includes/conditional/gitdir/mod.rs [hunk 1]
```diff
@@ -0,0 +1,159 @@
...
+        GitEnv::repo_name("worktree")?,
+    )
+}
+
+#[test]
+#[serial_test::serial]
+#[ignore]
+fn dot_slash_from_environment_causes_error() {
+    use git_config::file::from_paths;
+    // TODO: figure out how to do this, how do we parse sub-keys? Can git do that even?
```
svetli-n, line 53: @Byron git can do that:

➜  /Users/svetlin/IdeaProjects/gitoxide (cont_include_if_refact) GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=includeIf.gitdir:/Users/svetlin.path GIT_CONFIG_VALUE_0=/Users/svetlin/tmp/a/b/custom-git.cfg ~/CLionProjects/git/giT config --get core.filemode
includeif.gitdir:/Users/svetlin.path
true
➜  /Users/svetlin/IdeaProjects/gitoxide (cont_include_if_refact) GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=includeIf.gitdir:/Users/svetlin/.path GIT_CONFIG_VALUE_0=/Users/svetlin/tmp/a/b/custom-git.cfg ~/CLionProjects/git/giT config --get core.filemode
includeif.gitdir:/Users/svetlin/.path
false

The additional "includeif.gitdir:/Users/svetlin.path" is debug output of what git parsed for the key

git-config/tests/file/from_paths/includes/conditional/gitdir/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+    assert_section_value(
+        Condition::new("gitdir:WORKTREE/").expect_original_value(),
+        GitEnv::repo_name("worktree")?,
+    )
+}
+
+#[test]
+#[ignore]
+fn pattern_with_backslash() -> crate::Result {
+    // TODO: figure out what this is testing
```
svetli-n, line 109: I think? I saw git failing earlier when the pattern has \ but cant reproduce it now..

git-config/tests/file/from_paths/includes/conditional/gitdir/mod.rs
Byron: Thanks for trying it out and letting me know.

To help with scope, I think it's best to leave this as TODO in there, and for good measure I also added another task [in the tracking issue](https://github.com/Byron/gitoxide/issues/331). Eventually this needs to be implemented for correctness, but I don't think it's very pressing right now.

git-config/tests/file/from_paths/includes/conditional/gitdir/mod.rs
Byron: Thinking about it, maybe this wants to test what happens with escapes in paths, and/or backslashes in paths on unix specifically?

If so, this test can probably be split into its distinct cases (e.g. escaping and backslashes on unix) to see what happens compared to git.

Byron:
Thanks a lot for your continued work on this! I think I know how to continue here and will start processing the remaining TODO's along with the ignored tests from yet another continuation of this PR :).
Hence this PR can now finally be merged 🙂.

PR #437, opened 2022-06-04
Drop non-existent gitconfig paths before parsing

BlackHoleFox:
`git_config::File::from_env_paths` is currently fails too fast iterating git configuration locations, probably in part to its current UNIX-like biases assuming places like `/etc` usually exist (Git for Windows probably tweaks these in their fork). 

This poses a problem on Windows as in a "stock" setup (no manually set env variables in the shell) leads to `/etc/gitconfig` _always_ being added to the list. With that, this function can never work since it tries reading the hardcoded path and instantly fails and bails out without attempting the later items. The same problem exists for `.config/git/config` (I think?) but maybe is less noticeable day to day.

To fix both my active problem and make this code less problematic onwards, a quick filter was added that skips any paths that can't be `stat`ed. With it, it actually finds my user's `.gitconfig` now. 

----

### DIFF
--- a/git-config/src/file/from_env.rs
+++ b/git-config/src/file/from_env.rs
@@ -62,6 +62,12 @@ impl<'a> File<'a> {
             paths.push(PathBuf::from(git_dir).join("config"));
         }
 
+        // To support more platforms/configurations:
+        // Drop any possible config locations which aren't present to avoid
+        // `parser::parse_from_path` failing too early with "not found" before
+        // it reaches a path which _does_ exist.
+        let paths = paths.into_iter().filter(|p| p.exists());
+
         File::from_paths(paths, options)
     }
 


Byron:
Thanks a lot for your contribution, it's much appreciated!

Indeed this method is lacking a lot of tests, I think it has none, which is why it acts like it does and the proposed fix is sufficient (even without a test).

Note that we plan to think all the way through on how to instantiate a valid git configuration, which is when we will add tests and be sure that this won't break again.

Please let me know if you need a release for this, and it shall be done.

BlackHoleFox:
The fast review is appreciated, many thanks :) No rush on the release though, anytime works. After putting this up, I found that I needed `includeIf` support which is still on the roadmap so I'll be using a cobble of `git-config` and shelling out.

Byron:
IncludeIf support is already available and mostly functional [in this PR](https://github.com/Byron/gitoxide/pull/432).

PR #452, opened 2022-07-03
git-config: include-if support

Byron:
This PR is a continuation of the work done in #432 and focusses on asserting correctness in comparison to git itself.

### Tasks

* [x] parser and data structures use bstr internally, instead of `[u8]` or `str`
* [x] fix escape-tests which seem to hint at backslashes being handled incorrectly by the parser and during matching
* [x] dissolve the 'various-gitdir' test into individual tests in `gitdir` and `onbranch`
* [x] add git-baseline to `onbranch`, ideally so that git won't be executed every time
* [x] run ignored tests
* [x] assure all TODOs are gone


Byron:
Rebase onto main, thus merging wasn't recognized.

PR #453, opened 2022-07-04
git-features' walkdir: 2.3.1 -> 2.3.2

mzr:
[`sort_by_file_name`](https://docs.rs/walkdir/latest/walkdir/struct.WalkDir.html#method.sort_by_file_name) is not present in 2.3.1 version of walkdir. "caret" requirement (https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility) which is implicit here, sort of makes it work when the setup of crates is not so complicated, but when 2.3.1 is choosen the build fails.

----

- [ ] Record review and upload on YouTube publicly


Byron:

Issue #293, opened 2022-01-08
git-index reading


Byron:
Only interpret and make accessible all values and figure out how to bridge different formats in a single implementation ideally.

### Tasks

* V3 reading
* Extension
* V4 reading

Byron:
Moved to [this Tasks Discussion](https://github.com/GitoxideLabs/gitoxide/discussions/2799). This issue is being closed because task tracking now lives in Discussions.

_AI disclosure: This comment was written and posted by OpenAI Codex on behalf of Byron._

PR #473, opened 2022-08-03
Write V2 Index Files (with TREE and EOIE extensions)

SidneyDouw:
# Write V2 Index Files

- Link to git [Documentation](https://git-scm.com/docs/index-format)
- [git implementation](https://github.com/git/git/blob/master/read-cache.c#L2873)

## Features

**Version Support**
- [x] V2

**Extensions**
- [x] tree cache
- [x] end of index entry

## Out of Scope
- V3
- V4
- resolve undo
- split index / link
- untracked cache
- fs monitor
- index entry offset table

### DIFF
--- a/git-index/src/entry.rs
+++ b/git-index/src/entry.rs
@@ -1,3 +1,4 @@
+use crate::Version;
 use bitflags::bitflags;
 
 /// The stage of an entry, one of 0 = base, 1 = ours, 2 = theirs
@@ -42,6 +43,13 @@ pub(crate) mod at_rest {
         }
     }
 
+    impl Flags {
+        pub fn to_memory(self) -> super::Flags {
+            super::Flags::from_bits((self & (Flags::PATH_LEN | Flags::STAGE_MASK | Flags::ASSUME_VALID)).bits as u32)
+                .expect("PATHLEN is part of memory representation")
+        }
+    }
+
     bitflags! {
         /// Extended flags - add flags for serialization here and offset them down to u16.
         pub struct FlagsExtended: u16 {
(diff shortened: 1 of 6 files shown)


Byron:
### Session Notes

- `File::write_to()` to write with trailing hash
- outscope everything that you didn't start yet
- refactor tests roundtrip 
- pass `std::io::Write` to `write_to` as first argument, see other `write_to()` methods for reference
- write docs about `Options::hash_kind` - it's kind of intricate
- could there be baseline tests against all existing fixtures somewhat like this?
- definitely get rid of the index-table, you could try to wrap `std::io::Write` [like so](https://github.com/Byron/gitoxide/blob/234cd10ca55482ce1a840ce3244308d249895bcc/git-features/src/hash.rs#L151-L172) to track bytes written for example.

```patch
diff --git a/git-index/tests/index/file/write.rs b/git-index/tests/index/file/write.rs
index aa4342816..49d2129a1 100644
--- a/git-index/tests/index/file/write.rs
+++ b/git-index/tests/index/file/write.rs
@@ -1,5 +1,17 @@
 use git_index::write::Options;
 
+#[test]
+fn roundtrips() {
+    for version in [Version::V2] {
+        for fixture_path in ["all fixtures"] {
+            let mut buf = Vec::<u8>::new();
+            let actual = git_index::File::at(fixture_path)?;
+            actual.state.write_to(&mut buf, Options { version })?;
+            let foo = git_index::State::from_bytes(&buf)?;
+        }
+    }
+}
+
 #[test]
 fn v2_empty() {
     let path = crate::fixture_index_path("V2_empty");
```

Byron:
### Session Notes

- `File::write_to()` to write with trailing hash
- could there be baseline tests against all existing fixtures somewhat like this?

Byron:
Thanks a lot!

I took a couple of notes [in the tracking issue](https://github.com/Byron/gitoxide/issues/293) and plan to look at it a little more thoroughly to generalize the end-of-index extension a little more.

The main refactorings done here should help to…

- …put extension write implementations to the extension itself
- …add documentation to new bits and pieces as `allow(missing_docs)` was now removed. Doing that helped to recap what's actually present in this crate as I haven't touched it in a while.

What's currently in `main` should be a good basis for the read-tree work.

PR #476, opened 2022-08-07
Format `git-date::Time` with `time::format_description`.

svetli-n:
----


svetli-n:
@Byron WDYT?

Byron:
Thanks a lot, it's a great start!

Here are some suggestions for the next steps:

- add a method to format a string, like `%H%M` for supporting `time` format strings, and have a test or two for that
- use this format method to simplify the format used when disambiguating commits, the current ISO format is a bit noisy

After that I think we can talk about the nitty and the gritty around formatting :D.

svetli-n:
> * add a method to format a string, like `%H%M` for supporting `time` format strings, and have a test or two for that

Don't quite understand.. Time format strings look like `[year]-[month]-[day] [hour]:[minute]:[second] [offset_hour sign:mandatory]:[offset_minute]:[offset_second]`. 

Do you mean to have a method like `pub fn format(&self, format_str: &str) -> String {` and inside it we convert format_str like `%H%M` to `time` format string like `[hour]:[minute]` to create `impl Formattable` and use that instead of `Iso8601::DEFAULT`. 
Then we call it from e.g. `CandidateInfo::Commit` like `date.format("%Y-%m %H:%M")`?

svetli-n:
> * use this format method to simplify the format used when disambiguating commits, the current ISO format is a bit noisy

Shall we do like git, e.g. `2005-04-07` or something else?

Byron:
> > * add a method to format a string, like `%H%M` for supporting `time` format strings, and have a test or two for that
> 
> Don't quite understand.. Time format strings look like `[year]-[month]-[day] [hour]:[minute]:[second] [offset_hour sign:mandatory]:[offset_minute]:[offset_second]`.
> 
> Do you mean to have a method like `pub fn format(&self, format_str: &str) -> String {` and inside it we convert format_str like `%H%M` to `time` format string like `[hour]:[minute]` to create `impl Formattable` and use that instead of `Iso8601::DEFAULT`. Then we call it from e.g. `CandidateInfo::Commit` like `date.format("%Y-%m %H:%M")`?

Sorry for the ambiguity. The` %H%M` was just an example, let's use what ever `time` uses as it's a direct path-through only.

> Then we call it from e.g. CandidateInfo::Commit like date.format("%Y-%m %H:%M")?

I like that, as it's more powerful. And thanks to time it would probably more look like `date.format("[hour]:[minute]")`. I was thinking about this and wondered if it's really worth it to abstract away `time` from the API. I am wondering as it would be nice to be able to pre-parse these format strings so that formatting with them can be done quickly, which is a feature `time` recently introduced. Doing so needs some types from the time crate. Lastly, since we don't abstract away the `time`-specific format strings, it's safe to say that `time` is an open dependency.

If you can agree to using `time` openly, as opposed to what I said previously, let's embrace it and use its types to get best performance. You can also `pub use time` in `lib.rs` to export it for everyone to see and use its types more directly.

> > * use this format method to simplify the format used when disambiguating commits, the current ISO format is a bit noisy
> 
> Shall we do like git, e.g. `2005-04-07` or something else?

That would be a good start, but I encourage experimentation. My feeling is that what `git` does is alright as commit ambiguity is rare enough for these commits to be apart by days at least, so having greater detail than the day is probably not worth it. Maybe one day `gitoxide` can be smart enough to see which of these commits happened on the same day, and if so, display them with time precision even, but that's definitely out of scope for this PR :D.

I hope that helps, and that it's OK that this PR is a bit more explorative as I also don't know all the answers yet. But we are getting there :).

svetli-n:
> I hope that helps, and that it's OK that this PR is a bit more explorative as I also don't know all the answers yet. But we are getting there :).

I heard "explorative", so WDYT abt the last commit? :-)

Byron:
> > I hope that helps, and that it's OK that this PR is a bit more explorative as I also don't know all the answers yet. But we are getting there :).
> 
> I heard "explorative", so WDYT abt the last commit / [713784d](https://github.com/Byron/gitoxide/pull/476/commits/713784d5d9a7b236099c672c66d10deb4183ee17) ? :-)

I am open to exposing `time` types in the API, but strongly believe that the data structure itself must mirror what `git` is using without relying on external crates at all. If this refactoring is solving a particular problem, let's talk about it and see if there is another solution for it.

svetli-n:
> If this refactoring is solving a particular problem, let's talk about it and see if there is another solution for it.

I dint want to expose time outside the crate, but using deref was not good since it made the public api bigger, so I reverted these commits.

svetli-n:
> I am open to exposing `time` types in the API, but strongly believe that the data structure itself must mirror what `git` is using without relying on external crates at all

Can you elaborate (we can probably take it on sync meeting) on how and which types to expose as well as why "data structure itself must mirror what `git` is using without relying on external crates at all"?

Byron:
> > I am open to exposing `time` types in the API, but strongly believe that the data structure itself must mirror what `git` is using without relying on external crates at all
> 
> Can you elaborate (we can probably take it on sync meeting) on how and which types to expose as well as why "data structure itself must mirror what `git` is using without relying on external crates at all"?

Whenever possible, `gitoxide` exposes the data publicly for users to see and edit. This data is exactly what `git` itself is seeing so it represents a stable data model (as git itself is stable and doesn't break backwards compatibility). And that's a property currently present for `git_date::Time` and I think the benefits are worth keeping. Maybe more about that and how to expose the `time` types in the API in our sync meeting - looking forward.

Byron:
### Session Notes

* [ ] provide [standard date formats](https://git-scm.com/docs/git-log#Documentation/git-log.txt---dateltformatgt) as described in the git log docs and export them in `git_date::time::format::<name-like-git>`

### Next steps

* parse well-known/common formats (using `time`)
* custom-parse git-specific relative formats

PR #477, opened 2022-08-09
feat: support core.worktree option

davidkna:
This PR adds support for worktree overrides via `core.worktree` in `open_from_paths`. `core.worktree` appears to be interpreted as a path that is relative to the git dir.

I'm not sure if this is the best way to handle path absolutisation and joining here, and please tell me if this should be tested more heavily.

----


Byron:
Thanks for tackling this! I think the reason I didn't pick it up is that the docs for `core.worktree` are… discouraging :D. For now I noted the obvious deviation and added as many tests as I could think of to gain a better understanding of what git does, and what `gitoxide` should do. With those I feel confident that the feature won't cause trouble.

Something I thought about for a moment was a way to disable allowing such overrides, but then refrained from making the permissions more complex for that as it should be enough to ignore untrusted configuration values.

I am pretty sure that something intricate is still missing [around the handling of the common_dir](https://github.com/git/git/blob/master/setup.c#L661:L661), but also think that this should be orthogonal to plainly overriding the main worktree. If you can think of more tests around linked worktrees, I'd love to see them in the next PR, and hope that you will find the test-utilities for getting worktree information useful should you choose to tackle this.

PR #485, opened 2022-08-17
Show required features for items when compiling on docs.rs

poliorcetics:
- chore: remove default link to cargo doc everywhere
- feat: pass --cfg docsrs when compiling for https://docs.rs
- chore: uniformize deny attributes
- feat: use docsrs feature in code to show what is feature-gated automatically

This is my first PR, I'm interested by gitoxide and the work you have done is fantastic. As I
explored the code base through the docs I wondered what was available with which feature so I added
this to make it easier. To generate the docs with this locally one can do:

`RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --features=max --open`

----

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1302,6 +1302,7 @@ name = "git-mailmap"
 version = "0.3.1"
 dependencies = [
  "bstr",
+ "document-features",
  "git-actor",
  "git-testtools",
  "quick-error",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,13 +10,11 @@ default-run = "gix"
 include = ["src/**/*", "LICENSE-*", "README.md", "CHANGELOG.md"]
 resolver = "2"
 
-
 [[bin]]
 name = "ein"
 path = "src/ein.rs"
 test = false
 doctest = false
-
 
 [[bin]]
 name = "gix"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,6 @@ pretty-cli = [ "gitoxide-core/serde1", "prodash/progress-tree", "prodash/progres
 ## that appears after a short duration.
 prodash-render-line-crossterm = ["prodash-render-line", "prodash/render-line-crossterm", "atty", "crosstermion"]
 
-
 #! ### Convenience Features
 #! These combine common choices of the above features to represent typical builds
 
@@ -181,3 +180,4 @@ exclude = ["cargo-smart-release/tests/fixtures/tri-depth-workspace/a",
 
 [package.metadata.docs.rs]
 features = ["document-features", "max"]
+rustdoc-args = ["--cfg", "docsrs"]
--- a/cargo-smart-release/Cargo.toml
+++ b/cargo-smart-release/Cargo.toml
@@ -10,8 +10,6 @@ categories = ["development-tools::cargo-plugins"]
 keywords = ["cargo"]
 include = ["src/**/*", "README.md", "CHANGELOG.md"]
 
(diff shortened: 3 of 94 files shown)


poliorcetics:
First, here is a picture of the result:

Documentation showing the required features for the module in git-repository

Second:

Almost all the crates use `deny(unsafe_code, rust_2018_idioms)`, should I just move them into a `.cargo/config.toml` ?

I could use the following to ensure every rustc invocation picks them up:

```toml
[target.'cfg(any(test, not(test)))']
rustflags = [
    "-D", "unsafe_code",
    "-D", "rust_2018_idioms",
]
```

Byron (changes requested):
Thanks a lot for the kind words, this contribution is very much appreciated :)! I wasn't aware this was possible and it appears [`docs.rs` itself](https://docs.rs/git-repository/0.21.0/git_repository/) doesn't yet show feature toggles either. Maybe there is a configuration flag for that as well?

> Almost all the crates use deny(unsafe_code, rust_2018_idioms), should I just move them into a .cargo/config.toml ?

I'd rather keep them separate so each crates is self-describing. These flags are also part of a workflow as `missing_docs` for example won't be added until the crate moves up to the 'usable' state.

Also I'd love to see commit 691eb9a4cede7a38f3efdeb5d7b5e29c6b22e95a altered so it keeps `[forbid(unsafe_code)]` as is. Usually this is for `unsafe_code`, which will be highlighted by `cargo geiger` with a green checkmark or a lock, in any case, it will be highlighted more favorably than those with a mere `#[deny(unsafe_code)]`. Dropping it would be OK as well in case it's too much of a chore. This request really is only about `unsafe_code`, and I think it's OK to only `#[deny(…)]` anything else.

Thanks a lot for your help.

poliorcetics:
So https://docs.rs builds all documentation with the most recent nightly of Rust at the time the crate version is uploaded, which means using [nightly features](https://doc.rust-lang.org/rustdoc/advanced-features.html) is okay. That's how some crates, like [`time`](https://docs.rs/time/latest/time/) can show the required features for the modules: the `Cargo.toml` contains [some lines](https://github.com/time-rs/time/blob/d3581310451d26077224c3858297f22ad9f1924d/Cargo.toml#L23-L26) to communicate instructions to `docs.rs`.

The process to make it work is:

1. Pass a `--cfg somefeature` to `rustdoc` when compiling on `docs.rs`
2. Use `cfg_attr(somefeature, ...)` to only activate the nightly features on `docs.rs` (or manually via `RUSTDOCFLAGS="--cfg somefeature" cargo +nightly doc --open`)
3. The features to pass in the `cfg_attr` are `doc_cfg` and `doc_auto_cfg`

> keeps `[forbid(unsafe_code)]` as is

Yes that's on me, I thought a `deny` in a `lib.rs` could never be deactivated but `allow` still work in this context, moved back to `forbid` for `unsafe_code`

Byron:
Thanks a lot for the explanation, I see now that these features work because they are configured on a crate-by-crate basis, and not because they are special or enforced by `docs.rs`, at least not yet.

Besides that, the `#[forbid(unsafe_code)]` directives look good now, and I am looking forward to seeing the labels show up next time the crates are released.

Thanks again!

poliorcetics:
It worked correctly by the way, if you haven't had the time to check :)

https://docs.rs/git-transport/latest/git_transport/ has features displayed

Byron:
🎉 It's indeed everything as great (and helpful) as you claimed it to be and I am super-thankful :)!

poliorcetics:
> Please keep contributing

I would love to, if I found the time 🥲

PR #487, opened 2022-08-19
fix typo in git-repository doc

ultrasaurus:
This is a very tiny PR -- typo in docs

I noticed a typo when looking at docs.  Would not expect it will be useful to record and publish a video, but feel free :)

----

- [X] Feel free to record review and upload on YouTube publicly, if you think it will be useful to anyone
- [ ] Record review and upload on YouTube publicly


Byron:
Hehe, each time a new PR is opened I am exercising the thought of removing this template, especially since I haven't recorded a video in a while. But then I keep it thinking that it reminds me I should do more of these :D.

In any case, thanks for the fix.

PR #491, opened 2022-08-20
git-repository example creating new repo and writing to it

ultrasaurus:
I don't know if you even want this kind of example.  It took me a while to figure out this much, so I thought it might help other folks.  (I'm relatively new to Rust, so also feel free to give me feedback on idiomatic Rust language as well as gitoxide tips)

- makes a new repo
- writes one empty commit

----


ultrasaurus:
p.s. I wanted to write a commit with some actual data in it -- maybe a blob or even a file, but haven't figured out how to create objects from scratch yet.

Byron:
Thanks for sharing! I think `git-repository` should have more examples, if only as an API check for typical tasks with emphasis on usability. After all, all of this would also be tested, but a nice example can go along way and maybe help to get to the '99% right' state.

Maybe you can keep adding commits to this repository and see how that works. Assembling trees in-memory will work, but is cumbersome. Also these won't be visible in the worktree, everything will be marked as 'deleted'. For now, this is something we have to live with as `worktree` support is currently in development and more should be available to the end of this year.

ultrasaurus:
Appreciate the refactor!  thx for quick response 🎉

git-repository/examples/new.rs
```diff
@@ -1,35 +1,33 @@
...
+        author,
+        author,
         "initial commit",
         empty_tree_id,
         git::commit::NO_PARENT_IDS,
     )?;
-    println!("new object id: {:?}", id);
+    println!("new commit id with empty tree: {:?}", id);
+
+    tmp.into_path();
```
ultrasaurus: wondering whether `tmp.into_path();` is serving a purpose here... or unfinished?

Byron: That's a great point! I tried to make that clearer with [e540a76](https://github.com/Byron/gitoxide/pull/491/commits/e540a763a3e08bba669652116cb696ab51ebc21f).

Generally I want to refrain from deleting directories on the users disk assuming they are owned by us, and using temporary directories that are persisted for inspection is one way around that.

I could also imagine to take input from the user to point to an empty directory, but that would generally mean more checks to be performed or more complex code which doesn't really contribute to the intent of the example itself.

ultrasaurus: um... that doesn't make it clear (to me).  From rust docs:
> The [TempDir](https://docs.rs/tempdir/latest/tempdir/struct.TempDir.html) type creates a directory on the file system that is deleted once it goes out of scope.

so... with your change `_persisted` is still a local variable that goes out of scope when main() ends, why isn't the directory deleted?

Byron: Thanks for the feedback! I have just updated it to another version which I actually prefer. It resembles what app developers would do and shows off typical error handling as well. All this is safe to use as well and should overall be much easier to understand.

What do you think?

git-repository/examples/init-non-bare-repo.rs
```diff
@@ -0,0 +1,33 @@
+// cargo run -p git-repository --example new
```
ultrasaurus, line 1: maybe can remove this line -- I always forget the syntax, but might be better in the README.md 

Note: I learned from Rustacean podcast that git-repository was a good starting point, would be helpful if that were near the top of the README (maybe along with some supported use cases, maybe just point to this example folder 😄 )

git-repository/examples/init-non-bare-repo.rs
```diff
@@ -0,0 +1,33 @@
...
+
+fn main() -> anyhow::Result<()> {
+    let work_dir = std::env::args_os()
+        .nth(1)
+        .context("First argument needs to be the directory to initialize the repository in")?;
+    let repo = git::init(work_dir)?;
+
+    println!(
+        "Repo: {:?}",
+        repo.work_dir().expect("non-bare repositories have a work-dir")
```
ultrasaurus, line 14: by the way, I have been using git for about a decade (tho never with an API) and have no idea what a "bare" or "non-bare" repository is.  Maybe consider comment at the top with description of what the example does (and maybe explain that in context)... or a name like `init-repo-and-commit.rs`

git-repository/examples/init-non-bare-repo.rs
```diff
@@ -0,0 +1,33 @@
+// cargo run -p git-repository --example new
+
+use anyhow::Context;
+use git_repository as git;
+
+fn main() -> anyhow::Result<()> {
+    let work_dir = std::env::args_os()
+        .nth(1)
+        .context("First argument needs to be the directory to initialize the repository in")?;
+    let repo = git::init(work_dir)?;
```
ultrasaurus, line 10: I like this!  

FWIW, I was noodling with this approach:
```
    let work_dir = match (env::args().nth(1)) {
        Some(dir_name) => PathBuf::from(dir_name),
        None => {
            let tmp = tempfile::TempDir::new()?;
            tmp.path().join("repo-non-bare")
        }
    };
```
but your approach is more like a what a developer would write.

git-repository/examples/init-non-bare-repo.rs
```diff
@@ -0,0 +1,33 @@
+// cargo run -p git-repository --example new
+
+use anyhow::Context;
+use git_repository as git;
+
+fn main() -> anyhow::Result<()> {
+    let work_dir = std::env::args_os()
```
ultrasaurus, line 7: curious... why `args_os` and not `args`?  does that help with platform-specific path handling?

ultrasaurus (commented):
I like it!  added some detailed comments/questions inline

git-repository/examples/init-non-bare-repo.rs
Byron: The README can use an overhaul for sure, it's full of information, but it's more like a check-list of what's available than a way to learn where to start.

`gitoxide` is for application developers, for users of its CLI, and sooooon, for the server side. Application developers are completely neglected in the README right now and maybe this is something you could change by adding an MVP-like section that would help you get started with `git-repository`. I'd appreciate it for sure!

git-repository/examples/init-non-bare-repo.rs
Byron: That's fair! Please feel free to rename it, I like `init-repo-and-commit.rs` more as well as it's keeping things simple for the common case that most people know. There might be another example one day that deals with a bare repository, and that one could get some hints at what the difference is.

git-repository/examples/init-non-bare-repo.rs
Byron: It's fair to offer defaults though in case it shouldn't be an error. Tempdirs I'd not want to use anymore though, as `gitoxide` should be safe to use for examples as well (as in, "won't destroy your data").

Just being able to run an example while allowing some modification with additional arguments might be a way to go, and then ideally it informs about the possibility to set a value by passing arguments. Maybe that's something to try as well.

The work we are doing here can definitely help to bring all `git-repository` examples to a decent level that application developers would want to match at least.

git-repository/examples/init-non-bare-repo.rs
Byron: Yes, paths aren't necessarily UTF-8 encoded and thus can't be forced into a `String`. `gitoxide` does a lot to deal with paths correctly and not paint itself into a corner by assuming encodings that aren't there, to match the way git does it as a bare minimum and be just as flexible.

Funnily enough `git` will decode windows paths as UTF-8 and last time I checked just assumes that always works probably causing strange behavior. `gitoxide` will panic in these cases.

Byron:
Thanks a lot! I think this one is ready for being merged and hope that there will be other examples resulting from your trials with `gitoxide`. It's very valuable to me to learn how approachable the current API is and to produce a variety of useful examples to help others to get started.

PR #492, opened 2022-08-20
Update CI workflows to use recent/maintained actions

poliorcetics:
Currently the CI workflows use some outdated actions or unmaintained ones. This PR should fix that.
I don't know if there is a way to be warned when an action is outdated.

I'm marking as draft and doing separate pushes for the commit so I can see if I break CI on a 
specific one.

----

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - uses: Swatinem/rust-cache@v1
+    - uses: Swatinem/rust-cache@v2
     - name: Setup dependencies
       run:
         sudo apt-get install tree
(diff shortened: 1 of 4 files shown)


poliorcetics:
`actions/upload-release-asset` is not maintained anymore, but I don't understand if https://github.com/softprops/action-gh-release has the same capabilities  so I didn't make the change

poliorcetics:
https://github.com/EmbarkStudios/cargo-deny-action says in its README you can make a PR if you want to advertise using the action.

And with that, I believe everything is up to date 😄 marking as ready.

Byron:
First of all, thanks a lot for your efforts, they are much appreciated! I am absolutely no fan of editing YAML files so my stance here is to 'not touch' until it breaks and then grumble when it does, trying to get it back to work with the least amount of effort, hacking and all. By now this probably shows and I like that this PR cleans it up quite a bit.

> `actions/upload-release-asset` is not maintained anymore, but I don't understand if https://github.com/softprops/action-gh-release has the same capabilities so I didn't make the change

Yes, I am always happy whenever the release action decides to keep working. It doesn't run often and broke a couple of times in the past.

> https://github.com/EmbarkStudios/cargo-deny-action says in its README you can make a PR if you want to advertise using the action.
> 
> And with that, I believe everything is up to date 😄 marking as ready.

Thanks for the hint! I have [created a PR](https://github.com/EmbarkStudios/cargo-deny-action/pull/55) indeed :).

PR #493, opened 2022-08-22
write blob

ultrasaurus:
expand `init-repo-and-commit.rs` to write another commit a blob

It doesn't write a file to the local filesystem, just the commit.  I honestly didn't know how it was going to work before I wrote the code.  Then I thought this might be the way one might use git as data storage from an application where the user interacts with the app not the filesystem (and, if that's a valid use case, might change so it saves data that is less file-like).

interested in feedback!

----

- [x] Record if you want to

### DIFF
--- a/git-repository/examples/init-repo-and-commit.rs
+++ b/git-repository/examples/init-repo-and-commit.rs
@@ -3,7 +3,6 @@
 
 use anyhow::Context;
 use git::objs::tree;
-use git_odb::Write;
 use git_repository as git;
 
 fn main() -> anyhow::Result<()> {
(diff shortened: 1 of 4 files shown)


ultrasaurus:
I took some time out and learned about bare repositories.  I wonder if this example should be with a bare repository so it makes sense to not update local file system?

Byron:

> I took some time out and learned about bare repositories. I wonder if this example should be with a bare repository so it makes sense to not update local file system?

Indeed, at this point it's a bare repository and all objects are created entirely without the help of a work-tree, and I made the necessary modifications. Maybe the example could be renamed as well but it's not paramount just yet.

I also added a missing methods to avoid having to user lower level methods on `repo.objects` - there is nothing wrong with that, but I think typical workflows should be supported by `Repository` itself.

In time there can be another version of this that uses a worktree, and it will look entirely differently as it's more akin to what a user would do, i.e. write files to disk, add them to the index and committing that with the tree being auto-generated from it.

Byron (approved):
Looks good to me already.

ultrasaurus:
looks great. thanks!

PR #497, opened 2022-08-23
Fix doc comment

svetli-n:
----


Byron:
Thanks a lot, great catch!

PR #498, opened 2022-08-23
git-date parse draft

svetli-n:
### Tasks

* [ ] make `parse(input, Option<currenttime>)` pure, allowing the current time to be passed
* [ ] relative time i.e. `2 seconds ago`
* [ ]  `gix rev parse '@{1979-02-26 18:30:00}' -e` - should use a nicer format when displaying the _parsed_ time


git-date/src/parse.rs
```diff
@@ -1,11 +1,50 @@
...
 
 #[allow(missing_docs)]
 pub fn parse(input: &str) -> Option<Time> {
     // TODO: actual implementation, this is just to not constantly fail
     if input == "1979-02-26 18:30:00" {
         Some(Time::new(42, 1800))
     } else {
-        None
+        return if let Ok(val) = Date::parse(input, SHORT) {
+            let val = val.with_hms(0, 0, 0).expect("date is in range").assume_utc();
```
svetli-n, line 12: not sure if we can `assume_utc`

svetli-n:
Hey @Byron can I get early feedback, thanks!

Byron (commented):
Thanks for the early feedback request, much appreciated!

The structure of the `parse` module can scale to all the sub-parsers you might be writing, nice! 

Something that I'd definitely add to the test-suite is a `git` baseline. One way I found to trigger git's date parsing is ` git -c a.d=1979-02-26 config --expiry-date a.d`, which can be used in a fixture script to produce a baseline [like here](https://github.com/Byron/gitoxide/blob/4f6fa59d20b4e03663f3d7c3819a7d02b79ab982/git-pathspec/tests/fixtures/generate_pathspec_baseline.sh#L7-L6). This file can then be parsed back for quick comparisons [like here](https://github.com/Byron/gitoxide/blob/4b320e773368ac5e8c38dd8a779ef3d6d2d024ec/git-pathspec/tests/pathspec.rs#L34-L48).

With that you can validate that the date parsed is the same, and if not you could also mark it as such so it's clear we don't expect a perfect match. The reason is that I don't trust `git` parses dates correctly at all or handles them correctly, so `time` is probably better there.

However, with the hand-implemented parsers, there I would expect a perfect match against git. Ideally, each input tested in `git-date` is also run against the `git` baseline, and you can setup your test suite to enforce this. Note that it's not viable to call `git` each time, the baseline should be cached with a fixture script like in the example.

A lot to digest, I can elaborate if there are questions either here or in the session. Good luck!

git-date/src/parse.rs
```diff
@@ -1,11 +1,50 @@
+use crate::time::format::{RFC2822, SHORT};
 use crate::Time;
+use time::{Date, OffsetDateTime};
 
 #[allow(missing_docs)]
 pub fn parse(input: &str) -> Option<Time> {
```
Byron, line 6: Before I forget, the current time should be passed here. That way tests are deterministic.

git-date/tests/time/parse.rs
```diff
@@ -12,3 +13,43 @@ fn special_time_is_ok_for_now() {
...
+            offset_in_seconds: 28800,
+            sign: Sign::Plus,
+        }),
+        "could not parse with RFC2822 format"
+    );
+}
+
+#[test]
+fn relative() {
+    let two_weeks_ago = git_date::parse("2 weeks ago").expect("valid time");
```
Byron, line 45: These tests will be easier to write once `parse(…)` takes the current time and is 'pure'.

Byron (commented):
(Some more notes related to making `parse` a pure function)

svetli-n:
Regrading https://github.com/Byron/gitoxide/pull/498#discussion_r953249794 I think its local offset.

Byron:
### Session Notes

Ways to dig deeper into how date parsing is used:

- [ ]  `gix rev parse '@{1979-02-26 18:30:00}' -e` - should use a nicer format when displaying the _parsed_ time
- The above without `-e` shows how the date could be used (even though right now it isn't) to search the reflog

Byron (commented):
Thanks for the updates, I added some notes as well.

git-revision/src/spec/parse/function.rs
```diff
@@ -435,7 +435,7 @@ where
                 let time = nav
                     .to_str()
                     .ok()
-                    .and_then(git_date::parse)
+                    .and_then(|v| git_date::parse(v).ok())
```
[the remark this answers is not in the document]
svetli-n: Not sure if I understand, `and_then ` needs to return `Option` and `git_date::parse` returns `Result`

git-date/src/parse.rs
```diff
@@ -55,8 +58,11 @@ pub fn parse(input: &str) -> Result<Time, Error> {
         } else if let Some(val) = parse_raw(input) {
             // Format::Raw
             Ok(val)
-        } else if let Some(val) = relative::parse(input) {
-            Ok(Time::new(val.unix_timestamp() as u32, val.offset().whole_seconds()))
+        } else if let Some(val) = relative::parse(input, now.ok_or(Error::MissingCurrentTime)?) {
```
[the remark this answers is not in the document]
svetli-n: > This would cause unnecessary errors if no time is given but it's not a relative date either.

See https://github.com/Byron/gitoxide/pull/498/commits/1eac4de99de250e770b48054fd5bc6b806418e4d#diff-fe1f4921b662e3ab738c66e74ec7da02aa5ffde9cc94fe9c1a17e90326f54813R74 no error if passing None

Byron:
Thank you! I hope it's OK to handle the questions during our session, I am so swamped right now.

Byron:
I couldn't fix that git-lfs issue, too strange, so I am merging what we have and kindly ask you to move the checkboxes into a new PR :D.

Thank you

svetli-n:
> I couldn't fix that git-lfs issue, too strange, so I am merging what we have and kindly ask you to move the checkboxes into a new PR :D.
> 
> Thank you

Done, created https://github.com/Byron/gitoxide/pull/527 and https://github.com/Byron/gitoxide/pull/528

PR #501, opened 2022-08-24
restrict `sha1` `asm` to supported archs

davidkna:
It only compiles on aarch64, x86 and x86_64 (I also made an upstream PR, but I'm not sure how receptive they will be to that considering they still haven't disabled MSVC). I also switched the cargo files to use the new crate name `sha1`.

----

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1231,7 +1231,7 @@ dependencies = [
  "parking_lot 0.12.1",
  "prodash",
  "quick-error",
- "sha-1",
+ "sha1",
  "sha1_smol",
  "walkdir",
 ]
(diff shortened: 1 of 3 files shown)


Byron:
Thanks a lot! I did have some trouble with this in [`onefetch`](https://github.com/o2sh/onefetch/pull/752) already where it would fail on win-MSVC despite that seemingly being disabled. Thus I hope that this change here actually kicks in.

Do I assume correctly that you validated it? CI here isn't testing ARM32.

The name-change is much appreciated, too, I'd probably have stopped to receive updates because of it.

In any case, I am merging the change now in the hope it helps, please let me know if you need a new release after validating the change is effective. Thanks a lot.

davidkna:
> Do I assume correctly that you validated it? CI here isn't testing ARM32.

I only validated compilation via cross.

PR #502, opened 2022-08-24
Add feature to link to traditional zlib for dynamic linking support

davidkna:
Distro maintainers expressed a desire to use dynamic linking, which is not support for zlibng.

----


Byron:
Thanks a lot, and I am glad it was seemingly easy to implement this feature and that it makes distro maintainers happy :).

PR #511, opened 2022-08-30
Fix git_version_from_bytes to handle trailing newline

joelparkerhenderson:
----

- [x ] I give my permission to record review and upload on YouTube publicly


Byron:
Thanks a lot! Now I see - all the time I was implicitly relying on my git version to return `…2.32.1 (<something>)\n` so the newline would never become part of the last number.

PR #519, opened 2022-08-31
Index from Tree

SidneyDouw:
Create a `git-index::State` from a tree object.

#### TODOs

- [x] traverse tree, sort records and build the entry list
- [x] insert entries at their correct positions to avoid the sorting step
- ~~create a type for state entries that are not written to disk yet (`git-index::entry::Stat`)~~

#### Bonus

* [x] Add `Repository::rev_parse_single()` to make parsing revs as comfortable as in git2

#### Questions
- ~~What extensions should this support for now?~~

#### Next Up

- A method to write an entire index file (including trailing hash)
- `gix free index from-tree` to try this with big repos
    -  Check with `/usr/bin/time -lp` - does the peak mem usage in huge repos look different when doing `path_backing.shrink_to_fit()`? What does that do to runtime?
- TREE extension

#### Out of scope

- `Find` function returns result - there were problems in other places when doing that, fixable, certainly, but not as easy as hoped.


Byron:
Welcome back!

> What extensions should this support for now?

`TREE` would be most valuable as it will help with creating trees from an It _should_ be possible to generate this while creating the index as all data is available, but it's OK to postpone this if it creates any complications.

git-index/src/access.rs
```diff
@@ -1,6 +1,83 @@
...
+        let root = find(tree, &mut buf).expect("couldn't find a tree for given oid");
+        let state = breadthfirst::State::default();
+        let mut delegate = git_traverse::tree::Recorder::default();
+        breadthfirst(root, state, &mut find, &mut delegate)?;
+
+        // TODO: estimate size
+        let mut path_backing: PathStorage = Vec::new();
+
+        // TODO: insert entries directly into the correct position
+        // TODO: double check this way of sorting is actually the right one
```
jeffparsons, line 26: > Index entries are sorted in ascending order on the name field,
interpreted as a string of unsigned bytes (i.e. memcmp() order, no
localization, no special casing of directory separator '/'). Entries
with the same name are sorted by their stage field.

I believe this is how `BStr` sorts, so this should already be correct. However, I don't know if the intent is necessarily for `BStr` explicitly intends to always sort this way, so maybe it's worth either clarifying that upstream or otherwise explicitly dereferencing down to the byte slice and comparing that instead.

git-index/src/access.rs
```diff
@@ -1,6 +1,83 @@
...
+        let mut buf = Vec::new();
+        let root = find(tree, &mut buf).expect("couldn't find a tree for given oid");
+        let state = breadthfirst::State::default();
+        let mut delegate = git_traverse::tree::Recorder::default();
+        breadthfirst(root, state, &mut find, &mut delegate)?;
+
+        // TODO: estimate size
+        let mut path_backing: PathStorage = Vec::new();
+
+        // TODO: insert entries directly into the correct position
```
jeffparsons, line 25: If you did a depth-first traversal you'd naturally end up with entries in index order, wouldn't you? Are you doing depth first for any particular reason? (I'm not familiar with how other implementations do this — please excuse my ignorance if this is something well known. 😅)

SidneyDouw: I mainly used breadth-first traversal as this was the only algorithm available in the git-traverse crate at the moment 😄
But I believe you are right that depth-first would yield entries in the correct order directly, provided the TreeObject's data is already sorted.

Byron: I think I remember now why breadth-first was implemented. It was all about having only two buffers to store object data. Depth first should be no problem using a free-list of buffers for the objects to (re-)use. I still wonder if this 'special sorting' that entry-paths are using is going to spoil the fun, but it's most definitely worth finding out.

jeffparsons: I just remembered a gotcha that I ran into with libgit2 that may inform what you do here:

Even though the Git docs say the index is in memcmp order, in practice I have found that it differs by operating system / filesystem. On macOS with default configuration, for example, I found that the index I got out of libgit2 was ordered with a **case-insensitive** sort, whereas on Linux it was case-sensitive. This resulted in a really sneaky bug in a program I was writing that only showed up on macOS.

I'm afraid I can't remember whether this was just an issue with libgit2 or if the on-disk index created by Git itself was also not in the documented order. Either way, it seems worth being very careful here.

Byron: Thanks for sharing! This sounds like the case-sensitivity configuration of the repository is influencing this. Ultimately we typically look at the `git` source code to gain clarity, and definitely will do here to not potentially replicate `libgit2` bugs 😅.

Byron:
### Session Notes

- use `rev_parse_single()` in all the places where `rev_parse(…).single()` is used and let it return `Result<Id, …>`
- case insensitive paths are not a thing in `tree <-> index` interactions, but have to be handled in `index <-> worktree` interactions
- improve the test-case to run into todos for mode (e.g. symlinks)

Byron:
The merge video: https://youtu.be/DSodo3o01Vo

PR #520, opened 2022-08-31
git-discover: fix is_git check for macOS exFAT

### LINKED ISSUE #514: Regular Repositories can't be opened on macOS exFAT
### Duplicates

### Current behavior 😯

Repositories can't be opened on macOS exFAT Volumes. (starship/starship#4316)

Despite encountering a `.git` (directory) `File::open` succeeds, leading to the code path for opening `.git` files instead of directories. It contains file system data which cannot pass gitfile validation or could be rejected for size reasons.

### Expected behavior 🤔

Repository works on macOS exFAT

### Steps to reproduce 🕹

1. Try to open a git repository on a macOS exFAT Volume via `git_repository::open`

davidkna:
Closes #514

> Repositories can't be opened on macOS exFAT Volumes. ([starship/starship#4316](https://github.com/starship/starship/issues/4316))
> 
> Despite encountering a `.git` (directory) `File::open` succeeds, leading to the code path for opening `.git` files instead of directories. It contains file system data which cannot pass gitfile validation or could be rejected for size reasons.

Regarding the test, I am aware of the other `hdiutil` test (which I originally wrote). But for this I tried to handle the repository creation in a bash script, because as far as I can tell `git_testtools` performs some environment sanitization and should be a bit faster on subsequent runs.

----

### DIFF
--- a/git-discover/src/is.rs
+++ b/git-discover/src/is.rs
@@ -46,64 +46,52 @@ pub fn git(git_dir: impl AsRef<Path>) -> Result<crate::repository::Kind, crate::
         LinkedWorkTreeDir,
         WorkTreeGitDir { work_dir: std::path::PathBuf },
     }
-    #[cfg(not(windows))]
-    fn is_directory(err: &std::io::Error) -> bool {
-        err.raw_os_error() == Some(21)
-    }
-    // TODO: use ::IsDirectory as well when stabilized, but it's permission denied on windows
-    #[cfg(windows)]
-    fn is_directory(err: &std::io::Error) -> bool {
-        err.kind() == std::io::ErrorKind::PermissionDenied
-    }
+    let metadata = git_dir.as_ref().metadata().map_err(crate::is_git::Error::Metadata)?;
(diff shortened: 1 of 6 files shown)


Byron:
Thanks a lot for contributing a fix! The new way also looks so much better than the previous attempt to save IOPs at all cost, and I won't forget that relying on filesystem errors isn't portable unless you know which filesystem you will be running on.

>  But for this I tried to handle the repository creation in a bash script, because as far as I can tell git_testtools performs some environment sanitization and should be a bit faster on subsequent runs.

That's a great improvement, I love it :)! It appears to be twice as fast, clocking in at ~1s compared to the other test that takes 2s. I made a little refactor here 2cd1c007330749f69fc1f154ba1a6e7a0f6e23d0 but that's all I got.

Additionally I took a look at the `Repository::open(…)` code and thought that it probably works as it avoids duplicating work already done in git-discover.

Please let me know if (or when) a new `git-repository` release would be useful and it shall be done.

PR #527, opened 2022-09-05
gix rev parse uses ISO8601 time format.

svetli-n:
----
Tasks:
* [x]  `gix rev parse '@{1979-02-26 18:30:00}' -e` - should use a nicer format when displaying the _parsed_ time


Byron:
Thanks a lot, much appreciated! I made a bit clearer that all dependencies should come through `git` (aka `git-repository`) and removed a few of those extra dependencies that have been dragged in over time but now are superfluous.

PR #529, opened 2022-09-07
feat: add zlib-ng feature to allow linking against system libz-ng

jirutka:
Allow to use zlib-ng (zlib-ng-sys) with native API (no compat mode) that can co-exist with system libz (loaded by e.g. libcurl). This is used in gitoxide package on Alpine Linux.

----


Byron:
Great, thanks a lot for this contribution. Two lines with so much value.

PR #542, opened 2022-09-27
Make Content-Type header check case-insensitive

paulyoung:
Git allows lowercase (for example) header field names but prior to this change, `check_content_type` would return an error saying that the header couldn't be found unless the casing matched.

***

RFC 7230 (HTTP/1.1) says:

> Each header field consists of a case-insensitive field name followed by a colon (":"), optional leading whitespace, the field value, and optional trailing whitespace.

https://www.rfc-editor.org/rfc/rfc7230#section-3.2

***

RFC 7540 (HTTP/2) says:

> Just as in HTTP/1.x, header field names are strings of ASCII
characters that are compared in a case-insensitive fashion.

https://www.rfc-editor.org/rfc/rfc7540#section-8.1.2

----

### DIFF
--- a/git-transport/src/client/blocking_io/http/mod.rs
+++ b/git-transport/src/client/blocking_io/http/mod.rs
@@ -54,12 +54,12 @@ impl Transport<Impl> {
 
 impl<H: Http> Transport<H> {
     fn check_content_type(service: Service, kind: &str, headers: <H as Http>::Headers) -> Result<(), client::Error> {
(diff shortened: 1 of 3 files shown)


paulyoung:
I think the test failure is unrelated to my changes. Maybe some transient CI issue?

Byron:
Thanks a lot for the fix, _and_ the test! That's amazing work. Nothing to add :D.

PR #543, opened 2022-09-27
Introduce http-client feature to decouple from curl

paulyoung:
This is a backwards-compatible change that allows for using implementations of the `Http` trait other than `Curl`.

----


Byron:
That's great work, thanks a lot! I have put a minor refactor on top of it so that the `new` and `connect` methods don't compete anymore (0fd57c6b491a3c8d0127bc1d1f0eb958437edff9), and used it as vehicle to add a changelog message as well.

PR #546, opened 2022-09-28
Don't panic on parsing too short loose object

5225225:
Not sure about the error we return here. This was the easiest to return, but it's not technically true? Open to feedback here.

At least it doesn't panic anymore :)

----


Byron:
Thanks a lot for the fix, much appreciated! Returning with an error is what I would expect here as well.

PR #554, opened 2022-10-08
Make jwalk and rayon optional

pascalkuthe:
Gitoxide uses `jwalk` (which depends on rayon) for a parallel `walkdir` implementation.
`jwalk` is included if the `parallel` feature of `git-features` is enabled.
#481 made using `jwalk` optional while using the `parallel` feature to address problems like starship/starship#4251 by introducing the `fs-walkdir-single-threaded` feature flag.
However, if with this flag `jwalk` is still compiled which causes quite a bit of compile-time overhead, because optional dependencies can only be enabled (but not disabled) by a feature.

This PR replaces the `fs-walkdir-single-threaded` feature flag with the `fs-walkdir-parallel` feature flag.
`jwalk` is only enabled if this feature flag is used and always disabled otherwise.
This allows downstream crates to fully remove the dependency on `jwalk`/`rayon`.

This PR is a breaking change.
To avoid churn for downstream crates I have added the `fs-walkdir-parallel` feature flag as a dependency to the `max-performance` feature flag of `git-repository`

----

### DIFF
--- a/git-features/Cargo.toml
+++ b/git-features/Cargo.toml
@@ -18,16 +18,19 @@ default = []
 ## using facilities of the `prodash` crate.
 progress = ["prodash"]
 
-## If set, and with the `parallel` feature set, walkdir iterators will be single-threaded.
-## This feature exists to avoid [certain side-effects](https://github.com/starship/starship/issues/4251) of rayon threadpool configuration with `jwalk`.
-fs-walkdir-single-threaded = []
+## If set, walkdir iterators will be multi-threaded.
+## This feature has [certain side-effects](https://github.com/starship/starship/issues/4251) of rayon threadpool configuration with `jwalk`.
+fs-walkdir-parallel = ["parallel", "jwalk" ]
 
 ## Use scoped threads and channels to parallelize common workloads on multiple objects. If enabled, it is used everywhere
 ## where it makes sense.
 ## As caches are likely to be used and instantiated per thread, more memory will be used on top of the costs for threads.
 ## The `threading` module will contain thread-safe primitives for shared ownership and mutation, otherwise these will be their single threaded counterparts.
 ## This way, single-threaded applications don't have to pay for threaded primitives.
-parallel = ["crossbeam-utils", "crossbeam-channel", "num_cpus", "jwalk", "parking_lot"]
+parallel = ["crossbeam-utils",
+ "crossbeam-channel",
+ "num_cpus",
+ "parking_lot"]
 #* an in-memory unidirectional pipe using `bytes` as efficient transfer mechanism.
 io-pipe = ["bytes"]
 ## provide a proven and fast `crc32` implementation.
--- a/Makefile
+++ b/Makefile
@@ -94,7 +94,7 @@ check: ## Build all code in suitable configurations
 			   && cargo check
 	cd git-features && cargo check --all-features \
 			   && cargo check --features parallel \
-			   && cargo check --features parallel,fs-walkdir-single-threaded \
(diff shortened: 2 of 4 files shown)


pascalkuthe:
Forgot to commit the updated makefile and already turned off my PC.
I will fix this tomorrow, sorry about that

git-features/Cargo.toml
```diff
@@ -20,14 +20,17 @@ progress = ["prodash"]
...
-fs-walkdir-single-threaded = []
+fs-walkdir-parallel = ["parallel", "jwalk" ]
 
 ## Use scoped threads and channels to parallelize common workloads on multiple objects. If enabled, it is used everywhere
 ## where it makes sense.
 ## As caches are likely to be used and instantiated per thread, more memory will be used on top of the costs for threads.
 ## The `threading` module will contain thread-safe primitives for shared ownership and mutation, otherwise these will be their single threaded counterparts.
 ## This way, single-threaded applications don't have to pay for threaded primitives.
-parallel = ["crossbeam-utils", "crossbeam-channel", "num_cpus", "jwalk", "parking_lot"]
+parallel = ["crossbeam-utils",
```
Byron, line 30: Did you do this by hand or is it a tool? In case of the latter, I'd love to know :D. Currently I have no auto-formatter for toml files and am definitely not willing ever format files again 😁.

git-features/Cargo.toml
```diff
@@ -20,14 +20,17 @@ progress = ["prodash"]
 
 ## If set, and with the `parallel` feature set, walkdir iterators will be single-threaded.
 ## This feature exists to avoid [certain side-effects](https://github.com/starship/starship/issues/4251) of rayon threadpool configuration with `jwalk`.
-fs-walkdir-single-threaded = []
+fs-walkdir-parallel = ["parallel", "jwalk" ]
```
Byron, line 23: This is definitely an improvement and cleaner, as single-threaded directory walking is best done by `walkdir` without the added dependencies.

Byron (commented):
Thanks a lot for your help with this, it's a much appreciated improvement and cleans up a little blunder I committed just for a quick-fix.

I will wait for the makefile adjustment before merging. Please also let me know what you would need released, if this is the only change you needed to make for `helix`.

git-features/Cargo.toml
pascalkuthe: Oh I didn't even notice I have autoformat  turned on for toml files.
The formatting is done by the [tablo lsp](https://github.com/tamasfe/taplo). I use it with helix but it also has a vscode extension and a CLI (to run in CI for example).

It's really nice lsp for toml files because it picks up the json schema so you get autocomoletions for most things (that don't require querying crates.io)

pascalkuthe:
> Please also let me know what you would need released, if this is the only change you needed to make for `helix`.

Honestly I don't really need this for helix right now. The original PR I sent has not been merged yet and it will probably take quite a bit of time until it is merged. The maintainers of helix are rather busy and struggle to keep up with flood of PRs.

For that PR I do not yet need the `parallel` feature yet.
In followup PRs where I implement caching this will be more critical but a new gitoxide release will probably be out long before those get a review :D.
So just put this out whenever you would do a normal gitoxide release anyway.

git-features/Cargo.toml
Byron: Thanks a lot! `cargo install taplo-cli --locked` and `taplo fmt Cargo.toml` should be enough to get started and experiment.

Byron:
Thanks for sharing your plans, makes sense and I am glad there is no rush :).

PR #559, opened 2022-10-14
Fix "http not compiled in" w/ http-client-reqwest

paulyoung:
This fixes a runtime error; `Error: 'http' is not compiled in. Compile with the 'http-client-curl' cargo feature` when using `git-transport` with `features = ["http-client-reqwest"]`.

I wasn't sure how best to go about testing this.

I ran `make tests` locally and saw what appeared to be an unrelated failure.

----


Byron:
Thanks a lot and sorry for the hassle! This is clearly something that should have been part of the original reqwest PR, and a way to detect that would be to build a version of `gix` with `reqwest`. This hasn't been done yet I'd like https://github.com/seanmonstar/reqwest/issues/1646 to be resolved first. Ideally there would be a crate-internal test which shouldn't be too hard to do.

PR #561, opened 2022-10-19
git-transport: reqwest HTTP client configuration

paulyoung:
This introduces some ability to configure HTTP requests when using `http-client-reqwest`.

I think there are some things to discuss so I'm opening this as a draft for now. I'll make comments on the relevant pieces of code.

----

### DIFF
--- a/git-transport/src/client/blocking_io/http/reqwest.rs
+++ b/git-transport/src/client/blocking_io/http/reqwest.rs
@@ -1,3 +1,5 @@
+use crate::client::http::reqwest::remote::Config;
+
 pub struct Remote {
     /// A worker thread which performs the actual request.
     handle: Option<std::thread::JoinHandle<Result<(), reqwest::Error>>>,
@@ -5,9 +7,11 @@ pub struct Remote {
     request: std::sync::mpsc::SyncSender<remote::Request>,
     /// A channel to receive the result of the prior request.
     response: std::sync::mpsc::Receiver<remote::Response>,
+    config: Config,
 }
 
 mod remote {
+    use std::sync::Arc;
     use std::{any::Any, convert::TryFrom, error::Error, io::Write, str::FromStr};
 
     use git_features::io::pipe;
@@ -19,10 +23,17 @@ mod remote {
             let (req_send, req_recv) = std::sync::mpsc::sync_channel(0);
             let (res_send, res_recv) = std::sync::mpsc::sync_channel(0);
             let handle = std::thread::spawn(move || -> Result<(), reqwest::Error> {
-                for Request { url, headers, upload } in req_recv {
+                for Request {
+                    url,
+                    headers,
+                    upload,
+                    config,
+                } in req_recv
+                {
                     // We may error while configuring, which is expected as part of the internal protocol. The error will be
                     // received and the sender of the request might restart us.
-                    let client = reqwest::blocking::ClientBuilder::new()
+                    let client = config
+                        .configure(reqwest::blocking::ClientBuilder::new())
                         .connect_timeout(std::time::Duration::from_secs(20))
                         .build()?;
                     let mut req = if upload { client.post(url) } else { client.get(url) }.headers(headers);
@@ -130,6 +141,7 @@ mod remote {
                     url: url.to_owned(),
                     headers: header_map,
                     upload,
+                    config: self.config.clone(),
                 })
                 .expect("the remote cannot be down at this point");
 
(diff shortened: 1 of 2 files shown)


git-transport/src/client/blocking_io/http/reqwest.rs
```diff
@@ -193,18 +199,27 @@ mod remote {
...
     }
 
     #[derive(Default, Clone)]
     pub struct Config {
-        config: Option<String>,
-        configure: Arc<dyn FnMut(&mut reqwest::ClientBuilder) + Send + Sync + 'static>,
+        pub config: Option<String>,
+        pub configure: Option<
+            Arc<
+                Mutex<
```
paulyoung: I introduced a `Mutex` here to address `Cannot borrow data in an ``Arc`` as mutable`. I'm not sure if there are better or preferred alternatives.

git-transport/src/client/blocking_io/http/reqwest.rs
```diff
@@ -210,16 +210,7 @@ mod remote {
...
-            Arc<
-                Mutex<
-                    dyn FnMut(&mut reqwest::blocking::ClientBuilder) -> reqwest::blocking::ClientBuilder
-                        + Send
-                        + Sync
-                        + 'static,
-                >,
-            >,
-        >,
+        pub configure_request: Option<Arc<Mutex<dyn FnMut(&mut reqwest::blocking::Request) + Send + Sync + 'static>>>,
```
paulyoung, line 213: For my use case I need to configure `Request` instead of `ClientBuilder` so I made changes to only support that.

git-transport/src/client/blocking_io/http/reqwest.rs
```diff
@@ -181,15 +199,25 @@ mod remote {
...
+                self.config = config.clone();
+            }
             Ok(())
         }
     }
 
+    #[derive(Default, Clone)]
+    pub struct Config {
+        pub config: Option<String>,
+        pub configure_request: Option<Arc<Mutex<dyn FnMut(&mut reqwest::blocking::Request) + Send + Sync + 'static>>>,
```
paulyoung, line 213: I tried returning a `Result` from `configure_request` but I couldn't find a suitable error type.

`reqwest::Error` can't be constructed, and `Box<dyn Error + Send + Sync + 'static>` isn't compatible with the rest of the code here.

git-transport/src/client/blocking_io/http/reqwest.rs
```diff
@@ -19,16 +24,27 @@ mod remote {
             let (req_send, req_recv) = std::sync::mpsc::sync_channel(0);
             let (res_send, res_recv) = std::sync::mpsc::sync_channel(0);
             let handle = std::thread::spawn(move || -> Result<(), reqwest::Error> {
```
paulyoung, line 26: One option regarding errors would be to change the error type here to something more general.

git-transport/src/client/blocking_io/http/reqwest.rs
paulyoung: Since this doesn't return a `Result`, it limits the error handling options within the closure provided to `configure_request`.

paulyoung (commented):
This appears to work as intended for the first request I make but something seems to get stuck when calling `configure_request` the second time.

I can only assume this is related to shared ownership and the mutex but I'm not sure what's going wrong there.

git-transport/src/client/blocking_io/http/reqwest.rs
```diff
@@ -19,16 +24,27 @@ mod remote {
...
-                    let mut req = if upload { client.post(url) } else { client.get(url) }.headers(headers);
+                    let mut req_builder = if upload { client.post(url) } else { client.get(url) }.headers(headers);
                     let (post_body_tx, post_body_rx) = pipe::unidirectional(0);
                     if upload {
-                        req = req.body(reqwest::blocking::Body::new(post_body_rx));
+                        req_builder = req_builder.body(reqwest::blocking::Body::new(post_body_rx));
+                    }
+                    let mut req = req_builder.build()?;
+                    if let Some(mutex) = config.configure_request {
+                        let mut configure_request = mutex.lock().unwrap();
```
paulyoung, line 46: What’s the correct thing to do here? The `Mutex` documentation says:

> Most usage of a mutex will simply `unwrap()` these results, propagating panics among threads to ensure that a possibly invalid invariant is not witnessed

paulyoung:
Some additional info. My code looks like this:

```
let reqwest_config = http::Config {
    config: None,
    configure_request: Some(Arc::new(Mutex::new(
        |request: &mut reqwest::blocking::Request| {
            eprintln!("before");
            // code which mutates `request` is here
            eprintln!("after");
        },
    ))),
};
```

The first time a request is attempted, i.e. to list refs, both `before` and `after` are printed. I can see that the request is made and has been mutated as expected.

The second time a request is attempted, i.e. to fetch, `before` is printed and my application gets stuck.

paulyoung:
If I do the following I have no issues. I think I should turn something like this into a test.

```rust
static CUSTOM_HEADER_NAME: &'static str = "custom-header-name";
static CUSTOM_HEADER_VALUE: &'static str = "custom-header-value";
```

```rust
let reqwest_config = http::Config {
    config: None,
    configure_request: Some(Arc::new(Mutex::new(
        |request: &mut reqwest::blocking::Request| {
            eprintln!("before");
            let headers = request.headers_mut();
            let header_name = reqwest::header::HeaderName::from_static(CUSTOM_HEADER_NAME);
            let header_value = reqwest::header::HeaderValue::from_static(CUSTOM_HEADER_VALUE);
            headers.insert(header_name, header_value);
            eprintln!("after");
        },
    ))),
};
```

I think this probably means the issue is specific to another crate I'm depending on to modify the `request`.

Byron (approved):
Thanks a lot for turning the sketch into something usable!

What I like about it is that it can be upgraded easily, for instance I plan to have shared HTTP settings that every HTTP implementation should support in its own struct that can be added as field to the `reqwest::Options`. I think this will nicely layer with what's already here.

PR #565, opened 2022-10-22
fix: build error on Android

xJonathanLEI:
----

----

The library fails to build on Android due to a `cfg` misuse causing the `interpolate_user` function to be defined twice. This PR fixes it by using a correct `cfg` condition.

Discovered this issue when trying to use https://github.com/helix-editor/helix/pull/3890 on Android without success.

### DIFF
--- a/git-config-value/src/path.rs
+++ b/git-config-value/src/path.rs
@@ -170,7 +170,7 @@ impl<'a> Path<'a> {
         Err(interpolate::Error::UserInterpolationUnsupported)
     }
 
-    #[cfg(not(windows))]
+    #[cfg(not(any(target_os = "windows", target_os = "android")))]
     fn interpolate_user(
         self,
         home_for_user: fn(&str) -> Option<PathBuf>,


xJonathanLEI:
Wait.. Is the CI failure irrelevant? I should have changed nothing on targets other than Android...

Byron:
Sorry for the late reply, and thanks a lot for the the fix, it's much appreciated.

 The CI failure is a general issue on CI currently and triggered by [what I think is this bug](https://bugs.launchpad.net/ubuntu/+source/git/+bug/1993586) and it should be ~~fixed~~ worked around in `main` in a moment.

It looks like Helix isn't testing against Android on CI unless I am missing something (which I hope is the case), as I'd love to have the CI here to validate the change and warn about regressions in future. Otherwise I am pretty sure I will break android and not know about it until downstream breaks :/.

xJonathanLEI:
Thanks! Personally I think Android NDK and `cargo-ndk` is needed to cross-compile to Android and thus test it in the CI, but I have never tried doing that myself. Will try to put together something and submit another PR when I have time.

PR #569, opened 2022-10-26
Ensure body can be read before configure_request

paulyoung:
This fixes the issue described in https://github.com/PassFort/http-signatures/issues/17 where attempting to read the request body in `configure_request` would cause things to hang.

----


paulyoung:
There probably should be a test for this to prevent a regression.

Byron:
> There probably should be a test for this to prevent a regression.

I agree! For now I am willing to play it fast and loose as the reqwest backend is still something of your personal playground and will see many changes in the future provided it ends up working for you. Maybe in future, you could add `#[ignore]`d tests to indicate a test for particular cases is still missing. Besides, I'd hope that the existing mock server would be usable to trigger this particular code change as well in case you'd like to add such test in future.

PR #578, opened 2022-11-06
fix connections over the git:// protocol not specifying the host

pascalkuthe:
Fixes `gix clone` reporting a permission error when connecting to servers
over the `git://` protocol that use virtual hosts like `git.kernel.org`.

----


Byron:
Thanks so much for the fix (and the debugging that led to it)!

And I finally understood why CI sometimes fails with `git` terminating (when serving as remote during `file://*` access) unexpectedly. Some tests (those conducted with `test-serial`) change the environment of the current process, which affects the `git-upload-pack` process that is being spawned.

This is now fixed, as well, to make CI a little less flaky once again.

PR #580, opened 2022-11-06
Fix error when server omits service annoucement.

### LINKED ISSUE #577: Cloning linux from kernel.org fails
### Duplicates

### Current behavior 😯

Cloning the linux kernel from `kernel.org` instead of from github fails.
When using the git porotocol, the following error is produced:
```

Caused by:
    access denied or repository not exported: /pub/scm/linux/kernel/git/torvalds/linux.git
```

Cloning using `https` fails with 
```
Error: Expected to see "# service=git-upload-pack", but got "version 2\nagent=git/2.31.1\nls-refs=unborn\nfetch=shallow\nserver-option\nobject-format=sha1"`
```

### Expected behavior 🤔

Cloning the repository succedes (using both protocols).

### Steps to reproduce 🕹

Execute either
`gix clone git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git linux`
or
`gix clone https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git linux`

Both commands succeed with git.
I am not sure these issues are related but since they pertain to the same server I posted them together.
Cloning from other servers with these protocols work (`https` works for github and gitlab, `git` works for sourceforge) so its something specific to `kernel.org`.

pascalkuthe:
Fixes #577

----

### DIFF
--- a/git-transport/src/client/blocking_io/http/mod.rs
+++ b/git-transport/src/client/blocking_io/http/mod.rs
@@ -226,24 +226,12 @@ impl<H: Http> client::Transport for Transport<H> {
             .line_provider
             .get_or_insert_with(|| git_packetline::StreamingPeekableIter::new(body, &[PacketLineRef::Flush]));
 
-        let mut announced_service = String::new();
-        line_reader.as_read().read_to_string(&mut announced_service)?;
-        let expected_service_announcement = format!("# service={}", service.as_str());
-        if announced_service.trim() != expected_service_announcement {
-            return Err(client::Error::Http(Error::Detail {
-                description: format!(
-                    "Expected to see {:?}, but got {:?}",
-                    expected_service_announcement,
-                    announced_service.trim()
-                ),
-            }));
-        }
-
+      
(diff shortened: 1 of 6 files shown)


Byron:
Thanks a lot for setting up this PR! I am taking a look now to try and make it work - one issue was the fixture with had a newline in the end. Once removed, it doesn't hang anymore, but has other issues probably related to something else, or the fixture is still fishy. 

Fortunately there is a `curl` invocation in the Makefile to get data without `proxy`, I will try that next.

Byron:
It turned out that the fixture really wants a Content-Length header when it's provided by the mock server, for some reason, and 'chunked' encoding also was a problem, much like what I got through the proxy or through curl decodes the body so it's not really chunked.

With the fixture corrected, everything worked as it should. It's worth remembering that the test-setup with the mock http server acts a bit differently than the real world for reasons I don't understand (yet).

Byron:
Closing as it was rebased onto main due to me forgetting to pull before merging.

PR #586, opened 2022-11-08
configure http transport

Byron:
We should support the most common configuration flags for the widespread http transport.

### Tasks

* [x] analyse available `http` config keys
* [x] `http::Options` (shared by implementations) with most common configuration values.
* [x] Configure `curl` according to (simple) options
* [x] pass through the git user-agent to make it configurable
    * [x]  also add `gitoxide.userAgent` configuration for overrides.
* [x] Obtain 'simple' http options from configuration and pass them on
* [x] lenient mode for reading values
* [x] setup http proxy correctly (turn it into an actual URL like curl wants it)

### Next PR
* [ ] setup proxy-auth correctly as well, i.e. use credentials mechanism to fetch creds if username is present.
* [ ] apply remote overrides (probably best via `name` of the remote)
* [ ] find a way to apply environment variables and have them override (all?) values,
        probably via git-config. Be aware of lower-case versions of these vars (for historical reasons)
* [ ] Configure `curl` with all options available in `http::Options`
* [ ] try to implement complex `http.<url>.*` based option overrides


paulyoung:
@Byron that all sounds fine to me. Thanks for the heads up!

Byron:
@paulyoung Yet another change has landed which affects you. Now, to set transport options, your have to set an instance of `git_transport::client::http::Options` (with the `backend` field set) on the `remote::Connection` instance using `with_transport_options(…)`. That `backend` field is an instance of `git_transport::client::http::reqwest::Options`, which is the same as before. That way, all backends share a common set of options as provided by git and can implement them, with everything else being pushed into a custom type of the `backend`.

Furthermore, for debugging, it's now easy to make all traffic go through a proxy with `gix -c http.proxy=localhost:9090 remote ref-map` or similar. When looking at the docs of `reqwest` it appears that [it already reads the `HTTP_PROXY` environment variable](https://docs.rs/reqwest/latest/reqwest/index.html#proxies) which should do the same, even though I find it a bit much for library. It seems to be opt-out by calling `ClientBuilder::no_proxy()`.

PR #590, opened 2022-11-13
async-fetch

Byron:
Make it possible to fetch a pack with async transports.

### Tasks

* [x] port tests for async connect
* [x] port tests for async ref-map
* ~~port tests for async fetch/pack-receive~~ - tests can now easily be shared
* [x] port actual receive code, possibly with `blocking` - making it nicely interruptible is going
      to be the main feat here, and one test.
* [x] port all the other tests to run both async and sync versions, where feasible.


Byron:
@paulyoung I think this will work, and I am working my way through basic tests for what's already there, i.e. async ref-map, to an async pack-receive. The latter has custom code which will certainly have to be refactored to make it reusable between blocking and async, let's see how that will look like. Please note that I am happy to rewrite some boilerplate, but want to avoid duplicating code that has tests, as with the current setup I am unable to share tests between async and sync as it's possible in the much simpler `git-protocol` crate.

Byron:
I managed to figure out how to share all existing test code between async and sync implementations, which puts me right at the gates of finding a good way to get an unblocked pack-receive going. Let's see how that goes.

Byron:
This version of the implementation can run all pack-receive tests, and I took a shortcut to avoid more complex code for conditional use of the `blocking` crate. Now, `receive` is async, but blocks the executor of the client, but functions fine, to be suitable for typical client implementations that don't actually care about handling multiple connections at once (or if they do, hand these to threads anyway).

I have the disposition to merge this despite the shortcoming, but can be swayed to properly unblock [this part](https://github.com/Byron/gitoxide/blob/9b5f0eb06aa32ce3dd77781e85525a065525a7eb/git-repository/src/remote/connection/fetch/receive_pack.rs#L139-L156) if you think it's worth it (I don't think it takes long, just adds code as I have to duplicate this call I think).

PR #593, opened 2022-11-13
Added a notice that Perl needs to be installed on Fedora for building…

StandingPadAnimations:
… to work

----


Byron:
Thanks a lot! I took the paragraph and integrated into a more general heading to help dealing with all kinds of build issues. With that, the `max-pure` option was added which builds the fastest, most feature rich binary as possible with pure Rust implementations of Zlib and Sha1.

PR #596, opened 2022-11-16
push support in git-transport

Byron:
### Tasks

* [x] test for writing a typical push reqwest based on a real one
* [x] show how to receive server responses as well.

### Questions

* Should the response be read in parallel, which wouldn't work with HTTP anyway, or would it?
    - no, even though possible I don't think git does it either as the progress messages are from the step after having received the pack, the resolution phase. Thus progress can be provided while writing the pack, and then again while reading the response.


Byron:
@paulyoung [This test](https://github.com/Byron/gitoxide/blob/4d84a208e4d0a8bf3dbcfa6912b92ec1bdf4ec05/git-transport/tests/client/git.rs#L148-L210) shows how the transport part of a push would look like, both for sending the pack as well as for receiving the response.

`git-protocol` would probably wrap the this into its own abstractions similar to what was done for fetches, but it seems simple enough to do it by hand on transport level at least for a PoC. Please note that I didn't look at the underlying code of `git` at all, and `git-protocol` abstractions will probably have to know much more to capture all the nitty gritty details.

PR #597, opened 2022-11-17
read-worktree-config

SidneyDouw:
**How are configs loaded from different worktrees**

If `extensions.worktreeConfig` is enabled, then worktrees will load the worktree-specific config from the `$GIT_DIR/config.worktree` file, and the shared config from `$GIT_COMMON_DIR/config` file, where `$GIT_DIR` points to `path/to/main/worktree/.git/worktrees/linked-worktree-id` and `$GIT_COMMON_DIR` points to the main worktrees' git dir at `path/to/main/worktree/.git`.
When working in the main worktree `$GIT_DIR` and `$GIT_COMMON_DIR` point to the same path, namely the main worktrees git dir at `path/to/main/worktree/.git`.

**Tasks**

- [x] Detect if `extensions.worktreeConfig` is set and load and merge in the `config.worktree` file
  - [ ] make sure we are doing this in all the required places
- [ ] Use the `$GIT_DIR` and `$GIT_COMMON_DIR` env vars to load the configs

**Notes**

- gitoxide does not yet support linked worktrees, so we don't consider loading config files for those as they are handled differently than the main worktree
- `.git/config` is read before `config.worktree` meaning `config.worktree` overrides the shared config

**References**
- [git-worktree | config files](https://git-scm.com/docs/git-worktree#_configuration_file)
- [git-config | worktree extension](https://git-scm.com/docs/git-config#Documentation/git-config.txt-extensionsworktreeConfig)
- [repository layout | worktreeConfig](https://git-scm.com/docs/gitrepository-layout#_worktreeconfig)

### DIFF
--- a/git-config/src/file/init/comfort.rs
+++ b/git-config/src/file/init/comfort.rs
@@ -75,6 +75,7 @@ impl File<'static> {
     ///
     /// - globals
     /// - repository-local by loading `dir`/config
+    /// - worktree by loading `dir`/config.worktree
     /// - environment
     ///
     /// Note that `dir` is the `.git` dir to load the configuration from, not the configuration file.
--- a/git-config/src/file/init/comfort.rs
+++ b/git-config/src/file/init/comfort.rs
@@ -129,12 +129,12 @@ impl File<'static> {
         globals.resolve_includes(options)?;
         local.resolve_includes(options)?;
 
-        // TODO: Does the order matter here?
-        globals.append(local).append(Self::from_environment_overrides()?);
+        globals.append(local);
         if let Some(mut worktree) = worktree {
             worktree.resolve_includes(options)?;
             globals.append(worktree);
         }
+        globals.append(Self::from_environment_overrides()?);
 
         Ok(globals)
     }
--- a/git-config/src/file/init/comfort.rs
+++ b/git-config/src/file/init/comfort.rs
@@ -95,6 +95,20 @@ impl File<'static> {
             (local, path)
         };
 
+        let worktree = match local.boolean("extensions", None, "worktreeConfig") {
+            Some(Ok(worktree_config)) => worktree_config.then(|| {
+                let source = Source::Worktree;
+                let path = git_dir.join(
+                    source
+                        .storage_location(&mut |n| std::env::var_os(n))
+                        .expect("location available for worktree"),
+                );
+                Self::from_path_no_includes(path, source)
+            }),
+            _ => None,
+        }
+        .transpose()?;
+
         let home = std::env::var("HOME").ok().map(PathBuf::from);
         let options = init::Options {
             includes: init::includes::Options::follow(
@@ -114,7 +128,13 @@ impl File<'static> {
         globals.resolve_includes(options)?;
         local.resolve_includes(options)?;
 
+        // TODO: Does the order matter here?
         globals.append(local).append(Self::from_environment_overrides()?);
+        if let Some(mut worktree) = worktree {
+            worktree.resolve_includes(options)?;
+            globals.append(worktree);
+        }
+
         Ok(globals)
     }
 }
--- a/git-config/tests/fixtures/config_with_worktree_extension.sh
+++ b/git-config/tests/fixtures/config_with_worktree_extension.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+set -eu -o pipefail
+
+git init -q
+
+git config extensions.worktreeConfig true
+git config --worktree someSection.someSetting "some value"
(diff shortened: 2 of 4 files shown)


git-config/tests/fixtures/config_with_worktree_extension.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -eu -o pipefail
+
+git init -q
+
+git config extensions.worktreeConfig true
+git config --worktree someSection.someSetting "some value"
```
Byron, line 7: This would better be a meaningful section name and value, to indicate it comes from the worktree config. That way the test is more obvious and readable. _I mention this for adaptation in `git-repository`_.

Byron (approved):
Thanks a lot for getting started on this, and I am sure the research needed to get here will be useful further down the road.

Worktrees are certainly a beast on its own and I am thinking we should rather leave it out of `git-config` at first as it's considered `plumbing`. These _comfort_ methods are luring us into thinking otherwise though so I checked who uses them, and could only find `git-config` tests. This makes me think that these functions should rather be removed as their functionality is duplicated in a place that can do better in many ways: `git-repository`.

> gitoxide does not yet support linked worktrees, so we don't consider loading config files for those as they are handled differently than the main worktree

Actually it does, one can open worktree repos, and traverse worktrees as well. All functionality around worktrees is implemented with the knowledge about `GIT_COMMON_DIR` and `GIT_DIR` in mind. `git-ref` for instance has to be worktree aware too, and it is by having support for a `common_dir` on top of a `git_dir`.

~~Do you think that the removal of `comfort.rs` with all related tests and fixtures would be the way to go here in favor of moving worktree support to `git-repsitory`? If so, the removal can be done in this PR and the addition to `git-repository` can be done in another one. If you think there are other ways to avoid duplication of work between `git-config` and `git-repository`, please let me know. Generally I think that the methods provided thus far (except for `comfort`) are sufficient to set more logic on top.~~ - see next paragraph :D.

Before it's too late, there is an argument to be made for keeping `comfort.rs` as it does allow others to get a reasonable representation of git configuration without pulling in all of `git-repository`. However, unless `git-repository` can use these I think it's just duplication of work without `git-config` necessarily doing exactly the right thing, it's just an approximation. So I think those who use should rather not.

Checking the users via 'Dependents' on crates.io

* ✅ [cargo-generate](https://github.com/cargo-generate/cargo-generate/blob/235828ef6e6b3454c5ede35596441f02cfbe0a7c/src/git/gitconfig.rs#L4) only reads individual files
* ❌[watch-exec](https://github.com/watchexec/watchexec/blob/46fba794c1b18e9133c94cfc8ba04591b02847e3/crates/ignore-files/src/discover.rs#L56) uses `from_git_dir()` - this is probably where my inquiry has to end.

Ok, if `watchexec` is using it, let's leave the comfort API exactly for these cases but: let's not put more time in it for correctness. Instead, let's wait for complaints about missing features and we take it from there, and I see duplicate effort on the horizon to support this to work from linked worktrees (but should be doable with `git-discover`. But let's wait until people force us to and focus effort on `git-repository` instead.

That said, let's continue as follows: add `extensions.worktreeConfig` support to `git-repository` and test it with linked worktrees as well as the main one. The [respective fixture](https://github.com/Byron/gitoxide/blob/b46bff58e40bb9805af7ee7f96272f0dc19c0ac7/git-repository/tests/fixtures/make_worktree_repo.sh#L1-L0) might be helpful for that. I recommend using `git sparse-checkout` or the likes to have it configure the [main worktree accordingly](https://github.com/git/git/blob/a0789512c5a4ae7da935cd2e419f253cb3cb4ce7/builtin/sparse-checkout.c#L365), so we get a main worktree with worktree-specific configuration and a checkout with other sparse-related configuration, which is the way this is currently used anyway.

Thanks again for digging into this! I will merge this PR mostly as is and hope to see you again in `git-repository`. Please let me know if I am missing something fundamental though, I am sure you know these features better than me by now.

git-config/src/file/init/comfort.rs
```diff
@@ -95,6 +96,20 @@ impl File<'static> {
             (local, path)
         };
 
+        let worktree = match local.boolean("extensions", None, "worktreeConfig") {
+            Some(Ok(worktree_config)) => worktree_config.then(|| {
```
Byron, line 100: For reference in `git-repository`: We won't ignore errors there, instead there is `lenient` config toggle which allows to ignore errors like here only if it's set. Otherwise, it's an error to not be able to parse this a boolean. Handling `lenient` is very convenient by now.

PR #600, opened 2022-11-17
advanced http-transport configuration

Byron:
We should support the most common configuration flags for the widespread http transport.

### Tasks

* [x] setup proxy-auth correctly as well, i.e. use credentials mechanism to fetch creds if username is present. Also make it possible to approve these credentials, probably transparent in the transport as it's not a property of Transport (and shouldn't be I guess)!

### Next PR/left for later

* [ ] apply remote overrides (probably best via `name` of the remote)
* [ ] find a way to apply environment variables and have them override (all?) values,
        probably via git-config. Be aware of lower-case versions of these vars (for historical reasons)
* [ ] Configure `curl` with all options available in `http::Options`
* [ ] try to implement complex `http.<url>.*` based option overrides


Byron:
Thanks a lot. Here is [the new release](https://github.com/Byron/open-rs/releases/tag/v3.1.0).

PR #604, opened 2022-11-17
[git-object] Encode empty tags like git does

### LINKED ISSUE #603: [git-object] Encoding of an empty tag seems inconsistent with git's
### Duplicates

### Current behavior 😯

Currently, an annotated tag with an empty message gets encoded and decoded in the following format:
```
object 01dd4e2a978a9f5bd773dae6da7aa4a5ac1cdbbc
type commit
tag empty
tagger Sebastian Thiel <sebastian.thiel@icloud.com> 1592381636 +0800
```
as illustrated in the test fixture [empty.txt](https://github.com/Byron/gitoxide/blob/main/git-object/tests/fixtures/tag/empty.txt).

Note that the tagger line ends with an end-of-line character and there is no other end-of-line character between that and the end of the file.

I noticed this by using this library to process the tags on an existing git repo, and it would fail to parse some tags. I created a minimal repro from there (see steps to reproduce).

### Expected behavior 🤔

This seems inconsistent with the way `git` represents a tag with an empty message.

I would expect one more end-of-line in that file, like so:
```
object 01dd4e2a978a9f5bd773dae6da7aa4a5ac1cdbbc
type commit
tag empty
tagger Sebastian Thiel <sebastian.thiel@icloud.com> 1592381636 +0800

```

See the steps to reproduce to explain why I think this extra newline is needed for git compatibility. Am I missing something? Is there any way to generate a tag with an empty commit that git doesn't store with that extra newline? Please let me know if so.

### Steps to reproduce 🕹

* Create a new git repo with an empty commit
```
mkdir test_repo
cd test_repo
git init
git commit -m "commit" --allow-empty
```
* Create an annotated tag with an empty message
```
git tag -a tag -m ""
```
* Check git's representation of this tag (using bat here so I can show the output in a pretty way and there is no ambiguity about the extra newline character)
```
git cat-file -p tag | bat
```
Output:
```
───────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ STDIN
───────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ object b1a2ffad8b8b88e79765a412eba97b5c8eee6217
   2   │ type commit
   3   │ tag tag
   4   │ tagger Pierre Chevalier <pierrechevalier83@gmail.com> 1668709561 +0000
   5   │
───────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

pierrechevalier83:
A tag with an empty commit message is now displayed with an extra newline, consitently with the output of `git cat-file`.

The test fixture for `empty.txt` was updated, which means that the `round_trip` test now expects a correct round trip with the new format.

Note that the changes to tests in `immutable` reflect the new state of the code:
* `TagRefIter` now parses this body with an empty message.
* Where the length of the data was divided by 2, the point was to create an inconsistent dataset leading to an error. Unfortunately, with the extra character, it lines up so that the message can be parsed, so I had to change the magic number to keep the previous property.

Fixes #603

----

### DIFF
--- a/git-object/src/tag/decode.rs
+++ b/git-object/src/tag/decode.rs
@@ -53,7 +53,8 @@ pub fn message<'a, E: ParseError<&'a [u8]>>(i: &'a [u8]) -> IResult<&'a [u8], (&
     let (i, _) = tag(NL)(i)?;
     fn all_to_end<'a, E: ParseError<&'a [u8]>>(i: &'a [u8]) -> IResult<&'a [u8], (&'a [u8], &'a [u8]), E> {
         if i.is_empty() {
-            return Err(nom::Err::Error(E::from_error_kind(i, nom::error::ErrorKind::Eof)));
+            // Empty message. That's OK.
+            return Ok((&[], (&[], &[])));
         }
         // an empty signature message signals that there is none - the function signature is needed
         // to work with 'alt(…)'. PGP signatures are never empty
--- a/git-object/src/tag/write.rs
+++ b/git-object/src/tag/write.rs
@@ -29,8 +29,8 @@ impl crate::WriteTo for Tag {
             encode::trusted_header_signature(b"tagger", &tagger.to_ref(), &mut out)?;
         }
 
+        out.write_all(NL)?;
         if !self.message.is_empty() {
-            out.write_all(NL)?;
             out.write_all(&self.message)?;
         }
         if let Some(ref message) = self.pgp_signature {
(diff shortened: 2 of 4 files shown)


Byron:
Thanks a lot for your contribution, it's much appreciated.

A review-video can be found here: https://youtu.be/86OA4MrYaL4

Issue #292, opened 2022-01-06
Trial Rust edition 2021 upgrade


Byron:
Make it a major version bump for pre-release crates but hold back on released crates.

### Result

After running…

```bash
for name in 'git-hash'  'git-features'  'git-actor'  'git-config'  'git-object'  'git-diff'  'git-tempfile'  'git-chunk'  'git-traverse'  'git-pack'  'git-odb'  'git-packetline'  'git-url'  'git-transport'  'git-protocol'  'git-ref'  'git-repository'  'git-commitgraph'  'gitoxide-core'; do
    cargo fix --edition --package $name
done
```

…there was no meaningful lint so sticking with 2018 is probably OK for now.

Having a nicer prelude would be good, but it's unfortunately not checked by the migration guide apparently nor are there any other lints showing up.
Probably `gitoxide` will not upgrade for quite a while then until there is a way to make actual improvements to the codebase thanks to the edition jump.

### Tasks

* [ ] migrate `git-features` and `git-pack`, both of which suffer from a workaround not working anymore in edition 2021.

Byron:
The following crates can't be upgraded yet due to it moving a `*mut [T]` into threads for purpose of accessing items without aliasing. A workaround was used in edition 2018, but that doesn't work anymore in 2021 and I wonder how to do that then.

PR #605, opened 2022-11-18
Upgrade to 2021 edition

mrquantumoff:
----


Byron:
That's a good hint! Back in the days I already tried (see #292), but now with MSRV being 1.60 for `gitoxide` we are finally ready to do the switch.

PR #606, opened 2022-11-18
fix(describe): also sort entries lexigraphically

davidkna:
I noticed that lightweight git tags were sorted differently by `describe()`, compared to git. The difference seems to be that git will sort lightweight tags or tags created at the same time lexigraphically.

There was already a test that was supposed to check this, which I changed to actually verify this. The sorting code also first compared the time, followed by the priority, but as far as I can tell git will check the priority first, so I changed the order to match git.

----

### DIFF
--- a/git-repository/src/commit.rs
+++ b/git-repository/src/commit.rs
@@ -100,12 +100,12 @@ pub mod describe {
                     })
                     .collect();
                     // By priority, then by time ascending, then lexicographically.
-                    // More recent entries overwrite older ones.
+                    // More recent entries overwrite older ones due to collection into hashmap.
                     refs.sort_by(
                         |(_a_peeled_id, a_prio, a_time, a_name), (_b_peeled_id, b_prio, b_time, b_name)| {
-                            a_time
-                                .cmp(b_time)
-                                .then_with(|| a_prio.cmp(b_prio))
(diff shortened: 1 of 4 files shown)


Byron:

I couldn't believe that there is only one puny test for that anyway, and it appears that it is lacking a `git` baseline to be truly trustworthy, along with way more tests cases.

The reason I am saying this is that from what I can tell (please see my added commits) the proposed implementation also mixed up the values for the tuples so it probably worked accidentally. After fixing this, the test still succeeded and I hope it's correct now, but with one test it's hard to be sure.

Maybe there are more tests that can be obtained from the git test suite, even though I thought I already poached them, otherwise one would have to make some up.

With these changes, I am happy to merge this PR as I think it's definitely an improvement, and I am waiting for your OK in case you are OK with the lack of tests for this (and I think it's OK to be reactive for some parts and wait for the real world to come up with test-cases once something isn't as expected, but who would know if they don't know what `git` would do?). 

I am really looking forward on your thoughts here. Thank you.

Byron:
Thanks a lot for the tests, much appreciated! With these I feel confident that we can keep this feature in line with `git`, and validating that results match by hand seems alright here given the sample size.

PR #608, opened 2022-11-19
[refactor] Deduplicate Tag and TagRef

pierrechevalier83:
To do so, implement a conversion from a `&Tag` to a `TagRef`.

This wasn't possible with the existing reprensentation of `Tag` as the target was represented by an `ObjectId` which internally stored only 20 bytes, but the `TagRef` type used a `BStr` that represented the 40 bytes `oid` representation of that object.

To fix that issue, store the target as a 40 bytes `BString` in `Tag`.

Alternative: The existing conversion from `TagRef` to `Tag` could have been used instead, but that would have required allocating memory for the entire data-structure which felt heavy handed for a simple refactoring.

----

### DIFF
--- a/git-object/src/lib.rs
+++ b/git-object/src/lib.rs
@@ -159,8 +159,9 @@ pub struct TagRefIter<'a> {
 #[derive(PartialEq, Eq, Debug, Hash, Ord, PartialOrd, Clone)]
 #[cfg_attr(feature = "serde1", derive(serde::Serialize, serde::Deserialize))]
 pub struct Tag {
-    /// The hash this tag is pointing to.
-    pub target: git_hash::ObjectId,
+    /// The hash this tag is pointing to stored as a 40 bytes hexadecimal representation from which an `ObjectId`
+    /// can be constructed. Use `.target()` to access that `ObjectId`
+    pub target: BString,
     /// The kind of object this tag is pointing to.
     pub target_kind: crate::Kind,
     /// The name of the tag, e.g. "v1.0".
@@ -171,6 +172,13 @@ pub struct Tag {
     pub message: BString,
     /// A pgp signature over all bytes of the encoded tag, excluding the pgp signature itself.
     pub pgp_signature: Option<BString>,
+}
+
+impl Tag {
+    /// The hash this tag is pointing to.
+    pub fn target(&self) -> git_hash::ObjectId {
+        git_hash::ObjectId::from_hex(&self.target).expect("40 bytes hex sha1")
+    }
 }
 
 /// Immutable objects are read-only structures referencing most data from [a byte slice][crate::ObjectRef::from_bytes()].
--- a/git-object/src/object/convert.rs
+++ b/git-object/src/object/convert.rs
@@ -1,6 +1,6 @@
 use std::convert::TryFrom;
 
-use crate::{tree, Blob, BlobRef, Commit, CommitRef, Object, ObjectRef, Tag, TagRef, Tree, TreeRef};
+use crate::{bstr::BStr, tree, Blob, BlobRef, Commit, CommitRef, Object, ObjectRef, Tag, TagRef, Tree, TreeRef};
 
 impl From<TagRef<'_>> for Tag {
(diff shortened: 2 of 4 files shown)


pierrechevalier83:
Note: `Commit` and `CommitRef` can be similarly de-duplicated, but I will wait for feedback on this PR first to hear if it's acceptable to store an extra 20 bytes by hash or if we think that's something to be worried about.

pierrechevalier83:
Also, another alternative to storing a `BString` in `Tag` and a `BStr` in `TagRef` would be to store an `ObjectId` in `Tag` and a `&ObjectId` in `TagRef`. I guess I'm not sure where the discrepancy comes from in the first place, so very willing to get feedback on the best approach.

Byron (commented):

I took a moment to understand how it came to this myself and believe to have answers.

Generally, the `Ref` versions of objects should be able to be parsed from a buffer and refer to it with minimal overhead. Hence it will store `&BStr` as this represents the underlying data.

With the owned versions, there was no such requirement so the datatype was chosen that fit the data best and most efficiently, even if it comes at the cost of complexity or lack of compatibility with conversions.

Even though I can certainly live with making this change for `Tag` and `TagRef` and appreciate its benefits, applying the same logic to `Commit` and `CommitRef` would force many more allocations of `BString` due to the `tree` and `parents` fields, which seems like something I'd rather avoid.

Keeping `Tag` and `TagRef` now seems inconsistent with `Commit` and `CommitRef` so I'd rather deduplicate another way even though I didn't see one that wouldn't increase complexity elsewhere.

That said, I have recorded my own journey in commits even though they effectively undo the proposed changes in this PR for posterity, hoping that next time this raises questions it's more obvious why it is like this.

Since `gitoxide` is so cautions of allocations, I think a possible improvement would be to have a `SmallVec` like version of `BString` - something that doesn't enforce encodings _but_ can represent small strings without allocation with magic. It's not exactly critical to have that though as fewest uses actually use (or have to use) the owned versions of these object types, it's always easier and faster to keep the buffer around and parse on demand. This makes having this `SmallVec`-like optimization even more interesting as those who _do_ use the owned types may keep more objects around and can benefit.

Thanks again for contributing, please let me know if I missed something or you see other ways to improve the status quo. PRs are always appreciated.

pierrechevalier83:
Thanks a lot for the thoughtful explanations.

All you're saying makes a lot of sense to me.

> Keeping Tag and TagRef now seems inconsistent with Commit and CommitRef so I'd rather deduplicate another way even though I didn't see one that wouldn't increase complexity elsewhere.

Given the constraints, I don't see an obvious way here. Any win would have to be paid for with some extra complexity or allocation or space...

I think if we try to reduce the problem in this particular instance, it's that the `ObjectId` and `oid` abstractions are somewhat leaky: we end up being bound to their internal representations when trying to "dereference" an `ObjectId`.
If allocations are unacceptable, but storing twice the bytes isn't, one way around it might be to store a buffer of 40 bytes as an array in `ObjectId` for whichever variant of the enum. It does feel a bit clunky, though... We would loose locality if storing `Vec<ObjectId>`, for instance. The current `ObjectId` representation is minimal which feels right.

The current trade-off of sacrificing a tiny bit of code ergonomics is probably the correct one.

> Since gitoxide is so cautions of allocations, I think a possible improvement would be to have a SmallVec like version of BString

Yes, that sounds like a nice optimization regardless of this particular usecase.

Byron:
> If allocations are unacceptable, but storing twice the bytes isn't, one way around it might be to store a buffer of 40 bytes as an array in ObjectId for whichever variant of the enum. It does feel a bit clunky, though... We would loose locality if storing Vec<ObjectId>, for instance. The current ObjectId representation is minimal which feels right.

And not to mention that it's totally possible to keep tenth of millions of these `ObjectId`s in memory at a time, so doubling their size would be quite noticeable. Currently `gitoxide` can resolve deltas (for fetching, for instance) not only a lot (!!) faster than `git`, but it does so with significantly less memory as well. This didn't come for free and these data structures do use `ObjectId` fields as well.

PR #609, opened 2022-11-19
fix: Disable tag.gpgSign in test scripts

jpgrayson:
This is done for the same reason that commit.gpgsign is disabled for test scripts. It prevents test failures if the user has tag.gpgsign enabled in their global git config when invoking tests.

----


Byron:
Thanks a lot for this fix, much appreciated!

I took the liberty of removing a few of the explicit `git config` invocations as they shouldn't have been there in the first place - they must have survived a prior refactoring when `git-testtools` learned about `GIT_CONFIG_*` environment variables.

PR #610, opened 2022-11-19
Fix typo: contirbute -> contribute

precondition:


Byron:
Thanks a lot, good catch!

PR #615, opened 2022-11-20
tests: Set GIT_EDITOR in make_rebase_i_repo.sh

jpgrayson:
If the user has core.editor set in their global git config, then that value takes precidence over the EDITOR environment variable. The GIT_EDITOR environment variable, however, has higher precidence than core.editor. For this test, using GIT_EDITOR ensures that the desired sed command line is used.

----


Byron:
Thanks a lot for wading through this and leaving `gitoxide` better than before. Much appreciated.

PR #616, opened 2022-11-21
fix: Discover repo with relative path and ceiling

jpgrayson:
A couple of problems are repaired to allow discovering a repository from "." with a ceiling directory.

One problem was that find_ceiling_height() did the wrong thing when confronted with any relative search_dir. This is resolved by converting search_dir to be absolute if it is relative.

The other problem was that discover_opts() also mishandled relative paths. When the cursor started out as ".", cursor.pop() would be blindly called such that cursor would be "" for the second iteration. When a ceiling directory was in use such that there was a max_height, the current height would be burned going from ".", to "", and then to "<cwd>", before finally actually getting to a real parent directory. This problem is ameliorated by testing whether the cursor has a non-empty parent before popping.

N.B. the new test case relies on the test running from the git-discover directory such that the gitoxide repository will be found. This is a bit fragile and will fail if, for example, the test is run from an unpacked gitoxide source tarball.

----

### DIFF
--- a/git-discover/tests/fixtures/make_basic_repo.sh
+++ b/git-discover/tests/fixtures/make_basic_repo.sh
@@ -10,6 +10,7 @@ git commit -q -m c1
 echo hello >> this
 git commit -q -am c2
 
+mkdir subdir
 mkdir -p some/very/deeply/nested/subdir
 
 git clone --bare --shared . bare.git
--- a/git-discover/tests/isolated-01.rs
+++ b/git-discover/tests/isolated-01.rs
@@ -1,0 +1,46 @@
+use git_discover::upwards::Options;
+use std::path::Path;
+
+#[test]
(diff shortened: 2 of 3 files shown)


Byron:
Thanks a million! This portion of the code wasn't exactly loved by me and is one of the weaker areas of `gitoxide`, but with these improvements I gained a lot of trust and believe it's an improvement throughout.

I have added a standalone test, the first of its kind, which can actually `chdir` and not rely on a parent `gitoxide` repository, which I hope can lead the way towards having more realistic tests for some other aspects of the search in future which may lead to retiring some of the tests that have some expectation, which was done here as well.

Maybe by looking at `isolated-01` you get other ideas on how to get more test coverage, or just to test other cases that interest you, I will do the same.

In any case, this branch will be merged soon.

Byron:
Strange, this definitely has been merged.

PR #625, opened 2022-11-23
Path normalize

jpgrayson:
The semantics of the `absolutize()` function tripped me up, so here's a different name that may better fit its semantics.

----


Byron (changes requested):
Thanks! Names are hard 😅.

Could you fixup the commits so that the changes in `git-path` are in one commit with the subject `refactor!: …`, which is what's in cdaa8bf but with exclamation mark to help `cargo smart-release` pick it up as breaking change? Then there could be another commit adjusting all crates in need to those changes. I do this typically with `Adjust to changes in <crate-name>` . That way it's clear where the breaking changes are and version updates are handled correctly automatically.

That would help me a lot, thank you.

Byron:
Thanks a lot for the changes and sorry for the late response. Somehow I did not get notified or managed to consider my response done even though it wasn't.

In any case, it looks good now.

Before I go, one minor nit in case it can help with future PRs. Unless changes are user-facing, I don't prefix them with `git-conventional` style names to not have them show up in the changelogs. Commit 8ab47bb would be one without the `refactor: ` for instance as users don't have to care about that. Those who do will see a listing of all commits that affected the crate in question in the changelog, but it's much less prominent that way.

Writing the above I realized that this topic is clearly missing in the `DEVELOPMENT.md` file and I tried to remedy that in 932c353dc8f55722bdce91b1f62f06acb80ed4e6 . Your feedback on it is very welcome.

Issue #67, opened 2021-04-09
pack-generation MVP


Byron:
TBD

Byron:
Related to https://github.com/Byron/gitoxide/discussions/53

Byron:
In pursuit of better control pack generation and also pave the way for improved `async` integration, I figured having an `Iterator` interface would be a good idea.

Now it's possible to step through parallel computations.

However, the respective implementation has to expose `unsafe` due to the use of a scoped thread which exposes its join handle that can thus be leaked.

https://github.com/Byron/gitoxide/blob/15e47480054d9a517c28f47db3b5fa87968a307e/git-features/src/parallel/in_parallel.rs#L96

Depending on where this is exposed, `unsafe` might bubble up even further - after all, anything that holds the `SteppedReduce` can also leak it.

My intuition is to stop bubbling this up beyond `git-features` just to keep it practical, even though technically it's incorrect. What do you think, @joshtriplett?

joshtriplett:
It's a CPU-intensive operation; my first instinct would be to run it normally and use unblock or similar to run it on a blocking thread.

Trying to structure the computation so that it happens incrementally seems incredibly painful. And in particular, trying to adapt an operation that happens in a thread to happen incrementally seems like it's incurring all the pain of async without any language support for async.

I would suggest building the initial MVP in a synchronous fashion, on the theory that it can still be run in a background thread and controlled via an async mechanism.

I definitely don't think it's OK to use a scoped thread and hide the unsafe, if the unsafe isn't truly encapsulated to the point that you can't do anything unsound with the interface.

joshtriplett:
One other thought on that front: compared to the cost of generating a pack, one or two allocations to set up things like channels or Arc will not be an issue.

Byron:
Thanks for sharing. The main motivator for using scoped threads is to allow standard stack based operation without any wrapping - it's not at all about allocations, merely about usability and the least surprising behaviour. Truth to be told, I cannot currently imagine how traversal will play into static threads when arcs are involved especially along with traits representing an Object (an attempt to allow things like traversing directory trees).

What I take away is the following

* let's not hide `unsafe` unless it's encapsulated
* let's not make step-wise computation or extreme async friendliness a requirement for the MVP if something like `blocking` would work, too.

I hope to overcome my 'writers block' and just write the missing bits to be able to see through the whole operation and play with the parts more until I find a version of the API that feels right.

Byron:
[...]

PR #630, opened 2022-11-27
switch to custom Hasher implementation

pascalkuthe:
Speeds up pack creation by about 1% and ensures
that the `Hasher` will remain robust when
more variants (like sha256) are added to `ObjectId` in the future.

Looking at the flamegraph, the `HashMap` doesn't even show up tough.
It seems pack creation is almost entirely bound by zlip decoding during commit transversal.
So this might be a larger improvement in hashing performance but just doesn't matter as much overall.
Since git is still quite a bit faster during (single-threaded) pack creation it 
would definitely be interesting to find out why `gitoxide` is slower.
Since `gitoxide` uses a `zlib-ng` I would expect it to be faster than `git` if decoding is really the bottleneck.
Without that advantage the gap would be even larger.

Perhaps git somehow manages to do less work during transversal compared to `gitoxide`?

----


Byron:
> Perhaps git somehow manages to do less work during transversal compared to gitoxide?

I would not be surprised. Now, having forgotten most of the history collected [in the tracking issue](https://github.com/Byron/gitoxide/issues/67) I wonder why it decompresses entries to learn about the size instead of querying the index for that. Oh, right, because it first has to discover all objects and that tree traversal takes a long time even though it already skips over trees it has already seen.

git-pack/Cargo.toml
```diff
@@ -45,6 +45,7 @@ git-object = { version = "^0.23.0", path = "../git-object" }
 git-traverse = { version = "^0.19.0", path = "../git-traverse" }
 git-diff = { version = "^0.23.0", path = "../git-diff" }
 git-tempfile = { version = "^3.0.0", path = "../git-tempfile" }
+git-shamap = { version = "^0.1.0", path = "../git-shamap" }
```
Byron: I definitely like to take control over the hashmap implementation.

Byron (approved):
Thanks a lot for contributing! Every percent of performance is appreciated!

As a general note, it's very important to split commits and use the correct commit message to signal breaking changes. I have created a write-up here: https://github.com/Byron/gitoxide/commit/932c353dc8f55722bdce91b1f62f06acb80ed4e6 .

Besides that I think it's ready for being merged, but wanted to see if you are OK with the changes first.

Thank you

pascalkuthe:
Looks good to me, I will follow the commit conventions in the future. 
The documentation will certanly help with that.

A few comnments: 
> - pledge to not using unsafe code

Fine for now but custom collections (and hashtables in particular) are a prime candidate for unsafe code so we might have to remove that in the future.

> rename `git-shamap` to `git-hash-collections`

Bikeshed: Maybe `git-hashtable` (or `git-hashtables`) instead? 
All hash based collections I know of are `hashtable`s  internally and just have some extra abstractions on top so `hash-collections`  sounds  a bit odd to me (but it's also fine)

Byron:
> Fine for now but custom collections (and hashtables in particular) are a prime candidate for unsafe code so we might have to remove that in the future.

That sounds absolutely fair. All new crates (should) start out with these deny's (except for `missing_docs`), and that should help to also remind us that we try to avoid `unsafe`, and write safe code first if at all possible.

> Bikeshed: Maybe git-hashtable (or git-hashtables) instead?

Yes, let's do that instead! `git-hashtable` sounds good to me.

I think once the naming is fixed, we can merge this one.

Byron:
Closing as it's already merged, but I had to rewrite all commits but the first one in the end.

PR #634, opened 2022-11-30
avoid UTF8 decoding in git-transport/git-protocol

Byron:
Probably mainly for historical reasons we would only have a `BufRead` instance after a handshake
which makes it difficult to access the packet lines underneath and impossible to do it without using
`String` as line buffer.

This PR changes that and eithers reads all lines in advance or provides an ExtendedBufRead with
(newly added) readline() support.

### Tasks

* [x] capabilities V2 read from buffer
* [x] handshake ref lines V1 read line-by-line via `Vec<u8>` instead of `String` (for blocking only due to tests)
* [x] v2 refs read `Vec<u8>` instead of `String` (for blocking only due to tests)

**extra**
* [x] env var permissions should be by topic and fallback to by-prefix, not the other way around


paulyoung:
Thanks! 🙂

PR #635, opened 2022-12-01
`git-repository` loads worktree configs

SidneyDouw:
Give `git-repository` the capability to load worktree specific configuration files, if the appropriate setting (`extensions.worktreeConfig`) is set in the main config file.

**Tasks**
- [x] test fixture that creates multiple worktrees with individual `config.worktree` files
- [x] tests that check…
  - [x] …if shared config is being loaded
  - [x] …if worktree config is being loaded
  - [x] …if worktree config overrides shared config settings
- [x] implement loading of worktree configs in `git-repository`


git-repository/tests/fixtures/make_worktree_repo_with_configs.sh
```diff
@@ -0,0 +1,29 @@
...
+  git config override.setting "set in the shared config"
+)
+
+(cd wt-1
+  git config --worktree worktree.setting "set in wt-1"
+)
+
+(cd wt-2
+  git config --worktree worktree.setting "set in wt-2"
+  git config --worktree override.setting "override in wt-2"
```
Byron, line 28: Setting values so they describe their purpose and show up in the tests like this works really well! Thank you.

Byron (approved):
Thanks a lot, I absolutely like how spot-on the addition of worktree configuration support is in the bootstrapping sequence of the repository.

As a note and for future reference, it's important to properly classify configuration using their `git_config::Source` - each section has metadata attached by which it can be filtered and that's relevant to how much we can 'trust' these values.

It's great to see that finally `gitoxide` is able to properly start up and load everything that matters, laying the ground work for handling massive monorepos correctly soon enough :).

PR #645, opened 2022-12-07
Allow parsing arbitrary URL schemes

paulyoung:
Prior to this change, a URL with an unknown scheme would fail to parse. I think it makes sense to parse unknown schemes to `git_url::Scheme::Ext(String)` instead.

The [`bad_url_rewriting_can_be_handled_much_like_git` test](https://github.com/paulyoung/gitoxide/blob/4753e641eada72f4e944811ea85390481444b210/git-repository/tests/repository/remote.rs#L202) now fails because it [expects an error](https://github.com/paulyoung/gitoxide/blob/4753e641eada72f4e944811ea85390481444b210/git-repository/tests/repository/remote.rs#L218) when parsing `foo://dev/null`.

I wasn't sure what to do about that and hoped we could discuss it here.

----

### DIFF
--- a/git-repository/tests/fixtures/make_remote_repos.sh
+++ b/git-repository/tests/fixtures/make_remote_repos.sh
@@ -165,7 +165,7 @@ git init --bare bad-url-rewriting
 [remote "origin"]
   pushUrl = "file://dev/null"
 
-[url "foo://"]
+[url "invalid:://"]
   pushInsteadOf = "file://"
 
 [url "https://github.com/byron/"]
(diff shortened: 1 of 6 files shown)


Byron:
Thanks a lot! I totally forgot that the `rad` protocol was already merged into `Scheme::Ext`, and it seems like a step in the right direction to open it up to arbitrary protocols for use by custom implementations.

> I wasn't sure what to do about that and hoped we could discuss it here.

I played around with it and found another way to make the URL rewriting in `git-repository` fail, so the test can still work as before.

This PR will be merged once CI is green.

Issue #471, opened 2022-07-27
`git-date` towards 1.0


Byron:
We track features we consider necessary to release version 1.0 of the `git-date` crate. The following listing may not be complete, and doesn't have to be in order to qualify. 1.0 can be an minimal viable product despite `git` supporting additional details

### Features

* [ ] parsing
   - todo: list all the source formats
* [ ] flexible `Time` serialization
     - Time is actually a date-time and should be printable as date and/or time. Maybe integrate with the `time` crate which supports flexible formatting?

willstott101:
Having a look at this as it seems like an easy introduction. I assume the goal with `human` time is not to replicate git exactly, but to take the concept and run with it a bit?

WIP: https://github.com/willstott101/gitoxide/commit/2da937707d618ed25d224d394475dd8b73e2c3e4

Have some comments there re when git (2.37.2) chooses to show timezones seeming a bit weird to me.

Another question that might help me finish this off: Should I prevent the `time` crate from leaking into public API? I assume we do want to avoid leaking it, but checking as it might influence a few choices.

willstott101:
Final question re `local`, would we like that to be an additional boolean argument to `Time::format`, or more like `time.local().format()`?

Byron:
> Having a look at this as it seems like an easy introduction.

I am glad you are giving it a shot, thank you!

> I assume the goal with human time is not to replicate git exactly, but to take the concept and run with it a bit?

As baseline, `human` parsing should support what git does. If 'running with it' means to make improvements, like to fix shortcomings, or add more parseable input that will improve the user experience, I'd be open to that, too.

> Have some comments there re when git (2.37.2) chooses to show timezones seeming a bit weird to me.

I think it's important to do the same thing, unless it's clearly a bug. Here that doesn't seem to be the case though. However, there could be configuration options to not display the timezone, for example, to allow further customization. It's perfectly fine to do that as long as it's possible to get git behaviours.

> Another question that might help me finish this off: Should I prevent the time crate from leaking into public API? 

It's currently exposed and it can remain exposed. The `time` crate I consider 'safe' to rely on and would rather add fixes upstream than to try to abstract everything. In other words, `time` is there to stay and is a trusted building block.

> Final question re local, would we like that to be an additional boolean argument to Time::format, or more like time.local().format()?

[...]

PR #646, opened 2022-12-09
fix: always consider timestamps as UTC when loading from commits

willstott101:
Hello, whilst working on #471 (very slowly every now and then) I realised gitoxide is incorrectly using [`replace_offset`](https://docs.rs/time/0.3.17/time/struct.OffsetDateTime.html#method.replace_offset) when we should be using [`to_offset`](https://docs.rs/time/0.3.17/time/struct.OffsetDateTime.html#method.to_offset).

> Replace the offset. The date and time components remain unchanged.

Which sounds great, cause what git stores is a timestamp in UTC and an offset explaining in which timezone the commit was made. So we want to apply the offset without changing the time.. however the time crate doesn't store time as unix seconds! So replacing the offset without adjusting the stored day and time is not what we want.

```
let t = time::OffsetDateTime::from_unix_timestamp(123456789).unwrap();
let to = t.replace_offset(time::UtcOffset::from_whole_seconds(3000).unwrap());
assert_eq(t.unix_timestamp(), to.unix_timestamp()); // THIS FAILS
```

And just to verify for myself I wrote a small script which uses `git` to convert from the hardcoded timestamp in the tests of this repo, to various formats. https://gist.github.com/willstott101/ecbd08d359e7bbf8dbfa0ac26f25b367

This script outputs the following (which this patch makes the tests for formatting in this repo match)
```
human= Nov 30 1973        human-local= Nov 29 1973        iso= 1973-11-30 00:03:09 +0230    raw= 123456789
```

----

### DIFF
--- a/git-date/src/time/format.rs
+++ b/git-date/src/time/format.rs
@@ -63,6 +63,6 @@ impl Time {
     fn to_time(self) -> time::OffsetDateTime {
         time::OffsetDateTime::from_unix_timestamp(self.seconds_since_unix_epoch as i64)
             .expect("always valid unix time")
-            .replace_offset(time::UtcOffset::from_whole_seconds(self.offset_in_seconds).expect("valid offset"))
+            .to_offset(time::UtcOffset::from_whole_seconds(self.offset_in_seconds).expect("valid offset"))
     }
 }
--- a/git-date/tests/time/format.rs
+++ b/git-date/tests/time/format.rs
@@ -6,7 +6,7 @@ use time::macros::format_description;
 
 #[test]
 fn short() {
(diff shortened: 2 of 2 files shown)


Byron (approved):
Thanks so much for this contribution! Such a small change with massive impact, and good we catch this one now!

Seeing the git script to have a baseline, maybe this can be the inspiration for improving [the existing baseline tests](https://github.com/Byron/gitoxide/blob/70fe59f4a1cad25f687397206ee2cbe50e643181/git-date/tests/fixtures/generate_git_date_baseline.sh#L30) as well. _(And I don't even know what that would be specifically)_.

Byron:
(Merging, the CI failure is already fixed in `main` and unrelated)

PR #652, opened 2022-12-12
fix: A negative system timezone offset results in correct Time struct

willstott101:
Another really tiny PR fixing random parts of git-date

I don't know how to write tests for this, as the local timezone offset in tests appears to always be `+0000`, regardless of system time. This implies that tests have been considered by `time`... so presumably the local time is mockable somehow, but I couldn't find anything searching for `mock` in the time crate's github.

Sorry for the very small PRs, I'd just rather get easy things out of the way when I notice them, since I'm just tinkering when I have a few minutes rather than dedicating real time at the moment.

The actual affects of this are git commits being serialized incorrectly:
```
use git_date::time::{format, Format};

fn main() {
    println!("{:?}", git_date::Time::now_local_or_utc().format(Format::Custom(format::ISO8601_STRICT)));
    println!("{:?}", git_date::Time::now_local_or_utc().to_bstring());
}
```
results in (before the patch,  note the disagreeing timezones)
```
"2022-12-12T15:58:24-05:00"
"1670878704 +0500"
```
after this patch
```
"2022-12-12T16:00:04-05:00"
"1670878804 -0500"
```
----

### DIFF
--- a/git-date/src/time/init.rs
+++ b/git-date/src/time/init.rs
@@ -37,11 +37,11 @@ impl Time {
             .expect("this is not year 2038");
         // TODO: make this work without cfg(unsound_local_offset), see
         //       https://github.com/time-rs/time/issues/293#issuecomment-909158529
-        let offset = time::UtcOffset::local_offset_at(now).ok()?;
+        let offset_in_seconds = time::UtcOffset::local_offset_at(now).ok()?.whole_seconds();
         Self {
             seconds_since_unix_epoch,
-            offset_in_seconds: offset.whole_seconds(),
-            sign: Sign::Plus,
+            offset_in_seconds,
+            sign: offset_in_seconds.into(),
         }
         .into()
     }
@@ -62,7 +62,7 @@ impl Time {
         Self {
             seconds_since_unix_epoch,
             offset_in_seconds,
-            sign: Sign::Plus,
+            sign: offset_in_seconds.into(),
         }
     }
 }


paulyoung:
> presumably the local time is mockable somehow, but I couldn't find anything

From what I’ve seen the tests generally use [`GIT_COMMITTER_DATE`](https://github.com/Byron/gitoxide/search?q=GIT_COMMITTER_DATE) and `GIT_AUTHOR_DATE` in various places.

There’s a test helper that sets up a static environment, which is probably a good place to start before looking at other search results.

https://github.com/Byron/gitoxide/blob/35bbb15faae22852811f591d09f395ec7a9d5a05/tests/helpers.sh#L2-L11

Byron:
Thanks so much! I love these small PRs, please keep them coming. Whatever suits you best.

The problem with testing this particular function is not that the time is pinned as @paulyoung rightfully explained, but that these functions don't care about this environment variable at all, which is intended. Mocking would need to be supported by `time` as built-in feature for testing, or we'd have to fully isolate the test process to be able to meddle with its locale settings somehow. For now I am OK to keep fixing bugs as we find them, it seems like more effort than it's worth and ideally `time` will support testing better, maybe it does already. After all, they have to test this as well, so looking at their test-suite might be worth exploring. Maybe it's also `git-date` which should make its methods more testable, I can imagine that internally this method can be purely functional, and has unit-tests just for that, and it's just called with the result of `time::UtcOffset::local_offset_at(now)` which can be controlled in the test.

In any case, I have added a single locale-dependent test which always works on my machine (+2h) and should now also work on yours with -5h, to have at least something that could catch this issue.

willstott101:
Thanks, your test would make sense but from my testing it seemed like the `time` crate was always returning `+0000` in test code, but I don't see why that would happen from the `time` crate's source - unless cargo is doing something with env vars that affect libc. I might investigate further at some point, but for now I agree it's fine to merge without a reproducible test for this.

Byron:
Oh, right, one has to run these tests with `RUSTFLAGS="--cfg unsound_local_offset"  cargo test …` to even get the local time. It's a long-standing issue with an unsafe C implementation needed to read the time offset database of the system, and it's gated that way.

At least I could use this discovery to build release binaries with that flag enabled from now on 🎉.

PR #655, opened 2022-12-13
Read split index

SidneyDouw:
Currently, only the split index with the link extension is read, but the shared index is not, which means the in-memory representation is different when comparing a split index with a regular index.

Where to best place the `resolve_link_extension` function?
- should this only happen when we try to write a link extension or just happen by default everytime we load an index?
- does this always require a `git_index::File`?

**Tasks**
- [x] test comparing a split index with an identical "regular" index (in memory)
- [x] read and merge the shared index into the in-memory representation and discard the link extension
- [ ] ~~I am removing entries but not the corresponding path names in the path_backing~~
- [x] improve error handling - make sure to have distinct paths for the possible outcomes and dedicated errors with helpful messages (double check with [git](https://github.com/git/git/blob/57e2c6ebbe7108b35ba30184dcbcb6c34c929ad8/split-index.c#L112:L148))
  - [x] bound check for split & shared index entries
  - [x] entry marked for both replace & remove
  - [x] split entry path is not empty

**Notes**
- `git` does not seem to delete / update the entries immediately, but rather uses the `CE_REMOVE` and `CE_UPDATE_IN_BASE` flags


SidneyDouw:
I think this should be ready to have a look at.

This adds a new `resolve_link_extension` function which for now is being called when a `git_index::File` is being loaded. It checks for a link extension and tries to merge in the shared index into `git_index::State`

Byron (changes requested):
Thanks a lot for getting the ball rolling on the `link` extension, much appreciated!

It's amazing how thoroughly I misunderstood how it actually works, and as always it's much more complex than I thought it would be :D.

Besides improving the error handling (probably also by validating we have no less checks [than git seems to do](https://github.com/git/git/blob/57e2c6ebbe7108b35ba30184dcbcb6c34c929ad8/split-index.c#L112:L148)), I would love to have a more complex split-index that doesn't have 100% of all entries. When comparing both files, it felt like the shared and split index were very similar. That said, I also didn't validate exactly how entries are handled yet and will compare it to what git does in the next (and probably last) review round.

>  I am removing entries but not the corresponding path names in the path_backing

That's probably OK as we are definitely in special-case territory here. Wasting a little bit of memory seems fair.

> should this only happen when we try to write a link extension or just happen by default everytime we load an index?

I think it has to happen when loading the index to dissolve the `link`. This will naturally remove it when writing as it's not present in memory anymore. That's the bare-bone level of support we are going for :).

SidneyDouw:
I added some error checking, ensuring I check for the same errors [like git does](https://github.com/git/git/blob/57e2c6ebbe7108b35ba30184dcbcb6c34c929ad8/split-index.c#L112:L148).

At the moment I am checking the link extensions bitmaps for certain criteria in a separate function and returning a `VerifiedBitmaps` struct from that, which can then be used safely in the `Link::dissolve_into()` function. 
Not sure this is the right approach, but let me know what you think :)

Byron (changes requested):
Thanks for implementing the error handling, that's exactly what I had in mind.

I particularly like that it keeps the `Corrupt` state simple and uses a message only - ultimately there will be no code to match on it anyway, it's solely for some (unlucky) user.

After having studied the linked canonical code I believe that the verification can be done inline. The `CE_REMOVE` flag can be used just like git does, and just like `git` these entries should be removed. For us, this is costly, but it's OK (`Vec::retain(…)` should do it). Don't worry about updating the path backing, lets just have a few zombie paths in there, that's better than having to rebuild this one as well.

Lastly, I think that CE_UPDATE_IN_BASE is not relevant to us as we don't maintain the split index (we dissolve it).

Once all error handling is in-line without these extra allocations, I think this PR can be merged 🎉.

SidneyDouw:
`CE_REMOVE` and `Vec::retain` are now being used for the delete bitmap.
The allocating `VerifiedBitmaps` struct is now removed and the `verify_bitmaps` function now just returns a `Result<(), …>` instead. To make this a little nicer I thought I would try to make a version of `git_bitmap::ewah::access::Vec::for_each_set_bit` that can return a `Result` instead of an `Option`, but this turned out a little more complicated than expected:
I first tried with a `Result<(), Box<dyn Error>>`, but I couldn't quite make that work. Then I thought I would try with a generic Error type, `Result<(), E>`, but this needs a `E: From<TryFromIntError>` trait bound.
The latter is now the current version and works / compiles, but again I am unsure if this is a good approach, as it feels a little clunky.

Would love to know your thoughts here :)

Byron:
As I think this PR is ripe to be picked I took the liberty to try an alternative which doesn't involve modifying `git-bitmap` but passes errors out of the `FnMut` closure in a side-channel. Despite looking a bit clumsy (especially since `cargo fmt` is on strike) it does the job, is inline and makes obvious that we do bounds checking. 

Please note that whenever there is an API change in a crate, it should be done in a separate commit along with [conventional commit messages](https://www.conventionalcommits.org/en/v1.0.0/) to auto-update the respective changelog.

That said, I am looking forward to PRs that contain API improvements to make replacing `git2` in other projects look even more convincing. All the best.

PR #660, opened 2022-12-17
Use specific `Iter` constructors in stats example

Piturnah:
When looking through this example, I thought it might make sense to use the `Iter` constructors that are specifically for this use case, rather than manually do it using `Platform::prefixed`.

**NOTE**: Untested since I just did this quickly through GH editor. If I get a chance I'll run the example properly but I looked at the API docs and source and it looks like this should be a trivial change (touch wood)

----

### DIFF
--- a/git-repository/examples/stats.rs
+++ b/git-repository/examples/stats.rs
@@ -49,9 +49,9 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     // let num_branches = repo.branches()?;
     // let num_branches = repo.branches.remote("origin")?;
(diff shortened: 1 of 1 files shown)


Byron:
Thanks for the contribution, that's a great catch!

CI is happy, and so am I.

Byron:
By the way, I have added a [quick analysis of `gex`](https://github.com/Byron/gitoxide/issues/661) to see what it would take to do everything natively in `gitoxide`. Since `git` is invoked as well it would actually take much less, but `git status`-like functionality is definitely still missing but required. However, that should be available within the next 6 months.

PR #667, opened 2022-12-21
open::Options::open_path_as_is()

Byron:
The path to git repositories is well-known as they either end in `.git` or `.../.git`.
If this is not the case, by default we append `/.git` to the path.

With this new option enabled, no path transformations apply to open the given path as is,
which is preferable if you know it's a non-standard git repository folder name.


Byron:
Rebased on top, closing.

PR #669, opened 2022-12-22
fix(ssh): fixes SSH clones

noirbizarre:
This PR fixes SSH cloning handling by:

- Settings the default SSH port to 22 (was set to 21)
- Removing git-upload-pack extra parameters (rejected by both github and gitlab, the original `git` client is not sending those, can be verified by cloning any repository with github/gitlab `GIT_SSH_COMMAND="ssh -vvvv"`)
- Removing the double user inclusion in URL (resulting in `git@git@github:` URLs as well as broken paths given as SSH parameters)
- Properly handling relative paths in URL reconstruction on `connect()`

This PR follows #663 but while #663 was focusing on SCP-like URLs handling with some predefined test cases, this PR was done directly including the code in a project: results in the ability to clone directly from copy-pasted github and gitlab SCP-like provided repositories URLs 🎉 

----

### DIFF
--- a/git-transport/src/client/blocking_io/ssh.rs
+++ b/git-transport/src/client/blocking_io/ssh.rs
@@ -56,16 +56,16 @@ pub fn connect(
         _ => return Err(Error::UnsupportedSshCommand(ssh_cmd.into())),
     };
 
-    let host = match user.as_ref() {
-        Some(user) => format!("{}@{}", user, host),
-        None => host.into(),
-    };
-
     let path = git_url::expand_path::for_shell(path);
-    let url = git_url::Url::from_parts(
+    let url_parser = if path.to_string().starts_with('/') {
+        git_url::Url::from_parts
+    } else {
+        git_url::Url::from_parts_as_alternative_form
+    };
+    let url = url_parser(
         git_url::Scheme::Ssh,
         user.map(Into::into),
-        Some(host.clone()),
+        Some(host.to_string()),
         port,
         path.clone(),
     )
@@ -82,7 +82,7 @@ pub fn connect(
         None => blocking_io::file::SpawnProcessOnDemand::new_ssh(
             url,
             ssh_cmd.into(),
-            ssh_cmd_line.chain(Some(host.as_str())),
+            ssh_cmd_line.chain(Some(host)),
             None::<(&str, String)>,
             path,
             desired_version,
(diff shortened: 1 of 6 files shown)


git-transport/src/client/blocking_io/ssh.rs
```diff
@@ -56,16 +56,16 @@ pub fn connect(
...
     };
 
-    let host = match user.as_ref() {
-        Some(user) => format!("{}@{}", user, host),
-        None => host.into(),
-    };
-
     let path = git_url::expand_path::for_shell(path);
-    let url = git_url::Url::from_parts(
+    let url_parser = if path.to_string().starts_with('/') {
```
Byron, line 60: Keeping a function pointer is nice! I should remember that.

Byron (changes requested):
Thanks a million! SSH support is definitely under-tested and your work helps tremendously to get it where it needs to be.

Could you restructure the commits (now and in future) so that crates aren't mixed? Since conventional commits are used here, these will end up in the crate's changelogs and these should be specific to the crate to make sense.

Thanks a lot, happy to merge once the commit-split-by-crate is done.

git-transport/src/client/blocking_io/ssh.rs
noirbizarre: Thanks !
I am rather new at rust but I can definitively feel some Python influence and some patterns like this one feel rather intuitive.

noirbizarre:
PR updated, commit are separated by package.
I added a quick test for SSH default port and also a quick repository clone example.

Byron:
Thanks a lot for your help! 

When trying to run the SSH example I noticed that with my setup, it still doesn't work due to permission-denied errors even though it clones fine with `git clone` and the same scp-like URL. I will look into this later for sure as that shouldn't happen. It's also clear that without this PR, `gix` isn't able to clone at all due to mishandling these URLS, so it's already a huge step towards properly supporting SSH.

On top of that I noticed that the `clone` example suffers from not having the ability to show progress easily, like `gix` can, and it's probably something that could be provided in some shape or form. The `prodash` crate could make it easy to get a simple progress bar going, which could then be used in the examples.

JacobCoffee:
I was wondering if this was resolved? I'm running into a similar issue where `git clone` works fine but the gix version using the same scp-like URL is throwing the below.

Maybe I'm doing something dumb, though?

```sh
Repository not found
The requested repository does not exist, or you do not have permission to access it.

Caused by:
    failed to fill whole buffer
```

Byron:
It really depends on which version of `gix` you are using - it moves fast. Can you try with the version from `main`? It should be something like `cargo run --bin gix -- clone git@example.org/repo`.

PR #672, opened 2022-12-23
build(rust-toolchain): fix rust-toolchain.toml extension and format

noirbizarre:
This PR fixes the `rust-toolchain.yaml` file having a typo in the extension (`tml` instead of `toml`) and a format not handled by `Cargo` nor `rustup`: 
`channel` was `stable-1.65` and should be either `stable` or `1.65`, I choose to stick with `1.65` for the same Clippy issue than CI, see:
- https://github.com/Byron/gitoxide/blob/main/.github/workflows/ci.yml#L123
- 680c1436d0a8cb1d2f602369120c5bc0e36815b1

CI jobs command with a specific toolchain (Windows jobs) have been updated to force the toolchain from command line with `cargo +<toolchain>` syntax.

This fixes the `make tests` command for local development (aka. `clippy` more precisely)

----

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,9 +110,9 @@ jobs:
       - name: "Install prerequisites"
         run: vcpkg install  zlib:x64-windows-static-md
       - name: "Installation from crates.io: gitoxide"
-        run: cargo install --target ${{ matrix.target }} --target-dir install-artifacts --debug --force gitoxide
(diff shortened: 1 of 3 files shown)


Byron:
Thanks a million!

The `rust-toolchain` file was never working, and I didn't take the time to figure out its syntax and have immediately forgotten about it once CI worked again. So good to see this addressed and to have a version of a working `rust-toolchain` file, always better as `toml` as well :).

PR #673, opened 2022-12-24
fix: remove panic-assertions in `loose` `lookup_prefix`

davidkna:
(Potentially partial) fix for starship/starship#4755

The release assertions in `git-odb/src/store_impls/loose/find.rs` could lead to panics in corrupted git repositories. I replaced them with explicit error variants.

----

### DIFF
--- a/git-odb/src/store_impls/loose/find.rs
+++ b/git-odb/src/store_impls/loose/find.rs
@@ -4,15 +4,23 @@ use git_features::zlib;
 
 use crate::store_impls::loose::{hash_path, Store, HEADER_MAX_SIZE};
 
+#[derive(thiserror::Error, Debug)]
+#[allow(missing_docs)]
+pub enum InflateError {
+    #[error(transparent)]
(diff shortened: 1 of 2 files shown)


Byron:
Thanks a lot for contributing this fix 🎁! I see that every assertion will eventually be hit and it's probably worth it to proactively remove them at some point.

The `git-path` crate is full of intentional panics that happen if there is illformed UTF-8 coming from git repositories, and I am pleasantly surprised that `starship` didn't seem to have encountered them yet. Maybe by now, they are truly rare or even extinct.

Having had a look at the linked issue in `starship`, I hope that with these changes the file path will be revealed, and they could just make that one file available for integration into the test suite.

Byron:
Closing as the relevant portion of the commit is in `main` (_I altered it to only contain the changes needed in `git-odb`, removing those for `git-repository`. For future reference, one commit per crate would be the way to go if conventional commit messages are used, those affect the changelog of any crate the commit touches_.)

Issue #679, opened 2022-12-30
RFC2822 and DEFAULT date friction


jpgrayson:
### Duplicates

- [X] I have searched the existing issues

### Current behavior 😯

The git-date crate defines several well-known formats, including `git_date::time::format:RFC2822` and `git_date::time::format::DEFAULT`. These format descriptions are used both to parse and format `git_date::Time` objects. The problem is with the treatment of the day of month and specifically what happens with single-digit days of the month.

In the format descriptions, `[day]` is used. This is equivalent to `[day padding:zero]` ([reference](https://time-rs.github.io/book/api/format-description.html)).

When formatting, a zero-padded output is produced, e.g. `Sun, 01 Jan 2023 12:34:56 -0300` for `RFC2822` and `Sun Jan 01 12:34:56 -0300`.

### First problem

This is inconsistent with how `git` produces rfc2822 dates. For example, from a gitoxide repo:

```console
$ git log -n1 --pretty=%aD 21482848
Fri, 9 Dec 2022 19:31:38 -0600
#    ^-- single digit
```

So while [RFC 2822 itself](https://www.rfc-editor.org/rfc/rfc2822#section-3.3) calls for a "2DIGIT" day, `git` outputs a single digit when day-of-month < 10. Thus gitoxide will produce different rfc2822 date strings than `git` when using `git_date::time::format::RFC2822`.

### Second problem

When parsing dates, `git_date::time::format::RFC2822` rigidly _requires_ a two-digit day of month. Thus gitoxide will not parse rfc2822 dates originating from various `git` outputs. 

### Third problem

gitoxide's `git_date::time::format::DEFAULT` format is different than `git`'s default date output. E.g.:

`Fri Dec 09 2022 19:31:38 -0600`
vs.
`Fri Dec 9 19:31:38 2022 -0600`

I.e. the time and year components are swapped along with `git` using a non-padded day-of-month and gitoxide using a zero-padded day-of-month.

While the ordering of `git`'s default date is...unique, it is also ubiquitous. So it is a problem that `git_date::parse()` will not parse the most common date format output by `git`.

### Problem 1

It is not obvious whether gitoxide should output the same non-padded day of month that `git` outputs since `git` may not be 100% compliant with RFC 2822. Nonetheless it is somewhat surprising that gitoxide's rfc2822 output differs from `git`. This could be resolved by offering an alternative format description, e.g. `GIT_RFC2822` that produces the non-padded day-of-month like `git`.

```rust
[...]

PR #680, opened 2022-12-30
Issue 679

jpgrayson:
The commits in this PR collectively fix the git-date problems identified in #679.

----

### DIFF
--- a/git-date/src/parse.rs
+++ b/git-date/src/parse.rs
@@ -14,12 +14,12 @@ pub enum Error {
 pub(crate) mod function {
     use std::{convert::TryInto, str::FromStr, time::SystemTime};
 
-    use time::{Date, OffsetDateTime};
+    use time::{format_description::well_known, Date, OffsetDateTime};
 
     use crate::{
         parse::{relative, Error},
         time::{
-            format::{DEFAULT, ISO8601, ISO8601_STRICT, RFC2822, SHORT},
+            format::{DEFAULT, ISO8601, ISO8601_STRICT, SHORT},
             Sign,
         },
         Time,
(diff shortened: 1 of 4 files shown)


Byron:
Thanks so much, fantastic work, even if it wasn't accompanied with the issue describing it all in detail, thanks to the descriptive commit messages that I wish I would produce by default.

I particularly like the way the changes were done without breaking the existing API. One question I don't really have an answer for is if it should indeed be `GIT_DEFAULT` or `GIT_RFC…` or instead `GITOXIDE_DEFAULT` and `STANDARD_RFC…` so `RFC…` so the defaults would always be what git uses despite not being what `gitoxide` wants or what the standard wants.

After all, the crate is called `git-date` and one might expect it to implement things similarly to git by default.

I am happy to leave it as is, but if it seems like a worthy change to you, I would definitely welcome a PR to make these breaking name changes.

jpgrayson:
> One question I don't really have an answer for is if it should indeed be GIT_DEFAULT or GIT_RFC… or instead GITOXIDE_DEFAULT and STANDARD_RFC… so RFC… so the defaults would always be what git uses despite not being what gitoxide wants or what the standard wants.
> 
> After all, the crate is called git-date and one might expect it to implement things similarly to git by default.
> 
> I am happy to leave it as is, but if it seems like a worthy change to you, I would definitely welcome a PR to make these breaking name changes.

I think `GIT_RFC2822` and `RFC2822` is a good dichotomy. It correctly implies that there is an RFC-compliant format along with a format for what `git` does that is similar to the RFC.

For `DEFAULT`, you make a good point about "git-date" having "git" in the name and thus it should perhaps conform to `git`'s default date format. I note that `DEFAULT` is only used in tests and the `parse()` function within gitoxide, so it sort of exists as a new suggested date format for gitoxide users. To that end, changing `DEFAULT` to `GITOXIDE` and `GIT_DEFAULT` to `DEFAULT` would make sense to me.

Byron:
Thanks for your feedback, that's all I needed to take action :). I have renamed the formats according to your suggestion.

PR #682, opened 2022-12-31
fix: Stricter raw date parsing

jpgrayson:
The raw date parser (git_date::parse::function::parse_raw()) accepted some inputs that it should not have. Specifically, it would accept:

- Any character for the timezone offset's sign
- Trailing, non-whitespace characters after the timezone offset

Now either '+' or '-' is required for the timezone offset sign and only trailing whitespace is allowed.

Additional tests are added to cover both acceptable and unacceptable inputs.

N.B. the raw date parser is still accepting of whitespace leading, in the middle of, and trailing the date string. A yet stricter parser would only allow a single space character between the seconds-since-epoch and the timezone offset.

----

### DIFF
--- a/git-date/src/parse.rs
+++ b/git-date/src/parse.rs
@@ -71,10 +71,14 @@ pub(crate) mod function {
         let mut split = input.split_whitespace();
         let seconds_since_unix_epoch: u32 = split.next()?.parse().ok()?;
         let offset = split.next()?;
-        if offset.len() != 5 {
+        if offset.len() != 5 || split.next().is_some() {
             return None;
         }
-        let sign = if &offset[..1] == "-" { Sign::Minus } else { Sign::Plus };
+        let sign = match &offset[..1] {
+            "-" => Some(Sign::Minus),
+            "+" => Some(Sign::Plus),
+            _ => None,
+        }?;
         let hours: i32 = offset[1..3].parse().ok()?;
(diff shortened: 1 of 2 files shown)


Byron:
Thanks a million for your continued effort!

> N.B. the raw date parser is still accepting of whitespace leading, in the middle of, and trailing the date string. A yet stricter parser would only allow a single space character between the seconds-since-epoch and the timezone offset.

Would you be interested in adding more [baseline tests](https://github.com/Byron/gitoxide/blob/9f95f7fbbe5b56e65c00c26f580bf67a4001e146/git-date/tests/fixtures/generate_git_date_baseline.sh#L41-L40) to compare ourselves with `git`? That way we would know what we have to support, and can opt in to being less strict if we think this has advantages, and do so consciously.

Byron (approved):
Thanks a million for your continued effort!

> N.B. the raw date parser is still accepting of whitespace leading, in the middle of, and trailing the date string. A yet stricter parser would only allow a single space character between the seconds-since-epoch and the timezone offset.

Would you be interested in adding more [baseline tests](https://github.com/Byron/gitoxide/blob/9f95f7fbbe5b56e65c00c26f580bf67a4001e146/git-date/tests/fixtures/generate_git_date_baseline.sh#L41-L40) to compare ourselves with `git`? That way we would know what we have to support, and can opt in to being less strict if we think this has advantages, and do so consciously.

PR #684, opened 2023-01-03
gix clone improvements

### LINKED ISSUE #683: ux: clone: make `<DIRECTORY>` optional
### Summary 💡

`clone` command requires `<DIRECTORY>` argument, which I think it should be optional instead: `[<DIRECTORY>]`:

git way:
```
git clone [OPTIONS] <repository> [<directory>]
```

gix way:
```
gix clone [OPTIONS] <REMOTE> <DIRECTORY>
```

### Motivation 🔦

`git` by default uses the project name when creating folders. Whereas `gix`, we have to pass manually, which makes the UX a bit slower. (since we have to copy-paste the repository name)

The following command should just work:
```bash
$ gix clone git@github.com:Byron/gitoxide.git
error: The following required arguments were not provided:
    <DIRECTORY>
```

We can easily get the `<DIRECTORY>` from the repository name, which is `gitoxide` in this case.

Byron:
Fixes #683

### DIFF
--- a/git-transport/src/client/blocking_io/file.rs
+++ b/git-transport/src/client/blocking_io/file.rs
@@ -138,7 +138,23 @@ fn supervise_stderr(
                 Some(err) => Err(err),
                 None => match res {
                     Ok(n) if n == wanted => Ok(n),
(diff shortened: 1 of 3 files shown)


Byron:
Closing as the (rewritten commits) where merged to main.

PR #685, opened 2023-01-03
Signature resolution

jpgrayson:
This fixes the root cause of https://github.com/stacked-git/stgit/issues/256.

----

### DIFF
--- a/git-repository/tests/repository/config/identity.rs
+++ b/git-repository/tests/repository/config/identity.rs
@@ -128,3 +128,68 @@ fn author_and_committer_and_fallback() {
         }
     }
 }
+
+#[test]
+#[serial]
+fn author_from_different_config_sections() {
+    let repo = named_repo("make_signatures_repo.sh").unwrap();
+    let work_dir = repo.work_dir().unwrap().canonicalize().unwrap();
+
+    let _env = Env::new()
+        .set(
+            "GIT_CONFIG_GLOBAL",
+            work_dir.join("global.config").display().to_string(),
+        )
+        .set(
+            "GIT_CONFIG_SYSTEM",
+            work_dir.join("system.config").display().to_string(),
+        )
+        .set("GIT_AUTHOR_DATE", "1979-02-26 18:30:00")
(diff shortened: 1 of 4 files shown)


git-repository/tests/repository/config/identity.rs
```diff
@@ -128,3 +128,68 @@ fn author_and_committer_and_fallback() {
...
+        repo.author(),
+        Some(git_actor::SignatureRef {
+            name: "global name".into(),
+            email: "local@example.com".into(),
+            time: git_date::Time {
+                seconds_since_unix_epoch: 42,
+                offset_in_seconds: 1800,
+                sign: git_date::time::Sign::Plus
+            }
+        }),
```
Byron, line 177: In future, what I'd love to see is to add an`"explainer"` parameter as last argument to `assert*!()` to make clearer why this has to be the right value.

git-repository/tests/repository/config/identity.rs
```diff
@@ -128,3 +128,68 @@ fn author_and_committer_and_fallback() {
...
+            name: "local committer".into(),
+            email: "global-committer@example.com".into(),
+            time: git_date::Time {
+                seconds_since_unix_epoch: 320437800,
+                offset_in_seconds: 0,
+                sign: git_date::time::Sign::Plus,
+            }
+        })
+    );
+    assert_eq!(
```
Byron, line 191: Referring to my comment above, I can only assume why a seemingly unrelated method is tested, and an `"explainer"` would certainly help to make it clear.

Byron (approved):
My apologies for the late reply, but after looking through my emails it's clear I didn't get notified. Hopefully, that was just a GitHub issue that day and that it won't repeat.

Thanks for the fix, it's much appreciated! I will merge it momentarily and hope that maybe we can try GitHubs notifications by getting another PR with `"explainers"` for why the assertion is what it is.

Can you let me know when you want a new release with this fix? I can make one as you need it, but am happy to wait to get more fixes and improvements in, in the mean time.

PR #689, opened 2023-01-09
Fix default author and committer times

jpgrayson:
The commit message for a05b1c4d82bc6c7758989a3bbe326ea610903820 tells the whole story.

One aside, however, is that it is not apparent to me how much utility `Repository::user_default()` has. This public method was only used in `author_or_default()` and `committer_or_default()` and is now not used at all within gitoxide. I cannot think of any circumstances where a user of the git-repository crate would ever want to call `user_default()`. But the presence of `user_default()` could cause people to miss `author()` or `committer()`. Point being: consider whether to remove `user_default()` from `Repository`.

Second aside: the static default user name and email of "gitoxide" and "gitoxide@localhost" differ from git's fallbacks. If no user name is configured, it then interrogates the OS via `getpwuid()` before finally falling back to "Unknown". And in the case of no user email being configured, it interrogates the OS for the hostname via `gethostname()` and joins the previously discovered username or "unknown" with the hostname which might thus ultimately fall back to "unknown@localhost". I don't think gitoxide should be doing the same dance as git for divining user and email identity because it ends up being platform dependent, additional crate dependencies, and, even with the effort, almost certainly the wrong user identity. But the static "gitoxide"/"gitoxide@localhost" fallbacks are also always wrong. My argument here is to also consider removing the `author_or_default()`** and `committer_or_default()` methods from `Repository`. So just have `author()` and `committer()` and leave it the application to decide what to do if `author()` or `committer()` fail.

For StGit, I am doing what git also ultimately does if there is no identity configured: fail, showing the user an error message encouraging them to configure `user.name` and `user.email`.

----

### DIFF
--- a/git-config/tests/file/init/from_paths/includes/conditional/mod.rs
+++ b/git-config/tests/file/init/from_paths/includes/conditional/mod.rs
@@ -145,7 +145,7 @@ fn git_init(path: impl AsRef<std::path::Path>, bare: bool) -> crate::Result<git:
         bare.then(|| git::create::Kind::Bare)
             .unwrap_or(git::create::Kind::WithWorktree),
         git::create::Options::default(),
-        git::open::Options::isolated(),
+        git::open::Options::isolated().config_overrides(["user.name=gitoxide", "user.email=gitoxide@localhost"]),
     )?
     .to_thread_local())
 }
--- a/git-config/tests/file/init/from_paths/includes/conditional/onbranch.rs
+++ b/git-config/tests/file/init/from_paths/includes/conditional/onbranch.rs
@@ -287,7 +287,7 @@ value = branch-override-by-include
             git::lock::acquire::Fail::Immediately,
             git::lock::acquire::Fail::Immediately,
         )?
-        .commit(repo.committer_or_default())?;
+        .commit(repo.committer().expect("pre-configured"))?;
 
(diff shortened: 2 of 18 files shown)


jpgrayson:
I don't have any more gitoxide patches in my queue after this one. If a new git-repository release was made after this change is merged, that would allow me to close out a downstream StGit issue, which would be great.

Byron (approved):
Thank you! Thank you so much for making me review the identity API.

Sometimes one gets too used to the status quo and without challenging it, there is no improvement. It turns out that this API existed primarily for *my* convenience under the guise of `git` also having default values for the user.

Thanks to your intervention it became clear that for a library, even one that wants to be like `git`, it's a mistake to have a default. It would also be a mistake to reproduce `git`  guessing of a user as it's like you correctly state almost always *not* what one wants.

Thus I have removed all defaults and handle it as errors accordingly where these were used. Back then, having defaults made sense as it was hard to configure the user, but a lot has changed and now it's as easy as it should be.

I'd call this a wonderful addition for the next release. Please keep these suggestions coming, they are invaluable!

Byron:
A new release of `git-repository` is now available [here](https://github.com/Byron/gitoxide/releases/tag/git-repository-v0.32.0).

PR #690, opened 2023-01-10
Export `PostBodyDataKind`

paulyoung:
Unless I'm mistaken, without this change there's no way to implement the `Http` trait because `PostBodyDataKind` is effectively still private.

----


paulyoung:
Some additional context;

I've been depending on some `gitoxide` crates using `{ git = "...", rev = "..." }` in my `Cargo.toml` for a while. This is partly because I was using unreleased features and partly because it's simpler than figuring out compatible version numbers when depending on multiple `gitoxide` crates.

I'm unable to continue doing that in the short term because I set up CI using GitHub actions and it fails to clone those dependencies due to some problem with LFS files in the repo.

I can apparently work around this issue by depending on versions that have been published to crates.io instead. Unfortunately I ran into this issue with `PostBodyDataKind` when trying to do that.

If this is merged I'd appreciate a published release of any affected crates for the above reason. I'm depending on `git-features`, `git-repository`, and `git-validate`.

Byron:
That is definitely an oversight, thanks a lot for the fix. I took this PR as incentive to upgrade all dependencies to their latest versions as well, and `git-repository` is now available incorporating all of these at v0.33.0.

> I'm unable to continue doing that in the short term because I set up CI using GitHub actions and it fails to clone those dependencies due to some problem with LFS files in the repo.

That's unfortunate and I hope that this doesn't impede your ability to try out new things in your own fork. I'd hope there is a solution for it as well but wouldn't know what to look for just yet.

paulyoung:
Thanks!

PR #697, opened 2023-01-22
adjustments for cargo

Byron:
* [x] `http.extraHeader` tests
* [x] `gix progress` => `gix config-tree` 
* [x] `init.templateDir` is available as a key (even thought it's still planned to be implemented (need to disable it in the by setting it to "") - not planned anymore.
* [x] assure all keys in [the env-to-config function](https://github.com/Byron/gitoxide/blob/8737601d124f69fca4df85d160da024a1475e5f8/git-repository/src/config/cache/init.rs#L265) are the ones from the tree.

### DIFF
--- a/src/plumbing/progress.rs
+++ b/src/plumbing/progress.rs
@@ -2,6 +2,7 @@ use std::borrow::Cow;
 use std::fmt::{Display, Formatter};
 
 use crosstermion::crossterm::style::Stylize;
+use git_repository as git;
 use owo_colors::OwoColorize;
 use tabled::{Style, TableIteratorExt, Tabled};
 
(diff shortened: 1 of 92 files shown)


timscharfenort11 (commented):
Super

src/plumbing/progress.rs
```diff
@@ -14,10 +15,7 @@ enum Usage {
     /// We definitely want to implement this configuration value.
     Planned { note: Option<&'static str> },
     /// The configuration is already effective and used (at least) in the given module `name`.
```
fbstj, line 17: this comment is now wrong?

Byron: Great catch! Indeed it was out of date, this should be fixed next time I push.

timscharfenort11 (approved):

PR #699, opened 2023-01-26
Break cyclical dev-dependencies

pascalkuthe:
This PR has contains functional changes.
Instead, the point of this PR is to make gitoxide work with rust-analyzer.
Currently rust-analyzer is unable to resolve most gitoxide crates because they have
cyclical dev dependencies. That is crate `A` depends on crate `B` but crate `B`
has a `dev-dependency` on crate `A`.

Rust analyzer can't deal with repos like that because it treat `dev-dependencies` like normal
dependencies. Changing that would be very complex and the corresponding RA issue has been
open since 2019. The siutation is quite unofortunate but seeing that `rust-analyzer` is the
only offical rust ide and quite widespread it would be nice to support developing gitoxide with it.

Many cyclical dev-dependencies were just caused by gitoxide using itself as tooling for tests (like `testtools`).
These cases were easy to solve by pointing the dependencies at `crates.io` rather than local paths.

However, some crates like `git-pack`, `git-traverse` and `git-diff` require access to a repository or odb.
Breaking those dependencies by replacing them with crates.io dependencies does not work because
the crates.io dependencies are incompatible with the local crates being test. Instead,
I move these tests up in the dependency chain to `git-odb` and `git-repository` respectively.
In a sense these tests are integration tests, that also tests these crates, so it seemed fitting to move them here.

The migration was mostly painful, but there is one test that is failing.
I can't quite figure out why. Maybe you could give me a pointer @Byron.

----


Byron (changes requested):
Thanks a lot for going through the trouble! I'd love for RA to work as well.

Once `git-lfs` resources have been rewritten to remain `git-lfs` resources, I think I can put some work on top and get it merged.

Here is the list of files that were flagged:

```
Encountered 6 files that should have been pointers, but weren't:
        git-odb/tests/fixtures/generated-archives/make_diff_repo.tar.xz
        git-odb/tests/fixtures/generated-archives/make_pack_gen_repo.tar.xz
        git-odb/tests/fixtures/generated-archives/make_traversal_repo_for_commits.tar.xz
        git-odb/tests/fixtures/generated-archives/make_traversal_repo_for_commits_with_dates.tar.xz
        git-odb/tests/fixtures/generated-archives/make_traversal_repo_for_trees.tar.xz
        git-repository/tests/fixtures/generated-archives/make_repo_with_branches.tar.xz
```

There is a few things I would change:

* find a way to put integration tests into their own Cargo project to resolve the dependency cycle. I have seen this once or twice with examples, and hope it can work for integration tests as well in some shape or form.
* `hex_to_id` doesn't look good for public APIs in a plumbing crate and is strictly for testing. If one crate can't use the one in `git-testtools`, then it should be re-defined there. I am hoping that many crates can keep using the shared version.

pascalkuthe:
Closed in favor of #705

PR #705, opened 2023-01-30
Break cyclical dev dependencies

pascalkuthe:
Supercedes #699

This PR breaks cyclical dev-dependencies.
These cause problems with rust analyzer which makes gitoxide development with RA basically impossible.
git-testools is mostly used as a utility we can simply use the crates.io version of the respective gitoxide crates to circumvent that problem. However, most other crates require matching crate versions. Here the integration tests were simply moved to a subcrate in the same folder by adding a Cargo.toml file in the tests folder.

The hex_to_id function in git-testtools is a specical case because that function returns a git_hash::ObjectId instance. This ObjectId would belong to the crates.io version of git-hash and can not be used in tests. To work around that hex_to_id is moved into the individual tests as it's a very small one-line function.

----

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1192,8 +1192,9 @@ dependencies = [
  "bstr",
  "btoi",
  "document-features",
- "git-date",
- "git-features",
+ "git-date 0.4.1",
+ "git-features 0.26.1",
+ "git-hash 0.10.1",
  "git-testtools",
  "itoa",
  "nom",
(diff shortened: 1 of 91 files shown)


Byron (commented):

As it's a lot of changes I try to be as diligent as possible and approach it from a "it should still run all the tests" kind of perspective.

`make nextest` on `main` seems to run  1565 tests, on this branch it appears to be only 1440. A first check seems to indicate that `git-diff-tests` aren't picked up, but some others are. I am out of time now but will look into this later.

Byron (commented):
The issue with the missing tests in `cargo nextest` runs was found quickly, and was due to some single-threaded tests not being defined implicitly anymore. With that fixed, the amount of tests executed is the same once again.

Something else I noticed is that these 'internal-testing-feature-toggles' now moved into the dedicated testing crate, cleaning up the parent crate considerably. It's lovely that the tests just sit in the same place as before as well, and the merge-conflicts I expected when merging in main just didn't happen because of it.

Also I learned about the `auto*` manifest keys, to turn automatic test discovery off for the crates that now have a different layout. It's amazing how flexible `cargo` is.

Just one more read through all the lines here and I am ready.

PR #708, opened 2023-01-30
fix(git-refspec): Fixes build for parsing fuzz harness

silvergasp:
Fixes a build failure on the fuzz harness for git-refspec

----


Byron:
Thanks a lot for the fix!

That's bitrot right there, as I don't build the fuzz targets in CI (yet). Good this changes soon.

PR #709, opened 2023-01-31
feat(ci): Adds continuous fuzzing

silvergasp:
This is intended to catch shallow bugs before
they are merged into the repository. To do this
fuzzing will be run for ~10min. This is not
intended to be a complete fuzzing suite. For
more information on the complete fuzzing system
see the integration of gitoxide into
google/oss-fuzz.

----

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,24 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+ Fuzzing:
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a million!

I admit to not fully grasp how this will look like when active, but there is only one way to find out I suppose :D.

silvergasp:
No worries. If you run into any issues feel free to revert this change, and drop me a mention :)

PR #713, opened 2023-01-31
fix(git-transport): Use single quotes for ssh path arg

exactly-one-kas:
Git uses this method of quoting args for SSH transport too Some Git SSH servers require this method to be used (eg. BitBucket)

I have no idea how to test for the quoted string actually being used in `handshake()` without ripping the method's guts out. Do you have any pointers for me?

----


git-transport/src/client/blocking_io/file.rs
```diff
@@ -216,7 +231,8 @@ impl client::Transport for SpawnProcessOnDemand {
         if self.ssh_cmd.is_some() {
             cmd.args.push(service.as_str().into());
         }
-        cmd.args.push(self.path.to_os_str_lossy().into_owned());
+        cmd.args
+            .push(to_single_quoted(self.path.as_ref()).to_os_str_lossy().into_owned());
```
Byron, line 235: `SpawnProcessOnDemand` isn't only used by `ssh`, but also when spawning processes locally.
This transformation should only happen when `ssh_cmd` is set.

git-transport/src/client/blocking_io/file.rs
```diff
@@ -282,5 +298,18 @@ mod tests {
                 }
             }
         }
+
+        mod handshake {
+            use crate::client::file::to_single_quoted;
+
+            #[test]
+            fn quoted_strings() {
```
Byron, line 306: I think a couple of tests are missing to [implement this behaviour equivalently](https://github.com/git/git/blob/5dec958dcf965fc75e0f459f8e8ccf9c9f495b15/quote.c#L12-L44). CI seems to fail due to it as well.

Please also move it to [this module](https://github.com/Byron/gitoxide/blob/f7f136dbe4f86e7dee1d54835c420ec07c96cd78/git-quote/src/lib.rs#L6-L5) - in there it's easier to test as well. Try to break it with your tests :).

Byron (changes requested):
Thanks a lot, indeed [`git` quotes the path when using ssh](https://github.com/git/git/blob/5dec958dcf965fc75e0f459f8e8ccf9c9f495b15/connect.c#L1451).

Before I forget, please put changes to `git-quote` into their own commit, ideally prefixing it with `feat: …` mentioning the new quoting mechanism. That way the changelog is written with the note automatically.

Thanks again for your help.

Byron:
Thanks a lot.

I have a feeling the tests were written to reflect the function as it was already implemented, instead of what's [described in the git implementation](https://github.com/git/git/blob/2fc9e9ca3c7505bc60069f11e7ef09b1aeeee473/quote.c#L18-L24), as there was plenty to correct. Furthermore it's advisable to follow the structure of the crate closely to keep things as consistent as possible.

exactly-one-kas:
Ah, I didn't think of using the documented examples as tests
However, your implementation isn't correct - single quotes work somewhat like raw strings in rust - `r"a\"b"` wouldn't compile either since escapes get ignored. It would need to be akin to `String::new() + r"a" + "\"" + r"b"`, hence it becomes `'a'\''b'` for shells
image

Byron:
Thanks for pointing this out!

The function is still under-tested and it's probably one of these cases where it has to follow its C counterpart precisely, fortunate it's small. In theory it should be enough to implement the `need_bs_quote()` as well.

PR #714, opened 2023-02-01
feat(git-date): Adds fuzzer for date parser

silvergasp:
This fuzzer almost immediately crashes the parser.
----

### DIFF
--- a/git-date/fuzz/.gitignore
+++ b/git-date/fuzz/.gitignore
@@ -1,0 +1,4 @@
+target
+corpus
+artifacts
+coverage
--- a/git-date/fuzz/Cargo.toml
+++ b/git-date/fuzz/Cargo.toml
@@ -1,0 +1,27 @@
+[package]
+name = "git-date-fuzz"
+version = "0.0.0"
+publish = false
+edition = "2021"
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = "0.4"
+
+[dependencies.git-date]
+path = ".."
+
+# Prevent this from interfering with workspaces
+[workspace]
+members = ["."]
+
+[profile.release]
+debug = 1
+
+[[bin]]
+name = "parse"
+path = "fuzz_targets/parse.rs"
+test = false
+doc = false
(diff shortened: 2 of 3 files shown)


Byron:
We have to be careful with this as there is limited control over the quality of the parser. Most of it is implemented in the `time` crate, which is also where the first immediate crash is happening.

Can we avoid throwing this into `OSS-FUZZ` while it's clear there will be a lot of crashes, most of which are outside of our control?

Byron:
In any case, no way to hide, let's bring it in and deal with the fallout, maybe some crashes can be triggered within `gitoxide` itself, and from there we can see if fuzzing can be extended to `time` as needed.

Byron:
@silvergasp Unrelated to this PR, I just got informed that [the builds on OSS-FUZZ](https://oss-fuzz-build-logs.storage.googleapis.com/log-4b532f9e-3180-473c-a4be-d9e6017f3ba7.txt) fail continuously. I don't know if this was a temporary failure, or what to do about it. Apparently the coverage report (python script) is failing as it can't find source code files of the standard library.

silvergasp:
Hmm, I imagine we need to add the rust-src component using rustup.

`rustup component add rust-src`

I'm not sure if that needs to be in the dockerfile or in build.sh

I'll investigate tomorrow.

silvergasp:
It doesn't look like the build is failing. There is a post-processing stage after collecting code coverage that looks like it's failing. That stage, in particular, will need the original source code that made the binary. But if the rust-src component isn't installed then it won't be able to produce a code coverage report. By default rust is going to use a precompiled std library.

Byron:
Seems fair. What surprised me is that this happened now, and not immediately after setting up the fuzzing.

silvergasp:
> Seems fair, thanks for the analysis. What surprised me is that this happened now, and not immediately after setting up the fuzzing.

Yeah, I think code-coverage is calculated every other day or so. So the builds are working fine, so there were no immediate red flags there. Although I'm surprised this wasn't caught in the Pull Request CI for oss-fuzz.

silvergasp:
Moved this conversation to #719 as it doesn't seem to be related to this PR. I've also added a potential fix for this problem here https://github.com/google/oss-fuzz/pull/9581/

PR #717, opened 2023-02-02
fix single-quotes

Byron:
Thanks @exactly-one-kas, can you take another look?

These test cases have too many quotes in them for me to still be able to tell
if it's exactly what's needed. Maybe there are other, more realistic inputs that
would show that in practice, it does the right thing?

### DIFF
--- a/git-quote/src/single.rs
+++ b/git-quote/src/single.rs
@@ -1,4 +1,4 @@
-use bstr::{BStr, BString, ByteSlice};
+use bstr::{BStr, BString, ByteSlice, ByteVec};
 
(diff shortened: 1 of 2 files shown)


Byron:
I am merging this as it seemingly is more correct than before, but invite you to post comments here or submit a PR with fixes as you see fit. I am particularly interested in more sensible test-cases to show its meaning in the real-world a little better.

exactly-one-kas:
Since this is mostly used for folder paths there won't be many real-world examples other than stuff like `~miles.o'brien/inversed-polarity`
This functionality is mostly about keeping safe from malicious inputs like `'; rm -rf / #`

PR #718, opened 2023-02-02
fix(git-discover): split `GIT_CEILING_DIRECTORIES` by semicolon on Windows

Nerixyz:
Git splits `GIT_CEILING_DIRECTORIES` by `PATH_SEP` ([source](https://github.com/git/git/blob/2fc9e9ca3c7505bc60069f11e7ef09b1aeeee473/setup.c#L1260)). `PATH_SEP` is `;` on Windows (`:` anywhere else). 

This PR adds a check for the platform and uses `;` on Windows.

----

### DIFF
--- a/git-discover/src/upwards/types.rs
+++ b/git-discover/src/upwards/types.rs
@@ -1,6 +1,5 @@
-use std::{borrow::Cow, env, path::PathBuf};
-
-use bstr::{ByteSlice, ByteVec};
+use std::ffi::OsStr;
+use std::{env, path::PathBuf};
 
 /// The error returned by [git_discover::upwards()][crate::upwards()].
 #[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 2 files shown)


git-discover/src/upwards/types.rs
```diff
@@ -156,4 +158,42 @@ mod tests {
...
+
+        assert_eq!(ceiling_dirs.len(), 2, "Relative path is discarded");
+        assert_eq!(
+            ceiling_dirs[0].to_string_lossy(),
+            symlink_path
+                .canonicalize()
+                .expect("symlink path exists")
+                .to_string_lossy()
+                .strip_prefix(r#"\\?\"#)
+                .expect("on Windows, the extended length path syntax is used"),
```
Nerixyz, line 189: I don't really like this, but [`std::fs::canonicalize`](https://doc.rust-lang.org/stable/std/fs/fn.canonicalize.html) returns a path prefixed with `\\?\`, which can't be removed by `PathBuf::strip_prefix`.

Byron:
Thanks a lot, a great catch!

`git-discover` was among the first crates and it shows. So I found yet another related refactor and once CI is green, this PR will be merged.

PR #721, opened 2023-02-05
Permit ceiling dirs not to match if `apply_environment` is used

davidkna:
If values for `ceiling_dirs` are pulled from the environment instead of explicit `Options` configuration, I feel that it makes sense to permit non-matching `ceiling_dir` the same way git does.

----


Byron:
Thanks a lot, that makes sense.

For a moment I was thinking about moving this logic to `git-repository`, but reminded myself that plumbing crates are allowed and encouraged to mimic `git` where it makes sense for them as long as this behaviour can be altered or overridden. And that's clearly the case here.

PR #722, opened 2023-02-06
adjustments for cargo

Byron:
### Tasks

* [x] add a perform_fetch(cb) -> Result<…, OneCombinedError> method which combines all possible errors of related fetch calls into one with .is_spurious(). That way the set of errors is contained and spuriousness is maintained in gitoxide. Should be 'repo-corrupted' and 'is-spurious'

### DIFF
--- a/git-repository/src/env.rs
+++ b/git-repository/src/env.rs
@@ -34,3 +34,96 @@ pub fn args_os() -> impl Iterator<Item = OsString> {
 pub fn os_str_to_bstring(input: &OsStr) -> Option<BString> {
     Vec::from_os_string(input.into()).map(Into::into).ok()
 }
+
+/// Utilities to collate errors of common operations into one error type.
+///
+/// This is useful as this type can present an API to answer common questions, like whether a network request seems to have failed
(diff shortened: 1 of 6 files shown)


git-repository/src/env.rs
```diff
@@ -34,3 +34,86 @@ pub fn args_os() -> impl Iterator<Item = OsString> {
...
+
+/// Utilities to collate errors of common operations into one error type.
+///
+/// This is useful as this type can present an API to answer common questions, like whether a network request seems to have failed
+/// spuriously or if the underlying repository seems to be corrupted.
+/// Error collation supports all operations, including opening the repository.
+///
+/// ### Usage
+///
+/// The caller may define a function that specifies the result type as `Result<T, git::env::collate::<operation>::Error>` to collect
```
fbstj, line 46:
```suggestion
/// The caller may define a function that specifies the result type as `Result<T, git::env::collate::{operation}::Error>` to collect
```
maybe? since `collate::<operatoion>` could be a turbofish?

Byron: That's a great point! It's changed (next time I push).

PR #723, opened 2023-02-06
fix: ignore ceiling directories equal to cwd

Nerixyz:
[The git documentation](https://git-scm.com/docs/git) says about the ceiling directory: "It will not exclude the current working directory...". Currently, if a ceiling directory is equal the cwd, the [`max_height`](https://github.com/Byron/gitoxide/blob/f774027fcbd777df2ac60215aa47a11fb0066e54/git-discover/src/upwards/mod.rs#L67) will be `Some(0)` making the [check against the `current_height`](https://github.com/Byron/gitoxide/blob/f774027fcbd777df2ac60215aa47a11fb0066e54/git-discover/src/upwards/mod.rs#L82) always fail, and thus no repository will be discovered.

The git client, however, will discover a repo:

Repo location: `C:\Users\User\AppData\Local\Temp\.tmpZZuL2L`, `GIT_CEILING_DIRECTORIES="C:\Users\User\AppData\Local\Temp\.tmpZZuL2L\level1\level2"`
```powershell
> C:\Users\User\AppData\Local\Temp\.tmpZZuL2L\level1\level2
╰─ git rev-parse HEAD
35dc38f922e4b6d6596cc7c5a0bb5caf00f1db64
```

The git client [uses `longest_ancestor_length`](https://github.com/git/git/blob/a6a323b31e2bcbac2518bddec71ea7ad558870eb/setup.c#L1263) to check the ceiling height. This returns -1 (no ceiling) in the case above.

<details>
<summary>Reproduction</summary>

```c
#include <stdio.h>
#include <string.h>

// from git source
struct string_list_item {
  char *string;
  void *util;
};

struct string_list {
  struct string_list_item *items;
  size_t nr;
  // addidional fields omitted
};

int longest_ancestor_length(const char *path, struct string_list *prefixes) {
  int i, max_len = -1;

  if (!strcmp(path, "/"))
    return -1;

  for (i = 0; i < prefixes->nr; i++) {
    const char *ceil = prefixes->items[i].string;
    int len = strlen(ceil);

    /*
     * For root directories (`/`, `C:/`, `//server/share/`)
     * adjust the length to exclude the trailing slash.
     */
    if (len > 0 && ceil[len - 1] == '/')
      len--;

    if (strncmp(path, ceil, len) || path[len] != '/' || !path[len + 1])
      continue; /* no match */

    if (len > max_len)
      max_len = len;
  }

  return max_len;
}

int main() {
  struct string_list_item it = {
      .string =
          "C:/Users/User/User/Local/Temp/.tmpZZuL2L/level1/level2"};
  struct string_list lst = {.items = &it, .nr = 1};
  // this will print -1 (no ceiling)
  printf("%d\n",
         longest_ancestor_length(
             "C:/Users/User/AppData/Local/Temp/.tmpZZuL2L/level1/level2",
             &lst));
  return 0;
}
``` 
</details>

This PR adjusts the behavior to match.

----

### DIFF
--- a/git-discover/tests/isolated.rs
+++ b/git-discover/tests/isolated.rs
@@ -52,10 +52,7 @@ fn upwards_with_relative_directories_and_optional_ceiling() -> git_testtools::Re
         .unwrap_err();
 
         assert!(
-            matches!(
-                err,
-                git_discover::upwards::Error::NoGitRepositoryWithinCeiling { ceiling_height: 1, .. }
-            ),
+            matches!(err, git_discover::upwards::Error::NoMatchingCeilingDir),
             "limiting the ceiling to the CWD cannot work as it's just an empty dir"
         );
     }
--- a/git-discover/tests/upwards/ceiling_dirs.rs
+++ b/git-discover/tests/upwards/ceiling_dirs.rs
@@ -32,6 +32,24 @@ fn git_dir_candidate_within_ceiling_allows_discovery() -> crate::Result {
 }
 
 #[test]
+fn git_dir_candidate_with_cwd_ceiling() -> crate::Result {
+    let work_dir = repo_path()?;
+    let dir = work_dir.join("some/very/deeply/nested/subdir");
+    let (repo_path, _trust) = git_discover::upwards_opts(
+        dir.clone(),
+        Options {
+            ceiling_dirs: vec![dir],
+            match_ceiling_dir_or_error: false,
+            ..Default::default()
+        },
+    )
+    .expect("ceiling dir should be skipped");
+    assert_repo_is_current_workdir(repo_path, &work_dir);
+
+    Ok(())
+}
+
+#[test]
 fn ceiling_dir_limits_are_respected_and_prevent_discovery() -> crate::Result {
     let work_dir = repo_path()?;
     let dir = work_dir.join("some/very/deeply/nested/subdir");
--- a/git-discover/tests/upwards/ceiling_dirs.rs
+++ b/git-discover/tests/upwards/ceiling_dirs.rs
@@ -32,9 +32,11 @@ fn git_dir_candidate_within_ceiling_allows_discovery() -> crate::Result {
 }
 
 #[test]
-fn git_dir_candidate_with_cwd_ceiling() -> crate::Result {
+fn git_dir_candidate_with_cwd_ceiling_ignores_ceiling_if_no_ceiling_dir_may_be_matched() -> crate::Result {
     let work_dir = repo_path()?;
     let dir = work_dir.join("some/very/deeply/nested/subdir");
+    // the ceiling dir is equal to the input dir, which doesn't contain a repository.
+    // But we can ignore that just like git does (see https://github.com/Byron/gitoxide/pull/723 for more information)
     let (repo_path, _trust) = git_discover::upwards_opts(
         dir.clone(),
         Options {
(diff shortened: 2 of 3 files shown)


git-discover/tests/isolated.rs
```diff
@@ -52,10 +52,7 @@ fn upwards_with_relative_directories_and_optional_ceiling() -> git_testtools::Re
         .unwrap_err();
 
         assert!(
-            matches!(
-                err,
-                git_discover::upwards::Error::NoGitRepositoryWithinCeiling { ceiling_height: 1, .. }
-            ),
+            matches!(err, git_discover::upwards::Error::NoMatchingCeilingDir),
```
Nerixyz, line 55: I'm not sure if the description ("limiting the ceiling to the CWD cannot work as it's just an empty dir") is still right.

Byron: I have added a test which might be related. But indeed, I don't fully understand what's going on here.

git-discover/tests/upwards/ceiling_dirs.rs
```diff
@@ -31,6 +31,47 @@ fn git_dir_candidate_within_ceiling_allows_discovery() -> crate::Result {
...
+            ceiling_dirs: vec![dir],
+            match_ceiling_dir_or_error: true,
+            ..Default::default()
+        },
+    )
+    .unwrap_err();
+
+    assert!(
+        matches!(err, git_discover::upwards::Error::NoMatchingCeilingDir),
+        "this is a bit strange as it should have matched this directory, and said it's not within the ceiling"
```
Byron, line 70: I think something is fishy here and this test indicates the issue. It claims there is no ceiling dir match, even though there is - the dir is exactly at the ceiling dir but it's like it doesn't exist.

git-discover/tests/upwards/ceiling_dirs.rs
```diff
@@ -31,6 +31,47 @@ fn git_dir_candidate_within_ceiling_allows_discovery() -> crate::Result {
...
+fn git_dir_candidate_with_cwd_ceiling_ignores_ceiling_if_no_ceiling_dir_may_be_matched() -> crate::Result {
+    let work_dir = repo_path()?;
+    let dir = work_dir.join("some/very/deeply/nested/subdir");
+    // the ceiling dir is equal to the input dir, which doesn't contain a repository.
+    // But we can ignore that just like git does (see https://github.com/Byron/gitoxide/pull/723 for more information)
+    let (repo_path, _trust) = git_discover::upwards_opts(
+        dir.clone(),
+        Options {
+            ceiling_dirs: vec![dir],
+            match_ceiling_dir_or_error: false,
```
Byron, line 44: Thanks to this this should work, and due to the special case it ignores matching ceiling dirs if they are equal to one of the ceiling dirs.

Git does it like this, yes, and `gitoxide` should do the same. I just try to make it work within the logic framework that exists and right now *something* seems off.

Byron (commented):
Thanks for your help with this.

I see the point, and added a few comments in the hopes we both can get to a shared understanding on what's going on and/or what's right. This PR can probably be merged as is but I can't shake the feeling that something strange is happening when reporting errors now. It claims there is no match, but there clearly is. Maybe the error case needs to be adjusted to be more specific to make sense in this case - like 'we did match a ceiling dir but because of *reasons* we searched passed it'.

git-discover/tests/upwards/ceiling_dirs.rs
Nerixyz: I think the idea is that "you're standing on the ceiling":

```
repo_dir/
╰╴some/
  ╰╴subdir/  ← cwd   _________🡗ceiling
    ╰╴subsubdir/  from here, we'd hit the ceiling
```

So, there's no ceiling above the current location to match/hit.

Byron:
Thanks for the clarification, I think I got it now! Please note that I am closing this PR as it's already in `main`, but GitHub can't detect that due to my meddling with my own commit.

CC @davidkna who is using `git::Repository::discover_with_environment_overrides_opts(…)` which now should behave correctly in the presence of ceiling directories, as 'standing on the ceiling dir' will get to ignore the ceiling directory entirely without considering this an error.

PR #724, opened 2023-02-08
docs: fix typos

Lioness100:
Fixed one or two typos 😉 (this was a trial run for [a CLI I made](https://github.com/Lioness100/rspell))

----


git-config/tests/file/access/read_only.rs
```diff
@@ -64,7 +64,7 @@ fn get_value_for_all_provided_values() -> crate::Result {
             &[cow_str("")],
         );
 
-        assert_eq!(config.string("doesnt", None, "exist"), None);
+        assert_eq!(config.string("doesn't", None, "exist"), None);
```
Byron, line 67: This is interesting as I didn't dare putting `'` into section names for query. But it's OK to do that and has the same effect, good to see.

git-odb/src/traits.rs
```diff
@@ -45,7 +45,7 @@ pub trait Find {
     /// Returns true if the object exists in the database.
     fn contains(&self, id: impl AsRef<git_hash::oid>) -> bool;
 
-    /// Find an object matching `id` in the database while placing its raw, undecoded data into `buffer`.
+    /// Find an object matching `id` in the database while placing its raw, encoded data into `buffer`.
```
Byron, line 48: This isn't correct - it should remain `unencoded` or something semantically similar, not opposite.

Byron (commented):
Thanks a lot, it's a pretty decent result and a lot of mistakes removed. Much appreciated.

I noticed that it doesn't shy away from fixing spelling mistakes in enumeration variants, which technically is a breaking change. All of these changes where in errors so I think it's OK as people are unlikely to actually match on these cases.

On another note, it was happy to change `unencoded` to `encoded`, inverting the meaning in the process. I will change that back.

git-odb/src/traits.rs
Lioness100: Hi, sorry about that, I think the confusion came from that it's originally "un*de*coded", not "un*en*coded". So I think either way there was a typo? Either way, since both undecoded and unencoded aren't actually words, I think it would be better to use their positive counterpoint, encoded and decided respectively.

Lioness100:
My pleasure!

Lioness100:
@Byron, did you see my response to your comment?

Lioness100:
No problem, glad I could help!

PR #727, opened 2023-02-09
fix(fuzz): Fixes code coverage for oss-fuzz build

silvergasp:
The oss-fuzz coverage collection system requires paths to be relative to the root of the source directory. This means per-crate fuzzing harnesses is not currently supported. As a workaround this change moves all the per-crate fuzz harnesses into a single root level fuzz crate.

----


fuzz/Cargo.toml
```diff
@@ -0,0 +1,97 @@
...
+version = "0.0.0"
+publish = false
+edition = "2021"
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = "0.4"
+git-actor = {path= "../git-actor"}
```
Byron, line 12: Is there a need to list every single crate here? I'd expect this dependency list to only include crates which are fuzzed. 

Limiting the list also has benefits for compile times.
Thanks for the clarification.

Byron (commented):
Thanks for setting this up!

Even though I am a bit surprised that OSS-Fuzz is that opinionated or lacks configuration options, making this change doesn't seem like a disadvantage. After all, fuzz code isn't included in the crate package at all and having fuzz targets in a central place makes their discovery and maintenance easier.

I have one question left but when clarified, I think this PR is ready for merge.

fuzz/Cargo.toml
silvergasp: Yeah, that was the plan. It was easier for me to just `ls -1 git-*` + some regex to get all the dependencies. But I agree that this should just be the dependencies that are actually needed.

silvergasp:
> Thanks for setting this up!
> 
> Even though I am a bit surprised that OSS-Fuzz is that opinionated or lacks configuration options, making this change doesn't seem like a disadvantage. After all, fuzz code isn't included in the crate package at all and having fuzz targets in a central place makes their discovery and maintenance easier.
> 
> I have one question left but when clarified, I think this PR is ready for merge.

I've still been having the same issue with this PR and generating coverage from OSS-fuzz. So I don't think that this PR solves the issue either. I might escalate this upstream and see if I get any traction there.

I'm certain that code coverage works with rust, I'm just not sure why it's not working here. 

Here's a recent coverage report for serde_urlencoded https://storage.googleapis.com/oss-fuzz-coverage/serde_urlencoded/reports/20230208/linux/src/serde_urlencoded/report.html.

silvergasp:
I'm going to close this, as I don't think it helps at all :)

Byron:
Thanks for seeing this through, it's much appreciated.

Part of me is sad to see this PR go as well, having a central fuzz folder with easily accessible targets might reduce duplication and facilitate adding more fuzz tests due to lower overhead.

If you think there is something to it, maybe you can reopen the PR and remove all the unused crate references, and it will be merged.

Thanks and cheers

PR #728, opened 2023-02-11
rename crates

Byron:
Rename `git-repository` to `gix`


gix/CHANGELOG.md
```diff
@@ -5,15 +5,3994 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## v0.0.0 (2023-02-10)
+## 0.35.0 (2023-02-11)
 
-Initial release.
+This is the last release under this name and merely a notice to inform that `gix` from now on is `gix`.
```
Mathspy, line 10: It looks like during the find and replace in the last commit that this message was broken. It now says rename `gix` to `gix` and same for the note on line 16

Byron: 😅 Thanks for the pointer, these things are better not done in a rush. I think it's now fixed in `main` as well.

PR #731, opened 2023-02-11
Link git-url/serde1 feature to git-repository

bittrance:
For simplicity, I used `git_url::Url` directly in my config format, but it seems the git-repository crate does not propagate serde config. This PR adds it.

----


Byron:
Thanks a lot, good catch!

PR #733, opened 2023-02-13
git-discover with unc-paths support

### LINKED ISSUE #730: `git-discover` doesn't allow extended-length paths on Windows as ceiling directories
### Duplicates

### Current behavior 😯

Adding a ceiling directory with an extended-length path results in the directory being ignored when discovering upwards from a path that _doesn't_ use the extended-length path syntax. An extended-length path is a path prefixed with `\\?\` this is e.g. used in [`std::fs::canonicalize`](https://doc.rust-lang.org/std/fs/fn.canonicalize.html) (Windows docs: [Naming a file](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file), [Maximum Path Length Limitation](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation)).

### Expected behavior 🤔

The ceiling directory should be picked up regardless of which path syntax is used (this is the case for the regular git executable).

### Steps to reproduce 🕹

I created a test program that should run successfully, assuming the following directory layout:

```
directory-with-git-repo/
├─ .git
╰─ some/
   ╰─ child/
      ╰─ dir/ <- cwd
```

```rust
use git_discover::{upwards::Options, upwards_opts};

fn main() {
    let cwd = std::env::current_dir().unwrap();
    let parent = cwd.parent().unwrap();
    // all discoveries should fail, as they'll hit `parent` before finding a git repository.

    // dir: normal, ceiling: normal
    let r1 = upwards_opts(
        &cwd,
        Options {
            ceiling_dirs: vec![parent.to_path_buf()],
            match_ceiling_dir_or_error: false,
            ..Default::default()
        },
    );

    let parent = parent.canonicalize().unwrap();
    // dir: normal, ceiling: extended
    let r2 = upwards_opts(
        &cwd,
        Options {
            ceiling_dirs: vec![parent],
            match_ceiling_dir_or_error: false,
            ..Default::default()
        },
    );

    let cwd = cwd.canonicalize().unwrap();

    let parent = cwd.parent().unwrap();
    // dir: extended, ceiling: normal
    let r3 = upwards_opts(
        &cwd,
        Options {
            ceiling_dirs: vec![parent.to_path_buf()],
            match_ceiling_dir_or_error: false,
            ..Default::default()
        },
    );

    let parent = parent.canonicalize().unwrap();
    // dir: extended, ceiling: extended
    let r4 = upwards_opts(
        &cwd,
        Options {
            ceiling_dirs: vec![parent],
            match_ceiling_dir_or_error: false,
            ..Default::default()
        },
    );

    assert!(matches!(
        dbg!((r1, r2, r3, r4)),
        (Err(_), Err(_), Err(_), Err(_))
    ));
}
```

This should run on all platforms. Currently, on Windows, `r2` (the case I described above) will be `Ok(_)` (meaning a git repository was discovered). This is wrong.

It seems like [`Path::strip_prefix`](https://doc.rust-lang.org/std/path/struct.Path.html#method.strip_prefix) doesn't strip the prefix when `\\?\` is used in the prefix ([used here](https://github.com/Byron/gitoxide/blob/9e7d79273487abfcb99ed2d439c475a659cd25e6/git-discover/src/upwards/util.rs#L56)).

Byron:
Fixes #730

Related issue in `dunce`: https://gitlab.com/kornelski/dunce/-/issues/5

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -899,6 +899,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "dunce"
+version = "1.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0bd4b30a6560bbd9b4620f4de34c3f14f60848e58a9b7216801afcb4c7b31c3c"
+
+[[package]]
 name = "either"
 version = "1.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1440,6 +1446,7 @@ version = "0.13.0"
 dependencies = [
  "bstr",
  "defer",
+ "dunce",
  "git-hash 0.10.2",
  "git-path 0.7.1",
  "git-ref 0.24.0",
--- a/deny.toml
+++ b/deny.toml
@@ -39,6 +39,7 @@ allow = [
     "Apache-2.0",
     "BSD-3-Clause",
     "MIT",
+    "MIT-0",
     "ISC",
     "Unicode-DFS-2016",
     "LicenseRef-ring"
--- a/git-discover/Cargo.toml
+++ b/git-discover/Cargo.toml
@@ -21,6 +21,9 @@ git-hash = { version = "^0.10.2", path = "../git-hash" }
 bstr = { version = "1.0.1", default-features = false, features = ["std", "unicode"] }
 thiserror = "1.0.26"
 
+[target.'cfg(windows)'.dependencies]
+dunce = "1.0.3"
+
 [dev-dependencies]
 git-testtools = { path = "../tests/tools" }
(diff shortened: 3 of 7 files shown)


davidkna:
Your issue looks like a duplicate of https://gitlab.com/kornelski/dunce/-/issues/2 which was closed without fixing the bug (see also ajeetdsouza/zoxide#162).

Byron:
Thanks for digging this out! I noted that in the issue itself for good measure and am thinking that it's probably unlikely to get this fixed there then, maybe a good motivation to eventually integrate `dunce` into `git-path` in some way.

PR #740, opened 2023-02-16
fix link in docs for `sort_entries_by`

asottile:


Byron:
Thanks a lot for letting me know!

As I am currently renaming a lot I have manually ported this fix into my branch in commit eb48b162c3f9c449de0fabfc2e8a33001a2d14d0.

Hence I am closing this PR, and the fix will soon show up in `main`.

asottile:
I'm not sure why you didn't just merge? the patch you committed removes authorship and is identical (same file, same diff, same offset) so it would have merged cleanly

Byron:
Sorry for not having been more clear. I was unreasonably afraid of merge conflicts and chose to avoid them, even though in the end I encountered them anyway due to renaming everything from `git-*` to `gix-*`

I was thinking restoring some authorship by cherry-picking this commit onto `main` (ignoring that it would then be empty), but it's probably futile since it would also destroy the commit signature.

On the bright side, if you are perusing the docs, I am sure you will find more typos or other forms of brokenness, so there will be ample opportunity to remedy this in the future. Thank you.

asottile:
the easiest way to include attribution (even without cherry-pick) is to use [co-authored-by: ...](https://youtu.be/_-qNX8EbVYI) -- I don't really care about the signature personally (it was only signed because I made the commit from the github ui anyway)

Byron:
Great to hear the signature is only the one from GtiHub (I wasn't aware it does this now either), so cherry-picking won't be too bad. The new commit is here: 79d4461e787ec290c71d653b57d14564eacc15a1.

Impressive tech on your YouTube channel btw!

PR #744, opened 2023-02-18
feat: add `ObjectId::empty_blob()` to obtain the empty blob object.

### LINKED ISSUE #743: potentially useful: gix::ObjectId::empty_file(...)
### Summary 💡

I notice there's already a quick helper for `empty_tree` -- was wondering if one for the empty file would be useful as well

happy to submit a PR if this sounds like a good idea -- also completely fine with a "no" here -- I totally understand!

### Motivation 🔦

`git` uses this a bit internally, for instance for `--intent-to-add` (which is where I'm hitting it!)

no biggie, I can always use:

```rust
let oid = gix::ObjectId::from_hex(b"e69de29bb2d1d6434b8b29ae775ad8c2e48c5391").unwrap();
```

or whatever

asottile:
Resolves #743 

----

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1830,6 +1830,7 @@ name = "gix-hash"
 version = "0.10.2"
 dependencies = [
  "document-features",
+ "gix-features",
  "gix-testtools",
  "hex",
  "serde",
--- a/gix-hash/Cargo.toml
+++ b/gix-hash/Cargo.toml
@@ -26,6 +26,7 @@ document-features = { version = "0.2.0", optional = true }
 
 [dev-dependencies]
 gix-testtools = { path = "../tests/tools"}
+gix-features = { path = "../gix-features", features = ["rustsha1"] }
 
 [package.metadata.docs.rs]
 all-features = true
--- a/gix-hash/src/object_id.rs
+++ b/gix-hash/src/object_id.rs
@@ -99,6 +99,16 @@ impl ObjectId {
         }
     }
 
+    /// The hash of an empty blob
+    #[inline]
+    pub const fn empty_blob(hash: Kind) -> ObjectId {
+        match hash {
+            Kind::Sha1 => {
+                ObjectId::Sha1(*b"\xe6\x9d\xe2\x9b\xb2\xd1\xd6\x43\x4b\x8b\x29\xae\x77\x5a\xd8\xc2\xe4\x8c\x53\x91")
+            }
+        }
+    }
+
     /// The hash of an empty tree
     #[inline]
     pub const fn empty_tree(hash: Kind) -> ObjectId {
--- a/gix-hash/tests/object_id/mod.rs
+++ b/gix-hash/tests/object_id/mod.rs
@@ -53,12 +53,12 @@ mod empty {
     }
 
     #[test]
-    fn empty_blob() {
+    fn blob() {
         assert_eq!(ObjectId::empty_blob(Kind::Sha1), hash_contents(b"blob 0\0"));
     }
 
     #[test]
-    fn empty_tree() {
+    fn tree() {
         assert_eq!(ObjectId::empty_tree(Kind::Sha1), hash_contents(b"tree 0\0"));
     }
 }
--- a/gix-hash/tests/object_id/mod.rs
+++ b/gix-hash/tests/object_id/mod.rs
@@ -41,3 +41,24 @@ mod from_hex {
         }
     }
 }
+
+mod empty {
+    use gix_features::hash::hasher;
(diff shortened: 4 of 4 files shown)


asottile:
I have never written a rust test before so please guide me through this and if what I did was what you want -- I initially tried writing this as an inline "unit" test but I couldn't seem to get `cargo test` to find it? (unless I did `cargo test object_id` -- but I was afraid that wasn't picked up automatically?)

I also added another test for the empty tree (and learned a little bit about how these hashes are built in the first place!)

Byron:

> I initially tried writing this as an inline "unit" test but I couldn't seem to get cargo test to find it? (unless I did cargo test object_id -- but I was afraid that wasn't picked up automatically?)

I don't exactly know where the confusion is originating from, as in theory, when run from `gix-hash`, `cargo test` would pick up all tests automatically, no matter whether it's unit- or integration-tests. From what I could tell, all is well.

Adding tests as integration test is the way to go here, too, as we are testing the public API. Unit tests also have access to private functions, something we don't need or want here.

As it's a nice demo on how to make a small contribution, I did record my first video for the year as well: https://youtu.be/0ToR38ki2r4

asottile:
anyway -- here's what I was confused about with the unit tests:

```diff
diff --git a/gix-hash/src/object_id.rs b/gix-hash/src/object_id.rs
index d295fc5..347421d 100644
--- a/gix-hash/src/object_id.rs
+++ b/gix-hash/src/object_id.rs
@@ -227,3 +227,11 @@ impl PartialEq<&crate::oid> for ObjectId {
         self.as_ref() == *other
     }
 }
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn asdfasdfasdf() {
+        assert_eq!(1, 2);
+    }
+}
```

and then inside `gix-hash`:

```console
$ cargo test
    Finished test [unoptimized + debuginfo] target(s) in 0.19s
     Running tests/hash.rs (/tmp/gitoxide/target/debug/deps/hash-9ee2f6a1b6cf2a5f)

running 20 tests
test kind::from_hex_len::none_if_there_is_no_fit ... ok
test kind::from_hex_len::some_sha1 ... ok
test object_id::empty::blob ... ok
test object_id::from_hex::invalid::non_hex_characters ... ok
test object_id::empty::tree ... ok
test object_id::from_hex::invalid::too_short ... ok
test object_id::from_hex::valid::twenty_hex_chars_lowercase ... ok
test object_id::from_hex::valid::twenty_hex_chars_uppercase ... ok
test prefix::cmp_oid::it_detects_equality ... ok
test oid::to_hex_with_len::display_entire_range_sha1 ... ok
test prefix::cmp_oid::it_detects_inequality ... ok
test prefix::new::errors_if_hex_len_is_longer_than_oid_len_in_hex ... ok
test prefix::new::errors_if_hex_len_is_too_short ... ok
test prefix::try_from::id_8_chars ... ok
test prefix::try_from::id_9_chars ... ok
test prefix::try_from::id_to_long ... ok
test prefix::new::various_valid_inputs ... ok
test prefix::try_from::invalid_chars ... ok
test object_id::from_hex::invalid::too_long ... ok
test prefix::try_from::id_to_short ... ok

test result: ok. 20 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

```

whereas I expect a failure 🤷 

if I run the tests with a name -- then the new test fails :?:

```console
$ cargo test object_id
    Finished test [unoptimized + debuginfo] target(s) in 0.19s
     Running unittests src/lib.rs (/tmp/gitoxide/target/debug/deps/gix_hash-5d45f0e3c759dfe8)

running 1 test
test object_id::tests::asdfasdfasdf ... FAILED

failures:

---- object_id::tests::asdfasdfasdf stdout ----
thread 'object_id::tests::asdfasdfasdf' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `2`', gix-hash/src/object_id.rs:235:9


failures:
    object_id::tests::asdfasdfasdf

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

error: test failed, to rerun pass `--lib`
```

Byron:
Ah, right! I can reproduce this and the culprit is that [this setting](https://github.com/Byron/gitoxide/blob/96047839a20a657a559376b0b14c65aeab96acbd/gix-hash/Cargo.toml#L14-L13) in the `Cargo.toml`.

Crates without unit-tests better disable them as otherwise rustc will build an empty binary for these and run it, even though nothing will be in it. This time adds up.
Thus, if you find a crate that has no `cargo test --lib` tests, these should set `test = false` in their manifest.

PR #765, opened 2023-03-04
shallow support (Repository)

Byron:
### Tasks

* [x] make `GIT_SHALLOW_FILE` environment variable available via config - it's useful for hooks where this file isn't settled yet (and generally environment variables are used for this kind of thing)
* [x] inform about the shallow-ness of a repo based on presence of `.git/shallow`
* [x] read the `.git/shallow` file
* [x] ~~validate that the 'commit not founds' during iteration is actually a grafted commit and fail otherwise~~ remove special handling of guessed shallow commits - instead make it work properly.
* [x] assure we handle iteration on a complex shallow boundary correctly - after all we should be able to walk the whole graph


Byron:
already merged.

PR #784, opened 2023-03-17
Exploring API

svetli-n:
----

### DIFF
--- a/gix-archive/README.md
+++ b/gix-archive/README.md
@@ -1,0 +1,2 @@
+`gix-archive` creates an archive of files from a named tree
+


Byron:
Let's add READMEs once there is something substantial to say.

This reminds me that the existing plumbing READMEs also should be overhauled once crates shift to being production ready, as most of them are rather sparse as well.

Byron:
I just now saw that the PR title is Exploring API, so maybe more was to come. If that's the case, please reopen the PR and make it a draft so it's clear it's not actually to be merged.

Further, I recommend doing an actual analysis on the features that the API would entail in the tracking issue so PRs can be focussed on one particular topic.

svetli-n:
Sure, I though it was a draft, probably forgot about to set it so. 

Updated the tracking issue. I'll reopen this and push a change with the idea so we can discuss around it.

Issue #783, opened 2023-03-17
gix-archive: create an archive of files from a named tree support


svetli-n:
## Summary 💡
The goal of this issue is to track the tasks, requirements and analysis that will lead to adding functionality for archiving of files from a named tree to gitoxide.

## Background
git has archive command, allowing a user to to add a list of files to an archive. While the files normally originate from a named tree, identified by commit or tag ID/SHA1 hash, users can also add additional files by providing paths to these and/or their content. The files can be added to various archive types, such as tar, zip, gzip/tgz as well as to any custom archive type by utilising external tools/commands. The user can change file attributes prior to adding them to an archive, by looking for attributes in .gitattributes files, or explicitly setting modification time. Users can also select only certain parts of a named tree to be archived. Archives are written to stdout or to a file. The tree to be archived can be both local or from a remote repository.

## Method
Having the rich archive functionality described above, the plan to add support for it in gitoxide is as follows:
1. Define MVP functionality
2. Implement it and integrate it in the project, utilising external as well as internal/gitoxide crates
3. Iterate on potentially adding more features

## Scope

- **MVP for API** #796
- **gitattributes and archive specific filters**
   - possibility to query by path .gitattributes for [export-subst](https://git-scm.com/docs/git-archive#Documentation/git-archive.txt-export-subst) and [export-ignore](https://git-scm.com/docs/git-archive#Documentation/git-archive.txt-export-ignore) (to be implemented in another crate)
   - act on .gitattributes in the tree to be archived 
   - optionally act on .gitattributes in the work tree
   - `export-subst` filter expanding `git-log` [placeholders](https://git-scm.com/docs/git-log#Documentation/git-log.txt-emnem) (the expansion is to be handled by another to be created crate, e.g. `gix log`)
- **filter files of tree** - control what goes into the archive
- **add extra files**
  - [ ] as read file paths
  - [ ] By providing file paths and content
- **archive formats**
  - [x] zip
  - [x] tar
  - [x] convenient compression of `tar` archives
  - [ ] describe how to implement custom archives by leveraging the existing `tar` stream
  - [ ] extra options for archive backends
- [x] integrate into `gix` as optional dependency with optional API.
- [x] make available via `gix archive` on the CLI

## Q&A
- **How to support zip as well as tar?**
  - These are built-in and additional formats can be implemented by re-encoding the `tar` stream.
- **Do gitattributes always contribute to the archive, so not only when --worktree-attributes is set?**
[...]

PR #785, opened 2023-03-18
Explore api

svetli-n:
----

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1372,6 +1372,9 @@ dependencies = [
 [[package]]
 name = "gix-archive"
 version = "0.0.0"
+dependencies = [
+ "thiserror",
+]
 
 [[package]]
 name = "gix-attributes"
--- a/gix-archive/Cargo.toml
+++ b/gix-archive/Cargo.toml
@@ -12,3 +12,4 @@ rust-version = "1.64"
 doctest = false
 
 [dependencies]
+thiserror = "1.0.26"
--- a/gix-archive/README.md
+++ b/gix-archive/README.md
@@ -1,0 +1,2 @@
+`gix-archive` creates an archive of files from a named tree
+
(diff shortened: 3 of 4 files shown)


svetli-n:
An example with builder API, taking care of invariants, default and optional archive parameters. @Byron please feedback, thanks

Byron:
In order for me to be effective, I need to understand the feature. As a feature lead, you are responsible for helping me understand the feature - you are the expert. As `git-archive` is complex, at first there should be an understanding of what a minimum viable product (MVP) would be. 

Once the MVP is defined, it would certainly be good to also see the big picture by sketching out the final product as it may exist with all the bells and whistles, but that's also an opportunity to out-scope features that aren't a good fit for for `gitoxide` to begin with or seem more like "add when there is actual demand" (once there is users).

At this stage the whole undertaking might die if its unclear why people wouldn't just use `git archive` instead. Is there a unique selling point that our implementation provides. Is it faster, is it more flexible? If there is nothing, no need to spend time on it.

Something I don't want to endorse anymore is commits that chronicle the history but don't carry any value beyond their changeset. With the advent of [Stacked Git](https://github.com/stacked-git/stgit/) in my workflow I manage to be more organized, which also means commits are rewritten a lot. With the current workflow, before any merge can happen it will need a squash along with a proper commit message that you generate, but tool usage can make that part of the workflow instead of a chore (and post-processing step).

All this was a well-meant and long-winded way of saying: let's go to #783 and carve out the MVP, understand the value of the crate. Then let's cut this PR down to only contain the API surface needed for an MVP along with some tests that use *the entire API surface* just to try it out. And then I feel ready to provide feedback on it. I hope that makes sense.

svetli-n:
Here is MVP [suggestion](https://github.com/Byron/gitoxide/issues/783#issuecomment-1476902178)

PR #794, opened 2023-03-27
fix: uncategorized refs should not be per-worktree

jpgrayson:
When creating a new ref from using a Repository instantiated from a linked worktree, if the ref did not fit one of the well-known categories, it would be incorrectly stored as a per-worktree reference instead of a shared reference.

For example, the reference "refs/stacks/linked" does not fit any of the well known categories. According to git-worktree(1), this ref should be shared across all worktrees; i.e. located at $GIT_COMMON_DIR/refs/stacks/linked and *not* in $GIT_COMMON_DIR/worktrees/linked/refs/stacks/linked.

This change tweaks the control flow so that uncategorized refs correctly fallback to being relative to commondir while still placing refs that are supposed to be per-worktree in the per-worktree location.

----

### DIFF
--- a/gix-ref/src/store/file/find.rs
+++ b/gix-ref/src/store/file/find.rs
@@ -186,10 +186,10 @@ impl file::Store {
                             }
                         })
                         .unwrap_or((commondir.into(), sn)),
-                    PseudoRef | Bisect | Rewritten | WorktreePrivate => return None,
+                    PseudoRef | Bisect | Rewritten | WorktreePrivate => (self.git_dir.as_path().into(), name),
                 })
             })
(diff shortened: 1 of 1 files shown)


Byron:
A [new release of the `git-ref`](https://github.com/Byron/gitoxide/releases/tag/gix-ref-v0.27.2) crate is now available and should be pulled in by the latest `gix` automatically.

jpgrayson:
Thank you for merging this PR and for adding the new tests, @Byron.

With this issue resolved in gix-ref, I was able to resolve the remaining regressions to allow StGit to work from linked worktrees.

PR #800, opened 2023-03-30
fix $HOME detection on windows

pascalkuthe:
Fixes https://github.com/helix-editor/helix/issues/6467

Git manually fixes up the $HOME environment variable in compat/mingw.c because it is not set on windows (unless running inside unix emulation like git bash). This PR replaces all calls to `std::env::var_os("HOME")` with a function 
that matches the git implementation. Note that this is not necessarily the best  $HOME detection on windows (for example $HOMESHARE is not considered). I  purposefully exactly matched the git implementation here for compatibility.

----

### DIFF
--- a/cargo-smart-release/src/crates_index.rs
+++ b/cargo-smart-release/src/crates_index.rs
@@ -36,7 +36,7 @@ fn default_path() -> PathBuf {
         .or_else(|| {
             std::env::var_os("CARGO_HOME")
                 .map(PathBuf::from)
-                .or_else(|| std::env::var_os("HOME").map(|dir| Path::new(&dir).join(".cargo")))
+                .or_else(|| home::home_dir().map(|dir| Path::new(&dir).join(".cargo")))
         })
         .expect("one of these paths works")
         .join("registry/index")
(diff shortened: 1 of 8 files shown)


cargo-smart-release/src/crates_index.rs
```diff
@@ -36,7 +36,7 @@ fn default_path() -> PathBuf {
         .or_else(|| {
             std::env::var_os("CARGO_HOME")
                 .map(PathBuf::from)
-                .or_else(|| std::env::var_os("HOME").map(|dir| Path::new(&dir).join(".cargo")))
+                .or_else(|| home::home_dir().map(|dir| Path::new(&dir).join(".cargo")))
```
pascalkuthe, line 39: this is different because cargo-smart-release is not a git tool so I used `home::home_dir` here (which is already a dependency for `cargo-smart-release`). All other places use `gix_path::home`

Byron (approved):
Thanks a lot for fixing for this particularly subtle (and thus intractable) problem!

I will merge and create a new release once CI is green.

Byron:
And here it is: https://github.com/Byron/gitoxide/releases/tag/gix-v0.43.1

Issue #802, opened 2023-03-31
Revisions cannot contains the character @


lthms:
### Duplicates

- [X] I have searched the existing issues

### Current behavior 😯

The function `gix_revision::spec::parse::function::parse` (in `gix-revision/src/spec/parse/function.rs` does not support revision which contains the character `@`.

For instance, at work, we have the guidelines to prefix our branches names by our handle, so my branch looks like `lthms@fix-something`.

I’ve run into the issue because I am using Stacked Git, which has moved from its previous git library to `gitoxide` (see https://github.com/stacked-git/stgit/issues/306). Since I’ve upgraded, I get errors related to branch names with `@`, and after some digging, I was able to isolate the `parse` function.

### Expected behavior 🤔

`git branch` and `git rev-parse` handle branches with `@` in their name gracefully.

I would have expected that `gix` would do so too. But maybe it is not a goal? I’ve seen that there are tests explicitely checking that `gix` fails in these cases (in `gix-revision-tests`, in `refnames.rs`).

I am trying to get my head around the code, but so far it’s a bit hard to follow. Any help is welcome.

### Steps to reproduce 🕹

1. `git branch a@b`
2. `cargo run rev resolve a@b`

Result is

```
Error: The ref partially named "a" could not be found
```

Byron:
Thanks a lot for making me aware!

> I would have expected that gix would do so too. But maybe it is not a goal? 

`gix` desperately wants to be fully compatible to `git`, and this is clearly a shortcoming to be fixed. I didn't think of it, which explains why the code can't handle it.

Note that for testing `gix rev parse -e revspec` is most helpful as it visualizes the parser callbacks which ultimately drives the resolver.

PR #803, opened 2023-03-31
gix-revision: Parse revisions with the @ character

lthms:
Fixes https://github.com/Byron/gitoxide/issues/802

Happy to have some feedback. The solution seems to work, and makes sense to me, but of course I’m not very familiar with this codebase :sweat_smile: 

----

### DIFF
--- a/gix-revision/src/spec/parse/function.rs
+++ b/gix-revision/src/spec/parse/function.rs
@@ -344,15 +344,19 @@ where
     {
         let mut cursor = input;
         let mut ofs = 0;
+        const SEPARATORS: &[u8] = b"~^:.";
         while let Some((pos, b)) = cursor.iter().enumerate().find(|(pos, b)| {
             if **b == b'@' {
-                cursor.get(pos + 1) == Some(&b'{')
-                    || cursor
-                        .get(pos + 1)
-                        .and_then(|b| Some(b"~^:.".contains(b)))
-                        .unwrap_or(false)
-                    || (*pos == 0 && cursor.get(1) == None)
-            } else if b"~^:.".contains(b) {
+                if cursor.len() == 1 {
+                    return true;
+                }
+                let next = cursor.get(pos + 1);
+                let next_next = cursor.get(pos + 2);
+                if *pos != 0 && (next, next_next) == (Some(&b'.'), Some(&b'.')) {
+                    return false;
+                }
+                next == Some(&b'{') || next.map_or(false, |b| SEPARATORS.contains(b))
+            } else if SEPARATORS.contains(b) {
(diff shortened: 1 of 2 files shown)


Byron:
Thanks a lot for your contribution, I never thought about `@` being anything else than a special character, and it shows.

And guess who also it not familiar with the codebase (*anymore*) 😅?

Thus it's time to trust the tests, and I did manage to find more breaking examples by using ranges which might indicate that the right spot for handling `@` characters is at another place, or ranges also have to be handled specifically.

I am still looking into it to see what to do here.

Byron:
Ranges now work as well and in theory, this passes the fuzzer as well (practical to have it, thanks google-fuzz-project!). If so, and you can't find any other cases maybe by experimenting with `gix rev parse -e <revspec>` this should be good to go.

lthms:
Thanks for the refactoring! Looks like the CI does not pass actually! The output is a bit cryptic to me, but I will try to have a look.

lthms:
Looks like reintroducing the error for things like `.@.` is necessary.

Byron:
No worries about it. The idea is that the fuzzer finds an input that triggers a panic, then a test can be added that verifies that the panic doesn't occour anymore.

I'll be adding the test shortly.

Byron:
Yes, it's my first time the fuzzer found something in this CI and then the artifact can be downloaded from GitHub directly. I have added a test with that input, and if it goes through I'd think the implementation should be fine.

Thanks again for getting the ball rolling on this one!

lthms:
You are very welcome! I guess it might make sense to squash the two commits together, and mark both of us as co-authors. (:

Do you mind doing a minor release for `gix-revision`? So that I can push a fix to Stacked Git too?
(it is not that urgent, I can keep use my fixed version of stacked git built locally, but it would be nice to fix this regression at some point)

Byron:
> You are very welcome! I guess it might make sense to squash the two commits together, and mark both of us as co-authors. (:

A good point! It's done locally (don't want to trigger CI again), and it will be merged in instead of what's currently here.

> Do you mind doing a minor release for gix-revision? So that I can push a fix to Stacked Git too?

That's the plan - I will post here when done in a bit.

Byron:
The PR was brought into `main` and [a new `gix-revision` patch release](https://github.com/Byron/gitoxide/releases/tag/gix-revision-v0.12.2) make the fix available. 
CC @jpgrayson in case this warrants a new `stg` release.

lthms:
Thanks again @Byron. For the record, I was meaning to squash both our commits, not just claim co-authorship on you own :sweat_smile:

PR #811, opened 2023-04-11
fix: make `oid` `Hash` work on 32-bit targets

### LINKED ISSUE #810: `gix-hashmap` may panic on i686 targets
### Duplicates

### Current behavior 😯

starship/starship#5092
When running the starship `git_commit` tests on i686, a panic occurs in `gix-hashtable`:
```
thread '<unnamed>' panicked at 'range end index 8 out of range for slice of length 4', C:\Users\david\.cargo\registry\src\index.crates.io-6f17d22bba15001f\gix-hashtable-0.1.2\src\lib.rs:50:41
FAILED
````

### Expected behavior 🤔

It looks like `gix-hashmap` is used with a type that is too short on i686, but perhaps it should also be more error-resilient.

### Steps to reproduce 🕹

1. `git clone https://github.com/starship/starship.git`
2. `cd starship`
3. `cargo test --target i686-pc-windows-msvc -- --test-threads 1 --nocapture git_commit` (or apparently other i686 targets).

### PR DESCRIPTION
**Pull request by davidkna**
This could lead to panics on 32-Bit platforms
Closes #810

----

### CODE REVIEW

**Comment by pascalkuthe** [comment]
I am not sure that this is the right thing to do. The hasher in `gix-hashtable` is only intended to provide a fast hash implementation by cutting of a cryptographically secure hash (like sha1). Specifically, the Hasher in that crate is only intended for hashing `gix_hash::ObjectId`. You mentioned `write_len_prefix` so it sounds like you are hashing some kind of slice (or str?) somewhere. I grepped trough gitoxide and all uses of `gix_hashtable` only hash `ObjectId`s so this might be a downstream crate using the API incorrectly.

That might be the real source of the bug. You should never be hashing normal data with `gix-hashtable`. Could you provide some more details about where this crash actually occurred?

**Comment by davidkna** [comment]
The crash happens when an `ObjectId` is hashed inside gitoxide library code.
https://github.com/Byron/gitoxide/blob/abb8b9f3f11b09a614174447ed5bc1bb750bc0f0/gix-hash/src/object_id.rs#L26-L30
I tried changing the function to use `write_u64` instead, but I got the same problem, though I didn't test that thoroughly.

**Comment by pascalkuthe** [comment]
> The crash happens when an `ObjectId` is hashed inside gitoxide library code.
> 
> https://github.com/Byron/gitoxide/blob/abb8b9f3f11b09a614174447ed5bc1bb750bc0f0/gix-hash/src/object_id.rs#L26-L30
> 
> I tried changing the function to use `write_u64` instead, but I got the same problem, though I didn't test that thoroughly.

Can you provide a backtrace? The hash Implementation you linked (which I had in mind) just calls `write` directly which is the manual implemented in the `Hasher`. There is no reference to `write_usize` anywhere (which is exactly why this manual implementation exists instead of a derived implementation) so the call to `write_usize` has to come from somewhere else.

**Comment by davidkna** [comment]
This is the relevant part of a backtrace:

```thread '<unnamed>' panicked at 'range end index 8 out of range for slice of length 4', C:\Users\david\.cargo\registry\src\index.crates.io-6f17d22bba15001f\gix-hashtable-0.1.2\src\lib.rs:50:41
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library\std\src\panicking.rs:575
   1: core::panicking::panic_fmt
             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library\core\src\panicking.rs:64
   2: core::slice::index::slice_end_index_len_fail_rt
             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library\core\src\slice\index.rs:77
   3: core::slice::index::slice_end_index_len_fail
             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library\core\src\slice\index.rs:69
   4: core::slice::index::impl$4::index<u8>
             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0\library\core\src\slice\index.rs:409
   5: core::slice::index::impl$0::index<u8,core::ops::range::RangeTo<usize> >
             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0\library\core\src\slice\index.rs:18
   6: core::hash::Hasher::write_length_prefix<gix_hashtable::hash::Hasher>
             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0\library\core\src\hash\mod.rs:503
   7: gix_hash::borrowed::impl$14::hash<gix_hashtable::hash::Hasher>
             at C:\Users\david\.cargo\registry\src\index.crates.io-6f17d22bba15001f\gix-hash-0.10.3\src\oid.rs:22
   8: hashbrown::map::HashMap<enum2$<gix_hash::object_id::ObjectId>,enum2$<alloc::borrow::Cow<bstr::bstr::BStr> >,gix_hashtable::hash::Builder,hashbrown::raw::inner::alloc::inner::Global>::get_inner<enum2$<gix_hash::object_id::ObjectId>,enum2$<alloc::borrow::Co
             at C:\Users\david\.cargo\registry\src\index.crates.io-6f17d22bba15001f\hashbrown-0.13.1\src\map.rs:1395
   9: hashbrown::map::HashMap<enum2$<gix_hash::object_id::ObjectId>,enum2$<alloc::borrow::Cow<bstr::bstr::BStr> >,gix_hashtable::hash::Builder,hashbrown::raw::inner::alloc::inner::Global>::get_inner<enum2$<gix_hash::object_id::ObjectId>,enum2$<alloc::borrow::Co
             at C:\Users\david\.cargo\registry\src\index.crates.io-6f17d22bba15001f\hashbrown-0.13.1\src\map.rs:1395
  10: hashbrown::map::HashMap<enum2$<gix_hash::object_id::ObjectId>,enum2$<alloc::borrow::Cow<bstr::bstr::BStr> >,gix_hashtable::hash::Builder,hashbrown::raw::inner::alloc::inner::Global>::get<enum2$<gix_hash::object_id::ObjectId>,enum2$<alloc::borrow::Cow<bstr
             at C:\Users\david\.cargo\registry\src\index.crates.io-6f17d22bba15001f\hashbrown-0.13.1\src\map.rs:1350
  11: core::iter::range::impl$2::spec_next<usize>
             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0\library\core\src\iter\range.rs:621
  12: <unknown>
  13: gix::commit::describe::Platform::try_resolve
             at C:\Users\david\.cargo\registry\src\index.crates.io-6f17d22bba15001f\gix-0.43.1\src\commit.rs:209
  14: gix::commit::describe::Platform::try_format
             at C:\Users\david\.cargo\registry\src\index.crates.io-6f17d22bba15001f\gix-0.43.1\src\commit.rs:195
```

pascalkuthe:
the issue is not that we are hashing an `ObjectId` incorrectly but that we are using `get` with an `&oid`. The issue occurs because HashMaps assume that any type the implements `PartialEq<T>` has an equivalent hash implementation. That is pretty problematic for our usecase. To fix this we should just add a custom `Hash` implementation to `oid`  similar to `ObjectId`. The prefix isn't important since Sha1/Sha256 collisions are insanely unlikely (and would just result in a single byte comparison anyway).

To be more robust we should also panic on all `Hasher` functions except `write`.

davidkna:
@pascalkuthe I've updated the PR with your feedback now.

pascalkuthe (approved):
LGTM, thanks for looking into this

Byron:
Thanks a lot to everybody involved!

This looks good to me as well and once I managed to add a test that would hopefully trigger the issue, I can merge and create patch releases for both involved crates.

Byron:
I was only able to add a minimal 32bit test for now and didn't validate that this actually triggers the panic, so let's just assume it doesn't until it is revisited and improved upon. The reason I am willing to merge it even without assurance is that… there seems to be another issue around running `git` in docker which all of the sudden refuses to pick up configuration from environment variables which causes fixture scripts to fail. Despite considerable effort I wasn't able to understand why this would happen :/.

Byron:
A new patch release was made for `gix-hash` and `gix-hashtable` - maybe you can validate that this fixes the respective issue in `starship`?

PR #814, opened 2023-04-16
rename the serde1 feature to serde

vilgotf:
Uses Cargo weak-deps (stabilized in 1.60) to declare serde and thus avoid the awkward serde1 feature. 

----


Byron:
Thanks a lot for putting in the work needed to make the transition, it's much appreciated :)!

PR #434, opened 2022-05-28
Support native zlib-ng via flate2's zlib-ng feature

joshtriplett:
In addition to the zlib-ng-compat feature, flate2 now supports zlib-ng
via its native API and the libz-ng-sys crate. This avoids potential
symbol conflicts with dependencies on system C libraries that link to
system zlib. Using zlib-ng increases compatibility, avoids those
conflicts, and allows using zlib-ng in more circumstances.

(zlib-ng-compat is still useful when you have dependencies that need the
zlib C ABI and don't support zlib-ng, such as curl; git-transport still
uses zlib-ng-compat.)

----

Can [Byron](https://github.com/Byron) review the PR on video?

Please choose one - no choice means no recordings are made.

- [x] Record review and upload on YouTube publicly
- [ ] Record review and upload on YouTube, but keep video unlisted

If you ticked any of the above boxes, I will always share a link to the video in the PR.

PR #817, opened 2023-04-18
Support native zlib-ng via flate2's zlib-ng feature

Byron:
In addition to the zlib-ng-compat feature, flate2 now supports zlib-ng
via its native API and the libz-ng-sys crate. This avoids potential
symbol conflicts with dependencies on system C libraries that link to
system zlib. Using zlib-ng increases compatibility, avoids those
conflicts, and allows using zlib-ng in more circumstances.

(zlib-ng-compat is still useful when you have dependencies that need the
zlib C ABI and don't support zlib-ng, such as curl; git-transport still
uses zlib-ng-compat.)

----
**Rebased follow-up of**: #434

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,6 +107,7 @@ check: ## Build all code in suitable configurations
 			   && cargo check --features io-pipe \
 			   && cargo check --features crc32 \
 			   && cargo check --features zlib \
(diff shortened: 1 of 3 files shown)


joshtriplett:
Awesome! I'm glad this worked out.

PR #820, opened 2023-04-19
Port to rustix

joshtriplett:
Port the termios usage from nix to rustix.

Change `RestoreTerminalStateOnDrop` to a wrapper that owns the `File`.

----

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2073,8 +2073,8 @@ dependencies = [
  "gix-command",
  "gix-config-value",
  "gix-testtools",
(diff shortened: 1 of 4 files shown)


joshtriplett:
This needs a fix in rustix to work on macOS. Working on that fix in https://github.com/bytecodealliance/rustix/pull/614 .

Byron:
Thanks a lot for introducing `rustix`!

It was also my feeling that it is more future proof than `nix`, so I am glad to make the switch.

PR #824, opened 2023-04-22
Move to the smaller, cargo-team maintained `home` crate

utkarshgupta137:
----

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -949,16 +949,7 @@ version = "4.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ca3aa72a6f96ea37bbc5aa912f6788242832f75369bdfdadcb0e38423f100059"
 dependencies = [
- "dirs-sys 0.3.7",
-]
-
-[[package]]
-name = "dirs"
-version = "5.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dece029acd3353e3a58ac2e3eb3c8d6c35827a892edc6cc4138ef9c33df46ecd"
-dependencies = [
- "dirs-sys 0.4.0",
+ "dirs-sys",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


Byron:
That's awesome, thanks a lot!

I wonder if [this implementation in `gitoxide`](https://github.com/Byron/gitoxide/blob/5d2b5d02c3869e07dc2507a8f2519ee1df633df7/gix-path/src/env/mod.rs#L103-L125) would also be suitable to not even have to use the `home`. It seems that `home` uses the winapi which should make it more resilient, but then again, it seems that [the respective `git` implementation](https://github.com/git/git/blob/7580f92ffa970b9484ac214f7b53cec5e26ca4bc/compat/mingw.c#L2714) is worse than even the one of `gix-path` which handles more special cases.

So I am thinking that `home` can be removed entirely in favor of `gix-path`. What do you think?

utkarshgupta137:
Well, the [`home_dir`](https://doc.rust-lang.org/std/env/fn.home_dir.html) function in std was deprecated for using `HOME` on Windows. The git implementation you linked to seems to be only used under MinGW? In that case, it would be fine. If this crate is supposed to work outside MinGW/Cygwin, then I think it would be better to update gix-path too to use the `home` crate.

Byron:
That's a very valid point - I tend to make the mistake of looking at this minGW code and thinking it's for all of windows, even if it's still 'just' a compatibility layer of sorts which doesn't exist for applications using `gitoxide`.

Thus I agree, and will be using `home` in `gix-path` accordingly.

Byron:
I had to revert the change to `gix-path` as it is used in the WASM builds. Apparently [`home` doesn't yet build there](https://github.com/Byron/gitoxide/actions/runs/4772443930/jobs/8484956259). `gix-sec` seems to not be used there, as right now only a few crates are validated to work in this environment.

utkarshgupta137:
> I had to revert the change to `gix-path` as it is used in the WASM builds. Apparently [`home` doesn't yet build there](https://github.com/Byron/gitoxide/actions/runs/4772443930/jobs/8484956259). `gix-sec` seems to not be used there, as right now only a few crates are validated to work in this environment.

You could use the `HOME` variable for wasm targets & `home_dir` for everything else?

Byron:
Could you open a new PR and make it work there? Maybe it's as easy as using `home` on windows and asking for `HOME` everywhere else.

I am also investigating a failure in my current PR which seems to be due to `home` even though it shouldn't be used there, very puzzling.

utkarshgupta137:
> Could you open a new PR and make it work there? Maybe it's as easy as using `home` on windows and asking for `HOME` everywhere else.
> 
> I am also investigating a failure in my current PR which seems to be due to `home` even though it shouldn't be used there, very puzzling.

Created #826

PR #826, opened 2023-04-22
fix: use `home` in `env::home_dir()`

utkarshgupta137:
----


Byron:
Thanks a lot for the effort, it's much appreciated and helped me a lot.

PR #835, opened 2023-04-27
Add feature vendored-openssl which enables `crates-index/vendored-openssl`

NobodyXu:
Signed-off-by: Jiahao XU <Jiahao_XU@outlook.com>

----

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3181,6 +3181,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ff011a302c396a5197692431fc1948019154afc178baf7d8e37367442a4601cf"
 
 [[package]]
+name = "openssl-src"
+version = "111.25.3+1.1.1t"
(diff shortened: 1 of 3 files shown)


Byron:
I appreciate the initiative.

`max-pure` doesn't exist in `gix` though, and I wonder where `openssl` is supposed to come from since `gix` doesn't depend on anything without enabling additional features. There is also no need to enable support for HTTP.

```
# cargo tree -p cargo-smart-release -i openssl-sys
openssl-sys v0.9.86
└── libgit2-sys v0.14.2+1.5.1
    └── git2 v0.16.1
        └── crates-index v0.19.8
            └── cargo-smart-release v0.19.0 (/Users/byron/dev/github.com/Byron/gitoxide/cargo-smart-release)
```

As you can see, there is no need to make changes as `openssl` comes in with another dependency.

NobodyXu:
@Byron Thanks for pointing out!
I've changed this PR to contain a new feature `vendored-openssl` which enables `crates-index/vendored-openssl`.

PR #840, opened 2023-05-03
try `moonwalk` in `ein`

Byron:
`moonwalk` could speed up `ein tool find (and related) ` significantly,
and we give that an early shot here.

It turned out that `jwalk` was constrained from old times and unlocking it for all cores already makes a difference.
Can `moonwalk` be even faster? We will find out.

### Tasks

* [x] unleash `jwalk` for comparison
* [x] integrate `moonwalk`, loose sorting for now

### Problems

Without inspecting siblings its not easy to prevent it from, in parallel, descend into git repositories. Maybe synchronization on the user's side would work, but it seems hard.
If that would be solved, it would absolutely kill, for sure, as it perfectly saturates IO.

### DIFF
--- a/gitoxide-core/Cargo.toml
+++ b/gitoxide-core/Cargo.toml
@@ -16,7 +16,7 @@ default = []
 
 #! ### Tools
 ## Discover all git repositories within a directory. Particularly useful with [skim](https://github.com/lotabout/skim).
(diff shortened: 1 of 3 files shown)


Byron:
It's probably going to take while until `moonwalk` gets ready. We'll be back :).

PR #854, opened 2023-05-17
pack traverse memory

Byron:
- further optimize buffer usage when traversing packs
- feat: Use `jemalloc` for consistent memory performance


Byron:
merged

PR #863, opened 2023-05-24
Include license files in all crates that are published to crates.io

blinxen:
I want to package the `gix*` crates for fedora because [helix](https://github.com/helix-editor/helix/blob/master/helix-vcs/Cargo.toml#L20) depends on the `gix` crate and I would like to include `helix` to fedora. Looking at your crates I see that almost none of them contain license files. Is that on purpose? This PR tries to fix this by adding symlinks and including the files in `Cargo.toml` where necessary.

### DIFF
--- a/gitoxide-core/LICENSE-APACHE
+++ b/gitoxide-core/LICENSE-APACHE
@@ -1,0 +1,1 @@
+../LICENSE-APACHE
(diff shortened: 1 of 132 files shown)


Byron (approved):
Thanks for your help with this, it's much appreciated!

> Is that on purpose?

It's an oversight and I am glad its fixed.

PR #864, opened 2023-05-24
chore: inline format args

nyurik:
This is the result of running

```shell
cargo clippy --benches --examples --tests --lib --bins --workspace --fix \
          -- -A clippy::all -W clippy::uninlined_format_args
```

This should make reading code a bit simpler, and makes it a bit more concise (removes 25 lines)

It was not working as is until Ifixed the test file paths (I will submit this as a separate PR):

```diff
@ gix-features/src/zlib/stream/deflate/tests.rs:78 @ mod deflate_stream {
-            "../../../../tests/fixtures/objects/pack/pack-11fdfa9e156ab73caae3b6da867192221f2089c2.pack"
+            "../../../../../gix-odb/tests/fixtures/objects/pack/pack-11fdfa9e156ab73caae3b6da867192221f2089c2.pack"

@ gix-features/src/zlib/stream/deflate/tests.rs:92 @ mod deflate_stream {
-            "../../../../tests/fixtures/objects/pack/pack-11fdfa9e156ab73caae3b6da867192221f2089c2.pack"
+            "../../../../../gix-odb/tests/fixtures/objects/pack/pack-11fdfa9e156ab73caae3b6da867192221f2089c2.pack"
        );
```


Byron (approved):

I am trying to figure out what's happening with CI here as it now seems to fail consistently on one test, which unfortunately doesn't reproduce locally. Once I fixed that, I will happily merge this PR.

PR #865, opened 2023-05-25
Make cargo check to pass all tests

nyurik:
This command was failing:

```bash
cargo check  --benches --examples --tests --lib --bins --workspace
```


Byron (approved):

I am trying to figure out what's happening with CI here as it now seems to fail consistently on one test, which unfortunately doesn't reproduce locally. Once *I fixed that*, I will happily merge this PR.

PR #866, opened 2023-05-25
fix(docs): fix docs generation

nyurik:
URLs need to be escaped. Plus added doc build to CI without deps


Byron (approved):

I am trying to figure out what's happening with CI here as it now seems to fail consistently on one test, which unfortunately doesn't reproduce locally. Once *I fixed that*, I will happily merge this PR.

PR #867, opened 2023-05-25
Include custom clippy settings

nyurik:
Help needed: no idea why tests fail in all of my PRs :(

* All custom clippy settings are now centralized
* add `make fix-clippy` for automated fixes
* Modified some simple cases, e.g. use `_` instead of `true|false` in a match

Unrelated to this specific change, but I would highly recommend to use [justfile](https://github.com/casey/just#readme) -- much cleaner and easier.

Also, I think it would be better to use `cargo -p <project>` in the makefile and elsewhere - this way it will be cleaner to repeate the broken CI step without the `cd ...`, and also will not require the test to contain some non-buildable paths - see #865

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,6 @@
-# TODO: consider adding these clippy fixes and/or some other?
-#	 \
-#	-D  \
-#	-D clippy::range-plus-one \
-
 # Rejected for now, and why
 #	-D clippy::default-trait-access - sometimes makes imports necessary, just for a default value. It's good for more explicit typing though.
+#   -D clippy::range-plus-one - useful, but caused too many false positives as we use range types directly quite a lot
 CLIPPY_FLAGS = \
 	-D clippy::unnested-or-patterns \
 	-D clippy::explicit-iter-loop \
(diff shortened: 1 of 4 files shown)


Byron (approved):

I am loving the new clippy target and will merge once having sorted the `makefile` TODO. Also I will re-evaluate the `justfile`, even though last time I still preferred the simplicity and ubiquitousness of makefiles.

Byron:
Regarding `just`, I think it *could* work and I am keen to convert, despite the general shortcoming that it isn't installed automatically, but installing it should help to get there in the future :).

I check the manual to get some issues clarified:

* follow [these steps](https://just.systems/man/en/chapter_63.html) to get completions
* see how [it would get onto CI](https://just.systems/man/en/chapter_6.html)
* figure out how to change the default action to do a listing of possible commands.

It's a bit sad to see that the listing isn't as pretty as the one-liner that is used here with `make` :/, but I find comfort that I can contribute headlines myself once I am fed up enough 😁.

Screenshot 2023-05-25 at 10 43 12
Screenshot 2023-05-25 at 10 45 50

Yes, so I think I will have some fun porting the `makefile` to a `justfile` - thanks for reminding me!

Byron:
Thanks again for all your work here!

I will see if *not* changing directory `justfile` makes it more readable, but when trying here I thought it was preferable still. I have ideas though and think this can get better no matter what.

I will start a new PR for transcribing the makefile in a moment right after merging once CI is green, excitement!

PR #872, opened 2023-05-26
chore: fmt justfile

nyurik:
Ran `just --fmt --unstable` and added it to the `just fmt`

BTW, why is it not being used? When running `just fmt`, lots of changes are made?

P.S. Nice job with the justfile!!! :)

### DIFF
--- a/justfile
+++ b/justfile
@@ -3,7 +3,7 @@
 #   like a script, with `./justfile test`, for example.
 
 default:
-    {{just_executable()}} --list
+    {{ just_executable() }} --list
 
 alias t := test
 alias c := check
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot, it's good to see using `just` already pays. It also made me realize that the remaining `makefile` is probably happier as bash script, as right now it feels more like `make` was abused to get some parallelism and and do some file checks, for the cost of syntax which might be less readable than it could be.

> BTW, why is it not being used? When running `just fmt`, lots of changes are made?

It's run occasionally, ideally each time a PR is merged, but I keep forgetting 😅.

> P.S. Nice job with the justfile!!! :)

Thanks for making me! It felt like spring cleaning :).

PR #873, opened 2023-05-26
Apply -W clippy::implicit-clone and docs

nyurik:
Ran `just 'clippy-flags=-W clippy::implicit-clone' clippy-fix`

It auto-fixed a few related issues. Also sorted clippy-flags

### DIFF
--- a/cargo-smart-release/src/changelog/merge.rs
+++ b/cargo-smart-release/src/changelog/merge.rs
@@ -34,7 +34,7 @@ impl ChangeLog {
                 } => Some((idx, heading_level, version_prefix)),
                 _ => None,
             }) {
-                Some((idx, level, prefix)) => (idx, *level, prefix.to_owned()),
+                Some((idx, level, prefix)) => (idx, *level, prefix.clone()),
                 None => {
                     sections.extend(sections_to_merge);
                     return Ok(self);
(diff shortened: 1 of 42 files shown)


Byron (approved):
Thanks a lot - it's good to see that upon review, `to_owned()` can be removed completely or be replaced with `*thing` instead for copyable types.

It's a bit sad for me to see `to_owned()` go as it's a deliberate choice showing how I am thinking about what's going on there - 'making it owned' is what I want, with the side-effect of a clone, whereas `clone()` has a side-effect that I get something to own. In terms of complexity, it's probably simpler to just use `clone()` where possible though, so I am fine it especially if pedantic clippy thinks that this is more idiomatic 😁.

Byron:
So it looks like clippy is ping-ponging between `to_path_buf()` and `clone()` now :D - I have undone my previous quick-fix and leave taming it to you :).

nyurik:
Hi, thx for considering! Yeah, this is a weird one - I actually can't figure out why loose/iter is even complaining... perhaps there is a clippy bug?  (I have been fixing those sporadically)...  Regardless - if you feel `to_owned` here expresses your intent better, i will let it be as is

Byron:
> Regardless - if you feel to_owned here expresses your intent better, i will let it be as is

I can see how using less method names to effectively achieve the same thing is beneficial. Want to own something? Clone(). Want to copy something? *Something. That's it - previously we also had `to_owned()` mixed in and it tends to be more wordy as well. So let's go with what seems to be easier.

Byron:
As much as I appreciate the initiative, I keep looking at the PR and it takes attention.
Thus I am closing the PR now and hope it will be reopened once there is time to make progress.
Thanks a lot for your understanding.

PR #875, opened 2023-05-26
Apply -W clippy::cloned-instead-of-copied

nyurik:
Ran `just 'clippy-flags=-W clippy::cloned-instead-of-copied' clippy-fix`

Also sorted clippy-flags


Byron (approved):
Thanks a lot! Looks great!

PR #876, opened 2023-05-26
Cleaning up documentation

nyurik:
Applied this lint:

```
just 'clippy-flags=-W clippy::doc-markdown' clippy-fix
```

I also had to search/replace:

```
[`to_decimal`()]  ->  [`to_decimal()`]
```

I have never seen parenthesis used as part of the link, so this is perhaps why clippy handles it like that.

There are still a few warnings being produced, but they have to do with how some of the links are constructed, e.g. this link

```
[File::index_names().len()][File::index_names()]
```

shows a warning

> you should put bare URLs between `<`/`>` or make a proper Markdown link

Not certain how these should be fixed, but there are 10-20 of such warnings.  What are you linking to here?

### DIFF
--- a/gitoxide-core/src/hours/mod.rs
+++ b/gitoxide-core/src/hours/mod.rs
@@ -31,9 +31,9 @@ pub struct Context<W> {
 /// Estimate the hours it takes to produce the content of the repository in `_working_dir_`, with `_refname_` for
 /// the start of the commit graph traversal.
 ///
-/// * _working_dir_ - The directory containing a '.git/' folder.
-/// * _refname_ - The name of the ref like 'main' or 'master' at which to start iterating the commit graph.
-/// * _progress_ - A way to provide progress and performance information
+/// * `_working_dir_` - The directory containing a '.git/' folder.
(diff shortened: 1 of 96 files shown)


Byron (approved):
Incredible, I wasn't aware of all the tricks `clippy` has up its sleeves, and appreciate all the improvements that are made by it (and the industrious human who controls all that, thank you :)).

I managed to find a couple of other issues around these `function(<ellipsis>)` doc strings, and am happy about these massive improvements.

nyurik:
thx @Byron !  Do you actually want to include parens everywhere in the docs?  I haven't seen many usages like that. Most of the time I see either `[Some::Path::to_function]` or the Markdown style links like `[...text...](...link...)`

Byron:
The reason I do this is that I'd like to differentiate between fields and methods. Admittedly, most of the time the name takes care of that, but I felt like not taking chances.

If the current style disturbs tooling too much, I think that's enough to remove `(…)` and `()` in links even though I'd love to have it styled so that their nature is clear.

PR #878, opened 2023-05-30
Include missing changelog file in some crates

blinxen:
While packaging I noticed that some crates do not include their changelog file. Is that on purpose? I would like to have it included for the sake of completeness :D.


Byron:
Thanks a lot, it's a great catch as it's definitely not on purpose 😅.

PR #884, opened 2023-06-07
fix negotiation

Byron:
Seeing READY means a pack will follow, which is exactly the kind of knowledge we want.
Thus we now either listen to the client OR to what the server says.

This remedy relies on multi-ack-detailed though, so let's hope nobody with a very old server
comes along.


Byron:
Already merged via rewritten commit.

PR #887, opened 2023-06-07
traverse-with-commitgraph

Byron:
Refactor `gix-traverse` commit walks to provide more information about the commits it sees
and also support using the commitgraph for the same.

This can trickle up to `gix` which will yield certain information faster, naturally, without the need for
specialized walks for many cases, leveraging the features that are already present.

### Tasks

* [x] docs to help with what's currently there
* [x] rigorous cleanup and support for additional Commit information
* [x] `gix-traverse` support for commitgraphs
* [x] make the commit-graph easily usable from the `gix` iteration platform
* [x] `gix` uses new `gix-traverse` API for commit walks and wraps these.

Related to #853 

### Follow-up

Special-case for author-computation with commit-graph in `ein t h`. Only if one exists one (and we don't compute line/file changes), we can offload all work to a couple of threads.


Byron:
merged (with slightly different commits)

PR #891, opened 2023-06-10
improve tests related to the handling of shallow repos

Byron:
When commit-graphs are used, it's easily possible for it to point to commits that don't exist anymore
as the repo has been shallowed after generating it. Also we should be sure that we can deal with mixed
commit-graph present/not-present situations.


Byron:
merged with slightly different commit

PR #894, opened 2023-06-11
chore: Add `clippy::redundant-closure-for-method-calls` lint

poliorcetics:


gix-revision/src/spec/parse/function.rs
```diff
@@ -213,7 +213,7 @@ fn long_describe_prefix(name: &BStr) -> Option<(&BStr, delegate::PrefixHint<'_>)
         .and_then(|generation| {
             iter.next().map(|token| {
                 let last_token_len = token.len();
-                let first_token_ptr = iter.last().map_or(token.as_ptr(), |token| token.as_ptr());
+                let first_token_ptr = iter.last().map_or(token.as_ptr(), <[_]>::as_ptr);
```
Byron, line 216: Clearly, black magic ;), and amazing clippy is able to propose this.

Byron (approved):
Thanks for setting this up, much appreciated!

I'd love to have a way to make these clippy/check settings more general, too, and it's great to see there is a way in case this should become pressing.

For now I'd like to wait until there is a way to officially place these settings into some `Cargo.toml` instead of the `.config/cargo.toml` settings file which seems quite local (to me, at least).

poliorcetics:
>  instead of the `.config/Cargo.toml` settings file which seems quite local 

It's project local yes, isn't that what we want here ?

Byron:
What I mean is that I use this file to setup personal overrides that aren't supposed to be checked-in. These days it has mostly been `[patch."crates-io"]` sections, and I haven't used these in a while.

So maybe it's time to admit that centralising the warning/deny configuration is more important than me not risking to check-in personal edits to that file occasionally or rarely even.

Would you mind opening a PR that moves the extra-flags from the `justfile` to `.cargo/config.toml`?

poliorcetics:
Done in https://github.com/Byron/gitoxide/pull/898 :)

PR #898, opened 2023-06-13
chore: move clippy flags to .cargo/config.toml

poliorcetics:
This allows tools like `rust-analyzer` and regular `cargo clippy` calls to pick them up.

The same will work for `rustc` and `rustdoc` lints too, provided any are added.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,23 @@
+[target.'cfg(any(test, not(test)))']
+rustflags = [
+    # Rustc lints
+    # "-W", "warning_name"
+    # "-A", "warning_name"
+
+    # Clippy lints
+    "-W", "clippy::cloned_instead_of_copied",
+    "-W", "clippy::explicit_iter_loop",
+    "-W", "clippy::map_unwrap_or",
+    "-W", "clippy::redundant_closure_for_method_calls",
+    "-W", "clippy::unnested_or_patterns",
+    "-W", "clippy::uninlined_format_args",
+
+    # Rejected for now, and why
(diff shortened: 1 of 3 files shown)


.cargo/config.toml
```diff
@@ -0,0 +1,23 @@
+[target.'cfg(any(test, not(test)))']
```
Byron, line 1: I was already wondering last time I saw this, but put it off as example.
Now that what seems like a noop/`if true` kind of statement appears again, I wonder if it has a particular effect that I am unaware of?

Byron (approved):
That's awesome, thanks so much for contributing once more!

I assume the line `[target.'cfg(any(test, not(test)))']` is to signal that the configuration could be gated as well, without really having an effect right now, right?

.cargo/config.toml
poliorcetics: there is no `cfg(true)` in Rust so the only way to do it is to say `A or not(A)`, which this is doing.

Theorically I could do `cfg(not(this_cfg_does_not_exist))` but then someone could pass `--cfg this_cfg_does_not_exist` to cargo and the flags wouldn't be picked up: this is not possible when the condition is the one I used

Byron: And removing that line should work the same, right? (*even though I have a feeling that it's significant*)

poliorcetics: No, it wouldn't work at all I think, I could also have used `[build]` but `rustflags` needs to be in either a `target.<...>` or `build` table

poliorcetics: I'll check with `[build]`, it'll be simpler, I shouldn't have copied my conditions from another project

PR #901, opened 2023-06-14
use [build] instead of [target.<...>]

poliorcetics:
Ref https://github.com/Byron/gitoxide/pull/898

I checked it still works by commenting out an `#[allow(clippy::manual_map)]` and running `just clippy`

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,4 +1,4 @@
-[target.'cfg(any(test, not(test)))']
+[build]
 rustflags = [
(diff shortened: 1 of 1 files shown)


poliorcetics:
Ah, found it, I was using `target.` in my other project because I have other `[target.<...>]` tables and https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags says it only conflates flags if they're all `target`, not from both `target` and `build`.

This is not a problem here so the change is ok :)

PR #902, opened 2023-06-17
`gix corpus` improvements

Byron:
Tasks
=====

* [x] refactor tracing API to allow avoiding explicit drops
* [x] allow to record field values later
* [x] respect `--trace` even in corpus run (right now it overrides the dispatch impl)
* [x] `--dry-run` that just prints the selected repos once
* [x] `--repo-sql-suffix` and `--include-task`
* [x] an exclusive corpus task to verify the ODB
* [x] a corpus task to count packed objects


Byron:
Already merged with minor changes.

Issue #874, opened 2023-05-26
OCD: Satisfy some pedantic clippies


nyurik:
This is a (infrequently updated) list of the failing pedantic issues. Some probably should not be implemented.

To regenerate, use this bash command with `jq`:

```bash
cargo clippy --all-targets --workspace --message-format=json --quiet -- -W clippy::pedantic \
   | jq -r '.message.code.code | select(. != null and startswith("clippy::"))' \
   | sort | uniq -c | sort -h -r
```

- [ ] x709 `clippy::default_trait_access`
- [ ] x552 `clippy::doc_markdown` #876
- [x] x457 `clippy::semicolon_if_nothing_returned` #1546
- [ ] x447 `clippy::missing_panics_doc`
- [ ] x216 `clippy::cast_possible_truncation`
- [ ] x205 `clippy::needless_pass_by_value`
- [ ] x204 `clippy::return_self_not_must_use`
- [ ] x169 `clippy::unreadable_literal`
- [ ] x164 `clippy::items_after_statements`
- [ ] x162 `clippy::single_match_else`
- [ ] x161 `clippy::too_many_lines`
- [ ] x125 `clippy::module_name_repetitions`
- [ ] x110 `clippy::unnecessary_wraps`
- [ ] x99 `clippy::match_same_arms`
- [ ] x91 `clippy::cast_lossless` #1547
- [ ] x80 `clippy::ignored_unit_patterns`
- [ ] x75 `clippy::used_underscore_binding`
- [ ] x75 `clippy::needless_raw_string_hashes`
- [ ] x70 `clippy::implicit_clone` #873
- [ ] x64 `clippy::manual_let_else`
- [ ] x56 `clippy::cast_precision_loss`
- [ ] x43 `clippy::trivially_copy_pass_by_ref`
- [ ] x42 `clippy::redundant_else`
- [ ] x37 `clippy::if_not_else`
- [ ] x35 `clippy::match_wildcard_for_single_variants`
- [ ] x35 `clippy::cast_sign_loss`
- [ ] x32 `clippy::similar_names`
- [ ] x29 `clippy::struct_excessive_bools`
- [ ] x26 `clippy::cast_possible_wrap`
- [ ] x24 `clippy::explicit_iter_loop`
- [ ] x22 `clippy::explicit_into_iter_loop`
- [ ] x22 `clippy::explicit_deref_methods`
- [ ] x18 `clippy::inconsistent_struct_constructor`
[...]

PR #903, opened 2023-06-18
clippy: activate `clippy::enum-glob-use`

poliorcetics:
Helps with #874

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -5,6 +5,7 @@ rustflags = [
 
     # Clippy lints
     "-W", "clippy::cloned_instead_of_copied",
+    "-W", "clippy::enum_glob_use",
     "-W", "clippy::explicit_iter_loop",
     "-W", "clippy::map_unwrap_or",
     "-W", "clippy::redundant_closure_for_method_calls",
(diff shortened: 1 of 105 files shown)


poliorcetics:
I ran `just fmt` locally and it changed lots of imports, was that expected ?

poliorcetics:
The job `cargo-deny (bans licenses sources)` failing seems unrelated ? I added no deps nor changed any

Byron:
That should indeed be unrelated, I will see to fix it in `main` if it persists.

> I ran `just fmt` locally and it changed lots of imports, was that expected ?

Yes, as I run it only rarely and usually forget. Ideally it's run after each work-block, and last time I forgot.

Besides that, thanks a lot for tackling this, it's appreciated even though I have to postpone the review as I am still recovering from COVID. I'd hope I have the focus for that within a a couple of days though.

Byron:
I am sorry, but have to reject this one, as I am not a fan of not being allowed to use function-private enum-globs. They actually make the code more readable which becomes obvious now that I see it.

Secondly, `just fmt` shouldn't be mixed in which seems to be the case here - this is merely for future reference.

PR #919, opened 2023-07-01
integrate gix-filter

Byron:
Follow up of #914.

Filter pipelines and `gix` integration.

### Tasks

We should also think about `diffs` and if these need adjustments to diff what's in the worktree and not what's stored in git. Have to read up on what's desirable here.

* [x] pipelining support with complete configuration and streaming support where it matters
* [x] `gix` can configure filter pipelines, and uses trust levels to skip untrusted filters.
* ~~`gix` makes it easy to obtain a filtered version of object content, with streaming support.~~ - let's wait until we actually need such an API - filter pipelines are relatively low-level and for use with more than one file.
* [x] worktree checkouts use filters and support streaming.
* [x] execute worktree tests in nextest
* [x] support for delayed filter processing
* [x] delayed filter processing with multi-threading
* [x] set the driver context to something reasonable in `gix`
* ~~maybe an option to diff filtered content?~~ Seems like there is no use for that now, but it can be added whenever that changes
* [x] update `libgit2` API comparison

### Next PR

* basic `gix-archive` implementation that makes use of filter pipeline.


Byron:
Merged to main with minor changes.

PR #927, opened 2023-07-09
gix-archive 

Byron:
Follow-up of #919 

### Tasks

* [x] a full implementation of the `gix-archive` API
* [x] move `gix-archive` to `gix-worktree-stream` 
* [x] `gix-archive` with internal format only,and previously removed options
* [x] `gix-archive` with opt-in support for basic `tar`
* [x] integration into `gix` to easily create streams
* [x] integration into `gix` to easily create archives with progress support and interruptions
* [x] the `gix archive` command-line tool - MVP
* [x] `gix-archive` with opt-in support for `zip` and `tar.gz`
* [x] zip compression factor on CLI
* [x] `gix archive` with additional entries (from file and from string)
* [x] maybe a `corpus` task that archives `HEAD`?

Compare single-threaded performance and see if multi-threading would be worth the effort. Prefer not to invest more time than necessary though, but 5s to 6s for 1.3GB are to be beaten 16k files/s.

### Performance Comparison

It's a bit faster when handling the linux kernel, both in `.stream` and `.tar` mode, but about 1s slower when dealing with `WebKit`. System time is about 2s higher than git, and user time about 1s.
With some tuning, uncompressed tar output is now 23% faster, and compressed output is about 18% faster (even though the compressor seems slower and yields worse compression [it's not configured at all though]).

`zip` seems to be the winner, as under the hood it gets to use `zlib-ng` which by itself would be a massive speed boost. Thanks to the architecture, we also get to use more CPU in parallel, which leads to a whopping 1,9x speedup for `zip` archives as tested on WebKit. The produced container sizes are the same at 1.3GB.


Byron:
Merged already, but with changed last commit. Closing.

PR #932, opened 2023-07-16
A couple examples

metatoaster:
This includes a couple examples based on the `git ls-tree` and `git log` command.  I would have loved to include the `--patch` flag, however I am completely unfamiliar with the `imara-diff` library for generating diffs so I am going to leave that for someone else who might be more familiar with this, but I hope this is a good enough start to get that going.

I did include `structopt` as an additional `dev-dependencies` to make passing CLI arguments more ergonomic.


Byron (changes requested):
Thanks a lot for helping with the examples, it's much appreciated!

For this round, I only took a cursory look and liked what I saw. However, there are a couple of bigger requests I need to make before I can wave this through.

* Please do not start commits with `gix:` - this will make `cargo smart-release` unhappy. If it's a feature you'd like to advertise in the changelog, the `feat: ` prefix can be used.
* Please do not use `structuopt` - it has been integrated into `clap` for a while now and is used here already. Please take a look at `src/plumbing/options/mod.rs` to get an idea how to do this.

When done, I think I can make all remaining changes myself and merge.

Thanks a lot!

metatoaster:
Thanks for the very prompt review. I reworked (rebased the changes to where they belong, including the `clap` usage) onto every commit to ensure they can bisect cleanly with the current clippy settings (wait, guess there's more, I will get this done, probably should have read the full thing), and in the process I also fixed a minor bug that I didn't test with the latest changeset (left out the parentheses for the `||` of the `is_empty()` condition...).

Byron (approved):
Thanks a lot for the fixes!

I decided to put the examples to the top-level where they can use `clap` naturally, which is also a nice differentiator between tiny examples in `gix` and possibly larger examples in `gitoxide`.

In any case, once CI is done this PR can be merged. Thanks again for contributing, I hope you have more ideas for examples and find time (and motivation) to implement them :).

PR #946, opened 2023-07-24
Fix build on  by disabling sha1-asm

mati865:
Hello,
We have found `sha1-asm` to break build on i686-pc-windows-gnu and i686-pc-windows-gnullvm targets. Unfortunately `sha1-asm` crates are in maintenance mode so they are unlikely to be fixed: https://github.com/RustCrypto/asm-hashes/issues/60

EDIT: I'll hold back to see if upstream bug gets fixed.

### DIFF
--- a/gix-features/Cargo.toml
+++ b/gix-features/Cargo.toml
@@ -150,9 +150,9 @@ libc = { version = "0.2.119" }
 bstr = { version = "1.3.0", default-features = false }
 
 
-# Assembly doesn't yet compile on MSVC on windows, but does on GNU, see https://github.com/RustCrypto/asm-hashes/issues/17
+# Assembly doesn't yet compile on MSVC and x86 GNU on windows, but does on x86_64 GNU, see https://github.com/RustCrypto/asm-hashes/issues/17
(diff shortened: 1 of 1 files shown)


Byron:
Thanks for the fix! This configuration of `sha1-asm` is the biggest technical debt (that I am aware of) as it has the potential for failing builds while continuously giving me the feeling that this line isn't really working.
But maybe it is now, and I am looking forward to your feedback to see if it worked or not.

Ideally, of course, this CI here would be able to test for the issue, so if you have any hints on that it would be appreciated also.

Thanks for letting me know if it worked or not.

ognevnydemon:
the bug got fixed, so now you can just bump sha1-asm version

mati865:
There is no action required on `gitoxide` side since new versions are semver compatible. The update should happen in starship repository.

I'm hoping to take care of the CI within next two weeks (unless somebody beats me to it, which I wouldn't mind), probably in similar fashion as in https://github.com/RustCrypto/asm-hashes/pull/61

Byron:
I will be looking forward to your help with this due to lack of experience on my side, but will make sure that the PR is merged quickly along with a release afterwards. Thanks in advance.

PR #947, opened 2023-07-24
chore(release): Upgrade deps to get latest `winnow`

epage:
This is prep for porting from `nom` to `winnow` (with Byron's consent).

My expectation was to reduce duplicate dependencies in the workspace but I ended up duplicating `indexmap`.  No winning :)


epage:
Turns out I was on an old `main`

PR #951, opened 2023-07-27
Migrate config from nom 7 to winnow 0.5

epage:
Inspired by the talk of improvement in #948, between winnow 0.4 and 0.5, I refactored the code to make the migration easier.  After that upgrade, I went and did what felt like further improvements to me.

While I didn't benchmark, a couple of potential performance improvements in those changes include
- Moving the counting of newlines from the "hot loop" to only in  the error path.
- Removing the counting of indexes in `value_impl`, making it so we just do math when we need it

Further improvements that I held off on for time
- Switching the input type while parsing internals to `winnow::BStr` which makes debug output a lot nicer (especially when running with `cargo test -F winnow/debug`)
- Track the `ParseNode` in the context of a custom error type
- Move `dispatch` from being an explicit parameter to a part of the input type

By adjusting how those two other `&mut` parameters are handled, all of the functions would follow the normal parser structure / impl `Parser` which would make them easier to work with.

### DIFF
--- a/gix-config/src/parse/nom/mod.rs
+++ b/gix-config/src/parse/nom/mod.rs
@@ -3,16 +3,13 @@ use std::borrow::Cow;
 use bstr::{BStr, BString, ByteSlice, ByteVec};
 use winnow::{
     branch::alt,
-    bytes::complete::{take_till, take_while},
-    character::{
-        complete::{char, one_of},
-        is_space,
-    },
+    bytes::{one_of, take_till0, take_while0},
     combinator::opt,
     error::{Error as NomError, ErrorKind},
     multi::{fold_many0, fold_many1},
     prelude::*,
     sequence::delimited,
+    stream::AsChar,
 };
 
 use crate::parse::{error::ParseNode, section, Comment, Error, Event};
@@ -76,7 +73,7 @@ pub fn from_bytes<'a>(input: &'a [u8], mut dispatch: impl FnMut(Event<'a>)) -> R
 
 fn comment(i: &[u8]) -> IResult<&[u8], Comment<'_>> {
     let (i, comment_tag) = one_of(";#")(i)?;
-    let (i, comment) = take_till(|c| c == b'\n')(i)?;
+    let (i, comment) = take_till0(|c| c == b'\n')(i)?;
     Ok((
         i,
         Comment {
@@ -140,10 +137,10 @@ fn section<'a>(i: &'a [u8], node: &mut ParseNode, dispatch: &mut impl FnMut(Even
 fn section_header(i: &[u8]) -> IResult<&[u8], section::Header<'_>> {
     let (i, _) = '['.parse_next(i)?;
     // No spaces must be between section name and section start
-    let (i, name) = take_while(|c: u8| c.is_ascii_alphanumeric() || c == b'-' || c == b'.')(i)?;
+    let (i, name) = take_while0(|c: u8| c.is_ascii_alphanumeric() || c == b'-' || c == b'.')(i)?;
 
     let name = name.as_bstr();
-    if let Ok((i, _)) = char::<_, NomError<&[u8]>>(']')(i) {
+    if let Ok((i, _)) = one_of::<_, _, NomError<&[u8]>>(']')(i) {
         // Either section does not have a subsection or using deprecated
         // subsection syntax at this point.
         let header = match memchr::memrchr(b'.', name.as_bytes()) {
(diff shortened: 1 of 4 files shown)


Byron:
Thanks a million, and sorry for the late reply. I didn't have the focus time needed yet, but hope that will change soon so I can get this wonderful PR merged!

Byron:
Thanks so much for this wonderful PR! It's impressive how well `winnow` adapts and makes more readable even the most complex parts of the parser (I am looking at you, `value_impl` :)). But I am sure it's also due to your mastery of both `nom` and `winnow` that allows you to produce parsers that are using the `winnow` API perfectly. This makes me even more grateful that you chose to perform the upgrade yourself, as I know it will yield more maintainable and possibly better performing code while putting the bets on a crate that I think is better suited for the years to come.

When looking at the parsing code, it also becomes clear that a very conscious choice was made to get away from calling functions towards an explicit call to one of many possible ways a parser can be invoked. And even though I didn't understand by looking at `parse_peek` why it's called like that, I see where it's headed. (*Somehow I was expecting `peek` to be none-consuming, but could not see how that relates to the seeming difference in return type or how it's different from `parse_next()`*).

Besides that, there is not much more to say than: Thanks for the great work, and I am looking forward to the future PRs that continue (or even finish) the migration 🙏.

gix-config/src/parse/nom/mod.rs
```diff
@@ -158,303 +156,219 @@ fn section_header(i: &[u8]) -> IResult<&[u8], section::Header<'_>> {
...
-        Err(nom::Err::Error(NomError {
-            input: i,
-            code: ErrorKind::Eof,
-        }))
-    } else {
-        Ok((i, (v.as_bstr(), counter)))
-    }
+fn take_newlines1<'i>(i: &mut &'i [u8]) -> PResult<&'i BStr, NomError<&'i [u8]>> {
+    repeat(1.., alt(("\r\n", "\n")))
+        .map(|()| ())
```
epage, line 370: btw I figure this should be called out.

Nom's `many` always allocated into a `Vec`.

Winnow's `repeat` instead allocates into `trait Accumulate` which is kind of like `Default + Extend` with a `with_capacity` **but** we implement it for `usize` (to count parses) and `()` to ignore the output o the parser, removing the need for `count_` and `skip_` variants of these parsers.

`map(|()| ())` is a quick way of annotating the `repeat` which I feel is nicer than `repeat<_, _, (), _>` (or whatever the generic params are).  I'd love to have a `Parser::as_void` to make this more explicit but haven't figured out a way to do that.

epage:
> When looking at the parsing code, it also becomes clear that a very conscious choice was made to get away from calling functions towards an explicit call to one of many possible ways a parser can be invoked. 

To be clear, `parse_next` is the equivalent of doing a direct function call.  We just switched the built-in parsers from returning `impl FnMut` to `impl Parser`.  If nothing else, this helps avoid bugs where we accidentally return functions that don't match the blanket `Parser` impl which then can't be composed together anymore.

> And even though I didn't understand by looking at parse_peek why it's called like that, I see where it's headed. (Somehow I was expecting peek to be none-consuming, but could not see how that relates to the seeming difference in return type or how it's different from parse_next()).

`parse_next` accepts a `&mut Input`.  `parse_peek` instead has the old function signature (`fn(Input) -> Input`) so I felt that "peek" would help describe that it isn't modifying the existing parse state.  This is separate from the [`peek`](https://docs.rs/winnow/latest/winnow/combinator/fn.peek.html) combinator.  If people have ideas for better names, the earlier we fix them the better :)

gix-config/src/parse/nom/mod.rs
Byron: Thanks for pointing it out! It's definitely a major improvement already I can see why `winnow` is intriguing to (even) switch to.
And it's even *just* the beginning :).

Byron:
Triggered by the performance regressions in #956 I came back here to run the existing performance tests.

Here everything is how it should be, and it appears that `nom` is considered regressed now.

```
     Running benches/large_config_file.rs (/Users/byron/dev/github.com/Byron/gitoxide/target/release/deps/large_config_file-369d7e97635798b9)
GitConfig large config file
                        time:   [50.911 µs 50.925 µs 50.941 µs]
                        change: [+0.7544% +2.4336% +3.7484%] (p = 0.00 < 0.05)
Found 11 outliers among 100 measurements (11.00%)
  6 (6.00%) high mild
  5 (5.00%) high severe

Parser large config file
                        time:   [46.970 µs 47.393 µs 47.806 µs]
                        change: [+28.244% +31.972% +35.877%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

epage:
To verify, nom is the slower case here?

Byron:
Yes, the `nom` version of the parser was slower.

PR #953, opened 2023-07-29
build: rust-cache should happen after rustup

alexandervantrijffel:
As described on https://github.com/Swatinem/rust-cache, the Swatinem/rust-cache@v2 GitHub action should happen after installing Rust with rustup to use the current rustc version as the cache key.


Byron:
Absolutely, a great catch! Thanks so much!

PR #954, opened 2023-07-29
tests: Evaluate Cargo's target directory dynamically

niklaswimmer:
There are a few tests in Gitoxide that need to directly invoke the built binaries and previously this was achieved by hardcoding the path to those binaries to be in the local `target` directory. 

This PR changes ~~all~~ most (see below) cases to instead extract the target directory from the output of `cargo metadata`. This has the concrete advantage that the binaries can be found even if a developer (such as me) uses Cargo features like `CARGO_TARGET_DIR` to change the target directory to be somewhere else.

I realize that this might be kind of an edge case because not many people will just unconditionally set `CARGO_TARGET_DIR` to a central directory (like I do), but apart from the new dev dependency on `jq` (which is installed on all CI machines and probably most developer machines anyways) this change has no (to me obvious) disadvantages, so I went ahead and fixed the problem for everyone. Also, in the future this _might_ become more relevant gets https://github.com/rust-lang/rfcs/pull/3371 is accepted.

This change impacts
 - `make stress` (location of `gix`)
 - `make try-publish-all` (location of `cargo-smart-release`)
 - `just journey-tests-*` (location of `gix`, `ein`, `jtt` and `cargo-smart-release`)
 - `cargo test -p gix-prompt` (location of `use-askpass` and `credentials` examples)

The last two are commonly invoked via `just test`. I tested all those commands both with `CARGO_TARGET_DIR` set and unset and all worked (apart from `make stress` but that fails in both cases because it can't find `time` :shrug:).

There is one journey test for `cargo-smart-release` which interacts with the local target directory. I did not change that because I honestly don't quite understand what it does. It seems to be testing something with `.gitignore` which suggests to me that the test may depend on the assumption that there will always be a local target directory, even if their might be none?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,8 @@ debug-small: always ## minimal dependencies, at cost of performance
 
 ##@ Development
 
-target/release/gix: always
+gix := $(shell cargo metadata --format-version 1 | jq -r .target_directory)/release/gix
+$(gix): always
 	cargo build --release --no-default-features --features small
 
 ##@ Testing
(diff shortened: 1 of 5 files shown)


Byron:
Thanks a lot for this wonderful contribution, I appreciate it a lot!

It's also revealing an the issue of a 'catch-22' here. My preference would be to try to ditch `jq` and build a Rust program instead to extract the required information, but then how to get the executable path of said Rust program?
I checked and didn't see a way to tell `cargo metadata` to apply some JSON path right away, nor did I see a way to let `cargo build` output the main artifact path it produced. Too bad, so I think using `jq` is the best option for now.

Unless…thinking way outside of the box here… what if there was a tool like `jtt` that has the required features to extract information from `cargo metadata`, but is build explicitly using `cargo build --target-dir` to at least know the general location of the binary, which then could be searched. Then, with `jtt` one could replace `jq`. Yes, I know what you mean, it's an extra step and plenty of complexity particularly since 'searching the binary' isn't very portable, and `cargo build --out-dir` isn't stable yet even though it also would just *ease* the pain.

Oh well, let's live with it now and I for one am just thankful for the improved compatibility, while hoping that one day we can just do `cargo metadata --json-path .target_directory` and be independent of additional tooling.

niklaswimmer:
> Unless…thinking way outside of the box here…

I really like your thinking, although as you admit yourself, the added complexity is probably not worth in this case.

> while hoping that one day we can just do `cargo metadata --json-path .target_directory` and be independent of additional tooling

Ohhh yes that would be nice, well we will see I guess

PR #956, opened 2023-08-01
Finish migrating gitoxide from nom 7 to winnow 0.5

epage:
Follow up to #948 and #951

`gix-actor`, `gix-object`, `gix-ref`, and `gix-testtools` are interrelated, so this does it all at once

Things left out
- Removing `winnow` from the API of the above packages
- Switching `from_bytes` functions to using `parse` instead of (the implied) `parse_next`

BREAKING CHANGE: `nom`, and now `winnow`, is exposed in the public API of some of these packages

### DIFF
--- a/gix-actor/src/signature/decode.rs
+++ b/gix-actor/src/signature/decode.rs
@@ -29,9 +29,9 @@ pub(crate) mod function {
                 tag(b" "),
                 context("<timestamp>", |i| {
                     terminated(take_until(SPACE), take(1usize))(i).and_then(|(i, v)| {
-                        btoi::<SecondsSinceUnixEpoch>(v).map(|v| (i, v)).map_err(|_| {
-                            winnow::Err::Backtrack(E::from_error_kind(i, winnow::error::ErrorKind::MapRes))
-                        })
+                        btoi::<SecondsSinceUnixEpoch>(v)
+                            .map(|v| (i, v))
+                            .map_err(|_| winnow::Err::from_error_kind(i, winnow::error::ErrorKind::MapRes))
                     })
                 }),
                 context(
@@ -43,16 +43,16 @@ pub(crate) mod function {
                 ),
                 context("HH", |i| {
                     take_while_m_n(2usize, 2, is_digit)(i).and_then(|(i, v)| {
-                        btoi::<OffsetInSeconds>(v).map(|v| (i, v)).map_err(|_| {
-                            winnow::Err::Backtrack(E::from_error_kind(i, winnow::error::ErrorKind::MapRes))
-                        })
+                        btoi::<OffsetInSeconds>(v)
+                            .map(|v| (i, v))
+                            .map_err(|_| winnow::Err::from_error_kind(i, winnow::error::ErrorKind::MapRes))
                     })
                 }),
                 context("MM", |i| {
                     take_while_m_n(1usize, 2, is_digit)(i).and_then(|(i, v)| {
-                        btoi::<OffsetInSeconds>(v).map(|v| (i, v)).map_err(|_| {
-                            winnow::Err::Backtrack(E::from_error_kind(i, winnow::error::ErrorKind::MapRes))
-                        })
+                        btoi::<OffsetInSeconds>(v)
+                            .map(|v| (i, v))
+                            .map_err(|_| winnow::Err::from_error_kind(i, winnow::error::ErrorKind::MapRes))
                     })
                 }),
             )),
--- a/gix-actor/src/signature/decode.rs
+++ b/gix-actor/src/signature/decode.rs
@@ -5,12 +5,12 @@ pub(crate) mod function {
     use std::cell::RefCell;
     use winnow::{
         branch::alt,
-        bytes::complete::{take, take_until, take_while_m_n},
-        character::is_digit,
+        bytes::{take, take_until0, take_while_m_n},
         error::{ContextError, ParseError},
-        multi::many1_count,
+        multi::many1,
         prelude::*,
-        sequence::{terminated, tuple},
+        sequence::terminated,
+        stream::AsChar,
     };
 
     use crate::{IdentityRef, SignatureRef};
@@ -26,31 +26,31 @@ pub(crate) mod function {
             identity,
             b" ",
             (|i| {
-                terminated(take_until(SPACE), take(1usize))(i).and_then(|(i, v)| {
+                terminated(take_until0(SPACE), take(1usize))(i).and_then(|(i, v)| {
                     btoi::<SecondsSinceUnixEpoch>(v)
                         .map(|v| (i, v))
-                        .map_err(|_| winnow::Err::from_error_kind(i, winnow::error::ErrorKind::MapRes))
+                        .map_err(|_| winnow::error::ErrMode::from_error_kind(i, winnow::error::ErrorKind::MapRes))
                 })
             })
             .context("<timestamp>"),
             alt((
-                many1_count(b"-").map(|_| *tzsign.borrow_mut() = b'-'), // TODO: this should be a non-allocating consumer of consecutive tags
-                many1_count(b"+").map(|_| *tzsign.borrow_mut() = b'+'),
+                many1(b"-").map(|_: ()| *tzsign.borrow_mut() = b'-'), // TODO: this should be a non-allocating consumer of consecutive tags
+                many1(b"+").map(|_: ()| *tzsign.borrow_mut() = b'+'),
             ))
             .context("+|-"),
             (|i| {
-                take_while_m_n(2usize, 2, is_digit)(i).and_then(|(i, v)| {
+                take_while_m_n(2usize, 2, AsChar::is_dec_digit)(i).and_then(|(i, v)| {
                     btoi::<OffsetInSeconds>(v)
                         .map(|v| (i, v))
-                        .map_err(|_| winnow::Err::from_error_kind(i, winnow::error::ErrorKind::MapRes))
+                        .map_err(|_| winnow::error::ErrMode::from_error_kind(i, winnow::error::ErrorKind::MapRes))
                 })
             })
             .context("HH"),
             (|i| {
-                take_while_m_n(1usize, 2, is_digit)(i).and_then(|(i, v)| {
+                take_while_m_n(1usize, 2, AsChar::is_dec_digit)(i).and_then(|(i, v)| {
                     btoi::<OffsetInSeconds>(v)
                         .map(|v| (i, v))
-                        .map_err(|_| winnow::Err::from_error_kind(i, winnow::error::ErrorKind::MapRes))
+                        .map_err(|_| winnow::error::ErrMode::from_error_kind(i, winnow::error::ErrorKind::MapRes))
                 })
             })
             .context("MM"),
@@ -82,8 +82,8 @@ pub(crate) mod function {
         i: &'a [u8],
     ) -> IResult<&'a [u8], IdentityRef<'a>, E> {
         let (i, (name, email)) = (
-            terminated(take_until(&b" <"[..]), take(2usize)).context("<name>"),
-            terminated(take_until(&b">"[..]), take(1usize)).context("<email>"),
+            terminated(take_until0(&b" <"[..]), take(2usize)).context("<name>"),
+            terminated(take_until0(&b">"[..]), take(1usize)).context("<email>"),
         )
             .context("<name> <<email>>")
             .parse_next(i)?;
--- a/gix-actor/src/signature/decode.rs
+++ b/gix-actor/src/signature/decode.rs
@@ -5,9 +5,9 @@ pub(crate) mod function {
     use std::cell::RefCell;
     use winnow::{
         branch::alt,
-        bytes::{take, take_until0, take_while_m_n},
+        bytes::{take, take_until0, take_while},
+        combinator::repeat,
         error::{ContextError, ParseError},
-        multi::many1,
         prelude::*,
         sequence::terminated,
         stream::AsChar,
@@ -36,22 +36,20 @@ pub(crate) mod function {
             })
             .context("<timestamp>"),
             alt((
-                many1(b"-").map(|_: ()| *tzsign.borrow_mut() = b'-'), // TODO: this should be a non-allocating consumer of consecutive tags
-                many1(b"+").map(|_: ()| *tzsign.borrow_mut() = b'+'),
+                repeat(1.., b"-").map(|_: ()| *tzsign.borrow_mut() = b'-'), // TODO: this should be a non-allocating consumer of consecutive tags
+                repeat(1.., b"+").map(|_: ()| *tzsign.borrow_mut() = b'+'),
             ))
             .context("+|-"),
             (|i| {
-                take_while_m_n(2usize, 2, AsChar::is_dec_digit)
-                    .parse_next(i)
-                    .and_then(|(i, v)| {
-                        btoi::<OffsetInSeconds>(v)
-                            .map(|v| (i, v))
-                            .map_err(|_| winnow::error::ErrMode::from_error_kind(i, winnow::error::ErrorKind::Verify))
-                    })
+                take_while(2, AsChar::is_dec_digit).parse_next(i).and_then(|(i, v)| {
+                    btoi::<OffsetInSeconds>(v)
+                        .map(|v| (i, v))
+                        .map_err(|_| winnow::error::ErrMode::from_error_kind(i, winnow::error::ErrorKind::Verify))
+                })
             })
             .context("HH"),
             (|i| {
-                take_while_m_n(1usize, 2, AsChar::is_dec_digit)
+                take_while(1..=2, AsChar::is_dec_digit)
                     .parse_next(i)
                     .and_then(|(i, v)| {
                         btoi::<OffsetInSeconds>(v)
--- a/gix-actor/src/signature/decode.rs
+++ b/gix-actor/src/signature/decode.rs
@@ -5,7 +5,7 @@ pub(crate) mod function {
     use std::cell::RefCell;
     use winnow::{
         branch::alt,
-        bytes::complete::{tag, take, take_until, take_while_m_n},
+        bytes::complete::{take, take_until, take_while_m_n},
         character::is_digit,
         error::{context, ContextError, ParseError},
         multi::many1_count,
@@ -24,9 +24,9 @@ pub(crate) mod function {
         let tzsign = RefCell::new(b'-'); // TODO: there should be no need for this.
         let (i, (identity, _, time, _tzsign_count, hours, minutes)) = context(
             "<name> <<email>> <timestamp> <+|-><HHMM>",
-            tuple((
+            (
                 identity,
-                tag(b" "),
+                b" ",
                 context("<timestamp>", |i| {
                     terminated(take_until(SPACE), take(1usize))(i).and_then(|(i, v)| {
                         btoi::<SecondsSinceUnixEpoch>(v)
@@ -37,8 +37,8 @@ pub(crate) mod function {
                 context(
                     "+|-",
                     alt((
-                        many1_count(tag(b"-")).map(|_| *tzsign.borrow_mut() = b'-'), // TODO: this should be a non-allocating consumer of consecutive tags
-                        many1_count(tag(b"+")).map(|_| *tzsign.borrow_mut() = b'+'),
+                        many1_count(b"-").map(|_| *tzsign.borrow_mut() = b'-'), // TODO: this should be a non-allocating consumer of consecutive tags
+                        many1_count(b"+").map(|_| *tzsign.borrow_mut() = b'+'),
                     )),
                 ),
                 context("HH", |i| {
@@ -55,7 +55,7 @@ pub(crate) mod function {
                             .map_err(|_| winnow::Err::from_error_kind(i, winnow::error::ErrorKind::MapRes))
                     })
                 }),
-            )),
+            ),
         )(i)?;
 
         let tzsign = tzsign.into_inner();
@@ -83,10 +83,10 @@ pub(crate) mod function {
     ) -> IResult<&'a [u8], IdentityRef<'a>, E> {
         let (i, (name, email)) = context(
             "<name> <<email>>",
-            tuple((
+            (
                 context("<name>", terminated(take_until(&b" <"[..]), take(2usize))),
                 context("<email>", terminated(take_until(&b">"[..]), take(1usize))),
-            )),
+            ),
         )(i)?;
 
         Ok((
--- a/gix-actor/src/signature/decode.rs
+++ b/gix-actor/src/signature/decode.rs
@@ -4,13 +4,13 @@ pub(crate) mod function {
     use gix_date::{time::Sign, OffsetInSeconds, SecondsSinceUnixEpoch, Time};
     use std::cell::RefCell;
     use winnow::{
-        branch::alt,
-        bytes::{take, take_until0, take_while},
+        combinator::alt,
         combinator::repeat,
-        error::{ContextError, ParseError},
+        combinator::terminated,
+        error::{AddContext, ParserError},
         prelude::*,
-        sequence::terminated,
         stream::AsChar,
+        token::{take, take_until0, take_while},
     };
 
     use crate::{IdentityRef, SignatureRef};
@@ -18,7 +18,7 @@ pub(crate) mod function {
     const SPACE: &[u8] = b" ";
 
     /// Parse a signature from the bytes input `i` using `nom`.
-    pub fn decode<'a, E: ParseError<&'a [u8]> + ContextError<&'a [u8]>>(
+    pub fn decode<'a, E: ParserError<&'a [u8]> + AddContext<&'a [u8]>>(
         i: &'a [u8],
     ) -> IResult<&'a [u8], SignatureRef<'a>, E> {
         let tzsign = RefCell::new(b'-'); // TODO: there should be no need for this.
@@ -82,7 +82,7 @@ pub(crate) mod function {
     }
 
     /// Parse an identity from the bytes input `i` (like `name <email>`) using `nom`.
-    pub fn identity<'a, E: ParseError<&'a [u8]> + ContextError<&'a [u8]>>(
+    pub fn identity<'a, E: ParserError<&'a [u8]> + AddContext<&'a [u8]>>(
         i: &'a [u8],
     ) -> IResult<&'a [u8], IdentityRef<'a>, E> {
         let (i, (name, email)) = (
--- a/gix-actor/src/signature/decode.rs
+++ b/gix-actor/src/signature/decode.rs
@@ -31,7 +31,7 @@ pub(crate) mod function {
                     .and_then(|(i, v)| {
                         btoi::<SecondsSinceUnixEpoch>(v)
                             .map(|v| (i, v))
-                            .map_err(|_| winnow::error::ErrMode::from_error_kind(i, winnow::error::ErrorKind::MapRes))
+                            .map_err(|_| winnow::error::ErrMode::from_error_kind(i, winnow::error::ErrorKind::Verify))
                     })
             })
             .context("<timestamp>"),
@@ -46,7 +46,7 @@ pub(crate) mod function {
                     .and_then(|(i, v)| {
                         btoi::<OffsetInSeconds>(v)
                             .map(|v| (i, v))
-                            .map_err(|_| winnow::error::ErrMode::from_error_kind(i, winnow::error::ErrorKind::MapRes))
+                            .map_err(|_| winnow::error::ErrMode::from_error_kind(i, winnow::error::ErrorKind::Verify))
                     })
             })
             .context("HH"),
@@ -56,7 +56,7 @@ pub(crate) mod function {
                     .and_then(|(i, v)| {
                         btoi::<OffsetInSeconds>(v)
                             .map(|v| (i, v))
-                            .map_err(|_| winnow::error::ErrMode::from_error_kind(i, winnow::error::ErrorKind::MapRes))
+                            .map_err(|_| winnow::error::ErrMode::from_error_kind(i, winnow::error::ErrorKind::Verify))
                     })
             })
             .context("MM"),
@@ -188,7 +188,7 @@ mod tests {
                             .map_err(to_bstr_err)
                             .expect_err("parse fails as > is missing")
                             .to_string(),
-                        "Parse error:\nTakeUntil at:  12345 -1215\nin section '<email>', at:  12345 -1215\nin section '<name> <<email>>', at: hello < 12345 -1215\nin section '<name> <<email>> <timestamp> <+|-><HHMM>', at: hello < 12345 -1215\n"
+                        "Parse error:\nSlice at:  12345 -1215\nin section '<email>', at:  12345 -1215\nin section '<name> <<email>>', at: hello < 12345 -1215\nin section '<name> <<email>> <timestamp> <+|-><HHMM>', at: hello < 12345 -1215\n"
                     );
         }
 
@@ -199,7 +199,7 @@ mod tests {
                             .map_err(to_bstr_err)
                             .expect_err("parse fails as > is missing")
                             .to_string(),
-                        "Parse error:\nMapRes at: -1215\nin section '<timestamp>', at: abc -1215\nin section '<name> <<email>> <timestamp> <+|-><HHMM>', at: hello <> abc -1215\n"
+                        "Parse error:\nVerify at: -1215\nin section '<timestamp>', at: abc -1215\nin section '<name> <<email>> <timestamp> <+|-><HHMM>', at: hello <> abc -1215\n"
                     );
         }
     }
--- a/gix-actor/src/signature/decode.rs
+++ b/gix-actor/src/signature/decode.rs
@@ -2,10 +2,10 @@ pub(crate) mod function {
     use bstr::ByteSlice;
     use btoi::btoi;
     use gix_date::{time::Sign, OffsetInSeconds, SecondsSinceUnixEpoch, Time};
-    use std::cell::RefCell;
     use winnow::{
         combinator::alt,
         combinator::repeat,
+        combinator::separated_pair,
         combinator::terminated,
         error::{AddContext, ParserError},
         prelude::*,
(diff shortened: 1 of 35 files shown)


epage:
Since I'm seeing local test failures, I'm rely on CI for the cases I'm not easily hitting due to other errors

epage:
Huh, mac-fast is failing in CI differently than local (in `gix-worktree`) but all other test jobs passed.

gix-ref/src/store/file/log/line.rs
```diff
@@ -146,11 +146,11 @@ pub mod decode {
...
             if let Some(first) = message.first() {
                 if !first.is_ascii_whitespace() {
-                    return Err(winnow::Err::Backtrack(E::add_context(
-                        E::from_error_kind(i, winnow::error::ErrorKind::MapRes),
-                        i,
-                        "log message must be separated from signature with whitespace",
-                    )));
+                    return Err(
+                        winnow::Err::from_error_kind(i, winnow::error::ErrorKind::MapRes).map(|err: E| {
+                            err.add_context(i, "log message must be separated from signature with whitespace")
```
epage: When upgrading to winnow 0.5, we can switch to calling `add_context` without the `map`

gix-object/src/lib.rs
```diff
@@ -278,18 +278,18 @@ pub mod decode {
...
+                        winnow::Err::Backtrack(err) | winnow::Err::Cut(err) => winnow::error::VerboseError {
                             errors: err
                                 .errors
                                 .into_iter()
                                 .map(|(i, v)| (i.as_bstr().to_owned(), v))
                                 .collect(),
                         },
-                        nom::Err::Incomplete(_) => unreachable!("we don't have streaming parsers"),
+                        winnow::Err::Incomplete(_) => unreachable!("we don't have streaming parsers"),
                     },
```
epage: After 0.5 upgrade, we can switch to `into_inner().expect("not streaming")`

gix-object/src/lib.rs
```diff
@@ -321,12 +321,12 @@ pub mod decode {
...
-        impl<'a> From<nom::Err<ParseError<'a>>> for Error {
-            fn from(v: nom::Err<ParseError<'a>>) -> Self {
+        impl<'a> From<winnow::Err<ParseError<'a>>> for Error {
+            fn from(v: winnow::Err<ParseError<'a>>) -> Self {
                 Error {
                     inner: match v {
-                        nom::Err::Error(err) | nom::Err::Failure(err) => err,
-                        nom::Err::Incomplete(_) => unreachable!("we don't have streaming parsers"),
+                        winnow::Err::Backtrack(err) | winnow::Err::Cut(err) => err,
+                        winnow::Err::Incomplete(_) => unreachable!("we don't have streaming parsers"),
```
epage: After 0.5 upgrade, we can switch to `into_inner().expect("not streaming")`

tests/tools/src/lib.rs
```diff
@@ -689,10 +689,10 @@ fn extract_archive(
...
 
 /// Transform a verbose bom errors from raw bytes into a `BStr` to make printing/debugging human-readable.
-pub fn to_bstr_err(err: nom::Err<VerboseError<&[u8]>>) -> VerboseError<&BStr> {
+pub fn to_bstr_err(err: winnow::Err<VerboseError<&[u8]>>) -> VerboseError<&BStr> {
     let err = match err {
-        nom::Err::Error(err) | nom::Err::Failure(err) => err,
-        nom::Err::Incomplete(_) => unreachable!("not a streaming parser"),
+        winnow::Err::Backtrack(err) | winnow::Err::Cut(err) => err,
+        winnow::Err::Incomplete(_) => unreachable!("not a streaming parser"),
     };
```
epage: After 0.5 upgrade, we can switch to `into_inner().expect("not streaming")`

epage:
Already rebased and working on cleaning it up right now

gix-actor/src/signature/decode.rs
```diff
@@ -197,7 +170,7 @@ mod tests {
                             .map_err(to_bstr_err)
                             .expect_err("parse fails as > is missing")
                             .to_string(),
-                        "Parse error:\nVerify at: -1215\nin section '<timestamp>', at: abc -1215\nin section '<name> <<email>> <timestamp> <+|-><HHMM>', at: hello <> abc -1215\n"
+                        "Parse error:\nVerify at: abc -1215\nin section '<timestamp>', at: abc -1215\nin section '<name> <<email>> <timestamp> <+|-><HHMM>', at: hello <> abc -1215\n"
```
epage, line 173: Explanation: The old error reporting was capturing the input  after what was being validated.  Now its capturing it at the start of whats being validating.

gix-object/src/commit/decode.rs
```diff
@@ -21,52 +22,43 @@ pub fn message<'a, E: ParserError<&'a [u8]> + AddContext<&'a [u8]>>(i: &'a [u8])
                 .map(|err: E| err.add_context(i, "newline + <message>")),
         );
     }
-    let (i, _) = tag(NL)
+    preceded(NL, rest.map(|s: &[u8]| s.as_bstr()))
         .context("a newline separates headers from the message")
-        .parse_next(i)?;
-    Ok((&[], i.as_bstr()))
+        .parse_next(i)
```
epage: Explanation: Moving to declarative parsing where possible not because of any high moral ground or technical benefit but just because it makes the winnow 0.5 upgrades less intrusive

gix-object/src/commit/message/decode.rs
```diff
@@ -1,49 +1,44 @@
...
+                },
+            },
+            Err(_) => match i.get(1..) {
                 Some(next) => {
-                    consumed_bytes += 1;
-                    next
+                    i = next;
                 }
                 None => break,
             },
```
epage, line 36: After 0.5, we can simplify these with `let _ = i.next_token();`

Byron:
I am cheering you on from the side-lines, the goal is already in sight :)!

gix-ref/src/store/file/log/line.rs
```diff
@@ -123,34 +123,35 @@ pub mod decode {
...
+            gix_actor::signature::decode.context("<name> <<email>> <timestamp>"),
+            opt(b"\t"),
+            message.context("<optional message>"),
+        )
+            .context("<old-hexsha> <new-hexsha> <name> <<email>> <timestamp> <tz>\\t<message>")
+            .parse_next(bytes)?;
 
         if message_sep.is_none() {
             if let Some(first) = message.first() {
                 if !first.is_ascii_whitespace() {
```
epage, line 149: This is a little confusing.  We check for the separator being a tab but we actually allow any whitespace separator.  

I feel like it should either be:
```rust
opt((
    b"\t",
    message
))
```
or
```rust
opt((
    one_of(|b| b.is_ascii_whitespace()),
    message
))
```

unless we are intentionally capturing non-tab whitespace separators to be a part of `message`.

Byron: Sorry for the late reply. There should always be a test for special cases, and if nothing breaks with the simplified version this is probably an oversight.

gix-object/src/lib.rs
```diff
@@ -261,13 +261,13 @@ pub mod decode {
         use crate::bstr::{BString, ByteSlice};
 
         /// The type to be used for parse errors.
-        pub type ParseError<'a> = nom::error::VerboseError<&'a [u8]>;
+        pub type ParseError<'a> = winnow::error::VerboseError<&'a [u8]>;
```
epage, line 264: What is the goal with `verbose-object-parsing-errors`?

`VerboseError` is gone in 0.5, replaced with either
- `ContextError<Context>` where the default context allows `Label` (description of what you are parsing) and `Expected` (what is needed to parse it)
- `TreeError` which tracks all error state and is more useful for tests as the extra information might help when debugging

We could
- Use `ContextError<&'static str>` and provide our own rendering
- Use `ContextError<StrContext>` and switch the code to use `StrContext` instead of `&'static str`
- Create a custom error type

Byron: Thanks for bringing this up.

The original thinking here was that object parsing should be as fast as possible, so reducing errors to booleans seemed like the right path to take. I probably tested this and measured 1% performance improvement that way, but my memory is fuzzy.

With that said, I wouldn't mind going with 'high-performance' errors by default that leave the possibility to replace the error later with `()` or something similarly light. That way, one could benchmark it once more and see if it's worth the complexity to maintain both versions of the error.

Ideally, there would be no split and errors would be enabled and helpful if parsing fails, but fast enough to not make us want to disable them. I have a feeling that parsing is a little faster anyways now which leaves room for having decent errors without being slower than the no-error `nom` version in the end.

I'd absolutely leave it to your expertise to chose the right error type, and I will be sure to benchmark it against the no-error version on `main`. From there we can see if it can just stay like this.

For reference, `Commit` parsing has to be very fast, and tree-parsing already has its own fast-path anyway that implements it by hand. Commits are the only object type I am thinking about.

epage: Good to know; thanks.

- `basic` is 0.4's `Error` which was renamed to `InputError`
- `unit` is `()` (and hadn't previously been benchmarking it)
- `context` is 0.5's `ContextError` which replaced `VerboseError` and was pulled from `toml_edit`
```console
$ cargo bench --bench json -- canada # v0.4.9
json/basic/canada       time:   [10.473 ms 10.498 ms 10.525 ms]
json/verbose/canada     time:   [31.411 ms 31.488 ms 31.574 ms]

$ cargo bench --bench json -- canada # v0.5.7
json/basic/canada       time:   [9.9094 ms 9.9743 ms 10.040 ms]
json/unit/canada        time:   [8.7462 ms 8.7724 ms 8.7998 ms]
json/context/canada     time:   [14.040 ms 14.082 ms 14.128 ms]
```

In short, `()` still offers a significant performance improvement though `ContextError` does close the gap quite a bit compared to `VerboseError`.

Byron: The purpose of the error by default would just be to report error with boolean precision, but ideally there is a way to swap-in/toggle-on the best-possible error to help understanding what's wrong with a particular kind of object.

Thinking about it, it would be even better if this functionality could exist in the same binary (now that we are here to improve things). The workflow could be that parsing is very fast by default, anticipating no issues, but if there is an issue, the object could be parsed again with the sole purpose of generating an error to pinpoint the source of the issue.

Ultimately, I trust your judgement on this even if it means choosing one kind of error that isn't unit, because it might still be faster than is on `main` and I will be sure to test that.

epage:
I decided to hold off on removing `winnow` from the public APIs.  I feel like that involves a lot of gix decision making that might be left best for you.  One way of handling it is to create one or more `-syntax` packages, like `regex-syntax` that has the parsers.  The issue then is shared types.  A really hacky idea is to dual source the definitions and to transmute between them.

On a related note, some types have a `from_bytes` function that uses a parser and ignores some information.  Using `Parser::parse` instead would (1) ensure all of the input is consumed and (2) provide extra location information for errors.  The idea is that `Parser::parse_next` is for sub-parsers and `Parser::parse` is for the top-level parser.  Doesn't have to be used but its a convenience for when it makes sense.

Byron:
The first benchmark is in:

```
linux ( master) +369 -819 [!] took 9s
✦ ❯ hyperfine '/Users/byron/dev/github.com/Byron/gitoxide/target/release/ein t h' 'ein t h'
Benchmark 1: /Users/byron/dev/github.com/Byron/gitoxide/target/release/ein t h
  Time (mean ± σ):      9.740 s ±  0.906 s    [User: 11.264 s, System: 0.859 s]
  Range (min … max):    8.900 s … 11.614 s    10 runs

Benchmark 2: ein t h
  Time (mean ± σ):      9.089 s ±  0.218 s    [User: 10.620 s, System: 0.554 s]
  Range (min … max):    8.922 s …  9.552 s    10 runs

Summary
  ein t h ran
    1.07 ± 0.10 times faster than /Users/byron/dev/github.com/Byron/gitoxide/target/release/ein t h
```

The `ein t h` is the baseline using `nom` as just compiled from `main`, the other binary is the one from this branch. The workload is a real-world linux kernel clone which whose 1.1 million commits are decompressed and decoded. What makes it a good benchmark is that the decoding happens on a separate thread entirely.

When profiling it, both came out pretty much the same. Here is the `nom` version, which seems to spend most of its parsing time in decoding object ids actually.

Screenshot 2023-08-16 at 10 30 24

And here is the `winnow` version, which is also dominated by object ids as one would expect.
Screenshot 2023-08-16 at 10 30 07

That's inconclusive, and I think i'd rather try to get faster hex-decoding than anything else.

Looking into it more.

Byron:
I guess before it gets better, it has to get worse :/. What follows is the latest comparison, where `winnow` has the latest `main`  merged in alongside the `faster-hex` branch. That was compared to the `faster-hex` branch on `main`, which is `nom`.

```
✦ ❯ hyperfine --warmup 1  '/Users/byron/dev/github.com/Byron/gitoxide/target/release/ein t h' 'ein t h'  -N
Benchmark 1: /Users/byron/dev/github.com/Byron/gitoxide/target/release/ein t h
  Time (mean ± σ):     10.354 s ±  0.073 s    [User: 10.396 s, System: 1.834 s]
  Range (min … max):   10.289 s … 10.499 s    10 runs

Benchmark 2: ein t h
  Time (mean ± σ):      8.473 s ±  0.013 s    [User: 9.646 s, System: 0.461 s]
  Range (min … max):    8.446 s …  8.493 s    10 runs

Summary
  ein t h ran
    1.22 ± 0.01 times faster than /Users/byron/dev/github.com/Byron/gitoxide/target/release/ein t h
```

These numbers don't make much sense to me as the system time for the `winnow` `ein` is way too high. Also, the parsing thread doesn't have so much influence on the overall time so the ODB seems to now be slower somehow. 
It's still puzzling me.

Byron:
I have another indicator that commit parsing got slower, even though the compound effect of it shouldn't be more than 200ms (in practice, it's much more for unknown reason).

The stacks that follow were taken from a commit traversal (1.1m commits), and it seems `winnow` takes 1.63x longer to parse actor signatures.

Screenshot 2023-08-16 at 12 27 22

Here is the `nom` version.

Screenshot 2023-08-16 at 12 27 34

Maybe it's due to `winnow` doing less inlining?

If it was just these couple of milliseconds there wouldn't be an issue, but for some reason the effective difference is huge.

The following is done with a hot cache, the second invocation of the command.
```
✦ ❯ /Users/byron/dev/github.com/Byron/gitoxide/target/release/ein t h && ein t h
 12:45:44 traverse commit graph done 1.2M commits in 10.06s (116.3k commits/s)
 12:45:44        estimate-hours Extracted and organized data from 1169974 commits in 14.083542ms (83073848 commits/s)
total hours: 1119736.25
total 8h days: 139967.03
total commits = 1169974
total authors: 31645
total unique authors: 24103 (23.83% duplication)
 12:45:52 traverse commit graph done 1.2M commits in 8.22s (142.3k commits/s)
 12:45:53        estimate-hours Extracted and organized data from 1169974 commits in 15.748834ms (74289560 commits/s)
total hours: 1119736.25
total 8h days: 139967.03
total commits = 1169974
total authors: 31645
total unique authors: 24103 (23.83% duplication)

linux ( master) +369 -819 [!] took 18s
```

Byron:
As a summary, there seems to be a 20% performance regression and the best hint I have is `git_actor::signature::decode()`. But it might well be some sort of compound effect.

I will see to adding benchmarks to `main` which hopefully reproduce the performance difference so that it can be tackled.

Byron:
I have pushed the latest changes from `main`, which includes the `faster-hex` modifications, along with a cherry-picked commit to add simple benchmarks for `gix-object` parsing.

After having run it on `main`, when running it here, I get these values:

```
     Running benches/decode_objects.rs (/Users/byron/dev/github.com/Byron/gitoxide/target/release/deps/decode_objects-33a76fc9818b65fc)
CommitRef(sig)          time:   [3.2001 µs 3.2015 µs 3.2029 µs]
                        change: [+177.16% +177.99% +178.70%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) high mild
  4 (4.00%) high severe

CommitRefIter(sig)      time:   [3.3096 µs 3.3108 µs 3.3120 µs]
                        change: [+159.99% +160.37% +160.71%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 11 outliers among 100 measurements (11.00%)
  4 (4.00%) low mild
  1 (1.00%) high mild
  6 (6.00%) high severe

TagRef(sig)             time:   [1.6129 µs 1.6133 µs 1.6137 µs]
                        change: [+580.71% +581.47% +582.19%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 16 outliers among 100 measurements (16.00%)
  2 (2.00%) low severe
  3 (3.00%) low mild
  5 (5.00%) high mild
  6 (6.00%) high severe

TagRefIter(sig)         time:   [1.6176 µs 1.6185 µs 1.6196 µs]
                        change: [+590.26% +591.38% +592.51%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) high mild
  4 (4.00%) high severe

TreeRef(sig)            time:   [131.70 ns 132.21 ns 132.68 ns]
                        change: [-3.2794% -3.0601% -2.8339%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 13 outliers among 100 measurements (13.00%)
  7 (7.00%) low severe
  1 (1.00%) low mild
  2 (2.00%) high mild
  3 (3.00%) high severe

TreeRefIter(sig)        time:   [61.176 ns 61.338 ns 61.526 ns]
                        change: [-0.7091% -0.5238% -0.3479%] (p = 0.00 < 0.05)
Found 19 outliers among 100 measurements (19.00%)
  3 (3.00%) low mild
  4 (4.00%) high mild
  12 (12.00%) high severe
```

It looks like parsing did regress quite a bit, but I have a feeling that it's something small that can be fixed. I would't know how though as profiling granularity is rather coarse and maybe even spotty when things got inlined.

epage:
Thanks for the breakdown and the benches!

*Note: I tend to see two buckets of performance numbers, I'm assuming from P vs E cores*

353b1a788 (main): 165.49ns or 174.25 ns
421dbfd: 1.6198 µs or 1.7121 µs
2c73a9020 (w1): 1.7772 µs or 1.8231 µs
d1568c064 (w2): 1.7677 µs or 1.9179 µs
2f6d7a4b4 (w3): 1.5208 µs or 1.6010 µs
eb70d3750 (w4): 1.5179 µs
d081fe215 (w5): 1.5539 µs or 1.6467 µs
a8ae30fe5 (winnow04): 1.7295 µs or 1.8135 µs
20cd7e4db (w6): 1.7241 µs or 1.8318 µs
89ef38004 (w7): 1.7303 µs or 1.8172 µs
eab8780bd: 1.7257 µs or 1.8187 µs
33fce872c (w8): 1.7015 µs or 1.8058 µs
d9995efb9 (w9): 1.7279 µs or 1.8417 µs
07384f75c: 1.5977 µs or 1.6808 µs
e2b17bfac: 1.5882 µs or 1.6005 µs
5f92ea3bc: 1.5456 µs or 1.6427 µs
6f342c92e: 1.5417 µs or 1.6537 µs
45469c2c4: 1.5403 µs or 1.6478 µs
97eb45cd7 (winnow05): 1.8011 µs or 1.9520 µs

That is not a trend I expected.  I assumed it was going to initially be faster and then slow down as I did refactors.

epage:
> Thanks for the measurements! It appears winnow 0.5 really did something here. It's comforting to see that main is actually slower than the pre-winnow 0.5 branch has been and I am very much looking forward to trying ein tool estimate-hours with the new parser :).

To clarify, just switching from nom to winnow 0.3 made parsing take 10x the amount of time (ie winnow 0.3 is slower than nom within gix) and that has mostly stuck all the way through my changes up to and including winnow 0.5

epage:
Found the biggest source of slowdown: winnow 0.3 made memchr an optional dependency.  For most of my work, it offers no performance difference.  Apparently for gitoxide, it makes signature parsing take a **lot** longer.

Now things are back to the same order of magnitude

epage:
Now with `simd` enabled

Thanks for the breakdown and the benches!

**TagRefIter(sig)**
169228b12 (main): 174.01 ns
3d1fa31fb: 239.23 ns
5c5664dab (w1): 258.78 ns
a81070ef2 (w2): 267.65 ns
99f69aa2e (w3): 249.76 ns
c1c56d8cb (w4): 264.53 ns
941887c53 (w5): 260.41 ns
c08094d9a (winnow04): 244.65 ns
cbeebeb63 (w6): 242.33 ns
cc1db3533 (w7): 242.52 ns
25de35b9a: 241.31 ns
36c6605ff (w8): 245.65 ns
9b0b4b8ee (w9): 249.63 ns
a78d967df: 272.13 ns
517bb071c: 264.33 ns
72ee59f5d: 261.07 ns
c76604784: 272.63 ns
2fe271bd4: 257.55 ns
733e35825 (winnow05): 258.88 ns
cd5dc7400 (winnow): 255.74 ns

Better but still not the trend I was expecting.  More investigation is needed

epage:
And for across-the-board numbers of `main` vs `winnow` but without multiple runs so actual results might be off:
```console
$ cargo bench -p gix-object@0.34.0 --bench decode-objects
   Compiling gix-actor v0.24.2 (/home/epage/src/personal/gitoxide/gix-actor)
   Compiling gix-testtools v0.13.0 (/home/epage/src/personal/gitoxide/tests/tools)
   Compiling gix-object v0.34.0 (/home/epage/src/personal/gitoxide/gix-object)
    Finished bench [optimized] target(s) in 2.24s
     Running benches/decode_objects.rs (target/release/deps/decode_objects-257fa0caddf5c1ec)
CommitRef(sig)          time:   [682.97 ns 684.48 ns 686.13 ns]
                        change: [-20.805% -20.101% -19.032%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  1 (1.00%) low severe
  3 (3.00%) high mild
  3 (3.00%) high severe

CommitRefIter(sig)      time:   [796.96 ns 798.19 ns 799.63 ns]
                        change: [-9.6653% -9.4035% -9.1307%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  2 (2.00%) low mild
  6 (6.00%) high mild
  2 (2.00%) high severe

TagRef(sig)             time:   [266.80 ns 267.33 ns 267.91 ns]
                        change: [+54.025% +54.422% +54.863%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 8 outliers among 100 measurements (8.00%)
  5 (5.00%) high mild
  3 (3.00%) high severe

TagRefIter(sig)         time:   [271.03 ns 271.38 ns 271.78 ns]
                        change: [+61.824% +62.265% +62.670%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 6 outliers among 100 measurements (6.00%)
  2 (2.00%) high mild
  4 (4.00%) high severe

TreeRef(sig)            time:   [99.956 ns 100.07 ns 100.20 ns]
                        change: [-6.7856% -6.5919% -6.4002%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) high mild
  3 (3.00%) high severe

TreeRefIter(sig)        time:   [49.436 ns 49.491 ns 49.553 ns]
                        change: [+4.3995% +4.5843% +4.7813%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 6 outliers among 100 measurements (6.00%)
  2 (2.00%) low mild
  1 (1.00%) high mild
  3 (3.00%) high severe
```

Byron:
That's great, for me Commit parsing is now 10% faster in the benchmark.

```
CommitRef(sig)          time:   [1.0341 µs 1.0354 µs 1.0368 µs]
                        change: [-9.8560% -9.6710% -9.4759%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe

CommitRefIter(sig)      time:   [1.1505 µs 1.1524 µs 1.1547 µs]
                        change: [-9.9844% -9.7481% -9.4984%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  5 (5.00%) high mild

TagRef(sig)             time:   [694.97 ns 695.33 ns 695.78 ns]
                        change: [+193.29% +193.81% +194.30%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 16 outliers among 100 measurements (16.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild
  5 (5.00%) high mild
  9 (9.00%) high severe

TagRefIter(sig)         time:   [696.14 ns 696.43 ns 696.75 ns]
                        change: [+196.80% +197.29% +197.77%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 17 outliers among 100 measurements (17.00%)
  1 (1.00%) low severe
  2 (2.00%) low mild
  6 (6.00%) high mild
  8 (8.00%) high severe

TreeRef(sig)            time:   [128.25 ns 128.35 ns 128.46 ns]
                        change: [-6.5925% -6.2374% -5.8575%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) high mild
  5 (5.00%) high severe

TreeRefIter(sig)        time:   [61.302 ns 61.355 ns 61.412 ns]
                        change: [-0.1914% -0.0078% +0.1738%] (p = 0.94 > 0.05)
                        No change in performance detected.
Found 9 outliers among 100 measurements (9.00%)
  4 (4.00%) high mild
  5 (5.00%) high severe
```

Only tags are still considerable worse and the only reason I think that's interesting is that Tags share a lot of header-related parsing code. Maybe in that area there is still a 'handbreak' somewhere that needs to be lifted for the parser to fly. If that would be possible, I am sure commits would benefit as well and be even faster.

As for the none-synthetic benchmark, the results didn't improve quite as much (~10.6s vs ~10.5s) which makes me believe that maybe that Tag regression really has something to do with it. 

```
✦ ❯ hyperfine --warmup 1  '/Users/byron/dev/github.com/Byron/gitoxide/target/release/ein t h' 'ein t h'  -N
Benchmark 1: /Users/byron/dev/github.com/Byron/gitoxide/target/release/ein t h
  Time (mean ± σ):     10.510 s ±  0.191 s    [User: 10.491 s, System: 1.905 s]
  Range (min … max):   10.320 s … 10.785 s    10 runs

Benchmark 2: ein t h
 ⠴ Current estimate: 9.024 s      00:00:4
  Time (mean ± σ):      8.979 s ±  0.363 s    [User: 10.154 s, System: 0.641 s]
  Range (min … max):    8.597 s …  9.702 s    10 runs

Summary
  ein t h ran
    1.17 ± 0.05 times faster than /Users/byron/dev/github.com/Byron/gitoxide/target/release/ein t h
```

And yes, whenever I see these numbers I think that this really can't be that much of a difference overall, but it really is 116.9k commits/s (formerly ~115k) compared to ~140k).

epage:
Don't worry about me; I'm not satisfied until the winnow version is faster!

epage:
I backported general performance improvements to 0.3 and 0.4 so I can better compare over time how upgrades and code changes affect things.

**TagRefIter(sig)**
1b957c524 (main): 163.69 ns
ad3966245: 235.79 ns
79b32c422 (w1): 256.13 ns
bc75c4d9d (w2): 245.02 ns
56042843d (w3): 238.99 ns
ae296d8d6 (w4): 239.05 ns
d11e59ffb (w5): 234.40 ns
e698b5015 (winnow04): 245.44 ns
9be554a10 (w6): 259.99 ns
3de8864db (w7): 250.35 ns
e8f8aaf13: 243.84 ns
eebf07054 (w8): 241.05 ns
b8d895ab2 (w9): 243.87 ns
3089b4cdf: 256.58 ns
f1d72ace8: 258.12 ns
244a4c965: 258.98 ns
8c9993600: 256.84 ns
680ef16a4: 257.59 ns
2de799dbb (winnow05): 258.47 ns
4351eb504: 254.44 ns
62729b49b (winnow): 245.47 ns

Better but still not the trend I was expecting.  More investigation is needed

Byron:
It's amazing to see that whatever happened there is affecting performance so much for `gitoxide`, and I can imagine it to be quite some detective work.

Performance is definitely an unsolved problem for me as well right now as the only one preventing regressions, possibly, is me with a few specific runtimes that I remember so regressions are noticed. With `gix-corpus` it's already possible to collect a lot of data around an operation of choice, but I have no means of visualising it or get automated warnings like with local `criterion` runs. Maybe a good insurance here is to always run certain performance tests after upgrading key dependencies, like `zip`, `sha1` or `winnow`, by hand but better than nothing.

Anyway, I am quite confident that this experience will lead to many improvements that ultimately will be beneficial to the entire ecosystem.

epage:
> Maybe a good insurance here is to always run certain performance tests after upgrading key dependencies, like zip, sha1 or winnow, by hand but better than nothing.

I'm very interested in adopting callgrind-based benchmarks and using https://bencher.dev/ to track them so this is treated like code coverage in PRs

epage:
I think I've narrowed down the major slowdown to winnow-rs/winnow#86.

Byron:
https://bencher.dev looks like it could do the job indeed and is probably something to use until there are tools that can do that in a decentralised fashion.

> I think I've narrowed down the major slowdown to [winnow-rs/winnow#86](https://github.com/winnow-rs/winnow/pull/86).

I will try that tomorrow :).

epage:
With winnow-rs/winnow#316 and winnow-rs/winnow#317 merged into 0.5 and backported to 0.4 and 0.3

**TagRefIter(sig)**
510192e0e (main): 165.88 ns
ee75de1e6: 159.62 ns
86d7fd184 (w1): 157.66 ns
ac0e81c41 (w2): 140.30 ns
b37a909a5 (w3): 147.65 ns
fee441da8 (w4): 148.87 ns
b3f041829 (w5): 157.37 ns
86ea47f28 (winnow04): 157.88 ns
9ed7df0a1 (w6): 157.88 ns
f0cbf81a3 (w7): 158.08 ns
2fb4a54c9: 158.66 ns
12f03db64 (w8): 175.76 ns
3f8c91fa4 (w9): 155.50 ns
a07590cb4: 155.83 ns
266864f35: 155.77 ns
df226dd31: 159.27 ns
93fc441c6: 158.16 ns
62f7bc65c: 160.28 ns
be1f4b5d2 (winnow05): 155.82 ns
80d799177: 155.48 ns
b7ee9a1d0 (winnow): 162.52 ns

And for across-the-board
```console
$ cargo bench -p gix-object@0.34.0 --bench decode-objects
     Running benches/decode_objects.rs (target/release/deps/decode_objects-2030aacefe52a773)
CommitRef(sig)          time:   [536.77 ns 537.97 ns 539.26 ns]
                        change: [-34.148% -33.943% -33.752%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  7 (7.00%) high mild

CommitRefIter(sig)      time:   [648.26 ns 662.31 ns 684.63 ns]
                        change: [-30.898% -29.723% -28.166%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  1 (1.00%) high mild
  8 (8.00%) high severe

TagRef(sig)             time:   [170.18 ns 170.50 ns 170.86 ns]
                        change: [+4.3542% +4.7672% +5.1841%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe

TagRefIter(sig)         time:   [174.34 ns 174.55 ns 174.77 ns]
                        change: [+1.5350% +2.1648% +2.7832%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
  1 (1.00%) high severe

TreeRef(sig)            time:   [101.54 ns 101.70 ns 101.88 ns]
                        change: [+7.2389% +7.5092% +7.7863%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

TreeRefIter(sig)        time:   [46.948 ns 46.997 ns 47.054 ns]
                        change: [-6.4018% -6.2109% -5.9962%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  8 (8.00%) high mild
  3 (3.00%) high severe
```

Seeing some of those lows, I feel like we could do better

Byron:
It's strange that these benchmark improvements seem to have no bearing on the real-world test.

```
❯ /Users/byron/dev/github.com/Byron/gitoxide/target/release/ein t h
 08:25:29 traverse commit graph done 1.2M commits in 10.35s (113.0k commits/s)
 08:25:29        estimate-hours Extracted and organized data from 1169974 commits in 15.651417ms (74751952 commits/s)
total hours: 1119736.25
total 8h days: 139967.03
total commits = 1169974
total authors: 31645
total unique authors: 24103 (23.83% duplication)

linux ( master) +369 -819 [!] took 10s
❯ ein t h
 08:25:54 traverse commit graph done 1.2M commits in 8.50s (137.6k commits/s)
 08:25:54        estimate-hours Extracted and organized data from 1169974 commits in 15.961291ms (73300712 commits/s)
total hours: 1119736.25
total 8h days: 139967.03
total commits = 1169974
total authors: 31645
total unique authors: 24103 (23.83% duplication)
```

*(The cache was hot for both runs and I made sure the the `…/target/release/ein` binary is truly the one from b7ee9a1)*

I am absolutely puzzled by that. Is this something you can reproduce?

epage:
```console
$ git switch main
$ cd ..
$ cp -r gitoxide gitoxide1
$ cd gitoxide 1
$ git switch winnow
$ cd ../linux
$
$ cargo run --release --manifest-path ../gitoxide/Cargo.toml --bin ein -- t h
$ cargo run --release --manifest-path ../gitoxide/Cargo.toml --bin ein -- t h
$ cargo run --release --manifest-path ../gitoxide/Cargo.toml --bin ein -- t h
$
$ cargo run --release --manifest-path ../gitoxide1/Cargo.toml --bin ein -- t h
$ cargo run --release --manifest-path ../gitoxide1/Cargo.toml --bin ein -- t h
$ cargo run --release --manifest-path ../gitoxide1/Cargo.toml --bin ein -- t h
```
I found that both `main` and `winnow` had the same range of completion times, from ~13.5ms to 15ms.

Byron:
Thank you, this probably means that it's an issue with the way it is compiled. On x86 there seem to be optimizations that don't kick in anymore on ARM64.

What's more surprising is that after a `cargo clean`, the `main`/nom version at 510192e0e delivers just 112.6k commits/s, whereas the `winnow` version at b7ee9a1d0 gets up to 115k commits/s on a fresh build as well.

But here it comes. The version I produced with `cargo install` that is from 9ef69e14d delivers 139.1k commits/s. After `cargo install`ing the nom version at 510192e0e I am getting the same ~139 commit/s.

Now, after `cargo install`ing the `winnow` version at b7ee9a1d0 I also see numbers around 139k commits/s. Running  `./target/release/ein` after installing it directly also yields the good values. However, rebuilding it with `cargo build --release --bin ein` shows the bad values once again. 

The binary sizes are similar, so there must be something about `cargo install` that makes all the difference - maybe it builds something like `target-cpu=native`?

epage:
> The binary sizes are similar, so there must be something about cargo install that makes all the difference - maybe it builds something like target-cpu=native?

Looking at the code, I'm not seeing `--target-cpu` being used anywhere. The only difference I see with `cargo build` is that it hard codes `--release`.  Might be interesting to look at verbose output in case I missed anything.

Byron:
I didn't see it that with `-v` either, but tried to build with `-C target-cpu=native` - the resulting binary was just as fast as the bad case. In desperation, I wanted to see if `~/.cargo/bin` is a magic place, and fortunately after copying the 'bad' binary onto the 'good' one it was still slow when executing it through the `PATH`.

But then it struck me: `cargo install` runs on the latest available dependencies, whereas `cargo build` runs on `Cargo.lock`. After running `cargo update` it actually got as fast as it should be even after using `cargo build --release`.

I think this case is solved now 🎉.

Byron:
Performance-wise, it seems like the benchmarks were a great indicator after all and by the looks of it (comparing `winnow` to `nom`), all values are better or have regressed by less than 5% (`TagRef`).

```
CommitRef(sig)          time:   [1.1752 µs 1.1888 µs 1.2056 µs]
                        change: [+24.581% +25.593% +26.850%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 8 outliers among 100 measurements (8.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
  6 (6.00%) high severe

CommitRefIter(sig)      time:   [1.2681 µs 1.2690 µs 1.2699 µs]
                        change: [+20.260% +20.443% +20.635%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 8 outliers among 100 measurements (8.00%)
  7 (7.00%) high mild
  1 (1.00%) high severe

TagRef(sig)             time:   [245.54 ns 245.73 ns 245.95 ns]
                        change: [-5.2706% -4.9124% -4.6490%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  2 (2.00%) low mild
  7 (7.00%) high mild
  1 (1.00%) high severe

TagRefIter(sig)         time:   [243.32 ns 243.75 ns 244.42 ns]
                        change: [-6.7406% -6.4258% -6.0412%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  1 (1.00%) low mild
  4 (4.00%) high mild
  3 (3.00%) high severe

TreeRef(sig)            time:   [138.67 ns 139.40 ns 140.11 ns]
                        change: [+4.5696% +5.1061% +5.6137%] (p = 0.00 < 0.05)
                        Performance has regressed.

TreeRefIter(sig)        time:   [61.047 ns 61.104 ns 61.186 ns]
                        change: [-0.9602% -0.8084% -0.6524%] (p = 0.00 < 0.05)
Found 7 outliers among 100 measurements (7.00%)
  1 (1.00%) low mild
  5 (5.00%) high mild
  1 (1.00%) high severe
```

Thus I think performance won't be preventing a merge anymore, quite the opposite :).

epage:
I leave it to you.  For me, there are more things I want to look at but that can be done independent of this being merged.

epage:
I dropped b7ee9a1  because it just seemed to slow things down (oddly enough)

Byron:
> I leave it to you. For me, there are more things I want to look at but that can be done independent of this being merged.

Great, then I will work on merging this PR now as it seems ready in every way, while looking forward to seeing those ideas of yours to come to fruition.

Before I go: I have a feeling your CPU might be twice as fast as mine. With that said, what's the commits/s you get when running `ein t h` on the linux kernel? The best number I have seen is ~139k commits/s.

epage:
```console
$ cargo run --release --manifest-path ../gitoxide/Cargo.toml --bin ein -- t h
    Finished release [optimized] target(s) in 0.09s
     Running `/home/epage/src/personal/gitoxide/target/release/ein t h`
 14:45:32 traverse commit graph done 1.1M commits in 8.30s (131.6k commits/s)
 14:45:32        estimate-hours Extracted and organized data from 1092424 commits in 13.815333ms (79073304 commits/s)
total hours: 1050880.75
total 8h days: 131360.09
total commits = 1092424
total authors: 29848
total unique authors: 22742 (23.81% duplication)
```

Byron:
Aw, snap, I hoped to see something in the 200k range. Of course, the workload is bound by zlib performance and parsing isn't holding anything back here, so improvements are probably out of range unless object decompression gets faster over time. Oh well, I keep an eye on that number with a new baseline of ~140k commits/s expecting it to go up over time.

I guess the real benchmark is against `git shortlog`:

```
✦ ❯ /usr/bin/time -lp git shortlog >/dev/null
real 11.27
user 10.95
sys 0.31
          2335965184  maximum resident set size
                   0  average shared memory size
                   0  average unshared data size
                   0  average unshared stack size
              143036  page reclaims
                   7  page faults
                   0  swaps
                   0  block input operations
                   0  block output operations
                   0  messages sent
                   0  messages received
                   0  signals received
                   0  voluntary context switches
                 229  involuntary context switches
         75705099109  instructions retired
         36068289312  cycles elapsed
          1516510016  peak memory footprint

linux ( master) +369 -819 [!] took 11s
✦ ❯ /usr/bin/time -lp ein t h
 22:00:38 traverse commit graph done 1.2M commits in 8.49s (137.8k commits/s)
 22:00:38        estimate-hours Extracted and organized data from 1169974 commits in 14.844667ms (78814432 commits/s)
total hours: 1119736.25
total 8h days: 139967.03
total commits = 1169974
total authors: 31645
total unique authors: 24103 (23.83% duplication)
real 8.84
user 10.03
sys 0.56
          1024720896  maximum resident set size
                   0  average shared memory size
                   0  average unshared data size
                   0  average unshared stack size
               66234  page reclaims
                 240  page faults
                   0  swaps
                   0  block input operations
                   0  block output operations
                   0  messages sent
                   0  messages received
                   0  signals received
                   5  voluntary context switches
              228344  involuntary context switches
         67037989003  instructions retired
         32757758823  cycles elapsed
           209817216  peak memory footprint
```

Interestingly, besides memory usage, `ein` is just a little bit faster if it was single-threaded, but can truly get ahead thanks to trivial multi-threading (trivial because Rust makes it so).

Byron (approved):
Thanks a million for working on this topic! It's a great outcome and I couldn't have wished for more.

PR #957, opened 2023-08-02
Remove `todo!` for submodule entry in checkout

Jake-Shadle:
I'm trying to convert a project from git2 -> gix (PR https://github.com/EmbarkStudios/cargo-fetcher/pull/180), but gix currently doesn't have any functionality for submodules, so I rolled my own in the meantime since my needs are minimal (eg clone/fetch/checkout only), but unfortunately ran into this `todo!` that prevented me from using checkout. I realize why this is here, but IMO just skipping submodules is fine in this case as it then acts kind of like normal git CLI usage since submodule checkout has to be done separately from a clone/checkout anyhow.

I could also add an option to https://github.com/Byron/gitoxide/blob/daf338989e7d75d41efb1345ed95ee2a13c97007/gix-worktree/src/checkout/chunk.rs#L101-L107, that will default to panicing like the current code, if you prefer that approach  instead.

Basically, using this simple patch, I get full parity between the output of my program via gix as cargo does via git2.

### DIFF
--- a/gix-worktree/src/checkout/entry.rs
+++ b/gix-worktree/src/checkout/entry.rs
@@ -165,7 +165,7 @@ where
             obj.data.len()
         }
         gix_index::entry::Mode::DIR => todo!(),
-        gix_index::entry::Mode::COMMIT => todo!(),
+        gix_index::entry::Mode::COMMIT => 0,
         _ => unreachable!(),
     };
     Ok(Outcome::Written { bytes: object_size })


Byron:
I agree that skipping them is preferable, and `gix` should do better, by now, than leaving `todo!()`s that are avoidable.

I have also removed the `todo!()` when dealing with sparse directories and added a trace instead for some feedback should people use `gix --trace` for instance, or if it's enabled in the application using `gix` (the crate).

> Basically, using this simple patch, I get full parity between the output of my program via gix as cargo does via git2.

It's too bad that I was too slow as submodule support during checkout is next on the list. I will try to CC you when it's ready though.

I will merge this PR when CI is green and hope to create a new release with a set of fixes and improvements this week.

Byron:
I will take care of the failure and get it merged.

Byron:
> It's no big deal, it gave me a better understanding of this codebase and git submodules in general.

I will be sure to take a look at your implementation as well, it's great to have for reference (minimal or not).

Further, I will add a link to `tame-index` in `crates-index` :). I have a feeling that `tame-index` could easily have been `crates-index` 2.0 as well, amazing work. I can probably port over many of the improvements and everyone will be happier for it. Also, great name :)!

PR #958, opened 2023-08-02
gix-object: fix test for struct size for 32-bit architectures

decathorpe:
The size of gix_object::Data is 24 bytes only on 64-bit architectures, on 32-bit architectures it's exactly half that (12 bytes).


Byron:
Thanks a lot for the fix! I am glad that is the only place where 32/64 bit are an issue.
I am working on a new release and will be sure this PR is included.

PR #975, opened 2023-08-09
handlers mt

Byron:
- feat!: improve `interrupt::init_handler()` to be usable from multiple threads
- adapt to changes in `gix`


Byron:
Merged already

Byron:
For one it's already in the dependency tree, so (now) it's free. But the major selling point for me is to not have to deal with poisoned locks. Most just `.unwrap()` and I really don't want to have that. Of course, it could be an `.expect("not poisoned")`, but it's definitely more cumbersome than just relying on `parking_lot`.
I guess the short answer would then be: ergonomics :).

PR #976, opened 2023-08-13
Fix `git-transport` reqwest client: Support redirect

### LINKED ISSUE #974: Redirects are not followed (at least with reqwest)
### Duplicates

### Current behavior 😯

When you fetch from a URL that redirects, the redirect does not seem to be followed and an error occurs. For example, in gitlab.com, all repository urls _must_ end with `.git`, otherwise they get redirected, as seen in the output from git itself.

```
% git clone https://gitlab.com/gilrs-project/gilrs
Cloning into 'gilrs'...
warning: redirecting to https://gitlab.com/gilrs-project/gilrs.git/
remote: Enumerating objects: 4667, done.
remote: Counting objects: 100% (21/21), done.
remote: Compressing objects: 100% (21/21), done.
remote: Total 4667 (delta 5), reused 12 (delta 0), pack-reused 4646
Receiving objects: 100% (4667/4667), 819.00 KiB | 3.72 MiB/s, done.
Resolving deltas: 100% (2962/2962), done.
```

Instead, at least with the reqwest backend (unsure about curl), this leads to an error

```
An IO error occurred when talking to the server: Received HTTP status 422
```

This error code is...wrong, but I don't know if that's a "this is how the git protocol does redirects" or just a gitlab oddity, but at least git understands it.

### Expected behavior 🤔

The redirect should be followed and the repo fetch successfully. In this case it's a trivial fix to just add the `.git` extension so that a redirect is not needed, but just given the status code this was not really clear until I did the same operation in git which printed out the above warning.

### Steps to reproduce 🕹

1. Attempt to clone `https://gitlab.com/gilrs-project/gilrs` with gix
2. Get I/O error with 422 status code

NobodyXu:
Fixed #974

### DIFF
--- a/gix-transport/src/client/blocking_io/http/reqwest/remote.rs
+++ b/gix-transport/src/client/blocking_io/http/reqwest/remote.rs
@@ -42,32 +42,30 @@ impl Default for Remote {
         let handle = std::thread::spawn(move || -> Result<(), Error> {
             let mut follow = None;
             let mut redirected_base_url = None::<String>;
-
-            // This initial value is just a place holder, it will get modified
-            // before `client.execute` is called.
-            let current_follow_policy = Arc::new(atomic::AtomicBool::new(false));
-
-            let follow_policy = Arc::clone(&current_follow_policy);
+            let allow_redirects = Arc::new(atomic::AtomicBool::new(false));
 
             // We may error while configuring, which is expected as part of the internal protocol. The error will be
             // received and the sender of the request might restart us.
             let client = reqwest::blocking::ClientBuilder::new()
                 .connect_timeout(std::time::Duration::from_secs(20))
                 .http1_title_case_headers()
-                .redirect(reqwest::redirect::Policy::custom(move |attempt| {
-                    if follow_policy.load(atomic::Ordering::Relaxed) {
-                        let curr_url = attempt.url();
-                        let prev_urls = attempt.previous();
+                .redirect(reqwest::redirect::Policy::custom({
+                    let allow_redirects = allow_redirects.clone();
+                    move |attempt| {
+                        if allow_redirects.load(atomic::Ordering::Relaxed) {
+                            let curr_url = attempt.url();
+                            let prev_urls = attempt.previous();
 
-                        match prev_urls.first() {
-                            Some(prev_url) if prev_url.host_str() != curr_url.host_str() => {
-                                // gix does not want to be redirected to a different host.
-                                attempt.stop()
+                            match prev_urls.first() {
+                                Some(prev_url) if prev_url.host_str() != curr_url.host_str() => {
+                                    // git does not want to be redirected to a different host.
+                                    attempt.stop()
+                                }
+                                _ => attempt.follow(),
                             }
-                            _ => attempt.follow(),
+                        } else {
+                            attempt.stop()
                         }
-                    } else {
-                        attempt.stop()
                     }
                 }))
                 .build()?;
@@ -121,12 +119,8 @@ impl Default for Remote {
                 }
 
                 let follow = follow.get_or_insert(config.follow_redirects);
-
-                current_follow_policy.store(
-                    match *follow {
-                        FollowRedirects::None => false,
-                        FollowRedirects::Initial | FollowRedirects::All => true,
-                    },
+                allow_redirects.store(
+                    matches!(follow, FollowRedirects::Initial | FollowRedirects::All),
                     atomic::Ordering::Relaxed,
                 );
 
--- a/gix-transport/src/client/blocking_io/http/reqwest/remote.rs
+++ b/gix-transport/src/client/blocking_io/http/reqwest/remote.rs
@@ -3,14 +3,12 @@ use std::{
     convert::TryFrom,
     io::{Read, Write},
     str::FromStr,
+    sync::{atomic, Arc},
 };
 
 use gix_features::io::pipe;
 
-use crate::client::{
-    http,
-    http::{reqwest::Remote, traits::PostBodyDataKind},
-};
+use crate::client::http::{self, options::FollowRedirects, redirect, reqwest::Remote, traits::PostBodyDataKind};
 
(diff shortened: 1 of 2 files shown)


NobodyXu:
@Byron I've verified this locally and it seems to be working:

```
❯ ./gix clone https://github.com/NobodyXu/openssh-sftp-client

 18:21:26 read pack done 1.5MB in 3.09s (483.5KB/s)                                                      
 18:21:26  indexing done 5.3k objects in 3.06s (1.7k objects/s)                                          
 18:21:26 decompressing done 2.3MB in 3.06s (768.9KB/s)                                                  
 18:21:26     Resolving done 5.3k objects in 0.17s (32.0k objects/s)                                     
 18:21:26      Decoding done 14.9MB in 0.17s (89.9MB/s)                                                  
 18:21:26 writing index file done 149.2KB in 0.00s (71.2MB/s)                                            
 18:21:26  create index file done 5.3k objects in 3.23s (1.6k objects/s)                                 
 18:21:26           checkout done 62.0 files in 0.01s (5.9k files/s)
 18:21:26            writing done 270.3KB in 0.01s (25.9MB/s)
HEAD:refs/remotes/origin/HEAD (implicit)
        195c77ff87d6e3be094076dc07e35320534e70be HEAD symref-target:refs/heads/main -> refs/remotes/origi
n/HEAD [new]
+refs/heads/*:refs/remotes/origin/*
        366247497ed5fe03139df2914c28c301baaaa04f refs/heads/debug-sccache -> refs/remotes/origin/debug-sc
cache [new]
        195c77ff87d6e3be094076dc07e35320534e70be refs/heads/main -> refs/remotes/origin/main [new]
refs/tags/*:refs/tags/* (implicit, due to auto-tag)
        6bb14a8ba3ce09c381a98926497b221daa34e674 refs/tags/0.13.0 -> refs/tags/0.13.0 [new]
        22ef52a7b7928346183df4d6b568cf0b5d112f72 refs/tags/0.13.1 -> refs/tags/0.13.1 [new]
        e9135a66c158ac6d569826ea1d7f9eeee3e14707 refs/tags/0.13.10 -> refs/tags/0.13.10 [new]
        a3d9c578fa813228ed05764ee3d23f1f87cba000 refs/tags/0.13.2 -> refs/tags/0.13.2 [new]
        e11a8dc7bbdc6197d00d94a97fb36e9e118eae56 refs/tags/0.13.3 -> refs/tags/0.13.3 [new]
        ec5789501971ed6a5bd8382c2d02d185a21f9bb0 refs/tags/0.13.4 -> refs/tags/0.13.4 [new]
        a31798d30fce4024183a19810c987d48c2e7e459 refs/tags/0.13.5 -> refs/tags/0.13.5 [new]
        cd467dc70974562e3a6abeed00273ebb6d5a7f1c refs/tags/0.13.6 -> refs/tags/0.13.6 [new]
        7ac96c139ad1c72ab69cb2cff678d23b9aedf372 refs/tags/0.13.7 -> refs/tags/0.13.7 [new]
        5f4799a74d67f32c06d84ffe44d67b92c290768f refs/tags/0.13.8 -> refs/tags/0.13.8 [new]
        9f7243da2035273dde4e87145f1da0cf74a65af9 refs/tags/0.13.9 -> refs/tags/0.13.9 [new]

```

gix-transport/src/client/blocking_io/http/reqwest/remote.rs
```diff
@@ -3,14 +3,12 @@ use std::{
...
 use gix_features::io::pipe;
 
-use crate::client::{
-    http,
-    http::{reqwest::Remote, traits::PostBodyDataKind},
-};
+use crate::client::http::{self, options::FollowRedirects, redirect, reqwest::Remote, traits::PostBodyDataKind};
 
 /// The error returned by the 'remote' helper, a purely internal construct to perform http requests.
 #[derive(Debug, thiserror::Error)]
```
NobodyXu, line 14: P.S. I think it will be better if `#[non_exhaustive]` is added to all error types so that adding new variant won't be a breaking change.

Byron (approved):
Thanks a million, this is very substantial improvement for the backend!

I just wonder about one thing and maybe you can help. Would it be possible, with the current implementation, to be redirected indefinitely to different URLs of the same host?
Here with `reqwest` it seems to be trivial to check for `prev_urls` and stop the attempt if there is more than one. Should we do that?

It's unclear to me as well what the `curl` backend does in that regard, both here in this crate or [in `git`](https://github.com/git/git/blob/fac96dfbb1c24369ba7d37a5affd8adfe6c650fd/http.c#L2038).

NobodyXu:
> I just wonder about one thing and maybe you can help. Would it be possible, with the current implementation, to be redirected indefinitely to different URLs of the same host?

Yes, it's definitely possible with [`reqwest::redirect::Attempt::`](https://docs.rs/reqwest/latest/reqwest/redirect/struct.Attempt.html#method.previous) and it gives the full url, which is how I used it to stop from being redirected to another host.

> Here with reqwest it seems to be trivial to check for prev_urls and stop the attempt if there is more than one. Should we do that?

Perhaps I can open another PR for this?

> It's unclear to me as well what the curl backend does in that regard, both here in this crate or [in git](https://github.com/git/git/blob/fac96dfbb1c24369ba7d37a5affd8adfe6c650fd/http.c#L2038).

According to [here](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&cad=rja&uact=8&ved=2ahUKEwiu47W-q9mAAxXBxTgGHY8XCSsQFnoECBYQAQ&url=https%3A%2F%2Feverything.curl.dev%2Fhttp%2Fredirects&usg=AOvVaw0Qyd2KX4PMGctovyyXlxz9&opi=89978449):

> When following redirects is enabled, curl will follow up to 50 redirects by default. There is a maximum limit mostly to avoid the risk of getting caught in endless loops.

Byron:
Thanks again for digging this out! After checking again I convinced myself that `git` seems to be relying on the curl default, so for now I changed it to do the same. 50 seems like a mighty high number, but should this be abused we can always lower it or make it configurable.

PR #985, opened 2023-08-17
gix-config fuzz test with non-linear runtime

Byron:
This test-case was found by the oss-fuzz project and it shows long parsing times with `winnow`
of 10s in debug *and* release mode on my machine.
The `nom` time of this is 1ms, so this appears to be a side-effect of *maybe* the improvements to the parsing logic when switching to `winnow`.

Screenshot 2023-08-17 at 19 08 29
Screenshot 2023-08-17 at 19 09 29

@epage I put this reproducer directly onto the 'winnow' branch to validate that this is indeed still an issue. Maybe you have an idea on how to fix this. Thanks in advance for your help.


Byron:
As per request this reproducer was brought into `main`. If needed, 510192e0e5750bdfe461d701b3e124c03f22b7d9 can be reverted, which could be the case if CI times get too high.

PR #986, opened 2023-08-18
Improve gix-config parsing

epage:
The biggest improvement is in a degenerate case highlighted by 510192e0e5750bdfe461d701b3e124c03f22b7d9 which causes `d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=d[W]t=...\r` to be parsed as a value, failing when there isn't a `\n` after the `\r`, and then retrying just a couple more bytes in, slowly eating away at the input until it is empty.

This PR includes two flavors for fixing it.  The first reverts the behavior back to something more like pre-fe997c8a32b0a1dc3ad3613fe8dd48a987c3a5bf where we didn't advance `i` on failure, hitting a break if parsing didn't progress. The problem is we discard any error information in this case, so I then updated it to be slightly closer to how I would write a normal parser (if we didn't have `dispatch`).

Before I found that root cause, I found that `next_token` calls in `value_impl` were taken up significant amounts of time and that adding a fast-path for values dramatically improved it (from 8s to 5s on my machine).  I'm assuming this will carry forward to helping normal parsing cases so I left it in.

### DIFF
--- a/gix-config/src/parse/nom/tests.rs
+++ b/gix-config/src/parse/nom/tests.rs
@@ -808,7 +808,12 @@ mod key_value_pair {
     fn nonascii_is_allowed_for_values_but_not_for_keys() {
         let mut node = ParseNode::SectionHeader;
         let mut vec = Default::default();
-        assert!(key_value("你好".as_bytes(), &mut node, &mut vec).is_err());
+        // Verifying `is_ok` because bad keys get ignored
+        assert!(key_value("你好".as_bytes(), &mut node, &mut vec).is_ok());
+        assert_eq!(vec, into_events(vec![]));
+
+        let mut node = ParseNode::SectionHeader;
+        let mut vec = Default::default();
         assert!(key_value("a = 你好 ".as_bytes(), &mut node, &mut vec).is_ok());
         assert_eq!(
             vec,
--- a/gix-config/src/parse/nom/mod.rs
+++ b/gix-config/src/parse/nom/mod.rs
@@ -121,10 +121,7 @@ fn section<'i>(
             dispatch(Event::Newline(Cow::Borrowed(v.as_bstr())));
         }
 
-        let before_key = i.checkpoint();
-        if key_value_pair(i, node, dispatch).is_err() {
-            i.reset(before_key);
-        }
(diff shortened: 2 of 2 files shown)


gix-config/src/parse/nom/tests.rs
```diff
@@ -808,7 +808,8 @@ mod key_value_pair {
     fn nonascii_is_allowed_for_values_but_not_for_keys() {
         let mut node = ParseNode::SectionHeader;
         let mut vec = Default::default();
-        assert!(key_value("你好".as_bytes(), &mut node, &mut vec).is_err());
+        assert!(key_value("你好".as_bytes(), &mut node, &mut vec).is_ok());
```
Byron, line 811: Maybe it's explained in the PR body, but I don't understand how this can now parse OK. A second argument to `assert!()` could help clarify that especially in the presence of a test that claims that nonascii keys aren't allowed.

Besides that, I have no trouble with it as this is an internal parser and as long as the rest of the machinery works with it, all should be fine.

Byron (commented):
Thanks again for digging into this! With these parsers, I am now at a place where I hope I don't have to touch them again and that the test-suite covers everything 😅.

I left one remark of a possible improvement merely for clarification - if I happen to merge this PR before you get to it, it's totally OK to leave it as a comment here so I can add it after the merge.

Thank you

gix-config/src/parse/nom/tests.rs
epage: Before, `section` ignored the errors from `key_value_pair, so whether an error happened or not doesn't actually matter so long as the holistic behavior was correct.

Now, we ignore `config_name` errors so we can propagate `config_value` errors past `section`.

I've added a comment in the test to clarify things.

Byron: Perfect, thanks.

PR #988, opened 2023-08-19
fix(docs): fix incorrect s/git-config/gix-config/

not-my-profile:
3a861c8f049f6502d3bcbdac752659aa1aeda46a just blindly replaced any occurence of "git-config" or "git_config" with "gix-config"/"gix_config".

There is no such thing as a gix-config file.
gix-config is a git-config file parser.


Byron (approved):
A great catch, thanks so much!

PR #990, opened 2023-08-19
Rewrite parsing logic of gix-url

niklaswimmer:
The goal of this PR is to make the implementation of the url parser simpler. I consider this PR done when the test suite (without the baseline tests) succeeds again. Follow up PRs can gradually improve the results of the baseline tests until we can turn it on by default.

The old implementation is not yet removed, because I use it as a reference and that works way better if the file still exists in my local repository. It can be removed before merging this PR.

The first two commits of this PR I wanted to merge together with #965 but forgot to push them :)


Byron:
Is there anything I can do to help with this PR?
An option could also be to reduce its scope to get intermediate, but better-than-before, results merged.
Thanks for letting me know 🙏.

niklaswimmer:
I am sorry for the long silence, please know that I am still interested in this. And now that I have finally finished the Netflix series I have been spending my free time on for the past few weeks, I am confident that I will be finishing this PR soon :)

niklaswimmer:
It will probably take me another few days to finish Windows, especially because using a VM and everything slows down my workflow. But we will see how much needs fixing

Byron:
If the work with windows is too aggravating, I can also do it for you - with my setup it's still quite terrible, but for `gix-url` it will work well enough. In other words, I don't want you to loose momentum or happiness over windows, it's not worth it 😁.
Just let me know.

niklaswimmer:
No worries, I think my workflow will be quite workable - I just need to set it up first, which I am doing right now. If I have any problems I will let you know!

niklaswimmer:
There are only two failing tests left and they both seem to stem from backslash handling. So it seems like that is the last thing I need to figure out.

Byron:
Thanks for the update, that's great progress! For a moment I thought this also includes the baseline tests, but realized that this isn't the case yet. However, when enabling them, it already works better than what's on `main` which as 240 failures, compared to only 165 failures on this branch.

With that said, regarding the windows failures, it's probably OK to also see what git does in these moments to be sure they are expecting the right thing.

One of the failing tests talks about paths like `/C:/…` (d6f90beac37866f992a1714d38e5b320eea6f1bb) which seems legit. But even here one might realize that `git` handles this differently or not at all, and `gix-url` might have to follow suit then.

niklaswimmer:
Jeah, the main reason why it takes me so long is because I have to check Gits implementation and output to find the correct behavior. But I get there, slowly but surely

Byron:
Just to make you aware of a fix that will hopefully soon land in `main`: Running `gix clone 'ssh://-oProxyCommand=open$IFS-aCalculator/foo'` leads to code execution unless the URL is sanitized or other measures are taken. I yet have to investigate how `git` ended up solving the issue, and thus if `gix-url` should sanitize such URLs or not.

More context is here: https://secure.phabricator.com/T12961 

As it stands, valid hostnames can be interpreted as arguments if they start with a dash, and that's probably where the fix has to be: validate that the hostname isn't ambiguous before passing it to `ssh` for execution.

niklaswimmer:
I wonder whether that check should happen in gix-url or only in the transport layer?

Byron:
As such `-arg` hostnames are indeed valid hostnames, they may exist legitimately, probably the reason why the `url` crate doesn't do anything about it either.

A look at `git` revealed that [this function](https://github.com/git/git/blob/6807fcfedab84bc8cd0fbf721bc13c4e68cda9ae/path.c#L1544) is used to validate the hostname and basically all arguments that could be abused that way, i.e. originate from the URL. Thus I think `gix-transport` is the place to fix it, for both `ssh` and `git` protocols, I will look closely to not miss a spot.

niklaswimmer:
I think I am giving up. Git changes its behavior depending on whether it is called from bash or not and it seems to be because of some logic in mingw 

image

Byron:
Yes, please do not get bogged down by yet another windows-strangeness, and feel free to ignore these tests instead. Thanks to the rewrite, it's questionable whether these tests are still relevant anyway and I am happy to give it another look.

I hope you will regain your momentum and recover from this :).

Byron:
You probably want to rebase against `main` to get commit d80b5f69772a6e36b0131d3a538e896a8a6a29b1 , which calls to attention the dangers of using `host` and `path`.

niklaswimmer:
Do you want me to just rebase or to reserve the history/commit dates and merge main into this branch. I am happy to do either.

Byron:
You can do what's best for you, or what you see fit, I don't have a preference.

niklaswimmer:
Everything done now! I can write the changelog entry if you want - I probably have a much easier time summarizing the changes having worked on this for so long :) You would need to tell me how to do that correctly with cargo-smart-release though

Byron:
Great, I'd appreciate it!

I think the subject should be `feat: <subject>` if it's a non-breaking feature, or `feat!: <subject>` if it's a little breaking after all.

The body can contain whatever markdown you want for further explanation. This would have to go to at least one commit to be picked up by `cargo smart-release`.

Please feel free to rewrite commits as you see fit.

niklaswimmer:
I only focused on public API changes when marking my commits for inclusion in the changelog and explicitly left out any behavior changes that only affect what URLs we parse and how we parse them. My intend was to just write something along the lines of the following at the top of the changelog.

> This release contains many internal changes to the URL implementation that bring it closer to the reference implementation in Git. Gitoxide can now parse more URLs and be more correct while doing so! Highlights include the correct parsing of SSH aliases and an improvement to the handling of file URLs. Please note that there may be additional more subtle changes in the parsing behavior due to the rewrite. If you notice any derivations from how Git handles an URL please open an new issue!

Idk about the wording just did something quick. Or do you think we need to be more precise?

Byron:
Since the merge of `main` was the last thing that happened, I have removed it by resetting to https://github.com/Byron/gitoxide/pull/990/commits/49b278ae9c0a4927d38054c8f006cbd46e394d92 - please beware.

> Idk about the wording just did something quick. Or do you think we need to be more precise?

In theory, one would mark each feature or change as they happen to the respective crate by using conventional commit messages, exactly like you did. Adding a top-level paragraph in `gix` itself is a great idea - I think it's fair to put this underneath a new `## Unreleased` header in `CHANGELOG.md` and `gix/CHANGELOG.md` - that way different kinds of audiences are most likely to see it. Could you do that?

In the meantime, I will start reviewing and will probably push a few changes to this branch as well, so let's not force push anymore :).

Byron:
Oh, I realized that now after the reset, the changelogs are outdated. Then I think I have just discovered my preference for, in this case, rebasing on top of `main` instead of merging it. Then you can add the changelog entries, and when done, I can do the review and push smaller adjustments myself.

Thanks again 🙏

niklaswimmer:
Force pushed the rebase. No more forces from now on I agree :)

Byron:
Hehe, thanks! I think now you'd be able to make the CHANGELOG.md adjustments I mentioned earlier (just to have your name associated with the commit) and then I'd consider this ready-for-review if you agree :).

niklaswimmer:
Just to make sure, the top level changelog is for the binaries and the one in the `gix` folder for the lib, right?

Byron:
Yes, that's correct. In theory, you could also drop the paragraph into `gix-url/CHANGELOG.md` for completeness.

Byron (commented):
Thanks again for tackling this!

I am still in the process of reviewing everything, but wanted to share that I really like the new parser implementation. It's much cleaner for sure and great basis to get to a perfectly passing baseline test.

I even learned some tricks like the hidden and hard-to-discover `testing` module (which I hope one day we can get rid of), and the `parse` module trickery to be able to quickly swap out implementations.

My goal today is to complete the remaining tasks (beware: I *may* force push during the time) and then merge the PR. And while writing this, I hope you are still motivated drive the implementation to the level of perfection it deserves :).

### Tasks

* [x] run baseline tests, put them to work
     - baseline tests have been changed so that they are actually running and can work with what's there. Gradual improvements can be done that way, while protecting from regressions.
* [x] re-add all deleted tests in such a way that it's clear whether they should pass, or that they don't pass for good reason.
     - All manually added tests where *probably* added with a specific purpose in mind and that's value to try not to loose.
* [x] assure there is support for non-UTF-8 file-paths, assure there is a test for that.
* [ ] Add fuzzing (similar to pathspec) - will be picked up by google oss fuzz as well

niklaswimmer:
Thanks again for your patience with this PR, you can definitely expext more in the next weeks :)

Issue #987, opened 2023-08-19
Ideas on how to improve compile-time


NobodyXu:
## De-monomorphisation

Function like [`gix::clone::PrepareFetch::new`](https://docs.rs/gix/latest/src/gix/clone/mod.rs.html#67-118) can be trivally de-monomorphised by extracting an `inner` function.

@Byron It's possible we can do this without sacrificing code readability by creating our own variant of [`momo`](https://docs.rs/momo).

For [`gix::clone::PrepareFetch::fetch_then_checkout`](https://docs.rs/gix/latest/gix/clone/struct.PrepareFetch.html#method.fetch_then_checkout) it would be harder since [`Progress`](https://docs.rs/gix/latest/gix/trait.Progress.html) is not an `dyn`-safe trait.

I would like to add another provided method to `Progress`:

```rust
/// An opaque type so that `gix` can internally change its implementation details,
/// e.g. use an `Arc` instead of `Box`, or even inlining the progress.
pub struct DynProgress;

impl DynProgress {
    pub fn new<P: Progress>(p: P) -> Self;
}

impl Progress for DynProgress {
    type SubProgress = Self;

    // ...
}

trait Progress {
    // ...

    // Provided method
    fn into_dyn_progressive(self) -> DynProgress;
}
```

## Features

For starters, I think we can put `gix_negotiate`, `gix_ignore` and `gix_utils`.

Then we can put ability to fetch and update new repository into its own features since many crates might just want to discover local repository, e.g. [`vergen`](https://docs.rs/vergen).

### Motivation 🔦

As shown in https://github.com/cargo-bins/cargo-binstall/pull/1304#issuecomment-1683936506 , compilation of gix (not its dependencies) on GHA MacOS runner took more than 60s, which is a bit too long.

`simple-git`, which only uses a few `gix` interface, also takes 34s just to compile.

I believe this is caused by two factors:
 - too much monomorphization
 - too many unused code and dependencies pulled in

Byron:
Thanks for bringing up ideas for a way forward to reducing compile times, it's very much appreciated and I think it's good to start this sooner than later.

##### Monomorphization
First of all, I agree that in *most* cases, monomorphization can be considered a convenience rather than a necessity (i.e. performance). Thus I believe that unless a function or closure isn't going to be called millions of times per second, there is no *need* to monomorphize it. Most methods and functions provided here fulfil this criterium. This affects most places where callbacks and the `Progress` trait is used.

[...]

PR #992, opened 2023-08-20
feature: Add new crate `gix-macros`

NobodyXu:
To make de-monomorphizatioin easier without sacrification of code readability as mentioned in #987

## TODO

 - [x] `momo`: Put the inner function inside the `momo`ed function for any non-`self` function.
        E.g.
```
        fn f(s: impl AsRef<str>) {
            fn inner(s: &str) {}
            inner(s.as_ref())
        }
```
 - [x] Also Add tests for this on `this: Self` and without `Self`.
 - ~~Support parsing `impl` block so that we can apply `momo` to trait implementation~~
 - ~~`momo`: Remove unused generics in `TryInto`: Removed anything related to its `Error`~~
 - ~~`momo` `clone::PrepareFetch::new`~~
 - ~~`momo` `repository/object.rs`: `Repository::commit_as`~~
 - [x] `momo` discover.rs
 - [x] `momo` `init.rs`
 - [x] `momo` `open::repository.rs`
 - [x] Handle pattern in param name
 - [x] `momo` `create::into`
 - [x] (low-priority) fixed the unused mut errors: Remove all `mut` in the generic `fn` except for `AsMut`, and remove `mut` from `AsMut` target in inner `fn`. User can opt-out of this by annotating `#[keep]` in front of the `AsMut` param name.

## Expanded test code

Command for getting expanded code:

```bash
cargo expand  --test test
```

Actual expanded `tests/test.rs` code:

<details>

```rust
#![feature(prelude_import)]
#[prelude_import]
use std::prelude::rust_2021::*;
#[macro_use]
extern crate std;
use gix_macros::momo;
use std::pin::Pin;
struct Options;
#[allow(dead_code)]
fn test_open_opts_inner(
    _dir: impl Into<std::path::PathBuf>,
    _options: Options,
) -> Result<(), ()> {
    Ok(())
}
/// See if doc are kept
#[allow(dead_code)]
fn test_open_opts(
    directory: impl Into<std::path::PathBuf>,
    options: Options,
) -> Result<(), ()> {
    #[allow(unused_mut)]
    #[allow(dead_code)]
    fn _test_open_opts_inner_generated_by_gix_macro_momo(
        directory: std::path::PathBuf,
        options: Options,
    ) -> Result<(), ()> {
        test_open_opts_inner(directory, options)
    }
    _test_open_opts_inner_generated_by_gix_macro_momo(directory.into(), options)
}
#[allow(dead_code)]
fn test_with_try(
    t: impl TryInto<String, Error = ()>,
    _options: Options,
) -> Result<(), ()> {
    #[allow(unused_mut)]
    #[allow(dead_code)]
    fn _test_with_try_inner_generated_by_gix_macro_momo(
        t: String,
        _options: Options,
    ) -> Result<(), ()> {
        let t = t;
        t.strip_prefix('1').ok_or(())?;
        Ok(())
    }
    _test_with_try_inner_generated_by_gix_macro_momo(t.try_into()?, _options)
}
fn test_fn<E>(
    a: impl Into<String>,
    b: impl AsRef<str>,
    mut c: impl AsMut<str>,
    d: impl TryInto<String, Error = E>,
) -> Result<String, E> {
    #[allow(unused_mut)]
    fn _test_fn_inner_generated_by_gix_macro_momo<E>(
        a: String,
        b: &str,
        mut c: &mut str,
        d: String,
    ) -> Result<String, E> {
        let mut s = a;
        s += b;
        s += c;
        s += &d;
        Ok(s)
    }
    _test_fn_inner_generated_by_gix_macro_momo(
        a.into(),
        b.as_ref(),
        c.as_mut(),
        d.try_into()?,
    )
}
fn test_fn_call_style<E>(
    a: impl Into<String>,
    b: impl AsRef<str>,
    mut c: impl AsMut<str>,
    d: impl TryInto<String, Error = E>,
) -> Result<String, E> {
    #[allow(unused_mut)]
    fn _test_fn_call_style_inner_generated_by_gix_macro_momo<E>(
        a: String,
        b: &str,
        mut c: &mut str,
        d: String,
    ) -> Result<String, E> {
        let mut s = a;
        s += b;
        s += c;
        s += &d;
        Ok(s)
    }
    _test_fn_call_style_inner_generated_by_gix_macro_momo(
        a.into(),
        b.as_ref(),
        c.as_mut(),
        d.try_into()?,
    )
}
fn test_fn_where<A, B, C, D, E>(a: A, b: B, mut c: C, d: D) -> Result<String, E>
where
    A: Into<String>,
    B: AsRef<str>,
    C: AsMut<str>,
    D: TryInto<String, Error = E>,
{
    #[allow(unused_mut)]
    fn _test_fn_where_inner_generated_by_gix_macro_momo<E>(
        a: String,
        b: &str,
        mut c: &mut str,
        d: String,
    ) -> Result<String, E> {
        let mut s = a;
        s += b;
        s += c;
        s += &d;
        Ok(s)
    }
    _test_fn_where_inner_generated_by_gix_macro_momo(
        a.into(),
        b.as_ref(),
        c.as_mut(),
        d.try_into()?,
    )
}
struct TestStruct;
impl TestStruct {
    fn test_method<E>(
        self,
        a: impl Into<String>,
        b: impl AsRef<str>,
        mut c: impl AsMut<str>,
        d: impl TryInto<String, Error = E>,
    ) -> Result<String, E> {
        self._test_method_inner_generated_by_gix_macro_momo(
            a.into(),
            b.as_ref(),
            c.as_mut(),
            d.try_into()?,
        )
    }
    #[allow(unused_mut)]
    fn _test_method_inner_generated_by_gix_macro_momo<E>(
        self,
        a: String,
        b: &str,
        mut c: &mut str,
        d: String,
    ) -> Result<String, E> {
        let mut s = a;
        s += b;
        s += c;
        s += &d;
        Ok(s)
    }
    #[allow(clippy::needless_arbitrary_self_type)]
    fn test_method2<E>(
        self: Self,
        a: impl Into<String>,
        b: impl AsRef<str>,
        mut c: impl AsMut<str>,
        d: impl TryInto<String, Error = E>,
    ) -> Result<String, E> {
        self._test_method2_inner_generated_by_gix_macro_momo(
            a.into(),
            b.as_ref(),
            c.as_mut(),
            d.try_into()?,
        )
    }
    #[allow(unused_mut)]
    #[allow(clippy::needless_arbitrary_self_type)]
    fn _test_method2_inner_generated_by_gix_macro_momo<E>(
        self: Self,
        a: String,
        b: &str,
        mut c: &mut str,
        d: String,
    ) -> Result<String, E> {
        let mut s = a;
        s += b;
        s += c;
        s += &d;
        Ok(s)
    }
    fn test_fn<E>(
        a: impl Into<String>,
        b: impl AsRef<str>,
        mut c: impl AsMut<str>,
        d: impl TryInto<String, Error = E>,
        _e: (),
    ) -> Result<String, E> {
        #[allow(unused_mut)]
        fn _test_fn_inner_generated_by_gix_macro_momo<E>(
            a: String,
            b: &str,
            mut c: &mut str,
            d: String,
            _e: (),
        ) -> Result<String, E> {
            let mut s = a;
            s += b;
            s += c;
            s += &d;
            Ok(s)
        }
        _test_fn_inner_generated_by_gix_macro_momo(
            a.into(),
            b.as_ref(),
            c.as_mut(),
            d.try_into()?,
            _e,
        )
    }
    fn test_fn2<E>(
        _this: Self,
        a: impl Into<String>,
        b: impl AsRef<str>,
        mut c: impl AsMut<str>,
        d: impl TryInto<String, Error = E>,
        _e: (),
        _f: (),
    ) -> Result<String, E> {
        Self::_test_fn2_inner_generated_by_gix_macro_momo(
            _this,
            a.into(),
            b.as_ref(),
            c.as_mut(),
            d.try_into()?,
            _e,
            _f,
        )
    }
    #[allow(unused_mut)]
    fn _test_fn2_inner_generated_by_gix_macro_momo<E>(
        _this: Self,
        a: String,
        b: &str,
        mut c: &mut str,
        d: String,
        _e: (),
        _f: (),
    ) -> Result<String, E> {
        let mut s = a;
        s += b;
        s += c;
        s += &d;
        Ok(s)
    }
    fn test_fn3<E>(
        _this: Pin<&mut Self>,
        a: impl Into<String>,
        b: impl AsRef<str>,
        mut c: impl AsMut<str>,
        d: impl TryInto<String, Error = E>,
        _e: (),
        _f: (),
    ) -> Result<String, E> {
        Self::_test_fn3_inner_generated_by_gix_macro_momo(
            _this,
            a.into(),
            b.as_ref(),
            c.as_mut(),
            d.try_into()?,
            _e,
            _f,
        )
    }
    #[allow(unused_mut)]
    fn _test_fn3_inner_generated_by_gix_macro_momo<E>(
        _this: Pin<&mut Self>,
        a: String,
        b: &str,
        mut c: &mut str,
        d: String,
        _e: (),
        _f: (),
    ) -> Result<String, E> {
        let mut s = a;
        s += b;
        s += c;
        s += &d;
        Ok(s)
    }
}
struct S(bool);
impl TryInto<String> for S {
    type Error = ();
    fn try_into(self) -> Result<String, ()> {
        if self.0 { Ok(String::from("!2345")) } else { Err(()) }
    }
}
extern crate test;
#[cfg(test)]
#[rustc_test_marker = "test_basic_fn"]
pub const test_basic_fn: test::TestDescAndFn = test::TestDescAndFn {
    desc: test::TestDesc {
        name: test::StaticTestName("test_basic_fn"),
        ignore: false,
        ignore_message: ::core::option::Option::None,
        source_file: "gix-macros/tests/test.rs",
        start_line: 175usize,
        start_col: 4usize,
        end_line: 175usize,
        end_col: 17usize,
        compile_fail: false,
        no_run: false,
        should_panic: test::ShouldPanic::No,
        test_type: test::TestType::IntegrationTest,
    },
    testfn: test::StaticTestFn(|| test::assert_test_result(test_basic_fn())),
};
fn test_basic_fn() {
    match (
        &test_fn("12345", "12345", String::from("12345"), S(true)).unwrap(),
        &"123451234512345!2345",
    ) {
        (left_val, right_val) => {
            if !(*left_val == *right_val) {
                let kind = ::core::panicking::AssertKind::Eq;
                ::core::panicking::assert_failed(
                    kind,
                    &*left_val,
                    &*right_val,
                    ::core::option::Option::None,
                );
            }
        }
    };
    test_fn("12345", "12345", String::from("12345"), S(false)).unwrap_err();
    match (
        &test_fn_call_style("12345", "12345", String::from("12345"), S(true)).unwrap(),
        &"123451234512345!2345",
    ) {
        (left_val, right_val) => {
            if !(*left_val == *right_val) {
                let kind = ::core::panicking::AssertKind::Eq;
                ::core::panicking::assert_failed(
                    kind,
                    &*left_val,
                    &*right_val,
                    ::core::option::Option::None,
                );
            }
        }
    };
    test_fn_call_style("12345", "12345", String::from("12345"), S(false)).unwrap_err();
    match (
        &test_fn_where("12345", "12345", String::from("12345"), S(true)).unwrap(),
        &"123451234512345!2345",
    ) {
        (left_val, right_val) => {
            if !(*left_val == *right_val) {
                let kind = ::core::panicking::AssertKind::Eq;
                ::core::panicking::assert_failed(
                    kind,
                    &*left_val,
                    &*right_val,
                    ::core::option::Option::None,
                );
            }
        }
    };
    test_fn_where("12345", "12345", String::from("12345"), S(false)).unwrap_err();
}
extern crate test;
#[cfg(test)]
#[rustc_test_marker = "test_struct_method"]
pub const test_struct_method: test::TestDescAndFn = test::TestDescAndFn {
    desc: test::TestDesc {
        name: test::StaticTestName("test_struct_method"),
        ignore: false,
        ignore_message: ::core::option::Option::None,
        source_file: "gix-macros/tests/test.rs",
        start_line: 199usize,
        start_col: 4usize,
        end_line: 199usize,
        end_col: 22usize,
        compile_fail: false,
        no_run: false,
        should_panic: test::ShouldPanic::No,
        test_type: test::TestType::IntegrationTest,
    },
    testfn: test::StaticTestFn(|| test::assert_test_result(test_struct_method())),
};
fn test_struct_method() {
    match (
        &TestStruct
            .test_method("12345", "12345", String::from("12345"), S(true))
            .unwrap(),
        &"123451234512345!2345",
    ) {
        (left_val, right_val) => {
            if !(*left_val == *right_val) {
                let kind = ::core::panicking::AssertKind::Eq;
                ::core::panicking::assert_failed(
                    kind,
                    &*left_val,
                    &*right_val,
                    ::core::option::Option::None,
                );
            }
        }
    };
    TestStruct
        .test_method("12345", "12345", String::from("12345"), S(false))
        .unwrap_err();
    match (
        &TestStruct
            .test_method2("12345", "12345", String::from("12345"), S(true))
            .unwrap(),
        &"123451234512345!2345",
    ) {
        (left_val, right_val) => {
            if !(*left_val == *right_val) {
                let kind = ::core::panicking::AssertKind::Eq;
                ::core::panicking::assert_failed(
                    kind,
                    &*left_val,
                    &*right_val,
                    ::core::option::Option::None,
                );
            }
        }
    };
    TestStruct
        .test_method2("12345", "12345", String::from("12345"), S(false))
        .unwrap_err();
}
extern crate test;
#[cfg(test)]
#[rustc_test_marker = "test_struct_fn"]
pub const test_struct_fn: test::TestDescAndFn = test::TestDescAndFn {
    desc: test::TestDesc {
        name: test::StaticTestName("test_struct_fn"),
        ignore: false,
        ignore_message: ::core::option::Option::None,
        source_file: "gix-macros/tests/test.rs",
        start_line: 226usize,
        start_col: 4usize,
        end_line: 226usize,
        end_col: 18usize,
        compile_fail: false,
        no_run: false,
        should_panic: test::ShouldPanic::No,
        test_type: test::TestType::IntegrationTest,
    },
    testfn: test::StaticTestFn(|| test::assert_test_result(test_struct_fn())),
};
fn test_struct_fn() {
    match (
        &TestStruct::test_fn("12345", "12345", String::from("12345"), S(true), ())
            .unwrap(),
        &"123451234512345!2345",
    ) {
        (left_val, right_val) => {
            if !(*left_val == *right_val) {
                let kind = ::core::panicking::AssertKind::Eq;
                ::core::panicking::assert_failed(
                    kind,
                    &*left_val,
                    &*right_val,
                    ::core::option::Option::None,
                );
            }
        }
    };
    TestStruct::test_fn("12345", "12345", String::from("12345"), S(false), ())
        .unwrap_err();
    match (
        &TestStruct::test_fn2(
                TestStruct,
                "12345",
                "12345",
                String::from("12345"),
                S(true),
                (),
                (),
            )
            .unwrap(),
        &"123451234512345!2345",
    ) {
        (left_val, right_val) => {
            if !(*left_val == *right_val) {
                let kind = ::core::panicking::AssertKind::Eq;
                ::core::panicking::assert_failed(
                    kind,
                    &*left_val,
                    &*right_val,
                    ::core::option::Option::None,
                );
            }
        }
    };
    TestStruct::test_fn2(
            TestStruct,
            "12345",
            "12345",
            String::from("12345"),
            S(false),
            (),
            (),
        )
        .unwrap_err();
    match (
        &TestStruct::test_fn3(
                Pin::new(&mut TestStruct),
                "12345",
                "12345",
                String::from("12345"),
                S(true),
                (),
                (),
            )
            .unwrap(),
        &"123451234512345!2345",
    ) {
        (left_val, right_val) => {
            if !(*left_val == *right_val) {
                let kind = ::core::panicking::AssertKind::Eq;
                ::core::panicking::assert_failed(
                    kind,
                    &*left_val,
                    &*right_val,
                    ::core::option::Option::None,
                );
            }
        }
    };
    TestStruct::test_fn3(
            Pin::new(&mut TestStruct),
            "12345",
            "12345",
            String::from("12345"),
            S(false),
            (),
            (),
        )
        .unwrap_err();
}
#[rustc_main]
#[no_coverage]
pub fn main() -> () {
    extern crate test;
    test::test_main_static(&[&test_basic_fn, &test_struct_fn, &test_struct_method])
}


```

</details>


Byron (commented):
Great to see this :)! I think this crate will be very useful to produce landing pads *and* be a foray into proc-macro magic for `gitoxide`. With it it could be possible to perform static validation of values as well, for instance.

With that in mind, would you mind planning other kinds of macros as well and call it `gix-macros` instead? *(I you have a better name, please go for it though :))*

As for how to proceed with this PR, my preference is to wait until the crate can be used in a `gix` function or method somewhere so that we can see its real-world benefits.

gix-momo/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "gix-momo"
+version = "0.0.0"
+edition = "2021"
+description = "A gix proc_macro_attribute to outline conversions from generic functions"
+authors = ["Sebastian Thiel <sebastian.thiel@icloud.com>"]
```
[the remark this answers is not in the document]
NobodyXu: Thanks, I've added myself here and renamed this to `gix-macros`.

NobodyXu:
I will try to use it `gix`, but I think this PR is already ready for review and needs reviews since I'm not so familiar with proc-macros.

Byron:
I think I should get to this within the next couple of days. I myself have neary-zero experience with it either 😅.

Byron:
If you rebase now the `deny` issue should go away.

The [memory corruption issue](https://github.com/Byron/gitoxide/actions/runs/5939596007/job/16106304994?pr=992#step:6:10404) happens a lot these days and I don't know where that's coming from. Since I haven't been fiddling with `unsafe` lately, maybe it's due to a `zlib` update. As far as I can tell, this is only an issue with the `max` build, and never with the `pure` build which also [performs these clones](https://github.com/Byron/gitoxide/actions/runs/5939596007/job/16106304994?pr=992#step:6:8743). I may hope that it disappears as sudden as it appeared.

Byron:
I was wondering if this is the best upfront-investment of time given that monomorphism isn't a problem fi it's instantiated only once anyway. So I ran `cargo bloat -n 200` and noticed quite a few duplications in plumbing crates. And for plumbing crates, I wouldn't want to have `gix-macros` as dependency anyway if it can be avoided by replacing the generics with `dyn` by hand. Some of the functions I saw would make that possible.

In other words, I'd wish for this endeavour to be driven by data so the highest-value lowest effort changes can be made first. Maybe you have other means of determining where a problem actually is though.

With that said, please do feel free to continue on this track if it's fun and rewarding, as another approach would be to try `momo` things and compare the sizes of the build-artifacts and compile times.

NobodyXu:
> I was wondering if this is the best upfront-investment of time given that monomorphism isn't a problem fi it's instantiated only once anyway.

Yes, it will.

The cargo-binstall CI timings (generated by passing --timings to cargo) shows that on fresh build, simple-git takes ~45s on x86_64 MacOS, a lot longer than it should given that gix itself already takes ~45s.

The culprit here is the monomorphisation and it makes simple crates like that take incredibly long to compile.

Since simple-git is in workspace, it's not cached so every CI has to take the time to build it.

For local dev, that's more time waiting every time their crates using gix recompiles, even if it's trivial.

> And for plumbing crates, I wouldn't want to have `gix-macros` as dependency anyway if it can be avoided by replacing the generics with `dyn` by hand. Some of the functions I saw would make that possible.

Well, Into and TryInto for example, cannot be passed as dyn unless you Box them, but that will do more harm than good.

AsRef and AsMut can be passed as dyn but there's not much point doing that given that their implementation is most likely trivially inlinable.

That's where gix_macros::momo comes handy, avoid the duplication by slapping a proc-macro over it, while retaining the convenience and readability of the original implementation.

Admittedly though momo still has some problems with TryInto that I will fix.

> In other words, I'd wish for this endeavour to be driven by data so the highest-value lowest effort changes can be made first. Maybe you have other means of determining where a problem actually is though.

Adding features is definitely a low-hanging food compared to proc-macro, but it would require some extensive understanding of how gix internally uses dependencies.

Also I've already started this, so I'd like to finish this before doing other work.

> With that said, please do feel free to continue on this track if it's fun and rewarding, as another approach would be to try `momo` things and compare the sizes of the build-artifacts and compile times.

I will slap it on a few functions in gix tonight and slowly add them to more functions.

Will post before and after compile time on this

Byron:
> > And for plumbing crates, I wouldn't want to have `gix-macros` as dependency anyway if it can be avoided by replacing the generics with `dyn` by hand. Some of the functions I saw would make that possible.
> 
> Well, Into and TryInto for example, cannot be passed as dyn unless you Box them, but that will do more harm than good.
> 
> AsRef and AsMut can be passed as dyn but there's not much point doing that given that their implementation is most likely trivially inlinable.

When plumbing crates use `impl AsRef` or `impl TryInto` I usually do this just for my own convenience when writing tests, but I see how this is the wrong tradeoff. Also, I was not at all mindful about it. So to rephrase, these uses of generics in plumbing crates can be reviewed and probably replaced by the discrete type that is desired. Maybe that will have a positive effect as well.

There are rare cases where a trait is passed in, sometimes to `Progress`, sometimes to custom traits like in `gix-pack`, and only there it might be worth keeping them for maximum performance. At some point it's a trade-off and I would want to avoid putting `dyn` into all the places just so it's consistent.

Even though I don't like the idea of `gix-macros` in plumbing crates, I could imagine it not costing much as all plumbing crates already use `thiserror`, which pulls in the proc-macro dependency chain.

> 
> That's where gix_macros::momo comes handy, avoid the duplication by slapping a proc-macro over it, while retaining the convenience and readability of the original implementation.
> 
> Admittedly though momo still has some problems with TryInto that I will fix.
> 
> > In other words, I'd wish for this endeavour to be driven by data so the highest-value lowest effort changes can be made first. Maybe you have other means of determining where a problem actually is though.
> 
> Adding features is definitely a low-hanging food compared to proc-macro, but it would require some extensive understanding of how gix internally uses dependencies.
> 
> Also I've already started this, so I'd like to finish this before doing other work.

That makes sense, and maybe setting feature toggles is more my line of work given that I have created this component-graph to facilitate it. 

> 
> > With that said, please do feel free to continue on this track if it's fun and rewarding, as another approach would be to try `momo` things and compare the sizes of the build-artifacts and compile times.
> 
> I will slap it on a few functions in gix tonight and slowly add them to more functions.
> 
> Will post before and after compile time on this

I am looking forward to seeing those numbers - they will definitely help to make this more concrete. Right now to me all of it is theoretical.

Great work by the way, thanks for the initiative and engagement 🙏.

NobodyXu:
@Byron While there are still a few TODOs, I believe this PR is ready for review.

NobodyXu:
I benchmarked the compilation performance by running `cargo b --release --timings` in the root of `gitoxide`.

### Before

```
    Finished release [optimized] target(s) in 1m 12s
```

image

image

### After

```
    Finished release [optimized] target(s) in 1m 12s
```

image

image

### Findings

Using `momo` does increase compilation time of `gix` slightly but reduce time that uses it.

So far I have only `momo`ed a few functions, and due to the limit of `momo`, I wasn't able to put it over all generics that is eligable for it.

NobodyXu:
## Benchmark on `cargo-binstall`

Build command:

```
cd /path/to/cargo-binstall-repo-clone/crates/bin
cargo b --release --timings
```

### Before

```
    Finished release [optimized] target(s) in 2m 24s
```

image

### After

```
    Finished release [optimized] target(s) in 2m 21s
```

image

### Findings

Have similar findings here, however the change is still way too small and too nosy.
IMHO it's because the generics being `momo`ed are not enough and I haven't annotated function with `TryInto` yet due to `momo` limitation.

Byron:
Thanks for the data. I agree that the data probably isn't statistically significant at this point. I feel that if annotating a function with `momo`, ideally one will know if it gets duplicated due to monomorphization or not. Within `gitoxide`, there should be no such duplication at all in order to improve compile times, and `momo` can be used tactically to achieve that. After all, I don't expect the *blue* portion of the time to shrink, but the *purple* one. Strangely, the *blue* (i.e. rustc) part is huge for `cargo-binstall`, which is surprising. Can it be that codegen isn't displayed separately when compiling binaries?

If you rebase onto `main` the CI issues will go away. And while at that, could you do me a favor and split the change into one commit per crate?

gix/src/open/repository.rs
```diff
@@ -58,6 +59,8 @@ impl ThreadSafeRepository {
     ///
     /// Note that opening a repository for implementing custom hooks is also handle specifically in
     /// [`open_with_environment_overrides()`][Self::open_with_environment_overrides()].
+    #[momo]
+    #[allow(unused_mut)]
```
Byron, line 63: `#[allow(unused_mut)]` Is probably a side-effect of `momo` because it just passes `options` on without ever needing it to be `mut`. Since it would see the `mut` I suppose it should be the one automatically adding the lint configuration.

gix/src/pathspec.rs
```diff
@@ -92,6 +93,8 @@ impl<'repo> Pathspec<'repo> {
         alias = "matches_path",
         alias = "git2"
     )]
+    #[momo]
+    #[allow(clippy::needless_lifetimes)]
```
Byron, line 97: Probably this is a similar effect as observed with the `unused_mut` lint.

gix/src/repository/config/transport.rs
```diff
@@ -270,7 +274,7 @@ impl crate::Repository {
                         .proxy
                         .as_deref()
                         .filter(|url| !url.is_empty())
-                        .map(|url| gix_url::parse(url.into()))
+                        .map(|url| gix_url::parse(<&BStr>::from(url)))
```
Byron, line 277: It's interesting that this needed to change as it seems unrelated to the signature of the method affected by `momo`.

gix/src/repository/object.rs
```diff
@@ -173,11 +175,11 @@ impl crate::Repository {
         constraint: PreviousValue,
     ) -> Result<Reference<'_>, tag::Error> {
         let tag = gix_object::Tag {
-            target: target.as_ref().into(),
+            target: target.into(),
```
Byron, line 178: A very advanced version of `momo` would probably be able to remove `as_ref()` calls by itself one day. For now I find it acceptable to do this by hand though after `momo`ing a method.

gix/src/repository/remote.rs
```diff
@@ -75,45 +77,46 @@ impl crate::Repository {
...
+    }
+
+    #[momo]
+    #[allow(clippy::needless_lifetimes)]
     fn try_find_remote_inner<'a>(
         &self,
         name_or_url: impl Into<&'a BStr>,
         rewrite_urls: bool,
     ) -> Option<Result<Remote<'_>, find::Error>> {
-        fn config_spec<T: config::tree::keys::Validate>(
```
Byron, line 83: It's a bit unfortunate that inner functions aren't supported by `momo` presently so they have to be moved out.

In this particular case, I wonder if private methods should be `momo`'d at all - if they are only called from `momo`'d functions and methods, only one instance would exist anyway.

Lastly, if it is an internal method, I would probably just use concrete types. `momo` is only useful for the `public` API where we want the convenience for the caller.

Byron (commented):
Thanks a lot for all your hard work on this!

I took a first look at how `momo` is used and left a few comments. Some, as it seems, refer to a checkbox on the PR that indicates it's planned to improve it at some point, so please feel free to ignore these.

I think a major idea I brought up is to use `momo` only on the public API, and rather de-generalize internal functions by hand. Generics are either used for convenience, or for performance. And when convenience is involved, there is no need for that in the non-public-API to save compile time.

Generally, I really like where this is going and think `momo` will be a powerful took in the fight for smaller compile times. I also think that it *could* be that `cargo-binstall` won't profit much if its calls to `gix` didn't actually duplicate methods. Independently of that, I think a huge win could come from deduplicating `gix` internally - when running `cargo bloat` on the `gix` binary, there were a couple of duplicated methods, probably because the `gitoxide-core` code makes calls with different types. Maybe such calls are also done within `gix`, and removing such duplication would benefit everyone equally.

NobodyXu:
> Strangely, the *blue* (i.e. rustc) part is huge for `cargo-binstall`, which is surprising. Can it be that codegen isn't displayed separately when compiling binaries?

AFAIK, compiling, codegen and linking is all done at once for binary, that's why it's so long.

gix and ein is also quite long, that's why I always split stuff into library crates and the src/lib.rs part of the binary crate instead of src/main.ra

I also think gitoxide might benefit from supporting multi-call to avoid creating two binaries, which would reduce compilation time and the size of the binaries generated.


> If you rebase onto `main` the CI issues will go away. And while at that, could you do me a favor and split the change into one commit per crate?

Sure, I will do that.

gix/src/repository/config/transport.rs
NobodyXu: Yeah it's really a shame.
momo currently works by "search and replace", which is why this is a problem.

A better approach would be to build a syntax tree and understand which variable it's referring to, which is quite complex and not sure whether I can do it within one PR.

Even better would be to add this to rustc.

gix/src/repository/remote.rs
NobodyXu: Could have a try to see if I can fix this, though I feel that I need to change the way momo works.

NobodyXu:
> Generally, I really like where this is going and think `momo` will be a powerful took in the fight for smaller compile times. I also think that it *could* be that `cargo-binstall` won't profit much if its calls to `gix` didn't actually duplicate methods.

Yeah, it seems that cargo-binstall would benefit from having more features in gix instead.

> Independently of that, I think a huge win could come from deduplicating `gix` internally - when running `cargo bloat` on the `gix` binary, there were a couple of duplicated methods, probably because the `gitoxide-core` code makes calls with different types. Maybe such calls are also done within `gix`, and removing such duplication would benefit everyone equally.

Yes, that's why I come up with the idea of momo.

NobodyXu:
> If you rebase onto main the CI issues will go away. And while at that, could you do me a favor and split the change into one commit per crate?

I've rebased it against main and also split the change into commits.

gix/src/repository/remote.rs
NobodyXu: I've removed `momo` on `try_find_remote_inner` and instead put it over `try_find_remote` and `try_find_remote_without_url_rewrite`

gix/src/repository/config/transport.rs
Byron: Despite valuing the effort you would be willing to make, I think it can be better spent - making some adjustments by hand seems like a small price to pay to not have to write a lot of boilerplate.

Byron:
> gix and ein is also quite long, that's why I always split stuff into library crates and the src/lib.rs part of the binary crate instead of src/main.ra

`gitoxide-core` is the library that power both binaries, but they don't have their own lib but instead share some code via modules. I think that can and should be trivially shared via the `gitoxide` lib, which is currently unused. Thanks for the hint.

For everything else, my stance here is to wait for all the checkboxes to be ticked and ideally to see gix can compile faster. I understand that `prodash` might have to get its PR merged as well to be used here. What I am missing is an idea how to get the 'duplication information' for `gix` itself across all dependencies. Maybe `cargo bloat` so one can now what to work on. It looks like `cargo bloat` has a `--message-format` option which should make it easy to conceive a tool that helps pinpointing such duplicates sorted by priority.

```json
    {
      "crate": "gix_refspec",
      "name": "gix_revision::spec::parse::function::revision",
      "size": 5804
    },
    {
      "crate": "gitoxide_core",
      "name": "gix_revision::spec::parse::function::revision",
      "size": 5804
    },
    {
      "crate": "gix_revision",
      "name": "gix_revision::spec::parse::function::revision",
      "size": 5804
    },
    {
      "crate": "gix",
      "name": "gix::clone::fetch::<impl gix::clone::PrepareFetch>::fetch_only",
      "size": 5756
    },
    {
      "crate": "gitoxide_core",
      "name": "gitoxide_core::repository::fetch::function::print_updates",
      "size": 5748
    },
    {
      "crate": "gitoxide_core",
      "name": "gitoxide_core::repository::fetch::function::print_updates",
      "size": 5740
    },
```

By looking around a little bit I noticed these - it seems like it's pretty clear what the duplicates are even though some probably get different names of `impl` is used. `gix_revision::spec::parse::function::revision` could benefit from using `dyn` as I don't think monomorphism is needed here for performance. And from there, it's a long and grindy process, but I think each little step will have measurable reductions both in compile time and in size.

As `momo` also has a cost, in the worst case we will 'just' break even, so maybe it's more worth-while to go about this in a data-driven fashion?

It's nothing more than a question though, and you are welcome to proceed the way it feels right for you.

gix/src/repository/remote.rs
NobodyXu: I've removed use of `momo` and instead manually de-`momo` them since `try_find_remote_inner` is it's too trivial to use `momo`.

NobodyXu:
@Byron I'm now thinking of reducing the scope of `momo` a bit, removing support of `TryInto` for now and then added it back later in a separate PR.

Part of the reason is that `impl  TryInto` isn't as common as `AsRef`/`AsMut`/`Into`, but also because its error handling logic is much harder to parse and replace in the inner function.

By removing `TryInto`, we can simplify the `momo` macro by expanding the function the following way:

original:

```rust
pub fn outer(a: impl Into<String>, b: impl AsRef<str>, mut c: impl AsMut<str>) -> String {
    let mut s = a.into();
    s += b.as_ref();
    s += c.as_mut();
    s
}
```

```rust
pub fn outer(a: impl Into<String>, b: impl AsRef<str>, mut c: impl AsMut<str>) -> String {
    fn _outer_inner_generated_by_momo(a: impl Into<String>, b: impl AsRef<str>, mut c: impl AsMut<str>) -> String {
        let mut s = a.into();
        s += b.as_ref();
        s += c.as_mut();
        s
    }
    inner(a.into(), b.as_ref(), c.as_mut())
}
```

I took the inspiration from `Repositor::try_find_remote_inner` which is itself a generic, but it still avoid the monomorphization since it is monomorphized exactly once within the crate and cannot be accessed outside the crate.

This would dramatically simplify the `momo`'s proc-macro logic, making it much easier to maintain.

NobodyXu:
@Byron I've now updated `gix_macros::momo` according to the idea I described.

Not only does it dramatically simplifies `gix_macros::momo` implementation, but also removed need of `clippy::needless_lifetime` and also remove the need for manually removing `.as_ref()`/`.into()` in certain, making it truly painless to slap `gix_macros::momo` on any generics.

gix/src/repository/object.rs
NobodyXu: Resolved by the new internal https://github.com/Byron/gitoxide/pull/992#issuecomment-1697285009

gix/src/pathspec.rs
NobodyXu: Resolved by the new internals https://github.com/Byron/gitoxide/pull/992#issuecomment-1697285009

NobodyXu:
@Byron I've slapped `momo` over everywhere I can apply, please review again if you have time.

I will open a separate PR to de-`momo` functions using `TryInto` since they are way more complicated and better do as hand as they might use `.map_err`.

I also found several functions using `impl FnMut(...)` and wonder if it makes sense to use a `&mut dyn FnMut(...)` instead.

Byron:
> @Byron I've slapped momo over everywhere I can apply, please review again if you have time.

Finally I am getting around taking a look. Sorry for the delay.

> I also found several functions using impl FnMut(...) and wonder if it makes sense to use a &mut dyn FnMut(...) instead.

`gix` is (or should be) using callbacks sparingly, but when they are used I'd love them to be as easy to use as possible. In these few cases, I wouldn't mind de-momoing by hand. I think the same does apply for `TryInto` as well, even though it's probably more prominent and thus *might* be worth it. In general, I'd love for `gix-macros` to stay as small, simple and fast as possible.

I am taking a close look now and hope to see some improvements to either compile time or binary size as well. Let's see.

Byron:
An unscientific benchmark showed that the filesizes are a little lower when `momo` is used, whereas compile times are about the same.

###### Without Momo

-rwxr-xr-x  1 byron  staff   9894161 Sep  2 19:39 target/release/ein
-rwxr-xr-x  1 byron  staff  18991777 Sep  2 19:39 target/release/gix

###### With Momo

-rwxr-xr-x  1 byron  staff   9765057 Sep  2 19:37 target/release/ein
-rwxr-xr-x  1 byron  staff  18898161 Sep  2 19:37 target/release/gix

Of course, the mileage in other binaries may vary greatly.

One big piece of code to de-momo would certainly be those that use `Progress`, and getting there will be some work. As will be cutting down inter-crate duplication due to monomorphization.

gix-macros/src/lib.rs
```diff
@@ -0,0 +1,26 @@
...
+/// function or method for implementation.
+///
+/// The parameters using the following traits can be de-monomorphized:
+///
+/// * `Into`
+/// * `AsRef`
+/// * `AsMut`
+#[proc_macro_attribute]
+pub fn momo(_attrs: TokenStream, input: TokenStream) -> TokenStream {
+    //TODO: alternatively parse ImplItem::Method
```
Byron, line 22: Can this TODO be removed? Otherwise I have a feeling it could stay there for long, and I don't know what it means so couldn't easily follow up either.

Byron (commented):
Thanks a lot for this awesome crate!

I did a minor refactor and added docs as well as I could. While doing that, I imagined how I would use this macro and thought that the last thing I want is to accidentally increase the compile time, for example, by putting it on a method, function or place where it has no effect.

Would it be hard to make it emit a compile_error if it didn't do anything? I thought `ty_conversions` might be the key, but I wasn't sure if that misses something so didn't dare to try implementing it. Further, if it is easy enough, could there also be a test for that? I think I have seen a crate that can test compile-time characteristics, but forgot its name.

Once this is clarified, I think it's ready for merge on the merit of having demonstrated a light reduction in binary sizes, a trend that will certainly continue with additional work. I think the next steps could be to make `Progress` dyn-safe and drive forward the `gix-*` crates deduplication or (even explicit) `dyn`-ification.

gix-macros/src/lib.rs
NobodyXu: Yes, I've removed that now.

Byron: Thanks! Just to double-check: did you see my note about letting it fail if it couldn't apply any conversion?

NobodyXu: Thanks, I've read that comment but somehow missed that part.

NobodyXu:
> Would it be hard to make it emit a compile_error if it didn't do anything?

That should be possible.

> I thought ty_conversions might be the key, but I wasn't sure if that misses something so didn't dare to try implementing it. 

Yes, if you add an is_empty() check for it and then write something similar to gix-macros/src/momo.rs#88, then it should work as intended.

> Further, if it is easy enough, could there also be a test for that? I think I have seen a crate that can test compile-time characteristics, but forgot its name.

I think it can be done using a doc test, by marking it compile_fail:

    /// ```compile_fail
    /// let x = 5;
    /// x += 2; // shouldn't  compile!
    /// ```

Byron:
Thanks a lot for your help! I gave it a shot and found a good way to test error conditions of proc macros. While trying to implement [the `is_empty()` check](https://github.com/Byron/gitoxide/blob/c8e732430f3740348ccedd0dc1a9a28b06a0adee/gix-macros/src/momo.rs#L14-L21)  as suggested I noticed that there were a lot of failures where I wouldn't expect them. Either there is more to `ty_conversions`, or something else is going on.

Could you help?
Thank you 🙏

NobodyXu:
ty_conversions is where the conversions Into, AsRef, AsMut is detected and put into, you are using it the right way.

I can have a look 12h later.

Byron (approved):
Awesome, thanks again!

I was wondering if `momo`  should be even stricter and bail if it can't remove *all* generics of a function, but then contemplated that *less* generics is already a win as it reduces to set of possible instantiations.

On this branch, I am getting 24.0s for a debug build (`gix` + `ein`) and 48.7s for a release build.

```
❯ ls -l ./target/release/gix ./target/release/ein ./target/debug/gix ./target/debug/ein
-rwxr-xr-x  1 byron  staff  36282417 Sep  7 15:13 ./target/debug/ein
-rwxr-xr-x  1 byron  staff  68070609 Sep  7 15:13 ./target/debug/gix
-rwxr-xr-x  1 byron  staff   9765057 Sep  7 15:15 ./target/release/ein
-rwxr-xr-x  1 byron  staff  18898161 Sep  7 15:15 ./target/release/gix
```

On `main`, I am getting 23.8s for a debug build (`gix` + `ein`) and 44.9s for a release build.

```
❯ ls -l ./target/release/gix ./target/release/ein ./target/debug/gix ./target/debug/ein
-rwxr-xr-x  1 byron  staff  36358865 Sep  7 15:18 ./target/debug/ein
-rwxr-xr-x  1 byron  staff  66358817 Sep  7 15:18 ./target/debug/gix
-rwxr-xr-x  1 byron  staff   9734657 Sep  7 15:18 ./target/release/ein
-rwxr-xr-x  1 byron  staff  18750001 Sep  7 15:18 ./target/release/gix
```

I realize now that this branch isn't on top of main yet, and that apparently release builds already got faster.

Let me merge and see if it has any effect.

Now on `main`, I am getting 22.8s for a debug build (`gix` + `ein`) and 45.4s for a release build.

```
❯ ls -l ./target/release/gix ./target/release/ein ./target/debug/gix ./target/debug/ein
-rwxr-xr-x  1 byron  staff  36345649 Sep  7 15:22 ./target/debug/ein
-rwxr-xr-x  1 byron  staff  66302433 Sep  7 15:22 ./target/debug/gix
-rwxr-xr-x  1 byron  staff   9757153 Sep  7 15:23 ./target/release/ein
-rwxr-xr-x  1 byron  staff  18781889 Sep  7 15:23 ./target/release/gix
```

Debug-sizes seem to have gotten better, but release sizes got a little worse. Probably these aren't necessarily deterministic, and since it's all marginal I think simply isn't anything to see here. This makes me doubt the usefulness of `momo` at least for `gix` and `ein`, but they are possibly special as they only call each function once anyway, or use the same input types consistently.

I will be looking forward of anything improved for `cargo-binstall` already.

NobodyXu:
@Byron It seems that this PR indeed improves gitoxide CI a lot

Before [Merge branch `dyn`ification](https://github.com/Byron/gitoxide/actions/runs/6101181892) takes 44m, after [Merge branch 'feat/gix-momo'](https://github.com/Byron/gitoxide/actions/runs/6110383583) takes 33m

NobodyXu:
It mostly comes from CI test.yml, compiling tests on CI is speedup by 2m, documenting speedup by 10s, and not sure what the other 8m speedup comes from.

Byron:
Sometimes I have the same and CI fluctuates so that a 32min job takes 42min. I was already eyeing enterprise grade CI or adding an own runner even. In a few years once I get the next machine I think I will make it beefy enough be able to continuously provide some CI runner machines for MacOS, Linux and Windows, with each of them as strong as my current dev machine. Dreams… 😁.

NobodyXu:
> Sometimes I have the same and CI fluctuates so that a 32min job takes 42min.

Oops, I thought that it has improved the CI compilation time 😅 

> I was already eyeing enterprise grade CI or adding an own runner even.

I think it would be great if ee can form some organisation and provide CI for any Rust crate who needs them and unify our funds for larger instances.

cargo-binstall suffers the same thing though its CI is faster (usually aroud 15-20m), other crates would likely also want larger instances.

What makes it reasonable to have a dedicated runner for rust crates is that rust compilation is not embarrassingly parallel, this is especially true on release where binstall sets codegen-units to 1 and uses fat-lto.

It would improve with parallel frontend parallelisation, but cargo's compilation model will impose a limit on that, especially with build.rs (with compiling external libs in it blocking all dependents) and proc-macro.

> In a few years once I get the next machine I think I will make it beefy enough be able to continuously provide some CI runner machines for MacOS, Linux and Windows, with each of them as strong as my current dev machine. Dreams… 😁.

I am already considering sponsoring you, for new machine I think you could build one Intel i7-13900 or something better, AMD's good at single-core which doesn't work well with rust compilation and testing.

For MacOS, definitely go with Mac Mini with at most 12 core and 32G or Mac Studio with at most 24 cores and 192G.

I had to admit, for apple products, their GPU cores are a waste to me, unless we gave it to someone who can use them...

Byron:
> What makes it reasonable to have a dedicated runner for rust crates is that rust compilation is not embarrassingly parallel, this is especially true on release where binstall sets codegen-units to 1 and uses fat-lto.

This is the worst for performance, indeed! I had it too, but decided it's not worth the struggle, even locally. Of course, reproducible builds don't come from using more than one codegen units, but it's I will postpone this as long as possible.

> I am already considering sponsoring you, for new machine I think you could build one Intel i7-13900 or something better, AMD's good at single-core which doesn't work well with rust compilation and testing.

Actually, I love the idea of trying to get custom runners for Rust projects to use. Somehow I feel the Rust Foundation could provide these, they have all the data-center providers and already get some machines from them. Even a single server could already make a huge difference, even when shared across many projects.

I think it's worth pointing out that long CI times aren't a problem most of the time - they are *just* a lower bound on how faster one can make releases or merge PRs. Now the `gitoxide` CI is at 45min worst case, typically around 30m, and it has been for a while. Maybe that's just how it is and everything else is luxury.

But yeah, I don't really want a machine to stand at home and consume power all the time, while being mostly idle. So having a Mac Studio with maxed-out CPU cores seems like a way to have max performance while working, and enough spare to have many VMs to host ones own CI needs as well. But no matter how, it remains a luxury and I treat it as such :D.

>  had to admit, for apple products, their GPU cores are a waste to me, unless we gave it to someone who can use them...

My hope is that in a couple of years, when I can seriously think about updating hardware, there will be enough models available locally that run on the GPU. Then it might even be part of a normal workflow to run these locally, even though beefy machines are still required. Right now though, at least for my line of work, big GPUs are indeed a huge waste.

NobodyXu:
> Actually, I love the idea of trying to get custom runners for Rust projects to use. Somehow I feel the Rust Foundation could provide these, they have all the data-center providers and already get some machines from them. Even a single server could already make a huge difference, even when shared across many projects.

That's a good idea but not sure if feasible, I read https://kobzol.github.io/rust/rustc/2023/07/30/optimizing-rust-ci-2023.html and it seems that rust-lang/rust CI takes a lot time and their servers might also be overloaded.

> Maybe that's just how it is and everything else is luxury.

Yes, we also accept this in cargo-binstall.

> But yeah, I don't really want a machine to stand at home and consume power all the time, while being mostly idle.

Yeah, electricity bill price has increased dramatically, even in Australia where supposedly half of electricity is provided by renewable energy, it's more than 0.41 aud per kWh now in Sydney.

> So having a Mac Studio with maxed-out CPU cores seems like a way to have max performance while working, and enough spare to have many VMs to host ones own CI needs as well. But no matter how, it remains a luxury and I treat it as such :D.

> My hope is that in a couple of years, when I can seriously think about updating hardware, there will be enough models available locally that run on the GPU. Then it might even be part of a normal workflow to run these locally, even though beefy machines are still required. Right now though, at least for my line of work, big GPUs are indeed a huge waste.

Yes, for now I just want a GPu good enough for running desktop GUI.

PR #994, opened 2023-08-20
Checkout outcome now implements debug

bittrance:
As part of implementing a gitops-like feature, I found myself wanting to `dbg!()` the `Outcome` returned from `checkout()` during development. I'm guessing this is a relatively common use case, so I thought it worthwhile for it to `#[derive(Debug)]`.


Byron (approved):
That's definitely an oversight, thanks for the fix :)!

PR #1004, opened 2023-08-27
fixes and improvements

### LINKED ISSUE #1003: gix fetch return error 'Cannot find a remote for unborn branch' with bare repo
### Current behavior 😯

After some steps (see "Steps to reproduce"), `gix` returns the error:

```
$ gix fetch origin +refs/heads/*:refs/heads/*
Error: Cannot find a remote for unborn branch
```

with `gitoxide v0.29.0`

### Expected behavior 🤔

`git` success to run:

```
$ git fetch origin +refs/heads/*:refs/heads/*
from https://github.com/AOSC-Dev/oma
 * [new branch]          cratify-oma -> cratify-oma
$ echo $?
0
```

### Steps to reproduce 🕹

```
gix clone --bare https://github.com/AOSC-Dev/oma oma-bare
mkdir oma-not-bare
cd oma-not-bare
git init
git --git-dir ../oma-bare --work-tree . checkout -f e8f56b5
cd ../oma-bare
gix fetch origin +refs/heads/*:refs/heads/*
```

Byron:
### Tasks

* [x] fix #1003
* [x] `gix-index` can deal with null-hashes in index files
* [x] index writing can skip generating the hash, also wire it up in `gix` for benchmarks
* [x] see what happens if we would by default validate indices, just like git does to detect bit-rot


gix/src/repository/remote.rs
```diff
@@ -65,6 +66,38 @@ impl crate::Repository {
...
+    /// If `name_or_url` is `None`:
+    ///
+    /// * use the current `HEAD` branch to find a configured remote
+    /// * fall back to either a generally configured remote or the only configured remote.
+    ///
+    /// Fail if no remote could be found despite all of the above.
+    pub fn find_fetch_remote(&self, name_or_url: Option<&BStr>) -> Result<Remote<'_>, find::for_fetch::Error> {
+        Ok(match name_or_url {
+            Some(name) => {
+                if name.contains(&b'/') || name.contains(&b'.') {
```
niklaswimmer, line 85: I think checking for a `:` may be beneficial as well? Otherwise `server:repo` is used as remote name. Also, wouldn't it make more sense to first check whether we find a remote by name and only if not try with a url - or are we constrained by Gits implementation in that regard?

Byron: That's all valid points, thanks for chiming in. The truth is, I didn't check against `git` in terms of code, just in terms of behaviour, so this implementation probably isn't according to 'spec'. Now that it moved 'up' from `gitoxide-core` to `gix` it should be according to spec though.

I will take another look and see what needs to change to be what `git` does.

Byron: It looks like [this is how git decides how to interpret the `name_or_url`](https://github.com/git/git/blob/6807fcfedab84bc8cd0fbf721bc13c4e68cda9ae/remote.c#L673-L701). Some minor modifications can be made easily, but I also noticed that there are so called remotes-files and branches-files (!!) - yet another way to configure things apparently, and something I have never encountered or heard off. 

Ignoring that, I have just modified the detection routine to be in-line with what `git` does, and it happened to have been quite exactly what you suggested :). Thanks again!

niklaswimmer: > I also noticed that there are so called remotes-files and branches-files

I never had of those either, the amount of features Git has really is crazy. Anyways, glad I could help :)

Byron:
Here is the results of `--skip-hash` test runs with various settings.

```
❯ grv
origin  https://github.com/WebKit/WebKit (fetch)
origin  https://github.com/WebKit/WebKit (push)

WebKit ( main) via △
❯ gix rev parse @
91f7b27a2c241b8753006ec66762aacbd2c66089

❯ hyperfine --warmup 1 'gix free index from-list files.list' '/Users/byron/dev/github.com/Byron/gitoxide/target/release/gix free index from-list files.list --skip-hash' 'gix index from-tree @' '/Users/byron/dev/github.com/Byron/gitoxide/target/release/gix index from-tree @ --skip-hash' '/Users/byron/dev/github.com/git/git/git read-tree --index-output /dev/null @' '/Users/byron/dev/github.com/git/git/git -c index.skipHash=1 read-tree --index-output /dev/null @' -N -i
Benchmark 1: gix free index from-list files.list
  Time (mean ± σ):     139.7 ms ±   1.1 ms    [User: 118.9 ms, System: 18.7 ms]
  Range (min … max):   137.0 ms … 141.1 ms    21 runs

Benchmark 2: /Users/byron/dev/github.com/Byron/gitoxide/target/release/gix free index from-list files.list --skip-hash
  Time (mean ± σ):     100.5 ms ±   1.4 ms    [User: 80.9 ms, System: 18.4 ms]
  Range (min … max):    96.7 ms … 102.9 ms    29 runs

Benchmark 3: gix index from-tree @
  Time (mean ± σ):     278.1 ms ±   5.8 ms    [User: 242.2 ms, System: 33.6 ms]
  Range (min … max):   272.1 ms … 291.2 ms    10 runs

Benchmark 4: /Users/byron/dev/github.com/Byron/gitoxide/target/release/gix index from-tree @ --skip-hash
  Time (mean ± σ):     234.3 ms ±   3.6 ms    [User: 203.0 ms, System: 29.9 ms]
  Range (min … max):   229.7 ms … 243.1 ms    12 runs

Benchmark 5: /Users/byron/dev/github.com/git/git/git read-tree --index-output /dev/null @
  Time (mean ± σ):     482.8 ms ±   4.7 ms    [User: 435.4 ms, System: 42.3 ms]
  Range (min … max):   478.5 ms … 490.1 ms    10 runs

  Warning: Ignoring non-zero exit code.

Benchmark 6: /Users/byron/dev/github.com/git/git/git -c index.skipHash=1 read-tree --index-output /dev/null @
  Time (mean ± σ):     374.2 ms ±   5.4 ms    [User: 328.3 ms, System: 42.0 ms]
  Range (min … max):   368.1 ms … 386.1 ms    10 runs

  Warning: Ignoring non-zero exit code.

Summary
  /Users/byron/dev/github.com/Byron/gitoxide/target/release/gix free index from-list files.list --skip-hash ran
    1.39 ± 0.02 times faster than gix free index from-list files.list
    2.33 ± 0.05 times faster than /Users/byron/dev/github.com/Byron/gitoxide/target/release/gix index from-tree @ --skip-hash
    2.77 ± 0.07 times faster than gix index from-tree @
    3.72 ± 0.08 times faster than /Users/byron/dev/github.com/git/git/git -c index.skipHash=1 read-tree --index-output /dev/null @
    4.80 ± 0.08 times faster than /Users/byron/dev/github.com/git/git/git read-tree --index-output /dev/null @
```

This shows that as of now, `gix` is generally faster in writing the index file, and this even seems like an apples-to-apples comparison. It's notable that these tests also show how much time it takes to traverse a big tree, so `index-from-list` without hash can be as fast as 100ms (which generates a 53MB file).

I think once `gix` is in the business of reducing the time it takes to write the index, it will implement multi-threading and split out the tree-traversal into its own thread as well, while keeping the hash. Tree-traversal takes about 170ms, so I'd hope that with everything parallelized, 180ms would be the performance goal here, then just 80ms slower than the optimal case of reading paths from a file and skipping the hash entirely. If that were the case, it would nearly be twice as fast as git there.

Byron:
I also realize now that the massive performance improvements that `gix index entries` has over `git ls-files` is mainly due to not checking the hash.

```
❯ hyperfine --warmup 1 'gix index entries --no-attributes' '/Users/byron/dev/github.com/Byron/gitoxide/target/release/gix index entries --no-attributes' 'git ls-files' -N
Benchmark 1: gix index entries --no-attributes
  Time (mean ± σ):      31.1 ms ±   2.5 ms    [User: 24.3 ms, System: 32.0 ms]
  Range (min … max):    28.6 ms …  41.0 ms    91 runs

Benchmark 2: /Users/byron/dev/github.com/Byron/gitoxide/target/release/gix index entries --no-attributes
  Time (mean ± σ):      59.7 ms ±   0.7 ms    [User: 49.4 ms, System: 36.3 ms]
  Range (min … max):    58.2 ms …  61.2 ms    50 runs

Benchmark 3: git ls-files
  Time (mean ± σ):      72.3 ms ±   0.7 ms    [User: 70.0 ms, System: 27.1 ms]
  Range (min … max):    70.8 ms …  73.6 ms    42 runs

Summary
  gix index entries --no-attributes ran
    1.92 ± 0.16 times faster than /Users/byron/dev/github.com/Byron/gitoxide/target/release/gix index entries --no-attributes
    2.32 ± 0.19 times faster than git ls-files
```

With that benefit gone, it's just 20% faster. Maybe the implementation can be a little cleverer though, let's see.

Byron:
The take-away here is that hashing is very slow compared to reading the index file. The former happens at 1.5GB/s, the latter at 3.5GB/s, so even if done in parallel, the slowest denominator is the hashing.

```
❯ hyperfine --warmup 1 '/Users/byron/dev/github.com/Byron/gitoxide/target/release/gix index entries --no-attributes' 'git ls-files' -N

Benchmark 1: /Users/byron/dev/github.com/Byron/gitoxide/target/release/gix index entries --no-attributes
  Time (mean ± σ):      49.4 ms ±   0.7 ms    [User: 49.6 ms, System: 37.3 ms]
  Range (min … max):    47.8 ms …  51.0 ms    61 runs

Benchmark 2: git ls-files
  Time (mean ± σ):      69.8 ms ±   0.5 ms    [User: 68.2 ms, System: 26.5 ms]
  Range (min … max):    68.7 ms …  70.8 ms    42 runs

Summary
  /Users/byron/dev/github.com/Byron/gitoxide/target/release/gix index entries --no-attributes ran
    1.41 ± 0.02 times faster than git ls-files
```

With `index.skipHash` enabled, there is no need to check the hash (e.g. if it's a trailing null-hash), but in all other cases `gix-index` would have to validate the hash.

I think it's best to bend the rules a little and say that if `skipHash` is set as an option (or there is a null-hash), don't check the hash. That way it's possible to just turn that option on and safe a lot of time reading and writing the index.

Byron:
Merged

PR #1007, opened 2023-09-03
Add i686-pc-windows-gnu CI

mati865:
Sorry it took so long.
Addresses https://github.com/Byron/gitoxide/pull/946

Because of new versions of `sha1-asm` there are not code changes required.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,7 +110,7 @@ jobs:
   installation:
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot, it's appreciated no matter when it happens :).

It seems with this PR `gitoxide` will not accidentally break due to SHA1-asm anymore, while also providing a platform to test this kind of build by default. That's fantastic, as it roots out (one of ?) the last strangeness around builds that I remember.

PR #1009, opened 2023-09-05
Migrate gix-protocol to winnow

epage:
I don't know how I missed this package.  At least its port is more straightforward than the others.


Byron:
If it helps, I missed it too :D! Thanks for finalizing this awesome stream of work!

PR #1011, opened 2023-09-08
worktree reset

Byron:
Based on #1008 .

Implement a proper worktree reset, eventually leading to a high-level reset similar to how git supports it.

### Tasks

* [ ] worktree reset to match new index, with or without safety checks to prevent overwriting modified files or untracked files.
* [ ] figure out how this relates to the current `checkout()` method as technically that's a `reset --hard` with optional overwrite check. Could it be rolled into one, with pathspec support added?
* TBD

### Postponed

* [ ] insert entry into index
* [ ] traverse tree with pathspec matching and insert entry into index.

### Research

* How to integrate submodules
* How to deal with various modes like `merge` and `keep`? How to control `refresh`?


Skgland:
It doesn't look like closing this was intendet.

Byron:
No, it was not 😅. I published something while on this branch, instead of main, and now the tag is setting this history in stone and I had to bring it into `main`. Oh well…probably time to call it a night.

PR #1012, opened 2023-09-09
Optimize `gix`: de-momo `impl TryInto` by hand

NobodyXu:
Also:
 - Rm unnecessary lifetime annotation in `Repository::commit_as_inner`
 - Optimize `Repository::write_object`: Avoid dup momo
   By splitting function into two halfs: First half uses
   `object: impl gix_object::WriteTo` and the second half uses no generic
   to avoid dup codegen.
 - Apply `gix_macros::momo` to `Repository::write_blob`
 - Optimize `Repository::write_blob_stream`: Avoid dup codegen


Byron (changes requested):
Thanks a lot!

For this to be mergeable, can you please remove the changes to the binary files? These shouldn't be required and will only cause trouble 😅. Thanks.

NobodyXu:
I've reverted changes to the binaries, I have no idea why they would pop up like that

Byron (approved):
> I've reverted changes to the binaries, I have no idea why they would pop up like that

When running tests, probably (?) without git-lfs setup, the locally available caches will be placeholders. Then when running tests, these will be replaced with archives of the test-fixtures that have been generated with scripts. Now these files are modified and easy to commit by accident. At least… that's how it usually happens.

*In any case*, thanks a lot for your continued help. It's much appreciated and I think the biggest *coup* will be `therror`, a truly evolutionary step especially if it remains compatible with `thiserror` in terms of syntax. But I digress 😅

PR #1015, opened 2023-09-09
Optimize `clone::PrepareCheckout::main_worktree``

NobodyXu:
Use `DynNestedProgress` instead of `NestedProgress`, although this also require `<P as NestedProgress>::SubProgress: 'static` which is a breaking change.

### DIFF
--- a/gix/src/clone/checkout.rs
+++ b/gix/src/clone/checkout.rs
@@ -62,14 +62,14 @@ pub mod main_worktree {
         ///
         /// Note that this is a no-op if the remote was empty, leaving this repository empty as well. This can be validated by checking
         /// if the `head()` of the returned repository is not unborn.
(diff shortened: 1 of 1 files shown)


Byron (commented):
Thanks a lot! I remember having removed the generic version not too long again, but it's probably more convenient to go with generics here as landing pad, a good catch.

Byron:
Is there I reason that I shouldn't merge with this change: https://github.com/Byron/gitoxide/pull/1015/commits/ee9276f2a7789c20d88d40624ad648e44b604a27 ?

To me it's semantically the same, but a bit easier on the eyes. If there are no objections, I will merge. Thank you.

NobodyXu:
I don't think there's any difference @Byron

PR #1032, opened 2023-09-24
better transport argument sanitization

Byron:
See https://secure.phabricator.com/T12961 for more details.

### Tasks

* [x] fix the issue for `ssh`
* [x] fix issue for `file` 
* ~~fix issue for submodules~~ - let's not assume a command-line will be invoked that needs protection, with `gix` one day this won't be the case and until then the call-site must protect itself.
* [x] add journey tests to assure we see an error message

### DIFF
--- a/gix-transport/src/client/blocking_io/file.rs
+++ b/gix-transport/src/client/blocking_io/file.rs
@@ -211,6 +211,11 @@ impl client::Transport for SpawnProcessOnDemand {
         };
         cmd.stdin = Stdio::piped();
         cmd.stdout = Stdio::piped();
+        if self.path.first() == Some(&b'-') {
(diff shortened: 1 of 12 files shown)


niklaswimmer:
Seems like Git performs the check for:
- proxy connections with the `git` protocol (checks host and port separately)
- connections with the `ssh` protocol (checks host and path separately)
- the path of file urls and local paths
- before doing anything with submodule urls/paths

Byron:
Note that `gitoxide` takes a slightly different route as it will only validate arguments, like host and path, before trying to use them as command-line arguments. Otherwise, it allows these as valid hosts and valid paths to avoid unnecessarily restricting the range of input.

PR #1042, opened 2023-09-29
Improve output of the gix-url baseline test

niklaswimmer:
"Improve" may be subjective but I figured since I will be the one mainly working with it I can take the liberty to adjust it a bit for my needs :) 

The main functional change is that the test will now panic if there are less failures than expected (with a helpful message telling you to update the expected count). This will ensure we do not by fix a test case and then later regress it again.

### DIFF
--- a/gix-url/tests/baseline.rs
+++ b/gix-url/tests/baseline.rs
@@ -2,63 +2,87 @@ use std::any::Any;
 
 use bstr::ByteSlice;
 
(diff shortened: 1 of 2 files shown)


niklaswimmer:
fyi I am done here

Byron:
Thank you, this is definitely an improvement, feel free to keep adjusting so it works for you.

PR #1048, opened 2023-10-05
improvements

Byron:
Make sure `gix-url` does not stall.


Byron:
Rebased, commit is in `main`.

PR #1054, opened 2023-10-09
feat: add shell completions

### LINKED ISSUE #877: Shell completions support
### Summary 💡

What I mean is [shell tab completions](https://en.wikipedia.org/wiki/Command-line_completion).

I think those shells should be supported taking priority into account:
1. Bash
2. Fish
3. Zsh
4. Elvish

Then maybe `PowerShell` for windows, but I don't think any one care 🙃.

---

I thought about using [`clap_complete`](https://crates.io/crates/clap_complete) to auto-generate completion scripts, but after a simple check in source code, we may need to make some changes that could have some complexity to generate them during build-time. However, for run-time this can be easily done, although some additional bloat have to be added to the binary to provide an option like `--generate-completions` in the CLI interface. It can be added as a feature for those who want to enable it before building the tool.

We may face a problem with auto-generation, like in `git` we may need dynamic completions, for example: display available branches or list of commits or staged files. `clap_complete` doesn't support dynamic completions as in https://github.com/clap-rs/clap/issues/1232, but initially it may not be important for `gix`.

### Motivation 🔦

It will make using `gix` easy and more productive.

ShaunSHamilton:
Closes #877 

Adds shell completions, by adding a new `gix` command `generate-completions`:

```bash
gix generate-completions --shell bash --out-dir /tmp
# Source changes
gix <TAB>
```

---

I assume the version and CHANGELOG are automatically updated. Also, I am unsure if this should be under `feat` or `BREAKING!` - an exhaustive public enum has been changed, but I am not sure how/if it is used as a library.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -483,6 +483,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_complete"
+version = "4.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e3ae8ba90b9d8b007efe66e55e48fb936272f5ca00349b5b0e89877520d35ea7"
+dependencies = [
+ "clap",
+]
+
+[[package]]
 name = "clap_derive"
 version = "4.3.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1148,6 +1157,7 @@ version = "0.30.0"
 dependencies = [
  "anyhow",
  "clap",
+ "clap_complete",
  "crosstermion",
  "document-features",
  "env_logger",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,12 @@ default = ["max"]
 ## Everything, all at once.
 ##
 ## As fast as possible, tracing, with TUI progress, progress line rendering with auto-configuration, all transports based on their most mature implementation (HTTP), all `ein` tools, CLI colors and local-time support, JSON output, regex support for rev-specs.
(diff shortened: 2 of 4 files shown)


Skgland:
I don't think this command should assume that the user wants to install the completions system-wide, especially as that requires elevated permission (sudo). Most tools with an option to generate completions that I know output to stdout. This way the user can decide where to place the completions and no elevated permissions are required. I think a few of the tools I know by default add their completions by adding to bashrc and sourcing the output without ever writing to a file. Though that has other implications as the who knows what is sourced each time bashrc is run.

Skgland:
One example would be `rustup` e.g. `rustup completions bash`

The help text when running `rustup completions` without specifying a shell include:
`rustup completions bash cargo >> ~/.local/share/bash-completion/completions/cargo` for how to setup the completions.
Though I think that is typically also done automatically as part of the rustup installation.

Byron (changes requested):
Thanks! This contributions comes at a surprise but I think it can be made to work.

You could make it so it's very similar if not the same as [the same command for `atuin`](https://github.com/atuinsh/atuin/blob/main/atuin/src/command/mod.rs#L70).

In general, please do not attempt to auto-install these completions, being able to generate them is a good first step that already helps those who really want it.

@Skgland Please feel free to review as well, I think we want the same.

Skgland:
I think the latest changes look good :+1:

Byron:
Thanks a lot!

PR #1069, opened 2023-10-16
Allow changing URL password

bittrance:
When a GitHub app wants to clone a private repo, it is required to pass the access token as a part of the URL https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app-installation :unamused: . A `gix-url` user may want to use `Url` to store the validared Url internally, but the access token is retrieved at runtime, so such an application will want to clone the Url and set the password just before fetching the private repo.

### DIFF
--- a/gix-url/src/lib.rs
+++ b/gix-url/src/lib.rs
@@ -124,6 +124,13 @@ impl Url {
         self.user = user;
         prev
     }
+
+    /// Set the given `password`, or unset it with `None`. Return the previous value.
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot! Now the API is finally symmetric.

`gix-url` started out a little bit restrictive about password to the point where one couldn't even query them, but fortunately these dark ages are over now :).

PR #1070, opened 2023-10-18
trace packetlines

Byron:
Related to #1061 

The difficulty here is to make that configurable without the primary use of environment variables so that a flag can be passed down from the porcelain to plumbing.


Byron:
Merged after adjusting the last commit, hence GH doesn't recognize it.

PR #1076, opened 2023-10-25
Add feature to allow using rustls without trust-dns

rsdy:
Using `gitoxide` with bloop, we have observed that putting a laptop to sleep and waking it up often causes intermittent network errors. Using `trust-dns`, the DNS lookups never seem to recover, while using `hyper` directly works.

There's a related issue in [`reqwest`](https://github.com/seanmonstar/reqwest/pull/437) which seems to also point in this direction, and my own testing confirms disabling `trust-dns` resolves this issue.

It would be helpful to include this feature in the mainline repository so we aren't forced to use a fork or other workarounds.


Byron:
Thanks a lot!

Please note that I have made the 'no-trust-dns` version the default and have added a `blocking-http-transport-reqwest-rust-tls-trust-dns` feature for the previous behaviour. In theory, that is still pure-Rust as `getattrinfo` is `libc` and that should *still* be alright given that everything relies on `libc` anyway, at least in these early days.

If there is anything else, please let me know here or in #1077 .

rsdy:
Appreciate the super quick response on this. This looks great!

PR #1080, opened 2023-10-26
fix: prevent underflows while trying to `abs()` very large numbers.

Byron:
For example, "@{-9223372036854775808}" could trigger a panic previously.

### DIFF
--- a/gix-revision/src/spec/parse/function.rs
+++ b/gix-revision/src/spec/parse/function.rs
@@ -422,7 +422,7 @@ where
                 if n < 0 {
                     if name.is_empty() {
                         delegate
-                            .nth_checked_out_branch(n.abs().try_into().expect("non-negative isize fits usize"))
+                            .nth_checked_out_branch(n.unsigned_abs())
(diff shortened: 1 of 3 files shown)


gix-revision/src/spec/parse/function.rs
```diff
@@ -422,7 +422,12 @@ where
                 if n < 0 {
                     if name.is_empty() {
                         delegate
-                            .nth_checked_out_branch(n.abs().try_into().expect("non-negative isize fits usize"))
+                            .nth_checked_out_branch(
```
Skgland, line 425: Isn't the only negative `isize` for there is no positive `isize` value `isize::MIN` which could be special cased to turn into `isize::MAX as usize + 1`? So no need to turn it into an error?

Skgland: Or if msrv is at least 1.51 there is also [isize::unsigned_abs](https://doc.rust-lang.org/std/primitive.isize.html#method.unsigned_abs)

PR #1097, opened 2023-11-09
drop "MIT License" line from LICENSE-MIT

martinvonz:
Since jj now uses gitoxide, I need to import many gitoxide crates into Google's source control system. The automation at Google doesn't like the "MIT License" line at the beginning of the LICENSE-MIT file. AFAICT, the automation is correct in thinking that the line is not supposed to be there.


Byron (approved):
Thanks for the fix, much appreciated!

PR #1100, opened 2023-11-09
Count removed bytes correctly

caio:
Hello,

Thanks a lot for gix! I've been playing with it on and off for a while now and am very impressed with the speed and quality of your work.

One of the things I've done is implement my own version of `git log -- somefile`, which led me to doing my own rename detection and then my results started looking different from gix's

What follows is a patch to fix what I believe is the bug: removed_bytes is being overwritten every time the sink looks at a hunk of changes.

(Qualifying the `sum()` as `sum::<usize>()` is required due to multiple impls of std::ops::AddAssign on the type)


caio:
Glad to hear it! I still have many small things I plan on using gix for- will let you know if anything comes up :)

PR #1115, opened 2023-11-16
fix pointer aliasing in in_parallel.rs

martinvonz:
In `in_parallel_with_slice()`, there's a `&mut *input.0` that creates a `&mut [I]` of the whole slice in `Input`. The same value is created in multiple threads, thus potentially resulting in multiple threads having mutable references to the same value.

This patch fixes that by instead only getting mutable pointers to individual values. It does so by storing a `*mut T` in `Input` instead of the current `*mut [I]`, and then using `pointer::add()` to get pointers to individual elements.

Thanks to @Ralith for bringing this up and proposing the solution in our review of unsafe code at Google.

### DIFF
--- a/gix-features/src/parallel/in_parallel.rs
+++ b/gix-features/src/parallel/in_parallel.rs
@@ -229,11 +229,12 @@ where
                 .expect("valid name");
 
             let input_len = input.len();
-            struct Input<I>(*mut [I])
+            struct Input<I>(*mut I)
             where
                 I: Send;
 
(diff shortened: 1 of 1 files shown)


Ralith (approved):

Byron:
Thanks so much! I don't even know how I ended up with keeping a slice instead of a pointer, and I am very thankful for any improvements to the (few) unsafe spots in this codebase.

I really, really, don't like writing it and am not very familiar with it either, so please do keep this PRs coming :).

PR #1117, opened 2023-11-16
shell-avoidance on windows

Byron:
This should help to allow windows users to use more complex programs,
which might not have been possible previously.

It also allows certain areas of `gix` to indicate that they are safe to
use with this flag, which always is the case when `git` is prefixed, like
with credential helpers.

### DIFF
--- a/gix-command/src/lib.rs
+++ b/gix-command/src/lib.rs
@@ -20,6 +20,19 @@ pub struct Prepare {
     pub env: Vec<(OsString, OsString)>,
     /// If `true`, we will use `sh` to execute the `command`.
     pub use_shell: bool,
+    /// If `true` (default `true` on windows and `false` everywhere else)
+    /// we will see if it's safe to manually invoke `command` after splitting
+    /// its arguments as a shell would do.
+    /// Note that outside of windows, it's generally not advisable as this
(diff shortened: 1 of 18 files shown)


Byron:
@jalil-salame Today I finally found a way to achieve the same (or similar) program launch behaviour on windows as `git`, which doesn't use an intermediate shell there from what I can tell. This will help with any program invocation performed through `gitoxide`, like worktree-filters and diff-converters, on windows.

Previously this was a special case, albeit platform-independent, for credential helpers only.

gix-command/src/lib.rs
```diff
@@ -20,6 +20,19 @@ pub struct Prepare {
     pub env: Vec<(OsString, OsString)>,
     /// If `true`, we will use `sh` to execute the `command`.
     pub use_shell: bool,
+    /// If `true`, default `true` on windows and false everywhere else,
```
fbstj, line 23:
```suggestion
    /// If `true` (default `true` on windows and `false` everywhere else),
```

jalil-salame:
That's great! Thank you! 🎉

gix-command/src/lib.rs
Byron: Thanks, great catch! I will apply that.

Issue #1109, opened 2023-11-13
Respect `http.sslVerify`


jalil-salame:
### Summary 💡

I need to ignore the validity of SSL certificates (corporate proxy so security is "not a concern").

### Motivation 🔦

As mentioned in https://github.com/Byron/gitoxide/issues/1103#issuecomment-1808242110. I use git/gix behind a corporate proxy which has a self-signed certificate. To use gix there I need to either not verify the certificates, or to respect the configured certificates on windows (this is not done by git/curl so I would not personally pursue this path).

Byron:
It turns out that `gix` knows the following about that:

```
❯ gix config-tree | rg sslVerify
 🤔   http.sslVerify                                       not planned ℹ on demand ℹ  
```

It's indeed not implemented even for the curl backend, and doing so should be straightforward to wire up.
Please note that in the `max-pure` configuration, most of the `http` options aren't supported, as it needs a completely separate implementation given that it relies on `reqwest`.

dirthsj:
> or to respect the configured certificates on windows (this is not done by git/curl so I would not personally pursue this path).

Note that this *does* work with git under the following option;

`git config --global http.sslBackend schannel`

I would hope `gix` can also respect this setting eventually (maybe that's a separate issue though).

Byron:
Thanks for sharing!

I am quite sure that eventually this will be implemented as `gix` has a real opportunity to be more compatible to Windows than `git` is.

Byron:
Closing because a lot has changed, and I wouldn't be surprised if this is supported by now by one of the HTTP backends.

PR #1135, opened 2023-11-28
Add support to http.sslVerify

Alvenix:
I just started working on [1109](https://github.com/Byron/gitoxide/issues/1109). So work in progress. I am very new to the project so I may take time.

### DIFF
--- a/gix-transport/src/client/blocking_io/http/mod.rs
+++ b/gix-transport/src/client/blocking_io/http/mod.rs
@@ -179,6 +179,10 @@ pub struct Options {
     pub ssl_ca_info: Option<PathBuf>,
     /// The SSL version or version range to use, or `None` to let the TLS backend determine which versions are acceptable.
     pub ssl_version: Option<SslVersionRangeInclusive>,
+    /// Controls whether to perform SSL identity verification or not. Turning this off is not recommended and can lead to
+    /// various security risks. An example where this may be needed is when an internal git server uses a self-signed
+    /// certificate and the user accepts the associated security risks.
+    pub ssl_verify: bool,
     /// The HTTP version to enforce. If unset, it is implementation defined.
     pub http_version: Option<HttpVersion>,
     /// Backend specific options, if available.
--- a/gix/src/repository/config/transport.rs
+++ b/gix/src/repository/config/transport.rs
@@ -407,13 +407,31 @@ impl crate::Repository {
 
                     {
                         let key = "http.sslVerify";
-                        opts.ssl_verify = config
+                        let ssl_verify = config
                             .boolean_filter_by_key(key, &mut trusted_only)
                             .map(|value| config::tree::Http::SSL_VERIFY.enrich_error(value))
                             .transpose()
                             .with_leniency(lenient)
                             .map_err(config::transport::http::Error::from)?
                             .unwrap_or(true);
+
+                        let ssl_no_verify = config
+                            .boolean_filter(
+                                "gitoxide",
+                                Some("http".into()),
+                                gitoxide::Http::SSL_NO_VERIFY.name,
+                                &mut trusted_only,
+                            )
+                            .and_then(Result::ok)
+                            .unwrap_or_default();
+
+                        // ssl_no_verify take prority here because it is based on environment variable
+                        // and we try to match git behavior.
+                        if ssl_no_verify {
+                            opts.ssl_verify = false;
+                        } else {
+                            opts.ssl_verify = ssl_verify;
+                        }
                     }
 
                     #[cfg(feature = "blocking-http-transport-curl")]
--- a/gix/src/repository/config/transport.rs
+++ b/gix/src/repository/config/transport.rs
@@ -405,6 +405,17 @@ impl crate::Repository {
                         }
                     }
 
+                    {
+                        let key = "http.sslVerify";
+                        opts.ssl_verify = config
+                            .boolean_filter_by_key(key, &mut trusted_only)
+                            .map(|value| config::tree::Http::SSL_VERIFY.enrich_error(value))
+                            .transpose()
+                            .with_leniency(lenient)
+                            .map_err(config::transport::http::Error::from)?
+                            .unwrap_or(true);
+                    }
+
                     #[cfg(feature = "blocking-http-transport-curl")]
                     {
                         let key = "http.schannelCheckRevoke";
--- a/gix/tests/repository/config/transport_options.rs
+++ b/gix/tests/repository/config/transport_options.rs
@@ -55,6 +55,7 @@ mod http {
             verbose,
             ssl_ca_info,
             ssl_version,
+            ssl_verify,
             http_version,
             backend,
         } = http_options(&repo, None, "https://example.com/does/not/matter");
@@ -106,6 +107,9 @@ mod http {
                 max: version
             })
         );
+
+        assert!(ssl_verify, "SSL verification is enabled by default if not configured");
+
         assert_eq!(http_version, Some(HttpVersion::V1_1));
     }
 
@@ -313,5 +317,14 @@ mod http {
         let opts = http_options(&repo, None, "https://example.com/does/not/matter");
         assert_eq!(opts.proxy.as_deref(), Some("http://localhost:9090"));
         assert_eq!(opts.follow_redirects, FollowRedirects::Initial);
+    }
+
+    #[test]
+    fn no_ssl_verify() {
+        let repo = repo("no-ssl-verify");
+
+        let opts = http_options(&repo, None, "https://example.com/does/not/matter");
+
+        assert!(!opts.ssl_verify);
     }
 }
--- a/src/plumbing/progress.rs
+++ b/src/plumbing/progress.rs
@@ -409,10 +409,6 @@ static GIT_CONFIG: &[Record] = &[
         usage: NotPlanned { reason: "on demand" }
     },
     Record {
-        config: "http.sslVerify",
-        usage: NotPlanned { reason: "on demand" }
-    },
-    Record {
         config: "http.sslCert",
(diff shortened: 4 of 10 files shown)


Alvenix:
I have tested the feature in both max and max-pure features and it works as expected. I am waiting for review.

Should I add the modified fixtures to the commit?

gix-transport/Cargo.toml
```diff
@@ -78,7 +78,7 @@ base64 = { version = "0.21.0", optional = true }
 curl = { version = "0.4", optional = true }
 
 # for http-client-reqwest
-reqwest = { version = "0.11.12", optional = true, default-features = false, features = ["blocking"] }
+reqwest = { version = "0.11.12", optional = true, default-features = false, features = ["blocking", "rustls-tls"] }
```
Byron, line 81: `rustls-tls` can't be enabled here, it must be up to the user of `gix` to do that.

Doing so seems unrelated to the task at hand, but I see that it's needed to make the feature available in `reqwest`. This also means that in order to ever get there, one would have to wire all reqwest related features though `gix-transport`, which seems like a maintenance nightmare.

Thus, let's remove `reqwest` support for `sslVerify` until there is a solution.

gix-transport/src/client/blocking_io/http/mod.rs
```diff
@@ -179,6 +179,8 @@ pub struct Options {
     pub ssl_ca_info: Option<PathBuf>,
     /// The SSL version or version range to use, or `None` to let the TLS backend determine which versions are acceptable.
     pub ssl_version: Option<SslVersionRangeInclusive>,
+    /// Controls whether to perform ssl identity verification or not
```
Byron, line 182: Can you explain more about this? Why would one want to turn that on, or off. Are there security implications?

All that can be written in the doc-string. Please write in full sentences as well, which end in a `.` .

gix-transport/src/client/blocking_io/http/reqwest/remote.rs
```diff
@@ -44,38 +44,44 @@ impl Default for Remote {
...
-                }))
+                    }))
+            }
+
+            // We may error while configuring, which is expected as part of the internal protocol. The error will be
+            // received and the sender of the request might restart us.
+            let client_ssl_verify = setup_client_builder(allow_redirects.clone()).build()?;
+
+            let client_no_ssl_verify = setup_client_builder(allow_redirects.clone())
+                .danger_accept_invalid_certs(true)
```
Byron, line 84: Due to feature toggles of `reqwest` being invisible here, it's impossible to conditionally perform this operation. So let's remove reqwest support for this until there is a tenable solution - otherwise this crate will become a maintenance nightmare quickly.

gix-transport/src/client/blocking_io/http/reqwest/remote.rs
```diff
@@ -44,38 +44,44 @@ impl Default for Remote {
...
-                    }
-                }))
+                    }))
+            }
+
+            // We may error while configuring, which is expected as part of the internal protocol. The error will be
+            // received and the sender of the request might restart us.
+            let client_ssl_verify = setup_client_builder(allow_redirects.clone()).build()?;
+
+            let client_no_ssl_verify = setup_client_builder(allow_redirects.clone())
```
Byron, line 83: (*There is no need to create two clients for this. Instead, just re-assign the (one) client conditionally - this note is for reference only as reqwest support has to be removed for now.*)

gix/src/repository/config/transport.rs
```diff
@@ -405,6 +405,14 @@ impl crate::Repository {
                         }
                     }
 
+                    {
+                        let key = "http.sslVerify";
+                        opts.ssl_verify = config
+                            .boolean_filter_by_key(key, &mut trusted_only)
```
Byron, line 411: Please take a look around and note that each of the keys above in some way references a respective static version of the key they access. This must be done here as well, and there should be a `Http::SSL_VERIFY` constant.

src/plumbing/progress.rs
```diff
@@ -408,10 +408,6 @@ static GIT_CONFIG: &[Record] = &[
         config: "http.sslCipherList",
         usage: NotPlanned { reason: "on demand" }
     },
-    Record {
```
Byron, line 411: It's a good rule of thumb that whenever something is removed here, it should be added as a constant.

gix/tests/repository/config/transport_options.rs
```diff
@@ -106,6 +107,9 @@ mod http {
                 max: version
             })
         );
+
+        assert!(ssl_verify);
```
Byron, line 111: Please use the second argument to `assert!()` for a message that explains that the default is `true` here - this is the value of this assertion really, as an accidental 'false if unset' would be a terrible security issue.

Byron (changes requested):
Thanks a lot for tackling this. I like that it adds tests for portions of `gix` code that already have tests.

Please don't worry about integration tests for this, it's a topic of its own that can be tackled separately.

Generally, please split commits across crate boundaries. If it's a feature, please prefix the message with `feat: ` so the message gets into the changelog. You can take a look at other commits like it for reference.

Within each crate, also please split work into different commits if they touch different features. Commit messages could be more exhaustive as well.

If in doubt, you can create one commit per step and squash them later according to certain criteria.

While continuing to work with the codebase, I recommend to pay extra attention to the surrounding of the code you are adding, and to mimic everything down to small details like the usage of `.`.

gix/src/repository/config/transport.rs
Alvenix: I added a debug_assert to ensure the key match the constants, is that what you meant?

gix-transport/src/client/blocking_io/http/reqwest/remote.rs
Alvenix: I am not sure I understood what you meant. I did it this way, because the config is read from each Request after creating the client, so instead of recreating clients every time I did it like this. However, I removed reqwest support for now.

gix/src/repository/config/transport.rs
Byron: That's close. The question is why it shouldn't (potentially) fail if the value can't be read. [In this line](https://github.com/Byron/gitoxide/pull/1135/commits/778e597100740651fefff18c38c6522b681e180b#diff-9fe748aff5bf996a7a4dd05505aca13d725e2e6766613b1ccfbc3e22679120d8R344) there is an example for how to do this.

gix-transport/Cargo.toml
Alvenix: I think if this [feature](https://github.com/rust-lang/rust/issues/64797) is ever stabilized, it may solve the issue cleanly.

gix-transport/src/client/blocking_io/http/mod.rs
Alvenix: I added more explanation.

src/plumbing/progress.rs
Alvenix: I added the constants, but I am not sure if I should link it with the environment variable: GIT_SSL_NO_VERIFY, as I didn't add support for it and it is the negation of sslVerify.

gix/src/repository/config/transport.rs
Alvenix: Okay I understand what you mean, I added it.

Alvenix:
> Generally, please split commits across crate boundaries. If it's a feature, please prefix the message with `feat: ` so the message gets into the changelog. You can take a look at other commits like it for reference.
> 
> Within each crate, also please split work into different commits if they touch different features. Commit messages could be more exhaustive as well.

I split the commits and added feat: before them.

src/plumbing/progress.rs
Byron: Yes, that's a great point!

Keys can have an environment variable attached to it [using this method](https://github.com/Byron/gitoxide/blob/345712dcdfddcccc630bbfef2ed4f461b21550d3/gix/src/config/tree/keys.rs#L64). These then needs to be [initialized here](https://github.com/Byron/gitoxide/blob/8434aab5fb6ce32be2bf3b20e38c28c780bd5db9/gix/src/config/cache/init.rs#L74). However, as it is a negation, we'd need add our own version of that variable as `gitoxide.http.sslNoVerify`, link it with (with a `description()`) `GIT_SSL_NO_VERIFY`, initialize it, [test it](https://github.com/Byron/gitoxide/blob/8434aab5fb6ce32be2bf3b20e38c28c780bd5db9/gix/tests/gix-init.rs#L23), and then let it override any value that `http.sslVerify` may have [just like git](https://github.com/git/git/blob/cfb8a6e9a93adbe81efca66e6110c9b4d2e57169/http.c#L1293-L1294).

Hooking up variables is quite involved, but that separation is needed to avoid global state (like environment variables) to sneak in. This way they are controlled and used exactly in one or two places in the codebase, while being introspectable via `gix config` and controllable (as one can turn off certain kinds of environment variables from being read).

Byron (commented):
Thanks for bearing with me!

One more issue to resolve came up in the form of `GIT_NO_SSL_VERIFY`, but handling it is a good thing as it leads through other parts of the boot process that you probably want to be familiar with.

Byron (approved):

I have a made a few modifications and had to push them to a separate PR (#1142), which will be merged once CI passes. Thus, this PR can be closed as it was superseded.

Please note that next to my changes done in separate commits, I also changed one commit message from `feat:` to `feat!:` to indicate a breaking change - I forgot to mention this previously. 

There have been a few adjustments you might find interesting, and I recommend to re-test from my branch to be sure it still works. I'd expect that, as `gix` now/still does what `git` does as well in terms of `curl` configuration.

Byron:
Closing as this PR was superseded by #1142.

PR #1137, opened 2023-11-28
Document and restrict some unsafe code in gix-pack

martinvonz:
Thanks to @cramertj for highlighting these issues in our internal review of unsafe code at Google.


Byron:

This portion of the code has seen a lot of revisions, probably more than anything else, which once more shows how hard it is to get `unsafe` code right. And for better or worse, it's hard to get training with it as well and the Rustonomicon only gets you this far.

PR #1140, opened 2023-11-30
Connect new gitoxide.credentials subsection into section tree

bittrance:
I believe #1127 forgot to include the new `gitoxide.credentials` subsection into the section tree?


Byron:
Oh, true, that should have been there! Thanks for the fix!

PR #1141, opened 2023-11-30
Allow configuting the Repository object created by PrepareFetch

bittrance:
In the process of implementing terminal prompt configuration (see #1093) I cannot find a way to apply this configuration when using `gix::prepare_clone()`, that is, before calling `.fetch_only()`. There seems to be two alternatives to this:

- Add a `Remote::repo_mut()` method so that you can do `remote.repo().config_snapshot_mut()` in `PrepareFetch::configure_remote()`
- Add a `PrepareFetch::configure()` method which gives access to a `&mut SnapshotMut<'_>` object

I don't like the first solution. You are supposed to have recevied the `Remote` from a `Repository` and you should be able to do configuration from there. To my understanding the `configure_{remote,transport}` methods exist because you can do advanced stuff with the respective object that is not directly coverend by cofngiration.

I'm not too fond of the second option either (not least because there would be three different configuration methods on `PrepareFetch`), but I created this pull request to point out that there is currently no way to perform such configuration. 
 (You may of course claim that `gix::prepare_clone` is a convenience method and that if you need to configure your repo, you should init an empty repo and fetch into it.)


Byron:
Maybe it's a bit much to have to deal with two work-streams at the same time. If so, please feel free to close this PR and reopen once there is more time.

Byron:
> Config parse errors are not emitted until you fetch, which is a bit of a wart, but probably a minor one since in 9 cases out of 10, you will do both at the same time like the above example.

That's true, but it's supposed to be used like [like here in `cargo`](https://github.com/rust-lang/cargo/blob/e4e10f2393847071b5b35531cf182eb019edd281/src/cargo/sources/git/oxide.rs#L280). Going through `Keys` makes it safe as it is checked at the call-site.

Byron (approved):
Thanks a lot! This looks just like I imagined it :)!

And yes, a lot has to happen to make builders consistent, and further to change the API surface so that it can be stabilized. Right now, there are many open options structs that will break the API with each new addition. Further, each error enum technically breaks the API each time a new variant is added.

In any case, I have made minor adjustments, added a test, and will merge when CI is green.

bittrance:
Nice :+1:  Sorry about missing test cases; for some reason I looked for tests only inside `gix/src/clone` for existing test cases.

PR #1147, opened 2023-12-03
Recommend just over make in README.md and COLLABORATING.md

Alvenix:
Because the current makefile no longer support the mentioned args.

Side note:
I am not sure if check-size should be also still recommended because it fails on my device with:

```
  in gix-diff   →       Error: The actual estimated package size of 31.4 KB exceeded the limit of 15.0 KB by 16.4 KB
error: Recipe `check-size` failed on line 218 with exit code 1
```

But I assume it works on the CICD or maybe it is not executed.

Another side note:
index::file::read::sparse_checkout_non_sparse_index test fails which I had to ignore. I made sure to use git-lfs but maybe it is something related to my specific setup. So I will simply add #[ignore] to it locally. (All other tests work when I ignore it)

Here is the error:
```
---- index::file::read::sparse_checkout_non_sparse_index stdout ----
thread 'index::file::read::sparse_checkout_non_sparse_index' panicked at gix-index/tests/index/file/read.rs:267:13:
  left: Flags(EXTENDED | SKIP_WORKTREE)
 right: Flags(0x0)


failures:
    index::file::read::sparse_checkout_non_sparse_index
```

### DIFF
--- a/COLLABORATING.md
+++ b/COLLABORATING.md
@@ -10,7 +10,7 @@
      - …if collaboration on a feature and prior feedback is desired
    - **feel free to use the [project-board] to organize your issues, PRs or cards**
 - **`main` must never be broken or show warnings**
-   - An easy way to achieve this is to run `make tests check-size` before pushing or `make tests check-size && git push`.
+   - An easy way to achieve this is to run `just test check-size` before pushing or `just check-size && git push`.
    - If you're unsure about remembering to do this, we suggest using a pre-commit git hook.
 - **if `main` breaks on CI** _which can happen nonetheless_…
     - …and you _do know_ the cause, please fix it immediately. If necessary by reverting the offending commit until a more durable fix is possible.
--- a/README.md
+++ b/README.md
@@ -284,7 +284,7 @@ If what you have seen so far sparked your interest to contribute, then let us sa
 > ❗️Note❗️: For cloning, `gix-lfs` needs to be locally installed or the checkout will fail. `git lfs install` must have been called once, followed
   by `git lfs pull` to replace the `lfs`-pointer files.
 
-We recommend running `make tests check-size` during the development process to assure CI is green before pushing.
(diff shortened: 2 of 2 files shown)


Alvenix:
> This also means that if you'd be sure to use the files from [the archive](https://github.com/Byron/gitoxide/blob/27471e7f421446cc8f6d4543bdf85fe4e753b944/gix-index/tests/fixtures/generated-archives/v3_skip_worktree.tar.xz), it _should_ work. If the archive is present and an actual archive, deleting the whole `generated-do-not-edit` folder, it would extract the files from the archive accordingly.

It seems that the archive is ignored due to differences in the script identity.

Adding printing to checking the archive code as follows:
```
println!("Required script identity: {required_script_identity}");
println!("Archive identity: {archive_identity}");

if archive_identity != required_script_identity {
    return Err(std::io::ErrorKind::NotFound.into());
}
```

would print the following when the tests fail.
```
Required script identity: 1190435892
Archive identity: 4093570147
```

Even if I modify the code to force it to use the archive, the tests fails earlier due to older git version:
```
thread 'index::file::read::sparse_checkout_non_sparse_index' panicked at gix-index/tests/index/file/read.rs:259:5:
  left: V2
 right: V3
```
```
assert_eq!(file.version(), Version::V3);
```

Perhaps the provided archive in the repo is old somehow? I will ignore this test for now and rely on cicd for final checks.

Byron:
Thanks for bringing this up!

I have had a few changes related to archives in the pipeline and will get them to main now. In any case, you have discovered that some archives were stale, and indeed it depends on some factors if the scripts then evaluate to the same (or expected) result. When creating the archive-system I did it for performance, but it turns out to be vital for reproducible tests as well.

The archives in question will be updated, and from that point on I suspect it should work for you, too.

In order to prevent that from happening again, I'd have to write something that validates that all archives match their scripts, but I will skip this for now as it doesn't seem trivial enough.

PR #1154, opened 2023-12-06
Remove unsafe transmute of should_interrupt

maurer:
Adds a lifetime to the `ExtendedBufRead` trait to specify how long the callback provided must live.

cc @manishearth

### DIFF
--- a/gix-protocol/src/fetch/arguments/async_io.rs
+++ b/gix-protocol/src/fetch/arguments/async_io.rs
@@ -9,7 +9,7 @@ impl Arguments {
         &mut self,
         transport: &'a mut T,
         add_done_argument: bool,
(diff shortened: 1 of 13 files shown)


martinvonz (approved):
Looks good to me. Thanks!

Byron (approved):
Thanks so much!

The description of the transmute already makes clear how I struggled to make this work, and for some reason adding a lifetime is nothing I managed to do.

It's also great to see that I was awfully wrong thinking that this cannot currently be expressed, and that ultimately the solution was quite simple and definitely for the better.

Byron:
Closing, as this was merged with the commit rewritten as 88f8b34 , I needed the commit message to declare that this is a breaking change.

PR #1163, opened 2023-12-08
Fix one more unsafe code issue in the `gix` crate

martinvonz:


Byron:
Closing, as the commits were merged into main after rewriting one of them to prepend `fix!:` to the commit message. That way it's assured the breaking change will cause the version number to be incremented properly.

PR #1169, opened 2023-12-09
Add config section for mailmap

bittrance:
While working with #1125, I noticed that there were no config tree constants for `mailmap.{blob,file}`. This PR adds those keys. They will later be put to use in https://github.com/Byron/gitoxide/blob/58ed530da0eed05b6b5b97d4941a53e06b679918/gix/src/repository/mailmap.rs#L30 and following.


Byron (approved):
Thanks a lot for getting the ball rolling on these! I hope these are the last raw usages of configuration keys as well.

When looking at the mailmap code it was clear that it was from earlier times when configuration access was much more 'raw' than it is now - it surely has come a long way already, but also still has some way to go.

*This reminds me that `trusted_path()` and maybe even `path()` would still have to be overridden (or amended) in some way in order to get automatic path interpolation.*

PR #1170, opened 2023-12-10
fix: NetBSD doesn't have `st_mtime_nsec` and `st_ctime_nsec`

weihanglo:
fix: NetBSD doesn't have `st_mtime_nsec` and `st_ctime_nsec`

It has `st_mtimensec` and `st_ctimensec` instead.

* https://man.netbsd.org/NetBSD-8.0/stat.2
* https://docs.rs/libc/0.2/x86_64-unknown-netbsd/libc/struct.stat.html#structfield.st_mtimensec
* https://docs.rs/libc/0.2/aarch64-unknown-netbsd/libc/struct.stat.html#structfield.st_mtimensec
* https://docs.rs/libc/0.2/x86_64-unknown-netbsd/libc/struct.stat.html#structfield.st_ctimensec
* https://docs.rs/libc/0.2/aarch64-unknown-netbsd/libc/struct.stat.html#structfield.st_ctimensec

This blocks `cargo` submodule sync in rust-lang/rust btw: <https://github.com/rust-lang/rust/pull/118765#issuecomment-1848579803>

### DIFF
--- a/gix-index/src/fs.rs
+++ b/gix-index/src/fs.rs
@@ -56,7 +56,10 @@ impl Metadata {
         {
             Some(system_time_from_secs_nanos(
                 self.0.st_mtime.try_into().ok()?,
+                #[cfg(not(target_os = "netbsd"))]
                 self.0.st_mtime_nsec.try_into().ok()?,
+                #[cfg(target_os = "netbsd")]
+                self.0.st_mtimensec.try_into().ok()?,
             ))
         }
         #[cfg(windows)]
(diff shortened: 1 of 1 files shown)


weihanglo:
I have no NetBSD environment to test it though 😆.

Byron (approved):
Oh my!

I wasn't aware `gitoxide` is somehow pulled into the rust build chain!

This means that in theory, I would have to test all targets that Rust is being built on to prevent this from happening again. I guess it's possible, but I don't know about feasibility.

After merging, I will create a new patch release which maybe fixes the issue for you downstream, or at least makes it fixable.

Byron:
A `gix-index` patch was released, and I hope it works for you.

weihanglo:
Thanks for the release!

I don't think it tests all targets, but at this tiers with host build tools :)

PR #1172, opened 2023-12-11
allow organize to move repositories into a child directory of their current location

jcgruenhage:
I've run into a situation, where I had cloned a repo into a location and the canonicalized destination was a subpath of that location. `ein tool organize --execute` was unhappy about that:

```
~/dev 
❯ ein t organize --execute
 10:00:35 organize Moving ./example.org/foo to /path/to/destination/example.org/foo/foo
 10:00:35 organize Error when handling directory "./example.org/foo": Invalid argument (os error 22)
Error: Failed to handle 1 repositories
```

This patch moves directories twice, allowing for things like this to work out just fine:

```
~/dev 
❯ ein t organize --execute
 10:04:41 organize Moving ./example.org/foo to /path/to/destination/example.org/foo/foo
 ```

### DIFF
--- a/gitoxide-core/src/organize.rs
+++ b/gitoxide-core/src/organize.rs
@@ -198,9 +198,17 @@ fn handle(
             destination.display()
         )),
         Mode::Execute => {
-            std::fs::create_dir_all(destination.parent().expect("repo destination is not the root"))?;
+            if destination.starts_with(git_workdir) {
+                let tempdir = tempfile::tempdir_in(canonicalized_destination)?;
+                let tempdest = tempdir.path().join(destination.file_name().expect("repo destination is not the root"));
+                std::fs::rename(git_workdir, &tempdest)?;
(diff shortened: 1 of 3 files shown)


Byron:
Thanks a lot for the fix! I wasn't aware of this issue yet either.

Now I wonder if you'd consider making the tempdir-indirection dependent on whether or not `destination` is actually a subdir of the source worktree?

This would probably require to perform a `strip_prefix()` check on the canonicalized version of `workdir`, but overall that should be cheaper than creating temporaries all the time. As a side-effect, that would make really clear why the temporary directory is created in the first place, just by looking at the code.

Thanks for your consideration.

jcgruenhage:
Yep, sounds good to me, will look into it, should be easy enough.

Byron:
Thanks a lot!

I have added a journey test that actually made canonicalisation necessary. When CI is green, this PR will be merged, but please feel free to verify it locally once more as you see fit.

Byron (approved):

Issue #1167, opened 2023-12-09
update ring dependency


ognevnydemon:
### Summary 💡

currenlty ring at 0.16.20. update this dependency to 0.17

### Motivation 🔦

ring 0.17 added Windows on ARM support

Skgland:
On the point of dependency updates it might be of interest to include an update of the `url` crate to get the transitive `idna` dependency from 0.4.0 to 0.5.0, to get the fix for https://github.com/servo/rust-url/issues/870

Byron:
Thanks for letting me know! I wasn't aware a new release of `url` was available.

Otherwise, it seems like `ring` comes in with `reqwest`, which didn't see an upgrade yet.

```
❯ cargo tree -i ring --no-default-features --features max-pure
ring v0.16.20
├── rustls v0.21.7
│   ├── hyper-rustls v0.24.1
│   │   └── reqwest v0.11.22
│   │       ├── gix v0.56.0 (/Users/byron/dev/github.com/Byron/gitoxide/gix)
│   │       │   ├── gitoxide v0.32.0 (/Users/byron/dev/github.com/Byron/gitoxide)
│   │       │   └── gitoxide-core v0.34.0 (/Users/byron/dev/github.com/Byron/gitoxide/gitoxide-core)
│   │       │       └── gitoxide v0.32.0 (/Users/byron/dev/github.com/Byron/gitoxide)
│   │       └── gix-transport v0.39.0 (/Users/byron/dev/github.com/Byron/gitoxide/gix-transport)
│   │           ├── gitoxide-core v0.34.0 (/Users/byron/dev/github.com/Byron/gitoxide/gitoxide-core) (*)
│   │           ├── gix v0.56.0 (/Users/byron/dev/github.com/Byron/gitoxide/gix) (*)
│   │           └── gix-protocol v0.42.0 (/Users/byron/dev/github.com/Byron/gitoxide/gix-protocol)
│   │               └── gix v0.56.0 (/Users/byron/dev/github.com/Byron/gitoxide/gix) (*)
│   ├── reqwest v0.11.22 (*)
│   └── tokio-rustls v0.24.1
│       ├── hyper-rustls v0.24.1 (*)
│       └── reqwest v0.11.22 (*)
├── rustls-webpki v0.101.6
│   └── rustls v0.21.7 (*)
└── sct v0.7.0
    └── rustls v0.21.7 (*)
```

phanen:
[...]

PR #1174, opened 2023-12-12
chore: update lockfile

phanen:
Close https://github.com/Byron/gitoxide/issues/1167.


ognevnydemon:
`ring 0.16.20` is still in lock file :(

Byron:
`cargo update` doesn't work due to `time` being restricted to an older version while the MSRV is this old. It should be possible to use `cargo update -p <package>`, but it seems that `ring` can't really be updated to a new technically breaking version while other packages like `reqwest` still refer to the previous version.

Closing as it seems not to work like intended, feel free to reopen though if that changes.

PR #1178, opened 2023-12-12
Add `core.editor` key

lewiscowper:
Happy to adjust based on feedback, I wanted to add a feature to https://github.com/danobi/prr/pull/36 where we would use `$EDITOR` as a fallback, and use `core.editor` as the primary source, to match expectations with [`git` itself](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_core_editor), but the field isn't (yet) supported here, so this PR adds the `core.editor` field, I hope.

There is a test failing, but it seems like an unrelated failure, and it might be my local setup interfering, as it's about `file::init::comfort::from_git_dir`, so I want to open the PR to let the CI take a shot at it.

Looking forward to your feedback, feel free to edit and adjust as you see fit.

### DIFF
--- a/gix/Cargo.toml
+++ b/gix/Cargo.toml
@@ -61,6 +61,9 @@ comfort = ["gix-features/progress-unit-bytes", "gix-features/progress-unit-human
 #! A component is a distinct feature which may be comprised of one or more methods around a particular topic.
 #! Providers of libraries should only activate the components they need.
 
+## Provide a top-level `command` module that helps with spawning commands similarly to `git`.
+command = ["dep:gix-command"]
+
 ## Obtain information similar to `git status`.
 status = ["gix-status"]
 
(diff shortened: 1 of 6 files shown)


Byron:

I think this is exactly how it would be done, and seeing `Executable` used reminded me of a couple of fixes I had to apply to the existing code as I apparently got that wrong in some places :D. It also turns out that Git uses the shell for `core.editor`  ([see here](https://github.com/git/git/blob/564d0252ca632e0264ed670534a51d18a689ef5d/editor.c#L91)), so it would be identified as `Program` there - I made the fix already and will merge when CI is happy.

Please note that when using the value of `core.editor`, you'd want to assure the value comes from a trusted configuration file (an example could be [`trusted_path()`](https://docs.rs/gix/latest/gix/config/struct.Snapshot.html#method.trusted_path)) - unfortunately there is no such method yet for programs, but in the end it's a call to `string_filter()` with the filter being… it looks like the default filter isn't publicly available.

Let me fix that…

Byron:
Now there also is a `trusted_program()` method that can be used to obtain `core.editor`, and you will also find `command` in the top-level of the crate that helps with [executing these commands 'git-style'](https://docs.rs/gix-command/latest/gix_command/). 

I hope that will help.

PR #1185, opened 2023-12-17
fuzz: Add gix-object fuzzers

silvergasp:


silvergasp:
@Byron, Looks like the CI fuzzer found a bug in some of the [newly fuzzed code](https://github.com/Byron/gitoxide/actions/runs/7235563218/job/19713178558?pr=1185#step:5:317)... I'll file an issue once I've worked out the root cause.

Byron (approved):
Thanks a lot for bringing fuzzing to the next level!

I have addressed the one issue that was found immediately, let's see if CI gets through its first 10m fuzzing period this time :).

PR #1186, opened 2023-12-17
feat: Fuzz more of the gix-config api

silvergasp:


Byron (approved):
It's pretty neat how arbitrary contexts can be generated to try more of the API!

Thanks a lot!

PR #1187, opened 2023-12-17
gix-config: Add fuzzer for File

silvergasp:


silvergasp:
@Byron looks like the CIFuzzer has found [another bug](https://github.com/Byron/gitoxide/actions/runs/7240230486/job/19722979114?pr=1187#step:5:291), in newly fuzzed code. It seems like the there is a parser/serialisation mismatch. I'm not really familiar enough with git configs to say if this is truly a bug or not though.

gix-config/Cargo.toml
```diff
@@ -15,6 +15,7 @@ autotests = false
 [features]
 ## Data structures implement `serde::Serialize` and `serde::Deserialize`.
 serde = ["dep:serde", "bstr/serde", "gix-sec/serde", "gix-ref/serde", "gix-glob/serde", "gix-config-value/serde"]
+arbitrary = ["dep:arbitrary", "gix-sec/arbitrary"]
```
Byron, line 18: I'd love to stop the intrusion of `arbitrary` in its tracks at least as long as we don't have `dev-features` for exactly this purpose.

Is there other ways to deal with this?

gix-config/fuzz/fuzz_targets/fuzz_file_section.rs
```diff
@@ -0,0 +1,42 @@
...
+use gix_config::file::{Metadata, Section};
+use libfuzzer_sys::fuzz_target;
+use std::borrow::Cow;
+use std::hint::black_box;
+
+#[derive(Arbitrary, Debug)]
+struct Ctx<'a> {
+    name: Cow<'a, str>,
+    subsection: Option<&'a [u8]>,
+    meta: Metadata,
```
Byron, line 14: I think if meta data wasn't generated, `arbitrary` can be avoided as feature in `gix-config`. It's OK to not generate everything, and if needed to mindfully iterate the input data with a set of common meta-data configurations.

Byron (changes requested):
Thanks for taking fuzzing to the next level!

I left a comment, but besides that I am not so sure I want to let a fuzzer loose on `gix-config` round-tripping just yet. Even though it is valuable to have, I am just not in the position right now to deal with what it finds - sure, one or two low-hanging fruit should be fine, but there is a risk making the required fixes is huge effort or takes a lot of time.

Ideally, there would be a way to allow failures for particular fuzz targets, like the round-trip fuzzing, so I am not forced to make fixes right away or else have CI broken permanently. Maybe, in an ideal world, CI here would allow roundtrip-failures (but no other failures), while the Google-run fuzzer can open issues for each problem it finds.

What do you think?

silvergasp:
> Thanks for taking fuzzing to the next level!
> 
> I left a comment, but besides that I am not so sure I want to let a fuzzer loose on `gix-config` round-tripping just yet. Even though it is valuable to have, I am just not in the position right now to deal with what it finds - sure, one or two low-hanging fruit should be fine, but there is a risk making the required fixes is huge effort or takes a lot of time.

Yeah that's fair, I'm very grateful for all the work you have put into adding new features to gitoxide as I regularly use it in my personal project. So I don't want to completely bog you down with endless bug reports. 

Do you have a suggestion as to how you might want to proceed more generally. If I'm going to write a fuzzer I don't necessarily know ahead of time, what bugs it's going to find, or the associated cost of fixing said bug. Perhaps one suggestion would be that I add fuzz harnesses loosely in sync with you adding new features. That way there would be less accumulated 'technical debt' to wade through. The only downside of this approach is it might slow down new feature development.

> Ideally, there would be a way to allow failures for particular fuzz targets, like the round-trip fuzzing, so I am not forced to make fixes right away or else have CI broken permanently. Maybe, in an ideal world, CI here would allow roundtrip-failures (but no other failures), while the Google-run fuzzer can open issues for each problem it finds.

> What do you think?

Makes sense. I'll have a look to see if I can tag specific fuzzers as "can fail" or something. Hopefully that way it'll still warn for some bugs, but won't break the CI.

Perhaps a middle ground would be to continue round-trip fuzzing but remove the asserts. This would test that the configs parser/serializer can handle a round-trip without crashing. But doesn't enforce strict 1:1 equality of round-trip input/outputs. I could perhaps even leave the asserts but guard them behind a "strict_fuzzing" feature or something like that. Thoughts?

Otherwise I'll just close out the effort for roundtrip fuzzing, and we can reopen this when/if round tripping becomes a higher priority.

gix-config/Cargo.toml
silvergasp: Sure, arbitrary derives can be a quite viral, even when it's an optional dep, there is a non-zero of a maintenance overhead, especially when using third_party types. When implementing fuzzer's this is generally just a convenience thing, but I can definitely manually handle arbitrary types from each fuzzer, it just requires more boilerplate. I'll see what I can come up with.

gix-config/fuzz/fuzz_targets/fuzz_file_section.rs
silvergasp: Easy, I'll just start with `Metadata::default()` and go from there.

Byron:
You know, since this is getting more involved by the minute, I think it's better to focus on fuzzing more of the API rather then adding round-trip fuzzing. The reason for that is that, except for in this crate, I don't even know if it's necessary to have it anywhere.

Of course, if you *can* make it so that this CI does not fail, but the google one (the one that creates issues) does, then it would be perfect. Of course, that failure mode is special and only applies to the round-trip tests here.

If that sounds hard or too much of a hassle, it's better to not continue with `gix-config` round-trip fuzzing, I think.

silvergasp:
> You know, since this is getting more involved by the minute, I think it's better to focus on fuzzing more of the API rather then adding round-trip fuzzing. The reason for that is that, except for in this crate, I don't even know if it's necessary to have it anywhere.

No worries, I'll remove the roundtrip component of this PR and just leave it as a one-way fuzzer. But feel free to ping me if/when things change and you'd like to take another pass over it.

> Of course, if you can make it so that this CI does not fail, but the google one (the one that creates issues) does, then it would be perfect. Of course, that failure mode is special and only applies to the round-trip tests here.

Yeah it's not looking promising. Technically it's possible, but looks super complicated. I might just leave this for now.

silvergasp:
I ended up leaving it as a roundtrip fuzzer, but removed the strict assert. It doesn't seem to be crashing locally after 10min or so of fuzzing so assuming that it passes in CIFuzz, I think we should leave it as is. I've also removed the intrusive arbitrary derives.

gix-config/Cargo.toml
silvergasp: Done

Byron (approved):
Thanks a lot!

What I like about these tests is that they start with fully fuzzed inputs, like a generated configuration file, and from there validate a larger API surface. That way there is something to see.

The now removed `file_section` test for instance really only validated the validation of section and key names, while trying to interact with an empty section the rest of the time.

Thus I think these tests should be added with enough care to assure they are not no-ops, while checking that the code they test is reasonably complex.

I will merge this PR when CI is green 🎉.

Issue #703, opened 2023-01-30
Integration with google/oss-fuzz


silvergasp:
### Summary 💡

First of all thanks for all your excellent work here, I'm a huge fan of your work :) Google's [OSS-fuzz](https://github.com/google/oss-fuzz) is a free continuous fuzzing service. If you are interested in integrating this gitoxide with OSS-fuzz, the OSS-fuzz project will do the following each day;
- Build each of your fuzz tests,
- Run each of your fuzz tests on a distributed cluster,
- Report bugs back to you via email or via the private bug tracker.

I'd be happy to champion the effort for integration if you are keen on it?

### Motivation 🔦

I'd like to help with improving the stability of gix, by running all the fuzz tests using googles free cluster/fuzzing platform :)

Byron:
That sounds fantastic, please go ahead. There are already some fuzz setups, mostly for parsers, that you might be able to (re-)use. Adding more might also be a good idea, there are several parsers without a fuzzer setup. `git-index` comes to mind as a more complex one, and certainly the ones for parsing objects in `git-object` as well.

> Report bugs back to you via email or via the private bug tracker.

Feel free to report them here on GitHub issues if that's an option, otherwise my email address will work as well.

silvergasp:
> That sounds fantastic, please go ahead. There are already some fuzz setups, mostly for parsers, that you might be able to (re-)use. Adding more might also be a good idea, there are several parsers without a fuzzer setup. `git-index` comes to mind as a more complex one, and certainly the ones for parsing objects in `git-object` as well.
> 
> > Report bugs back to you via email or via the private bug tracker.
> 
> Feel free to report them here on GitHub issues if that's an option, otherwise my email address will work as well.

Not a problem. I'll configure OSS-fuzz to report to the GitHub issue tracker. The only downside to configuring OSS-fuzz in this way is that **if** a vulnerability is found in gitoxide, then it will be publicly reported. Feel free to let me know if you change your mind and I'll tweak the oss-fuzz integration.

A couple of things I'll need from you;
- An email address: [This should preferably be one associated with a google account](https://google.github.io/oss-fuzz/faq/#why-do-you-require-a-google-account-for-authentication). Your `@icloud.com` email address in the git logs will work for receiving bug reports but you won't be able to get full access to the fuzzing statistics dashboard or the private bug tracker. Also, something to keep in mind is that the email address you choose will be stored in plain text in the oss-fuzz repository.
[...]

PR #1188, opened 2023-12-17
gix-ref: Add fuzz-harness's

silvergasp:


gix-ref/fuzz/fuzz_targets/fuzz_packed_buffer.rs
```diff
@@ -0,0 +1,54 @@
...
+use libfuzzer_sys::fuzz_target;
+use std::fs;
+use std::hint::black_box;
+use std::path::PathBuf;
+
+struct FuzzedBufferPath(PathBuf);
+
+impl FuzzedBufferPath {
+    fn new() -> Self {
+        Self(PathBuf::from(format!("/tmp/fuzz_packed_buffer_{}", std::process::id())))
```
Byron, line 16: I think this really must be `tempfile::NamedTempFile` or else it will be racy (or might become racy) depending on the way the fuzzer parallelises. Right now it assumes process ids are enough and I don't know if that's always going to be true.

For safety, using clash-free tempfile seems like the better choice.

Byron (commented):
Thanks a lot!

There is just a question about the safety of relying on the process ID alone for filenames.

Further, seeing the workspace adjustment, there is a [related question](https://github.com/Byron/gitoxide/issues/703#issuecomment-1859643133) that fits in here quite well.

Thanks for your help!

gix-ref/fuzz/fuzz_targets/fuzz_packed_buffer.rs
Byron: > How do you feel about me refactoring [open](https://github.com/Byron/gitoxide/blob/0fe20e80145419e1662f869657dabf689786395f/gix-ref/src/store/packed/buffer.rs#L33) into two functions one that opens from a `&[u8]` and is memory backed and the other which maintains the same api i.e. dynamically chooses memory Mapped/InMemory backing.

A great idea! This is the way to go :).

silvergasp: Done.

gix-ref/src/store/packed/buffer.rs
```diff
@@ -84,6 +68,35 @@ pub mod open {
...
+            };
+            Self::open_with_backing(backing, path)
+        }
+
+        /// Open a buffer from existing memory.
+        ///
+        /// In order to allow fast lookups and optimizations, the contents of the packed refs must be sorted.
+        /// If that's not the case, they will be sorted on the fly with the data being written into a memory buffer.
+        #[cfg(fuzzing)]
+        pub fn open_from_memory(bytes: &[u8]) -> Result<Self, Error> {
```
Byron, line 96: As this adds a feature, could you put this change into a separate commit and prefix the subject line with `feat:` ? The subject itself could say which method was added, with the body being a short description of what it does. 

That automatically goes into the changelog.

Byron (changes requested):
Thanks a lot! 

As I am asking for a commit to be inserted, could you also avoid having `gix-ref:` prefixes? They should be removed without replacement, as it would cause trouble with the changelog.

Thanks again.

gix-ref/src/store/packed/buffer.rs
silvergasp: Done

silvergasp:
> They should be removed without replacement, as it would cause trouble with the changelog.

Oops I misread this. And replaced the prefixes with 'test'. If that's going to cause problems I'll completely remove the prefixes.

Byron (approved):
Thanks a lot!

I have made a few adjustments myself, with one of them being https://github.com/Byron/gitoxide/pull/1188/commits/5828f3796031de8d9f50bd6536697754da0b3d8c that you might want to have a look at in case there a good reason to keep using Cargo.lock files for fuzzers as well.

The problem I see with these lock files is that I will never update them, so they are bound to use outdated and possibly broken dependencies of crates that can also cause failures. Maybe I am missing something though.

PR #1189, opened 2023-12-18
gix-attributes: Add a search fuzzer

silvergasp:


Byron (approved):
Thanks a lot! I love that this check also finds a big immediately.

I will merge once CI is green.

Byron:
As a side-note: the fuzzer (impressively) quickly finds a glob pattern that runs `gix-glob` for non-linear time, which interestingly is a 1:1 conversion from the respective function in C. This means we have now found a shortcoming of that algorithm, and maybe even a shortcoming in all such general glob engines.

Screenshot 2023-12-18 at 10 36 48

Let's see how long it takes until it decides the runtime is too long so it provides the glob pattern that caused this, so one can attempt a fix somehow.

PR #1191, opened 2023-12-18
Remove reset of `match_ceiling_dir_or_error` in `apply_environment()`

kdarkhan:
In https://github.com/Byron/gitoxide/pull/721, this reset was added which makes using
[ThreadSafeRepository::discover_with_environment_overrides_opts]( https://github.com/Byron/gitoxide/blob/f7adf97bc6aa2bf82f906c4463b1cbb3f9cddae3/gix/src/discover.rs#L67) inconvenient.

If the old behaviour of `match_ceiling_dir_or_error = false` desired, then it can be achieved by
setting it on `options` just before calling `apply_environment()`.

The inverse is not true, and `ThreadSafeRepository::discover_with_environment_overrides_opts`
cannot enable this flag because `apply_environment` overrides it.

I am not sure if this change is the best approach to accomplish what I need.
Let me know if you think it is better to change `discover_with_environment_overrides_opts`
so that `options` parameter is used as override instead of default.

### DIFF
--- a/gix-discover/src/upwards/types.rs
+++ b/gix-discover/src/upwards/types.rs
@@ -87,16 +87,12 @@ impl Options<'_> {
     ///
     /// Note that `GIT_DISCOVERY_ACROSS_FILESYSTEM` for `cross_fs` is **not** read,
     /// as it requires parsing of `git-config` style boolean values.
-    ///
-    /// In addition, this function disables `match_ceiling_dir_or_error` to allow
-    /// discovery if an outside environment variable sets non-matching ceiling directories.
     // TODO: test
     pub fn apply_environment(mut self) -> Self {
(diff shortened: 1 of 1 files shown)


Byron:
Oh, that's interesting! It seems then you will want to talk to @davidkna to see how it remains configurable instead of hardcoding the same in `starship` with the patch above. *(But maybe I am misunderstanding something)*

kdarkhan:
Right, I will reach out to him. I am planning to add a new config option in `starship` if they accept.

The reason I want to have the `err` case is because I want to prevent `starship` git discover logic execute on a slower network drive. When `match_ceiling_dir_or_error = true`, it will accomplish the behavior I need. I can make this behavior non-default to not break other people's workflows.

PR #1192, opened 2023-12-19
test: Add fuzz_commit seed corpus builder

silvergasp:


silvergasp:
This corresponds to an oss fuzz change https://github.com/google/oss-fuzz/pull/11386

If you are not familiar with the terminology a seed corpus is a set of known good inputs that gives the fuzzer a good starting point. It'll then mutate the inputs from the seed corpus to maximize code coverage.

Byron:
Thanks a lot!

I presume it works since the upstream PR was merged, but it's always a bit hard for me to tell as a lot is going on in the logs.
Merging, also with the hope more fuzz tests can be equipped with these starting points.

silvergasp:
Yeah I test this all out locally. But you can see it working here. https://github.com/Byron/gitoxide/actions/runs/7279035543/job/19834505560#step:4:510

I'll be following up with more corpus builders soon.

PR #1193, opened 2023-12-20
Snap created and published

kz6fittycent:
I have created and published a working [snap](https://snapcraft.io/gitoxide) for `gitoxide`. I wanted to provide this PR as a means to offer a couple of options for you and your team:

1. I can continue to manage the snap for the project, as a form of contribution.
2. I can work with you to transfer the snap to you fully and you can manage it yourself.

Option 2 would require creating an account in the Snapcraft [forum](https://forum.snapcraft.io/), and we can start a thread to perform that transfer. Typically a fast process. 

Not sure what a snap is? Check out this [link](https://snapcraft.io/about) to learn more. The TL;DR is that it provides a "universally" installable package for applications in Linux. Updates are automatic and automated, based on commits to the upstream repo(s).


Byron (changes requested):
Thanks a lot for making me aware and for making `gitoxide` available in the Snap ecosystem!

The way I see it, it's a package and thus could remain in your custody just like deb/rpm packages in some distributions.

This also means that this PR would have to be reduced to the dedicated Snap installation instructions in the README, if you are OK with it.

Lastly, I recommend removing the logo because the Rust logo is technically trademarked and it looks like it's now associated with `gitoxide`. Having no logo seems to be preferable to me.

Thanks for your consideration.

PR #1194, opened 2023-12-20
fix: use correct fields for ctime and mtime on AIX

ecnelises:
On AIX, ctime and mtime are structs containing seconds and nanoseconds.

Reference: https://github.com/rust-lang/libc/blob/f05d6a35b25288b143241ebb6b4b071618945431/src/unix/aix/powerpc64.rs#L63

### DIFF
--- a/gix-index/src/fs.rs
+++ b/gix-index/src/fs.rs
@@ -55,11 +55,16 @@ impl Metadata {
         #[cfg(not(windows))]
         {
             Some(system_time_from_secs_nanos(
+                #[cfg(not(target_os = "aix"))]
                 self.0.st_mtime.try_into().ok()?,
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot!

When looking at these seemingly arbitrarily renamed field names over three operating systems, I feel like they are a cruel joke introduced for the sake of incompatibility 😅.

ecnelises:
Thanks. Maybe we can reuse code from rustix or std? I remember there's similar code.

Byron:
It would certainly be wonderful to be able to 'inherit' all the compatibility layers they must have put in already. Right now each time there is a change here, there is also a risk to break another platform, of which none is currently tested in CI.

Maybe that could be an additional approach - know the platforms to support and try to build `gix` on them.

PR #1196, opened 2023-12-20
fix gix-config stack-overflow (found by fuzzer)

Byron:
Instead of recursion, we now use a simple loop.


Byron:
@silvergasp I had to disable CI fuzzing as it [now fails](https://github.com/Byron/gitoxide/actions/runs/7276393627/job/19826172487?pr=1196#step:4:666) - probably related to the new corpus settings. I also validated that the removed Cargo.lock files aren't the problem (by reverting the respective commit temporarily).

PR #1205, opened 2023-12-23
Fuzz more of existing api's

silvergasp:


gix-config/fuzz/fuzz_targets/fuzz_file_corpus_builder.sh
```diff
@@ -0,0 +1,23 @@
...
+ROOT=$1
+OUTPUT_CORPUS=$2
+FIXTURES_DIR=$(readlink -f $ROOT/gix-config/tests/fixtures)
+
+echo $ROOT
+echo $FIXTURES_DIR
+find $FIXTURES_DIR -name "*.config" -exec zip -j $OUTPUT_CORPUS {} \;
+
+# Generate configs.
+REPO=/tmp/config_repo
```
Byron, line 16: Would you be able to use `mktemp` or the likes to get a temp directory that is more unique? I can imagine this to work on CI where the state is fresh, but would expect this not to be idempotent otherwise.

gix-config/fuzz/fuzz_targets/fuzz_file.rs
```diff
@@ -1,32 +1,84 @@
...
-fuzz_target!(|ctx: Ctx| {
+fn fuzz(ctx: Ctx) -> Result<()> {
     let meta = Metadata::default();
     let options = Options::default();
-    let file = unwrap_or_return!(File::from_bytes_no_includes(&ctx.input, meta.clone(), options.clone()));
+    let file = File::from_bytes_no_includes(&ctx.input, meta.clone(), options.clone())?;
+
+    let mut triples = ctx.section_subsection_key_triples.iter();
+
+    let (section_name, subsection_name, key) = triples.next().ok_or(Catchall)?;
```
Byron, line 40: Would it not be better to assure the whole API surface is always tested in full? Right now there could be too few keys to get them all. Further, there could be too many keys that are left unused (despite having cost power to generate).
Lastly, I fear that most of this code only encounters the 'key not found' case, unless the fuzzer truly is smart enough to figure out the relationship between `input` and `section_subsection_key_triples`.
From that point of view, it might be easier for the fuzzer to only use one well-known key and let it discover that one in `input` - but it's just an uneducated guess which is probably wrong.

Byron (commented):
Thanks a lot!

A few questions came up and I am very interested to hear what you think. Thanks again!

gix-config/fuzz/fuzz_targets/fuzz_file_corpus_builder.sh
silvergasp: I'm admit I'm not familiar with mktemp. But I'll look into it.

gix-config/fuzz/fuzz_targets/fuzz_file.rs
silvergasp: > Would it not be better to assure the whole API surface is always tested in full?

I'm not sure I understand your question. Could you expand on what you mean by this.

> Right now there could be too few keys to get them all.

Yeah, this is a potential problem, and "The fuzzer won't generate enough keys" is a pretty reasonable hypothesis. My general approach to problems like this is to just fuzz as much stuff as possible, with a fuzz harness that is as simple as possible. Let it run for a few hours, see what actually happens and then optimise it as necessary.

> Further, there could be too many keys that are left unused (despite having cost power to generate).

This is less of an issue than too little data. The input data will get longer incrementally via mutation i.e. libfuzzer is  a genetic optimiser (so when I say "mutation" I mean that the fuzzer is modelled on evolution, and mutation represents a new generation) where the goal state is 100% code coverage. So if libfuzzer mutates an input to create a new longer input, but then that doesn't result in an increase in code-coverage that new mutated input will just die off and not keep growing. Happy to expand on this if you are interested.

> Lastly, I fear that most of this code only encounters the 'key not found' case, unless the fuzzer truly is smart enough to figure out the relationship between input and section_subsection_key_triples.
From that point of view, it might be easier for the fuzzer to only use one well-known key and let it discover that one in input - but it's just an uneducated guess which is probably wrong.

I would say that's a pretty reasonable fear to have, and a good hypothesis. I would say that the fuzzer IS smart enough to work out an input that will trigger correctly. However, I don't know how long it'll take the fuzzer to do that. It could be 2min or it could be 10,000yrs, really hard to know without trying it. As mentioned I typically like to give the fuzzer at least a few hours and then evaluate the code coverage after that. My thought was just to run this on oss-fuzz over night, but if you prefer I can set this up on my laptop overnight before merging.

There are a few approaches to give the fuzzer a helping hand.
1. You've mentioned we could just hard code the key. This is a good option, and is simple enough. It might just miss some things e.g. what happens if I have a key like `"foo[a]bar"`. I'm sure you would have more an intuition if bugs triggered by the key are an actual problem, I'm certainly going to fumble around a bit more while I wrap my head around the gigantic amount of work that you've done.
2. We could use a [libfuzzer dictionary](https://www.llvm.org/docs/LibFuzzer.html#id32) to specify a set of key->config pairs. For example we could give the fuzzer a hint that the strings `"core"` and `"[core]"` are important strings to treat as singular tokens when mutating. e.g. our dictionary might look something like the following;

```
# String key
"core"
# Config key
"[core]"
```

Option 2. gives the fuzzer more flexibility in how it mutates data (i.e. it can randomly insert a dictionary entry anywhere in the fuzzed input). But at the cost of additional complexity. Complexity that depending on your intuition may not be justified.

silvergasp: Actually re-reading your comment. I might be able to do a hybrid of the two e.g.
```rust
    // This;
    let (section_name, subsection_name, key) = triples.next().ok_or(Catchall)?;
    // Could be something like this;
    let (section_name, subsection_name, key) = triples.next().unwrap_or(("section", "subsection", "key"))?;
```

That way if the fuzzer "chooses" it can either fuzz any of these fields OR it can just create an empty vector, or one that is too small. I'm pretty certain that the fuzzer would have an easier time optimizing for coverage that way.

silvergasp: I've opted for the ` let (section_name, subsection_name, key) = triples.next().unwrap_or(("section", "subsection", "key"));` approach let me know what your thoughts are, I'm certainly not against your approach of hard-coding all of these.

Byron: Thanks so much for the elaborate reply - I definitely learned something while reading it!

Understanding your approach better also helps me for future reviews, and I hope this will save you time. Generally, I am OK to leave all that to you as you have it under control, even though I'd also be open to learn everything about your workflow and apply it myself. For that matter I'd definitely be interested in your top-5 or top-10 fuzzer-related actions you apply. For instance, you seem to evaluate effectiveness using coverage reports, and I don't think I have ever done that. If you'd consider to share the top-level invocations with me, that would definitely be appreciated - I can take it from there then to not take more of your time.

Regarding keys and what's valid, I'd certainly use *some* knowledge of the code to optimize the fuzzer a little. I.e. it's OK to know that `Key` is always used to parse input keys into `section[.subsection].key`, thus one can fuzz it as primitive to assure it works and focus higher-level APIs on all the other code they run and feed them mostly keys that are supposed to exist. However, this is just what I would do in an attempt to 'optimize', without knowing if that actually works *or* is a good idea 😁.

Byron:
Thanks so much for the fixes, all is great now!

gix-config/fuzz/fuzz_targets/fuzz_file.rs
silvergasp: Just as a bit of a follow up on your query. After running on oss-fuzz for a few days now we have some real world results to compare our hypotheses. It looks like the fuzzer is finding valid inputs that work with the above fuzzer. But in some cases the number of valid inputs is eclipsed by the number of invalid inputs. Looking at [the coverage for `raw_value_filter_inner`](https://storage.googleapis.com/oss-fuzz-coverage/gitoxide/reports/20231227/linux/src/gitoxide/gix-config/src/file/access/raw.rs.html#L60) we can see that the fuzzer has only found 60 inputs out of 31.7k that meet the ("section", "subsection", "key") triple. I wouldn't say that this is a great result, but it's not terrible either. In an ideal scenario we'd want the ratio between Err/Ok to be roughly 1:1, or at least in single digit percentages. I'm inclined to just let this run in the background and check back in a week or so to see if we are getting a higher count on the call buckets.

> Understanding your approach better also helps me for future reviews, and I hope this will save you time. Generally, I am OK to leave all that to you as you have it under control, even though I'd also be open to learn everything about your workflow and apply it myself.

Of course, I don't mind sharing so feel free to ask as many questions as you'd like.

> For that matter I'd definitely be interested in your top-5 or top-10 fuzzer-related actions you apply. For instance, you seem to evaluate effectiveness using coverage reports, and I don't think I have ever done that. If you'd consider to share the top-level invocations with me, that would definitely be appreciated - I can take it from there then to not take more of your time.

Yeah so for some context the reason why it's important to use code coverage as an evaluation tool is that it's often super easy to write a fuzzer that looks great but is basically useless. e.g. 

```rust
fn my_png_parser(png_file: &[u32]) -> Result<Vec<Vec<u8>>,  String> {
    if png_file.len() < MIN_PNG_SIZE {
        Err("Corrupt image")
    }
   // Check magic number that identifies this file as a PNG.
   if png_file[0] == 0x718398402 {
     Err("Corrupt image")
   }
   // ... The rest of your parser
}

fuzz_target!(|png_file: &[u32]| {my_png_parser(png_file); });
```

In this case there is a super small chance (i.e. 1/2^32) that the fuzzing engine will guess the correct magic number at index 0. So the input will almost always cause an error. So I'll typically run the fuzzer for a while and then evaluate the coverage. It's usually reasonably obvious if the fuzzer is being blocked by something like this. There are a couple of strategies that can be used here;
- Provide a seed corpus with inputs that will pass the validation checks.
- Provide a dictionary to help the fuzzer work out how to get past the validation checks.

To collect code coverage locally I'd typically run the fuzzer typically forking with however many cores I have e.g.
```bash
# If you don't need to use your computer for anything else run;
$ cargo +nightly fuzz run my_fuzzer -- -fork=$(nproc)
# If you still want to be able to use a web browser or anything else on your PC run with `nproc` minus one e.g.
$ nproc
8
$ cargo +nightly fuzz run my_fuzzer -- -fork=7
```

You'll get some basic coverage information on the fly when you run a fuzzer e.g.

```
#86358: cov: 13058 ft: 11294 corp: 1993 exec/s 4459 oom/timeout/crash: 0/0/0 time: 7s job: 5 dft_time: 0
```

`cov`: Is a representation of how many code blocks are executed. Worth noting that `blocks!=lines`

`ft`: Is a representation of other intsrumented features e.g. these can be different values in control flow statement i.e. if/match/loop etc.

When both `cov` and `ft` are stalled and not increasing for more than an hour or two it's a pretty safe bet that your fuzzer is stuck somewhere and/or it's reached everything that is possible to reach. To find out where it's stuck you'll need full source based code coverage.

This isn't strictly necessary but when measuring code coverage with `cargo fuzz` I'd strongly recommend installing from git e.g. `cargo install --git https://github.com/rust-fuzz/cargo-fuzz`. [I made some changes upstream](https://github.com/rust-fuzz/cargo-fuzz/pull/336) as the original code-coverage collection (in the latest release) was unbearably slow.

To collect code coverage based on your generated corpus (generated when you `cargo fuzz run`), you simply repeat the command changing run->coverage, e.g. `cargo +nightly fuzz coverage parse` this will rebuild the fuzzer with different instrumentation needed to collect source based code coverage. It'll print out a path to a `.profdata` file, you'll need to use [llvm tooling](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html) and the original coverage instrumented elf file to get human readable code coverage information. Unfortunately you'll need to exact version of llvm that rustc uses. This is a bit of a pain e.g.
```bash
$ # Install llvm tooling
$  rustup component add --toolchain nightly llvm-tools
$ # find where it installed llvm-cov
$ find ~/.rustup -name llvm-cov
/home/nathaniel/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/bin/llvm-cov
$ # Store that path for conveniance (it won't be in your PATH but you can add it if you want)
$ export LLVM_COV=$(find ~/.rustup -name llvm-cov)
$ # Use llvm cov to generate a coverage report
$ $LLVM_COV report ./target/x86_64-unknown-linux-gnu/coverage/x86_64-unknown-linux-gnu/release/parse -instr-profile=/home/nathaniel/projects/github.com/silvergasp/gitoxide/gix-url/fuzz/coverage/parse/coverage.profdata
```
There are other formats e.g. html it you want something a bit more browsable. You can read about that [here](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html).

I've been meaning to open a PR upstream with the cargo-fuzz project to make this process a bit easier/update the docs, but haven't got around to it yet.

Byron: Thanks so much for sharing - I learned a lot! I found the explanations for `cov` and `ft` particularly interesting - I had no idea how to interpret these

For starters, I installed the latest upstream of `cargo-fuzz` directly from git along with the matching LLVM tools. Even though it is quite involved, I think this description will be all people need to get started with coverage-fuzzing!

Thus I hope you will eventually get around to upstreaming a version of what you wrote here for great effect. Please do feel free to CC me on the PR when that happens as I have a feeling I could 'user-test' it so one can reasonably assume that if it works for me as non-expert, it will work for others as well.

---- 

> I wouldn't say that this is a great result, but it's not terrible either. In an ideal scenario we'd want the ratio between Err/Ok to be roughly 1:1, or at least in single digit percentages. I'm inclined to just let this run in the background and check back in a week or so to see if we are getting a higher count on the call buckets.

It's definitely interesting to see that it manages to run into these branches, but apparently it doesn't necessarily make use of that information. For 'guided evolution' I would have expected it to be aware that it's seeing the same code over and over again, but in the moment it finally finds a key, it should definitely 'focus' on that as it now sees code that it never or rarely saw. Maybe it's even able to do that, but it's exponentially harder if multiple generated 'items' of input are involved. Maybe that already shows that it needs more guidance to more effectively hit the interesting parts.

In any case, I am very thankful that you are running the fuzzers and take the time to analyse coverage reports, as otherwise complex fuzzer setups seem to have the potential to be quite wasteful, as in wasting power but also not finding problems it otherwise could have found. Of course, compute time is cheaper than human time and that should be a driving factor as well, so all I am saying is that I hope there is a simple set of 'rules' that we can follow here that give a good ratio of hitting the interesting code-paths and the complexity/simplicity of the setup.

PR #1206, opened 2023-12-23
Give the CI fuzzer more time

silvergasp:
Currently the CI fuzzer isn't on the critical path to completion. So we can extend the amount of fuzzing time without impacting the critical path.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -15,7 +15,7 @@ jobs:
      with:
        oss-fuzz-project-name: 'gitoxide'
        language: rust
-       fuzz-seconds: 600
+       fuzz-seconds: 840
    - name: Upload Crash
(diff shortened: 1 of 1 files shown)


silvergasp:
For context the "ci / test (pull_request)" seems to be the critical path.

silvergasp:
Ah, it looks like I increased the time too much. I'll drop 40s off this tomorrow.

Byron:
Thanks for the proposition. However, seeing the fuzzing stage to now dominate the runtime, I'd like to not pursue even if the runtime was reduced by a minute.

Instead, I'd like to see the current difference between CI-test and CI-fuzz as safety-buffer. The reason for not wanting to maximise fuzz time is the unpredictability of CI runtimes in general, while experience shows that the fuzzer finds obvious errors even within a very short time, while taking much longer than the time we have for follow-ups.

PR #1208, opened 2023-12-24
Add gix-commitgraph::File fuzzer

silvergasp:


Byron (approved):
Thanks a lot for increasing coverage!

I took a look to see if one could avoid the tempfile, and it seems like it's possible to use an anonymous mapping which could be provided using an alternative contructor. I will give that a shot and merge when it works.

PR #1209, opened 2023-12-24
Fuzz more of gix_url::Url

silvergasp:


Byron (approved):
Thanks again for this initiative!

I like the style of placing as much as possible outside of the macro by the way, remembering `gix-config` fuzzing where that's not the case (yet).

PR #1217, opened 2023-12-30
Refactor gix-config fuzzer

silvergasp:
- Add mutable API's to fuzzer.
- Reading key's first and then using them to drive everying. this should improve fuzzing performance significantly.


Byron (approved):
Using the actually present keys for testing the API is a great idea! Loving it :). Thanks so much for the continued improvements!

----

On an unrelated note, I saw that Git also has an OSS-Fuzz configuration by now, and they are testing date parsing. I will keep an eye out particularly when the fuzzer finds something, in the hopes that these test cases can also be adopted here for additional validation.

PR #1220, opened 2023-12-30
Fuzz more of mutable gix_config::File API

silvergasp:


silvergasp:
Some additional context on the dictionary commit;
- oss-fuzz https://github.com/google/oss-fuzz/pull/11424
- Libfuzzer docs https://www.llvm.org/docs/LibFuzzer.html#dictionaries

silvergasp:
I was expecting some minor improvements with the dictionary but it turns out that it's a fairly drastic improvement.

Byron:
Thanks a lot!

I see - dict's are like the sibling of a corpus archive, and maybe both can even be combined. It would be interesting if `gix-object` could benefit from it as well even though I don't see how right now. Maybe having more corpus archives will already help, with file format parsers for instance.

This reminds me, `gix-pathspec` might most easily benefit from dicts given its keywords and all.

silvergasp:
> I see - dict's are like the sibling of a corpus archive, and maybe both can even be combined. It would be interesting if gix-object could benefit from it as well even though I don't see how right now. Maybe having more corpus archives will already help, with file format parsers for instance.

Yeah the fuzzer will use both the dicts and the seed-corpus. The seed corpus is run at startup to collect baseline coverage with known good inputs. Then the seed corpus is mutated using the dictionary mutator (in addition to other mutator operations), to produce new inputs.

> This reminds me, gix-pathspec might most easily benefit from dicts given its keywords and all.

No worries I'll look into it.

PR #1221, opened 2023-12-30
Fuzz more of pathspec patterns

silvergasp:
- Fuzz more of the API
- Add associated dictionary


Byron (approved):
Thanks a million for making the fuzzer more efficient!

PR #1230, opened 2024-01-03
make ItemSliceSync::get_mut() check that the index is in range

martinvonz:
This removes some of the unsafety from
ItemSliceSync::get_mut(). There's still the unsafety that the caller needs to make sure a single index is not used concurrently.


Byron:
Thanks a lot!

Since this is an internal usage at some point the code has to 'trust itself' to indeed be correct. However, I have left the added check for use with debug assertions compiled in, so if there was a bug related to slice access now or in future it would be caught by this addition.

Byron:
Sorry, I forgot to push my local merge earlier and now had to rebase to get it in. GH doesn't detect that, thus I have to close the PR.

PR #1232, opened 2024-01-03
RFC: remove "Copyright Sebastian Thiel ..." from LICENSE files

martinvonz:
The "Copyright Sebastian Thiel ..." line in the LICENSE files confuse our tooling and reviewers at Google, making it take longer than necessary to import new versions of Gitoxide into our monorepo. IIUC, the copyright is supposed to go in each source code file instead. Maybe it will become necessary to move it out of the LICENSE file as soon as the project decides to copy/vendor some code from somewhere else anwyay, because then we probably can't claim copyright for the entire repo? I'm not a lawyer and far from an export on this.

If we remove the copyright line from the LICENSE files, does that mean that we should add them in each source file? I think I've heard that that's not necessary because the information about who wrote the code is available in the commit log..

### DIFF
--- a/LICENSE-APACHE
+++ b/LICENSE-APACHE
@@ -176,8 +176,6 @@
 
    END OF TERMS AND CONDITIONS
 
-   Copyright 2018-2021 Sebastian Thiel, and [contributors](https://github.com/byron/gitoxide/contributors)
-
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
    You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


Byron:
I am happy to help with this, I probably copied the license file from somewhere as well.

When it comes down to trying the copyright, from all I know one will look at detailed line-by-line authorship anyway and indeed use Git for that. And since each source tarball can always be traced back to its source commit or tag, there should be no need to mention names anywhere either.

martinvonz:
Thanks!

PR #1235, opened 2024-01-05
greatly reduce worst-case gix-config memory consumption

Byron:
- fix!: greatly reduce peak memory usage when reading files.
- add test for memory usage in worse-case scenario


Byron:
Closing - it was merged but with changed commit messages.

PR #1236, opened 2024-01-05
Config access uses `impl Key` to allow type-safe keys

bittrance:
This PR changes the API of `gix_config::File` to use a trait (gix_config::Key) as key when requesting or setting configuration. This will allow `gix` to use its tree of structured and validating typed keys (`gix::config::tree::sections::*`) as keys directly, and also means that consumers of the `gix` crate to use type-safe config access. gix-config also provides an implementation for &str and &bstr so that constant strings can be used as keys. This is part of the work addressing #1125.

The PR is draft, as there remains various question marks, see specific comments in code below.

gix contains a lot of strings and `format!()` operations that could be changed to use type-safe keys, but that is left for a separate PR.

### Tasks

* [x] initial refactor
* [x] get repository tests to work and remove TODOs around mutations (see `as_config_key()`)
* [x] call `key` parameters `value_name` instead to be in sync with `KeyRef::value_name`, also see `keys()` methods
* [x] proper split into commits with breaking-change signalling and documentation of breaking changes


gix-config/src/file/access/raw.rs
```diff
@@ -434,43 +401,35 @@ impl<'event> File<'event> {
...
-    ) -> Result<Option<Cow<'event, BStr>>, crate::file::set_raw_value::Error>
-    where
-        Key: TryInto<section::Key<'event>, Error = E>,
-        section::key::Error: From<E>,
-    {
-        let mut section = self.section_mut_or_create_new_filter(section_name, subsection_name, filter)?;
-        Ok(section.set(key.try_into().map_err(section::key::Error::from)?, new_value.into()))
+    ) -> Result<Option<Cow<'event, BStr>>, crate::file::set_raw_value::Error> {
+        let mut section = self.section_mut_or_create_new_filter(key.section_name(), key.subsection_name(), filter)?;
+        let key = key.name().to_owned().try_into().map_err(section::key::Error::from)?; // TODO Borrow, plz
```
bittrance, line 431: This `to_owned()` is not optimal, but it is not obvious to me how to formulate the required lifetimes.

gitoxide-core/src/repository/config.rs
```diff
@@ -57,12 +57,12 @@ struct Filter {
...
 impl Filter {
     fn new(input: BString) -> Self {
-        match gix::config::parse::key(<_ as AsRef<BStr>>::as_ref(&input)) {
-            Some(key) => Filter {
-                name: key.section_name.into(),
-                subsection: key.subsection_name.map(ToOwned::to_owned),
+        match ((&input).section_name(), (&input).subsection_name()) {
+            (section, subsection) if !section.is_empty() => Filter {
+                name: section.into(),
+                subsection: subsection.map(ToOwned::to_owned),
```
bittrance, line 63: I'm not sure exactly how the filter is used, so I'm not 100% sure this change retains the exact behavior of the filter in all corner cases.

gix-config/src/file/access/comfort.rs
```diff
@@ -2,48 +2,30 @@ use std::{borrow::Cow, convert::TryFrom};
...
     ///
     /// As strings perform no conversions, this will never fail.
-    pub fn string(
-        &self,
-        section_name: impl AsRef<str>,
-        subsection_name: Option<&BStr>,
-        key: impl AsRef<str>,
-    ) -> Option<Cow<'_, BStr>> {
-        self.string_filter(section_name, subsection_name, key, &mut |_| true)
+    pub fn string<'a>(&self, key: impl Key<'a>) -> Option<Cow<'_, BStr>> {
```
bittrance, line 12: Should we rename these lifetimes to `'lookup'?

gix-config/src/traits.rs
```diff
@@ -0,0 +1,11 @@
+use bstr::BStr;
+
+/// Parsed elements of a Git config key, like `remote.origin.url` or `core.bare`.
+pub trait Key<'a> {
+    /// The name of the section key, like `url` in `remote.origin.url`.
+    fn name(&self) -> &'a str;
```
bittrance, line 6: Is `name()` to generic given that we implement it on `&[b]str`? `.key_name()`  might be better?

gix-config/src/file/access/raw.rs
```diff
@@ -85,25 +74,21 @@ impl<'event> File<'event> {
...
-        key: &'lookup str,
+        key: impl Key<'lookup>,
         filter: &mut MetadataFilter,
     ) -> Result<ValueMut<'_, 'lookup, 'event>, lookup::existing::Error> {
         let mut section_ids = self
-            .section_ids_by_name_and_subname(section_name, subsection_name)?
+            .section_ids_by_name_and_subname(key.section_name(), key.subsection_name())?
             .rev();
-        let key = section::Key(Cow::<BStr>::Borrowed(key.into()));
+        let key = section::Key(Cow::Borrowed(key.name().into())); // TODO Borrow, plz
```
bittrance, line 91: Oops, this TODO is already addressed.

Byron (changes requested):
Thanks for sharing your amazing progress!

Before I can look at anything in detail, I'd need the commits split up so that…

* …the `Key` related changes in `gix-config` are in one commit, whose subject is prefixed with `fix!: ` or `feat!: `. Please write the subject and body so they can go into the changelog.
* …all other changes to adapt to this change can go into its own commit without any special message requirements.

There is also this one big request of only modifying the `string_by_key` based access with the new trait. After all, the natural way of accessing the API is by passing triplets of section, subsection and 'key' (definitely open to a better term here).

That way, only `string_by_key()` has to change at all and that change would probably be compatible to most existing uses depending on the available implementations of `Key` for `&BStr` for example. I don't think there is a need to break the world here, and hope this could even work without making any breaking change.

Is that feasible at all?

bittrance:
> …the Key related changes in gix-config are in one commit, whose subject is prefixed with fix!:  or feat!: . Please write the subject and body so they can go into the changelog.

Is the purpose simply to have a `feat!` commit in the history? If so, I'd just split out adding the Key trait into a separate commit, no problem. The current commit looks like it does because I'm always striving for every commit to be viable (i.e. buildable and git-bisect:able) and still be meaningful.

> There is also this one big request of only modifying the `string_by_key` based access with the new trait. After all, the natural way of accessing the API is by passing triplets of section, subsection and 'key' (definitely open to a better term here). That way, only `string_by_key()` has to change at all and that change would probably be compatible to most existing uses depending on the available implementations of `Key` for `&BStr` for example. I don't think there is a need to break the world here, and hope this could even work without making any breaking change.

I was thinking we should keep the API surface small by removing all the `_by_key` method so that only the `impl Key` versions remained (I prepared  https://github.com/bittrance/gitoxide/commit/c54bb3addfa9e8a2fbcd0fad82bf1f1b3fe62d28 for this purpose). My thought is that `gix` config API should encourage using type-safe access and I think `config.string(format!("foo.{submod}.bar"))` is more readable than `config.string("foo", submod, "bar")`, but I defer to your judgement on when it is worth making a breaking API change.

Byron:
> Is the purpose simply to have a `feat!` commit in the history? If so, I'd just split out adding the Key trait into a separate commit, no problem. The current commit looks like it does because I'm always striving for every commit to be viable (i.e. buildable and git-bisect:able) and still be meaningful.

No, the purpose is to have a changelog that applies to the `gix-config` crate exclusively - otherwise `cargo smart-release` would think there are breaking changes in all of the crates whose code was adjusted, and it will claim new features in their changelogs, too.

That takes priority over each commit being green on CI, it's all trade-offs.

> I was thinking we should keep the API surface small by removing all the `_by_key` method so that only the `impl Key` versions remained (I prepared [bittrance@c54bb3a](https://github.com/bittrance/gitoxide/commit/c54bb3addfa9e8a2fbcd0fad82bf1f1b3fe62d28) for this purpose). My thought is that `gix` config API should encourage using type-safe access and I think `config.string(format!("foo.{submod}.bar"))` is more readable than `config.string("foo", submod, "bar")`, but I defer to your judgement on when it is worth making a breaking API change.

Thanks for sharing your thoughts.

For a plumbing crate, it would already be more than fine. However, it's exposed in `gix` and thus it needs to provide more support for a nicer API, but that shouldn't supersede the 'API that is natural for the data', i.e. triplets as it's without overhead.

But since you would be touching all of these methods anyway, let's encourage the use of the potentially typesafe and nicer API by making their methods the shorter ones. This means, for example, to keep `File::string(key: impl Key)` as is, but have a longer method that takes the triplet, like `File::string_by(section, subsection, key)`. The `_by` suffix is an example but something that seems to work nicely, and it seems better than my initial thought `_by_triplet()` which is punishingly long.
Please do feel free to experiment (and bikeshed) with the suffix though.
Finally, this also means that `string()` would call `string_by()`.

I hope all that makes sense, please do let me know about any concern you might have as well.

gix-config/src/file/access/comfort.rs
```diff
@@ -2,29 +2,29 @@ use std::{borrow::Cow, convert::TryFrom};
...
     /// As strings perform no conversions, this will never fail.
-    pub fn string(
+    pub fn string_by(
         &self,
         section_name: impl AsRef<str>,
         subsection_name: Option<&BStr>,
         key: impl AsRef<str>,
     ) -> Option<Cow<'_, BStr>> {
-        self.string_filter(section_name, subsection_name, key, &mut |_| true)
+        self.string_filter_by(section_name, subsection_name, key, &mut |_| true)
```
NobodyXu, line 18: To reduce unnecessary monomorphprisation of generics:

```suggestion
        self.string_filter_by(section_name.as_ref(),  subsection_name, key.as_ref(), &mut |_| true)
```

gix-config/src/file/access/comfort.rs
```diff
@@ -52,18 +47,18 @@ impl<'event> File<'event> {
...
     /// As paths perform no conversions, this will never fail.
-    pub fn path(
+    pub fn path_by(
         &self,
         section_name: impl AsRef<str>,
         subsection_name: Option<&BStr>,
         key: impl AsRef<str>,
     ) -> Option<crate::Path<'_>> {
-        self.path_filter(section_name, subsection_name, key, &mut |_| true)
+        self.path_filter_by(section_name, subsection_name, key, &mut |_| true)
```
NobodyXu, line 56:
```suggestion
        self.path_filter_by(section_name.as_ref(), subsection_name, key.as_ref(), &mut |_| true)
```

gix-config/src/file/access/comfort.rs
```diff
@@ -84,33 +79,28 @@ impl<'event> File<'event> {
...
     /// Like [`value()`][File::value()], but returning `None` if the boolean value wasn't found.
-    pub fn boolean(
+    pub fn boolean_by(
         &self,
         section_name: impl AsRef<str>,
         subsection_name: Option<&BStr>,
         key: impl AsRef<str>,
     ) -> Option<Result<bool, value::Error>> {
-        self.boolean_filter(section_name, subsection_name, key, &mut |_| true)
+        self.boolean_filter_by(section_name, subsection_name, key, &mut |_| true)
```
NobodyXu, line 94:
```suggestion
        self.boolean_filter_by(section_name.as_ref(), subsection_name, key.as_ref(), &mut |_| true)
```

NobodyXu (changes requested):

gix-config/src/file/access/comfort.rs
bittrance: Fixed.

gix-config/src/file/access/comfort.rs
bittrance: Fixed.

gix-config/src/file/access/comfort.rs
bittrance: Fixed.

gix-config/src/impls.rs
```diff
@@ -0,0 +1,87 @@
+use bstr::{BStr, BString, ByteSlice};
+
+use crate::traits::Key;
+
+impl Key for &String {
```
Byron, line 5: Now that I see it, the way `Key` is designed requires some duplicate work. In order to make any call to one of the `*_by(section, subsection, key)` methods, one will have to do extra work in order to get the subsection name, even though what one is really doing is to split a string across `.` two or three times.

That operation, along with the potential for failure, are already captured in `parse::key()`. Thus `Key` would probably be better named `ParseKey` while providing only one method to obtain a `parse::Key`.

Doing this would probably also do away with the duplication in implementation, as it then all boils down to getting a `&BStr` for parsing a key.

gix-config/src/tests.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    #[test]
+    fn key_impl_for_str() {
+        assert_eq!("", "".section_name());
+        assert_eq!(None, "".subsection_name());
+        assert_eq!("", "".name());
+
+        assert_eq!("", "foo".section_name());
+        assert_eq!(None, "foo".subsection_name());
+        assert_eq!("foo", "foo".name());
```
Byron, line 14: Related to my earlier comment, this is a good example on where the API of `Key` isn't currently suitable to capture what a `Key` is. Top-level keys like `"foo"` are not valid, they need a section name, and empty section names are not permitted (nor are empty keys).

Thus I think it's important to be able to express a parsing failure or 'invalid key' as part of the `Key` trait (or `ParseKey` trait as proposed).

Byron (commented):
Thanks so much for the changes!

It's wonderful to see how the standard-API of `gix-config` can be enhanced by the `Key` trait. While looking at it more closely, I also noticed something else that I think can be fixed without too much overhead.

But if that work should ever appear overwhelming or boring because the API surface is quite massive, please let me know and I can help doing the refactoring. The last thing I'd want is for you to get exhausted due to all these comments of mine 😅.

Just let me know how I can help and I will try to make it happen.
Thanks again!

Byron:
As a note, I plan to finish this PR myself once I get a chance.

Byron:
The content of this PR was merged to main, but I forgot to push before changing the commit message, hence GitHub doesn't see the merge happened.

Thanks again for getting this started!

PR #1239, opened 2024-01-08
Add `cargo-binstall` support for gitoxide

NobodyXu:
So that `cargo-binstall` can download official pre-built binaries from the github release.

Tested locally using `cargo binstall --manifest-path Cargo.toml gitoxide --force --strategies crate-meta-data`

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,10 +161,10 @@ Our [stability guide] helps to judge how much churn can be expected when dependi
 
 ### Download a Binary Release
 
-Using `cargo quickinstall`, one is able to fetch [binary releases][releases]. You can install it via `cargo install cargo-quickinstall`, assuming 
+Using `cargo binstall`, one is able to fetch [binary releases][releases]. You can install it via `cargo install cargo-quickinstall`, assuming 
 the [rust toolchain][rustup] is present.
 
-Then install gitoxide with `cargo quickinstall gitoxide`.
+Then install gitoxide with `cargo binstall gitoxide`.
 
 See the [releases section][releases] for manual installation and various alternative builds that are _slimmer_ or _smaller_, depending
 on your needs, for _Linux_, _MacOS_ and _Windows_.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -161,10 +161,10 @@ Our [stability guide] helps to judge how much churn can be expected when dependi
 
 ### Download a Binary Release
 
-Using `cargo quickinstall`, one is able to fetch [binary releases][releases]. You can install it via `cargo install cargo-quickinstall`, assuming
```
Byron, line 164: Could you provide more context to help me understand why `quickinstall` is replaced?
To me it seems it's a viable alternative, and both will work equally well.

The way I understand it is that `quickinstall` will download binaries that it somehow obtained or built itself. To my mind, what `binstall` does is certainly preferable, even though it seems that some might still benefit from `quickinstall`.

To me it seems it would be better to mention `binstall` first, but also mention `quickinstall` and why some might want to use that instead.

Please note that my knowledge is definitely spotty here and includes a lot of guessing, so I'd be happy to be educated.

Byron (changes requested):
Thanks a lot for making this work! It's about time, too 😁.

> It would be great if you can provide:
> 
> * `aarch64-apple-darwin` builds
> * `universal-apple-darwin` builds
> * provide *-musl builds for all *-gnu builds
> * provide a new type of builds which uses `zlib-ng` for max perf and linked with it statically, uses `rustls` to avoid external `openssl` dep.

I agree, it would be great, and I'd be happy if someone with more knowledge or persistence with CI builds could take a stab at this. My current level of pre-built binaries is "Oh my, it works, let's hope it will stay that way so I don't have to touch it ever again".

What's here was basically copied from `ripgrep` at some point, and now is held together with duct-tape :D.

README.md
NobodyXu: > The way I understand it is that quickinstall will download binaries that it somehow obtained or built itself.

Quickinstall only uses binaries it built in its own repository, we have a mechanism for building popular crates (all binstall and quickinstall would HEAD an API to report usage).

Binstall can download from official releases and fallback to quickinstall if there's none official binary for this target, binstall also has a complex target fallback chain (e.g. x86_64-unknown-linux-gnu, fallback to musl).

PR #1244, opened 2024-01-12
Add a max-performance-zlib-ng-compat flag

joshtriplett:
This allows users to use zlib-ng-compat without having to add a
dependency on gix-features, and without having to keep track of which
options max-performance depends on other than max-performance-safe.

### DIFF
--- a/gix/Cargo.toml
+++ b/gix/Cargo.toml
@@ -171,6 +171,14 @@ pack-cache-lru-dynamic = ["gix-pack/pack-cache-lru-dynamic"]
 ## Note that some platforms might suffer from compile failures, which is when `max-performance-safe` should be used.
 max-performance = [ "max-performance-safe", "gix-features/zlib-ng", "fast-sha1" ]
 
+## Activate other features that maximize performance, like usage of threads, `zlib-ng-compat`, and access to caching in object databases.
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot, it makes perfect sense to bubble up variants of certain features that are useful in downstream projects.

Personally I also hope that it will be possible to avoid 'a million variations', but instead allow the downstream to enable `gix-features` features without having to depend on it, clearly that's a crutch.

joshtriplett:
@Byron I hope so as well, and I don't like to cause feature proliferation like this.

One other alternative option would be to add a max-performance-no-zlib option that doesn't depend on *any* zlib variant, and then *separately* offer a zlib-ng-compat option on `gix` so people don't have to dive down to `gix-features`. That would be just as easy for downstream projects to work with, without having to depend on `gix-features`, and without potentially missing out on anything new that gets added to `max-performance`.

Byron:
That's probably a little more scalable, even though it also looks like the end-game of this technique is to repeat all `gix-features` features in `gix` for maximum flexibility. Choosing practical tradeoffs here is key, of course, and I think I am willing to do what it takes until `cargo` offers nicer ways to handle this problem.

PR #1249, opened 2024-01-14
Avoid using #[cfg] on multiple individual function arguments

joshtriplett:
Attaching #[cfg] to individual arguments makes it look like the function
has five conditionally present arguments, and doesn't make it
immediately apparent that the first two are for the first argument and
the last three are for the second argument.

Split them into separate `let` statements for clarity.

In the process, factor out the common `.try_into().ok()?` from each.

### DIFF
--- a/gix-index/src/fs.rs
+++ b/gix-index/src/fs.rs
@@ -54,17 +54,21 @@ impl Metadata {
     pub fn modified(&self) -> Option<SystemTime> {
         #[cfg(not(windows))]
         {
+            #[cfg(not(target_os = "aix"))]
+            let seconds = self.0.st_mtime;
+            #[cfg(target_os = "aix")]
+            let seconds = self.0.st_mtim.tv_sec;
+
+            #[cfg(not(any(target_os = "netbsd", target_os = "aix")))]
+            let nanoseconds = self.0.st_mtime_nsec;
+            #[cfg(target_os = "netbsd")]
+            let nanoseconds = self.0.st_mtimensec;
+            #[cfg(target_os = "aix")]
+            let nanoseconds = self.0.st_mtim.tv_nsec;
+
             Some(system_time_from_secs_nanos(
-                #[cfg(not(target_os = "aix"))]
-                self.0.st_mtime.try_into().ok()?,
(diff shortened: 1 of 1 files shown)


gix-index/src/fs.rs
```diff
@@ -54,17 +54,21 @@ impl Metadata {
...
-                self.0.st_mtime.try_into().ok()?,
-                #[cfg(target_os = "aix")]
-                self.0.st_mtim.tv_sec.try_into().ok()?,
-                #[cfg(not(any(target_os = "netbsd", target_os = "aix")))]
-                self.0.st_mtime_nsec.try_into().ok()?,
-                #[cfg(target_os = "netbsd")]
-                self.0.st_mtimensec.try_into().ok()?,
-                #[cfg(target_os = "aix")]
-                self.0.st_mtim.tv_nsec.try_into().ok()?,
+                seconds.try_into().ok()?,
```
Byron, line 70: I see, and here is where the conversion happens from signed to unsigned 🤦‍♂️.

Byron (approved):
Thanks a lot, definitely so much easier to read and thus easier to review, maintain and change. Now that I see the change it's also clear how hard it was to know what is what, exactly like the PR text is describing.

Interestingly, when I reviewed this the last time I just gave in assuming that this kind of `libc` based code has to be on the terrible side, without thinking of alternatives.

RoDmitry:
What about `cfg!` macro?:
```rust
let seconds = if cfg!(target_os = "aix") {
    self.0.st_mtim.tv_sec
} else {
    self.0.st_mtime
};

let nanoseconds = if cfg!(target_os = "aix") {
    self.0.st_mtim.tv_nsec
} else if cfg!(target_os = "netbsd") {
    self.0.st_mtimensec
} else {
    self.0.st_mtime_nsec
};
```

Byron:
I think this style would work with [`cfg-if`](https://docs.rs/cfg-if/latest/cfg_if/) for sure, but shouldn't work like this as it boils down to `if true { x.foo } else { x.bar }` and Rust would complain, even though the optimizer would later be able to remove one branch entirely.
If you tell me this works, that would be cool, but also would give me the shivers.

PR #1250, opened 2024-01-14
Fix stat on files with timestamps before the epoch

joshtriplett:
stat converted the seconds to a `u64` before converting them to
SystemTime. This would cause timestamps before the epoch to wrap around
to timestamps near `u64::MAX`.

Instead, pass the seconds as an i64. Convert it to SystemTime by
computing an absolute Duration and then adding or subtracting as
appropriate.

### DIFF
--- a/gix-index/src/fs.rs
+++ b/gix-index/src/fs.rs
@@ -186,6 +186,11 @@ impl Metadata {
 }
 
 #[cfg(not(windows))]
-fn system_time_from_secs_nanos(secs: u64, nanos: u32) -> SystemTime {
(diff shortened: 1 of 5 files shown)


Byron:
Now there is https://github.com/bytecodealliance/rustix/pull/990 which, if accepted, would probably alleviate the need to cast to `i64` explicitly. However, doing so also makes sense to protect from future issues of this kind, so I will probably leave it as is even if the PR gets merged.

Byron:
And finally, the there is a new release with the fix: https://github.com/Byron/gitoxide/releases/tag/gix-index-v0.28.2 .

PR #1254, opened 2024-01-18
Update h2

Jake-Shadle:
https://rustsec.org/advisories/RUSTSEC-2024-0003


Byron:
Thanks a lot!

I also turned on dependabot security updates, maybe it will catch issues like these in future.

PR #1257, opened 2024-01-20
Disable faster-hex's default `serde` feature

joshtriplett:
faster-hex has a default-enabled `serde` feature, but gitoxide doesn't
use that functionality. Disable that feature, so that builds of gitoxide
that otherwise disable serde don't end up pulling it in anyway.


Byron:
That's a great catch! Thanks a lot for the fix.

PR #1260, opened 2024-01-22
entrymode canonicalization

### LINKED ISSUE #1259: Discretisation of EntryMode to EntryKind is almost certainly incorrect
Looked at 3ac5d0b7feed9554ddbf249748cb1591a7977d9b after hitting the privatisation of EntryMode's state[^1] and I might be misreading the code but I think the normalisation behaviour is not correct: https://github.com/Byron/gitoxide/blob/main/gix-object/src/tree/mod.rs#L77-L90

This converts exact tree, link, and commit (submodules) modes to the corresponding kind, then everything else is converted to an executable blob if 0o100 is set, and to a non-executable blob otherwise.

Git's [own mode canonicalisation](https://github.com/git/git/blob/e02ecfcc534e2021aae29077a958dd11c3897e4c/object.h#L140-L149) works very differently though: it uses the `S_IS*`  macros from stat.h to parse the various modes. To my understanding this means it masks in the upper nibble, then checks *that* exactly against one of the type constants. So e.g.

    0o40534

is considered a directory, because the permission bits are discarded before dispatch. In gix-object, I think this currently would parse as an executable blob.

Git checks for BLOB, LINK, then TREE but that should not matter overly much as all of those are equality checks, however *it then falls back to GITLINK* (`Commit`) so e.g. 0o10000 or 0o140000 should be interpreted as `Commit`, not `Blob`.

That would also make the first four `is_` functions incorrect as they check for exact values rather than follow normalisation.

[^1]: incidentally the ability to create an arbitrary `EntryMode` might be useful to test git implementations themselves, but the payload is not `pub` and there seems to be no way to convert a `u16` to an `EntryMode`.

Byron:
Fixes #1259

This means that `mode.kind()` and `mode.is_*` will also work correctly if the underlying mode wasn't canonicalized, which could happen when looking at trees that were stored with very old versions of Git.

### DIFF
--- a/gix-object/src/tree/mod.rs
+++ b/gix-object/src/tree/mod.rs
@@ -72,46 +72,50 @@ impl std::ops::Deref for EntryMode {
     }
 }
 
+const IFMT: u16 = 0o170000;
+
 impl EntryMode {
     /// Discretize the raw mode into an enum with well-known state while dropping unnecessary details.
     pub const fn kind(&self) -> EntryKind {
-        match self.0 {
-            0o40000 => EntryKind::Tree,
-            0o120000 => EntryKind::Link,
-            0o160000 => EntryKind::Commit,
-            blob_mode => {
-                if blob_mode & 0o000100 == 0o000100 {
-                    EntryKind::BlobExecutable
-                } else {
-                    EntryKind::Blob
-                }
+        let etype = self.0 & IFMT;
+        if etype == 0o100000 {
+            if self.0 & 0o000100 == 0o000100 {
+                EntryKind::BlobExecutable
+            } else {
+                EntryKind::Blob
             }
+        } else if etype == EntryKind::Link as u16 {
+            EntryKind::Link
+        } else if etype == EntryKind::Tree as u16 {
+            EntryKind::Tree
+        } else {
+            EntryKind::Commit
         }
     }
 
     /// Return true if this entry mode represents a Tree/directory
     pub const fn is_tree(&self) -> bool {
-        self.0 == EntryKind::Tree as u16
+        self.0 & IFMT == EntryKind::Tree as u16
     }
 
     /// Return true if this entry mode represents the commit of a submodule.
     pub const fn is_commit(&self) -> bool {
-        self.0 == EntryKind::Commit as u16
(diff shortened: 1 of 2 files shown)


Byron:
@xmo-odoo  Please take a look at the fix, do you think that works for #1259?

Byron:
Alright, I will hold this PR for now while it awaits your review. Thanks for your help with this.

Byron:
> I also discovered recently that at one point git had an experimental _binary_ loose object header patterned after pack entry headers, but that ended up being dropped).

Oh that's interesting! I never saw it, and hope that was never released to avoid having to deal with this new loose object format 😅.

> [..] the kinda textual nature of git makes for a lot of lurking bugs and inconsistencies, it's pretty maddening, [..]

Generally, `gix` has to be just as compatible, even if it's unpleasant. There are options around having a 'strict' mode though if there is demand, of course, but even if there was right now that would have to be postponed to the future.

Byron:
Actually, it fits my workflow better to merge this, and instead while reviewing you could create a new PR with changes should you find anything.

Byron:
> > > I also discovered recently that at one point git had an experimental _binary_ loose object header patterned after pack entry headers, but that ended up being dropped).
> > 
> > 
> > Oh that's interesting! I never saw it, and hope that was never released to avoid having to deal with this new loose object format 😅.
> 
> I only know of it because [libgit2 has a codepath to parse them](https://github.com/libgit2/libgit2/blob/25e2b9d8c217e92eb70cde85bb4007a3dec0428d/src/libgit2/odb_loose.c#L222-L274) so I'd assume it did leak in the wild at one point. However [the commit originally adding it is not exactly informative](https://github.com/libgit2/libgit2/commit/3d3552e8fd709214634435ee188e6fb4930e7b39#diff-910cc5446b9d8dfea661a1f98517b27d6803c82787938d6de77cfa07078f2f01R288-R327). Seems reasonable to ignore it I think.

Thanks for digging this out, very interesting indeed!

And I agree, it's probably fine to just let it be forgotten and assume that no repository that uses these objects is still around. If that changes, that is somebody raises an issue with such an object as example, I'd be happy to implement it though - let's see :D.

PR #1267, opened 2024-02-01
Update to Winnow 0.5.36

epage:
This resolves a couple of deprecations.  I'm thinking of doing 0.6 soon and I expect this to be the end of the deprecations that will happen in the lead up.  So for people on 0.5.36 with all deprecations resolved, i expect it to be a trivial version bump to 0.6 for most people.


Byron (approved):
Thanks a lot! I really do appreciate this level of care 🙏.

PR #1277, opened 2024-02-05
Use chunks_exact where possible

kornelski:
I've noticed a few functions use `chunks(len)`, but assume they will actually get `len`-long chunks, even though there could be a shorter remainder chunk.

I suspect there should be more strict length checks for the inputs to `chunks_(exact)`, and `zip` iterators. For example `read_fan` would tolerate inputs shorter than the number of bytes read it returns.

### DIFF
--- a/gix-commitgraph/src/file/init.rs
+++ b/gix-commitgraph/src/file/init.rs
@@ -259,8 +259,10 @@ impl TryFrom<&Path> for File {
 
 // Copied from gix-odb/pack/index/init.rs
 fn read_fan(d: &[u8]) -> ([u32; FAN_LEN], usize) {
+    assert!(d.len() >= FAN_LEN * 4);
+
     let mut fan = [0; FAN_LEN];
-    for (c, f) in d.chunks(4).zip(fan.iter_mut()) {
+    for (c, f) in d.chunks_exact(4).zip(fan.iter_mut()) {
         *f = u32::from_be_bytes(c.try_into().unwrap());
     }
     (fan, FAN_LEN * 4)
--- a/gix-index/src/decode/mod.rs
+++ b/gix-index/src/decode/mod.rs
@@ -96,9 +96,10 @@ impl State {
                     let entries_res = match index_offsets_table {
                         Some(entry_offsets) => {
                             let chunk_size = (entry_offsets.len() as f32 / num_threads as f32).ceil() as usize;
-                            let num_chunks = entry_offsets.chunks(chunk_size).count();
+                            let entry_offsets_chunked = entry_offsets.chunks(chunk_size);
+                            let num_chunks = entry_offsets_chunked.len();
                             let mut threads = Vec::with_capacity(num_chunks);
-                            for (id, chunks) in entry_offsets.chunks(chunk_size).enumerate() {
+                            for (id, chunks) in entry_offsets_chunked.enumerate() {
                                 let chunks = chunks.to_vec();
                                 threads.push(
                                     gix_features::parallel::build_thread()
--- a/gix-pack/src/index/access.rs
+++ b/gix-pack/src/index/access.rs
@@ -31,7 +31,7 @@ impl index::File {
     fn iter_v1(&self) -> impl Iterator<Item = Entry> + '_ {
         match self.version {
             index::Version::V1 => self.data[V1_HEADER_SIZE..]
-                .chunks(N32_SIZE + self.hash_len)
+                .chunks_exact(N32_SIZE + self.hash_len)
                 .take(self.num_objects as usize)
                 .map(|c| {
                     let (ofs, oid) = c.split_at(N32_SIZE);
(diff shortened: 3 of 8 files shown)


gix-commitgraph/src/file/init.rs
```diff
@@ -260,7 +260,7 @@ impl TryFrom<&Path> for File {
 // Copied from gix-odb/pack/index/init.rs
 fn read_fan(d: &[u8]) -> ([u32; FAN_LEN], usize) {
     let mut fan = [0; FAN_LEN];
-    for (c, f) in d.chunks(4).zip(fan.iter_mut()) {
+    for (c, f) in d.chunks_exact(4).zip(fan.iter_mut()) {
```
Byron, line 263: Let's keep using `chunks` here, as that will cause a panic which is better than silently working with a `fan` that isn't fully initialised. Please apply this to all other instances of this function as well.

gix-index/src/decode/mod.rs
```diff
@@ -96,7 +96,8 @@ impl State {
                     let entries_res = match index_offsets_table {
                         Some(entry_offsets) => {
                             let chunk_size = (entry_offsets.len() as f32 / num_threads as f32).ceil() as usize;
-                            let num_chunks = entry_offsets.chunks(chunk_size).count();
+                            // Lengths of the chunks may vary
```
Byron, line 99: If a test failed when trying `chunks_exact()` here, this comment can be removed. That would be my expectation.

gix-pack/src/index/access.rs
```diff
@@ -49,9 +49,9 @@ impl index::File {
         let pack64_offset = self.offset_pack_offset64_v2();
         match self.version {
             index::Version::V2 => izip!(
-                self.data[V2_HEADER_SIZE..].chunks(self.hash_len),
```
Byron, line 52: Let's leave this at `chunks` as well, it should panic if the count is off (which it will).

gix-pack/src/index/access.rs
```diff
@@ -164,7 +164,7 @@ impl index::File {
                 let offset32_start = &self.data[self.offset_pack_offset_v2()..];
                 let pack_offset_64_start = self.offset_pack_offset64_v2();
                 offset32_start
-                    .chunks(N32_SIZE)
+                    .chunks_exact(N32_SIZE)
```
Byron, line 167: Let's leave this at chunks as well, it should panic if the count is off (which it will).

src/plumbing/main.rs
```diff
@@ -226,8 +226,8 @@ pub fn main() -> Result<()> {
                         add_paths: add_path,
                         prefix,
                         files: add_virtual_file
-                            .chunks(2)
-                            .map(|c| (c[0].to_owned(), c[1].clone()))
+                            .chunks_exact(2)
+                            .map(|c| (c[0].clone(), c[1].clone()))
```
Byron, line 230: Let's leave this at chunks as well, it should panic if the count is off (which it will due to indexed access).

Byron (changes requested):

I will definitely keep `chunks_exact()` in mind as it's certainly less of a footgun than `chunks()`, even though both have their strength.

Regarding these, I think there are occasions where this is an improvement. In the other cases, I think it's better to go back to `chunks()` - the idea is that it should panic if there is unexpected input, which typically is checked beforehand.

gix-pack/src/index/access.rs
kornelski: This could truncate `izip!` too, so maybe an explicit check of all 3 lengths would be better?.

src/plumbing/main.rs
kornelski: This one is already checked at the beginning of the function.

gix-index/src/decode/mod.rs
kornelski: I've meant to comment that this code handles chunks of varying lengths by design, so that next person wondering if this should be chunks_exact instead won't need to wonder.

gix-pack/src/index/access.rs
kornelski: I've added an explicit length check instead. I think that makes intention clearer. It lowers risk of performing some work with side effects before panicking, and may be easier to diagnose bad inputs than panicking somewhere down the line.

gix-pack/src/index/access.rs
kornelski: I've added length check, which helps with zip too.

gix-commitgraph/src/file/init.rs
kornelski: `.chunks()` would not catch cases of broken input when when it's multiple of 4, e.g. 8-byte long `d` initializes 2 bytes, and keeps 254 bytes zeroed. I've added an assert instead, which catches all length mismatches.

kornelski:
gix-pack uses offsets into `self.data`, such as `&self.data[self.offset_pack_offset_v2()..]`. 

Is this necessary?

Looking at implementation of offset_crc32_v2, offset_pack_offset_v2, offset_pack_offset64_v2 I can see they're one after another, but this relationship is unclear elsewhere that just uses the offset.

The offsets are also used with an open range `[offset..]`, rather than being slices of their own fragment of data only. This makes it tricky to check if usage of `chunks` or `chunks_exact` is correct, because the open-range data slice definitely doesn't have the right length, and it's not clear if it's meant to have, or whether the pack format allows arbitrary data at the end or the different parts to overlap.

Instead of providing offsets, could these functions return right-sized slices of `data`?

gix-pack/src/index/access.rs
Byron: Thank you, very readable and a much stronger assertion.

Byron (approved):
Great! Thanks a lot! I will definitely be more careful about the usage of `chunks` and `chunks_exact()` from now on.

Byron:
Apologies, I forgot to reply.

> gix-pack uses offsets into `self.data`, such as `&self.data[self.offset_pack_offset_v2()..]`.
> 
> Is this necessary?

I don't now what you are alluding to. Please feel free to submit a PR with improvements if this can clear this up.

> 
> Looking at implementation of offset_crc32_v2, offset_pack_offset_v2, offset_pack_offset64_v2 I can see they're one after another, but this relationship is unclear elsewhere that just uses the offset.
> 
> The offsets are also used with an open range `[offset..]`, rather than being slices of their own fragment of data only. This makes it tricky to check if usage of `chunks` or `chunks_exact` is correct, because the open-range data slice definitely doesn't have the right length, and it's not clear if it's meant to have, or whether the pack format allows arbitrary data at the end or the different parts to overlap.
> 
> Instead of providing offsets, could these functions return right-sized slices of `data`?

In the index, locations of all portions of data can be known in advance, I *think*, which should make it possible to determine the exact range a slice should occupy. A PR for this would definitely be welcome, maybe this could help avoid running into 'traps' laid by malicious actors. Note though that such an actor would need to be able to distribute full repos, with packs *and* indices, as typically all indices are build from scratch when cloning, a vital part of validating the received pack. From that point of view, there might be little benefit to further restricting the regions. Definitely your call though, as I wouldn't expect right-sizing to greatly increase the code's complexity.

kornelski:
What I've meant was about Rust API/design pattern, instead of:

```rust
let slice_of_crcs_and_more = self.data[self.offset_crc32_v2()..];
```

refactoring it to make the offset an implementation detail, and have:

```rust
let slice_of_crcs_only = self.crc32_v2();
```

where `crc32_v2` would return `self.data[start..end]` knowing the start offset and num_elements, so that callers of this function wouldn't need to use`take(num_elements)`.

PR #1288, opened 2024-02-13
Fix into_{blob,tag} panic messages

caio:
Scratched my head for a few seconds when I got a panic with "Tried to use X as commit, but was commit" :grin:

### DIFF
--- a/gix/src/object/mod.rs
+++ b/gix/src/object/mod.rs
@@ -77,8 +77,8 @@ impl<'repo> Object<'repo> {
     /// Transform this object into a blob, or panic if it is none.
     pub fn into_blob(self) -> Blob<'repo> {
         match self.try_into() {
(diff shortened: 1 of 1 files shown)


caio:
Got a couple of check failures but both are related to cargo-deny-action:

> 2024-02-13 09:23:54 [ERROR] failed to start `cargo metadata`: No such file or directory (os error 2): No such file or directory (os error 2)

I'd guess it's a CI environment issue (permissions or smth?), not related to my patch. I've no interest in github ci yaml spaghetti so I'll leave it at that. Let me know you if you need me to do something.

Byron (approved):
😅, thanks a lot for the fix! I totally did not copy-paste that ;).

Regarding CI, I am looking into it - my guess is that `cargo deny` is broken for everyone right now.

PR #1290, opened 2024-02-13
Update winnow to 0.6

epage:
This required using `winnow::stream::BStr` in test errors because `TreeError` now needs `impl Stream for I`.

### DIFF
--- a/gix-actor/src/signature/decode.rs
+++ b/gix-actor/src/signature/decode.rs
@@ -162,7 +162,7 @@ mod tests {
                             .map_err(to_bstr_err)
                             .expect_err("parse fails as > is missing")
                             .to_string(),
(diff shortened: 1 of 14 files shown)


gix-actor/src/signature/decode.rs
```diff
@@ -162,7 +162,7 @@ mod tests {
                             .map_err(to_bstr_err)
                             .expect_err("parse fails as > is missing")
                             .to_string(),
-                        "in slice at ' 12345 -1215'\n  0: expected `<email>` at ' 12345 -1215'\n  1: expected `<name> <<email>>` at ' 12345 -1215'\n  2: expected `<name> <<email>> <timestamp> <+|-><HHMM>` at ' 12345 -1215'\n"
+                        "in slice at ' 12345 -1215'\n  0: expected `<email>` at ' 12345 -1215'\n  1: expected `<name> <<email>>` at 'hello < 12345 -1215'\n  2: expected `<name> <<email>> <timestamp> <+|-><HHMM>` at 'hello < 12345 -1215'\n"
```
epage, line 165: 0.6 included a change that allows `TreeError`  to behave closer to how nom behaved with errors in capturing the start location for each stage things are processed rather than always attaching it to only the original error location

Byron (approved):
Thanks a lot for the great support 🙏, I appreciate it!

Issue #729, opened 2023-02-11
slow compile time, compared to git2 (in a crate without proc macros)


benmkw:
### Duplicates

- [X] I have searched the existing issues

### Current behavior 😯

I have a super small program which opens the github repository of a cloned repo in the browser.

https://github.com/benmkw/githome

After switching from git2 to gitoxide, the compilation time went roughly from 12s to 30s. This seems high to me as everything I need is basically search for a git repo and read a text file. The resulitng binary is also larger with 1.8MB compared to 1.2MB (on macos). (I activated the usual nobs for reducing binary size in Cargo.toml)

### Expected behavior 🤔

None of this is a practical problem for me but I wanted to report it because I think that gitoxide could be faster than git2 because it needs to compile less code (its already split into many crates).

### Steps to reproduce 🕹

clonse https://github.com/benmkw/githome and compare 0db0221 (gitoxide) to b828bf4 (git2)

Byron:
Thanks for bringing it up and for providing a way to reproduce the issue.

My quick and unscientific measurements mostly confirm the findings:

```
    Checking git-refspec v0.7.2
    Checking git-worktree v0.12.2
    Checking git-discover v0.13.0
    Checking git-config v0.16.0
    Checking git-odb v0.40.2
    Checking git-repository v0.35.0
    Checking githome v0.3.0 (/Users/byron/dev/githome)
    Finished dev [unoptimized + debuginfo] target(s) in 5.83s

githome (0db0221) took 5s
❯ gco b828bf4
Previous HEAD position was 0db0221 deps
HEAD is now at b828bf4 up deps and better error message

githome (b828bf4)
❯ cargo clean

githome (b828bf4)
❯ cargo check
   Compiling libc v0.2.76
   Compiling pkg-config v0.3.18
   Compiling autocfg v1.0.1
    Checking tinyvec v0.3.4
    Checking matches v0.1.8
   Compiling log v0.4.11
   Compiling bitflags v1.2.1
    Checking cfg-if v0.1.10
    Checking percent-encoding v2.1.0
    Checking opener v0.4.1
    Checking unicode-bidi v0.3.4
    Checking unicode-normalization v0.1.13
   Compiling jobserver v0.1.21
    Checking idna v0.2.0
   Compiling cc v1.0.59
    Checking url v2.1.1
   Compiling libz-sys v1.1.0
   Compiling openssl-sys v0.9.58
   Compiling libssh2-sys v0.2.19
[...]

PR #1292, opened 2024-02-15
inline btoi to reduce compile times

benmkw:
This PR inlines the calls to btoi into gix-utils to reduce compile times (removes num-traits as well).
Compared to the previous diff I extracted basic tests from the doc comments and created a macro to make it easier to add new types/ reduce noise.
If needed I can squash the commits.
I have https://github.com/tamasfe/taplo installed and it reformatted the touched Cargo.toml files, I can revert that as well if wanted.

ref discussion in https://github.com/Byron/gitoxide/issues/729#issuecomment-1941515655

### DIFF
--- a/gix-utils/src/btoi.rs
+++ b/gix-utils/src/btoi.rs
@@ -1,5 +1,17 @@
-// ported from https://github.com/niklasf/rust-btoi version 0.4.3
-// see https://github.com/Byron/gitoxide/issues/729#issuecomment-1941515655
+/// A module with utilities to turn byte slices with decimal numbers back into their
+/// binary representation.
+///
+/// ### Credits
+///
+/// This module was ported from <https://github.com/niklasf/rust-btoi> version 0.4.3
+/// see <https://github.com/Byron/gitoxide/issues/729> for how it came to be in order
+/// to save 2.2 seconds of per-core compile time by not compiling the `num-traits` crate
+/// anymore.
+///
+/// Licensed with compatible licenses [MIT] and [Apache]
+///
+/// [MIT]: https://github.com/niklasf/rust-btoi/blob/master/LICENSE-MIT
+/// [Apache]: https://github.com/niklasf/rust-btoi/blob/master/LICENSE-APACHE
 
 /// An error that can occur when parsing an integer.
 ///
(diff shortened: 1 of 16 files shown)


Byron:
Thanks a lot for your tremendous help with this!

I will merge once CI is green after my refactor round.

gix-utils/src/btoi.rs
```diff
@@ -0,0 +1,316 @@
...
+                    kind: ErrorKind::Underflow,
+                })
+            }
+        };
+    }
+
+    Ok(result)
+}
+
+/// minimal subset of traits used by btoi_radix and btou_radix
```
fbstj, line 271:
```suggestion
/// minimal subset of traits used by [`to_signed_with_radix `] and [`to_signed_with_radix`]
```
or however it is that doclinks (or just generally improve this doc?)

Byron: That's a great catch, thanks for the suggestion.
I didn't apply the fix because it uses `to_signed_with_radix` twice, but applied the intended fix locally and force-pushed.

gix-utils/src/btoi.rs
```diff
@@ -0,0 +1,316 @@
...
+
+    Ok(result)
+}
+
+/// minimal subset of traits used by [`to_signed_with_radix`] and [`to_unsigned_with_radix`]
+pub trait MinNumTraits: Sized + Copy {
+    ///
+    fn from_u32(n: u32) -> Option<Self>;
+    ///
+    fn zero() -> Self;
```
fbstj, line 276: have you considered changing this to
```suggestion
    /// the 0 value of this type
    const ZERO: Self;
```

fbstj: ahhh apologies for doing this post-merge, the browser hadn't refreshed

Byron: Oh, I see! Some of these could probably be const, maybe const-fn in traits is allowed too? 

In any case, if you would like to submit a PR, that would be very welcome.

benmkw: I just took this from num-traits with minor changes (changed &x parameters of `checked_*` to pass by value cause its only used for small copy types here). Making it an associated type seems reasonable to me as well.

fbstj: I realized later in the evening that it was likely to do with MSRV that such consts were not used, but maybe that is 'too conservative'?

I'll see about a PR if that's ok with benmkw :)

PR #1293, opened 2024-02-16
remove tabled

benmkw:
follow up from https://github.com/Byron/gitoxide/pull/1292

I did another profiling run on gitoxide main with `cargo build --release --timings --no-default-features` and was wondering if tabled and its (derive) dependencies are really needed.

Screenshot 2024-02-16 at 08 36 55

It turns out there is really just one place where its used and while I don't yet understand the exact output semantics (I feel like a manual implementation could be more clear here) it seems that with clever use of the rust [format specifiers](https://doc.rust-lang.org/std/fmt/#syntax) one could probably go a long way.

tabled usage:
```rust
impl Tabled for Record {
    const LENGTH: usize = 3;

    fn fields(&self) -> Vec<Cow<'_, str>> {
        let mut tokens = self.config.split('.');
        let mut buf = vec![{
            let name = tokens.next().expect("present");
            if name == "gitoxide" {
                name.bold().green()
            } else {
                name.bold()
            }
            .to_string()
        }];
        buf.extend(tokens.map(ToOwned::to_owned));

        vec![
            Cow::Borrowed(self.usage.icon()),
            buf.join(".").into(),
            self.usage.to_string().into(),
        ]
    }

    fn headers() -> Vec<Cow<'static, str>> {
        vec!["icon".into(), "key".into(), "info".into()]
    }
}


let mut table = tabled::Table::new(sorted);
    let table = table.with(Style::blank()).with(Extract::rows(1..));
    println!(
        "{}",
        if let Some((terminal_size::Width(w), _)) = terminal_size::terminal_size() {
            table.with(Width::wrap(w as usize).keep_words().priority::<PriorityMax>())
        } else {
            table
        }
    );
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1248,7 +1248,6 @@ dependencies = [
  "owo-colors",
  "prodash 28.0.0",
  "serde_derive",
- "terminal_size",
  "time",
  "tracing",
  "tracing-forest",
@@ -4591,16 +4590,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ff1bc3d3f05aff0403e8ac0d92ced918ec05b666a43f83297ccef5bea8a3d449"
 dependencies = [
  "winapi-util",
-]
-
-[[package]]
-name = "terminal_size"
-version = "0.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "21bebf2b7c9e0a515f6e0f8c51dc0f8e4696391e6f1ff30379559f8365fb0df7"
-dependencies = [
- "rustix 0.38.31",
- "windows-sys 0.48.0",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -52,31 +52,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
-name = "ansi-str"
-version = "0.8.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1cf4578926a981ab0ca955dc023541d19de37112bc24c1a197bd806d3d86ad1d"
-dependencies = [
- "ansitok",
-]
-
-[[package]]
 name = "ansiterm"
 version = "0.12.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4ab587f5395da16dd2e6939adf53dede583221b320cadfb94e02b5b7b9bf24cc"
 dependencies = [
  "winapi",
-]
-
-[[package]]
-name = "ansitok"
-version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "220044e6a1bb31ddee4e3db724d29767f352de47445a6cd75e1a173142136c83"
-dependencies = [
- "nom",
- "vte",
 ]
 
 [[package]]
@@ -138,12 +119,6 @@ name = "arc-swap"
 version = "1.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bddcadddf5e9015d310179a59bb28c4d4b9920ad0f11e8e14dbadf654890c9a6"
-
-[[package]]
-name = "arrayvec"
-version = "0.5.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "23b62fc65de8e4e7f52534fb52b0f3ed04746ae267519eef2a83941e8085068b"
 
 [[package]]
 name = "arrayvec"
@@ -447,12 +422,6 @@ name = "bumpalo"
 version = "3.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f30e7476521f6f8af1a1c4c0b8cc94f0bee37d91763d0ca2665f299b6cd8aec"
-
-[[package]]
-name = "bytecount"
-version = "0.6.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e1e5f035d16fc623ae5f74981db80a439803888314e3a555fd6f04acd51a3205"
 
 [[package]]
 name = "byteorder"
@@ -1279,7 +1248,6 @@ dependencies = [
  "owo-colors",
  "prodash 28.0.0",
  "serde_derive",
- "tabled",
  "terminal_size",
  "time",
  "tracing",
@@ -3487,12 +3455,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6877bb514081ee2a7ff5ef9de3281f14a4dd4bceac4c09388074a6b5df8a139a"
 
 [[package]]
-name = "minimal-lexical"
-version = "0.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "68354c5c6bd36d73ff3feceb05efa59b6acb7626617f4962be322a825e61f79a"
-
-[[package]]
 name = "miniz_oxide"
 version = "0.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3542,16 +3504,6 @@ dependencies = [
  "libc",
  "memoffset",
  "pin-utils",
-]
-
-[[package]]
-name = "nom"
-version = "7.1.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d273983c5a657a70a3e8f2a01329822f3b8c8172b73826411a55751e404a0a4a"
-dependencies = [
- "memchr",
- "minimal-lexical",
 ]
 
 [[package]]
@@ -3709,19 +3661,6 @@ name = "owo-colors"
 version = "4.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "caff54706df99d2a78a5a4e3455ff45448d81ef1bb63c22cd14052ca0e993a3f"
-
-[[package]]
-name = "papergrid"
-version = "0.11.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9ad43c07024ef767f9160710b3a6773976194758c7919b17e63b863db0bdf7fb"
-dependencies = [
- "ansi-str",
- "ansitok",
- "bytecount",
- "fnv",
- "unicode-width",
-]
 
 [[package]]
 name = "parking"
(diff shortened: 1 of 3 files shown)


Byron:
Thanks for researching this!

If removing the derive in favour of a manual implementation works while really only adding the code demonstrated above, I am all for it.

But also to provide some guidance, to me compile time optimisations should be focussed on `gix` (the library crate), and personally I consider `gix`  (the binary) to be a development tool where anything goes as long as it facilitates or accelerates development. That doesn't mean to be wasteful, but generally I wouldn't optimise here.

With all that said, if you do find an opportunity that seems worth it, I'd always want to bring it in.

benmkw:
> works while really only adding the code demonstrated above, I am all for it

As far as I can see the new (or more or less your old) implementation does only lack the line wrapping (noticeable for smaller terminals). Because this seems to only be used when running `config-tree` this does not seem like a large limitation.

Unfortunately I could not get the config name column to be padded to 50 characters (`: <50` does not change the output for me, not sure why, [this works just fine](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=dab4c0650626bce21ba48d7fda0c5a1f)). Maybe something about terminal escapes (?).

Given that this removes 14 packages it seems somewhat worth it to me.

> I consider gix (the binary) to be a development tool where anything goes as long as it facilitates or accelerates development. That doesn't mean to be wasteful, but generally I wouldn't optimise here.

That makes sense. I can only add to that that while I had to recompile some times to tweak the format strings the compile times sure do add up, but thats also just rust dev in general :D

Byron:
> Unfortunately I could not get the config name column to be padded to 50 characters (`: <50` does not change the output for me, not sure why, [this works just fine](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=dab4c0650626bce21ba48d7fda0c5a1f)). Maybe something about terminal escapes (?).

All this formatting can get complicated and sometimes it feels a little like luck if it actually looks like it should 😅.

> Given that this removes 14 packages it seems somewhat worth it to me.

I wasn't aware, but yes, that sounds valuable especially if that removal also reduces compile time meaningfully.

> That makes sense. I can only add to that that while I had to recompile some times to tweak the format strings the compile times sure do add up, but thats also just rust dev in general :D

Agreed, and it's particularly bad when working with `gix` (the binary), and it won't get better as code just accumulates there. Even `gix`, the library, takes some time to compile.

I suppose that makes the split into plumbing crates even more valuable, as these typically compile very fast and make testing simpler. Thus far I am quite happy with the overall performance though, and think it can be good for another 3 years or so (without having to upgrade hardware).

benmkw:
Looking around more I also found itertools as a dependency which is only used for join and tuple_windows in a total of 3 places and already has the izip macro [vendored](https://github.com/Byron/gitoxide/commit/834050878b43bae677287767332adc746a8aa2ed) (although that's only used once so far) (I think this really should be in std, I have had the need for that many times as well). I don't want to drive this in a direction which creates more noise but thought I'd mention it briefly.

Screenshot 2024-02-16 at 10 20 33

One noticeable benefit of these changes in btoi I think is the opportunity for better naming as you did/ more domain specific adaptions. In the tabled case I think the shorter code is actually easier to read.

Byron:
Oh my! Itertools with about 10 seconds! I actually never really want to use it by feel its convenience forces my hand. Indeed, maybe it's feasible to inline the few things that are used here, but I have a feeling it's going to be harder and might add large swaths of code.

> I don't want to drive this in a direction which creates more noise but thought I'd mention it briefly.

Yes, let's take one step at a time while keeping other opportunities in mind as well.

benmkw:
> Yes, let's take one step at a time while keeping other opportunities in mind as well.

The main issue with this PR from my POV is that the alignment does not yet work, but I'm not yet sure how to best debug this. The blinking is not the reason at least (tried without it).

> but I have a feeling it's going to be harder and might add large swaths of code.

I would try, I have a padded iterator inlined here which was not much code https://github.com/benmkw/p2r/blob/eae398a8388b7ce6b1d1663b5499fdcf7fe68cb6/p2r/src/util.rs e.g..

For the two uses of join one can probably collect in a vec and [join afterwards](https://doc.rust-lang.org/std/primitive.slice.html#method.join) cause its not in any hot code.

The one usage of `tuple_windows` could be replaced with a regular for loop and a `prev` variable instead of a fold.

Byron:
> The main issue with this PR from my POV is that the alignment does not yet work, but I'm not yet sure how to best debug this. The blinking is not the reason at least (tried without it).
> 

It's surprising that the formatting is affected at all given that the only think I thought would change is the addition of the manual implementation of what the `tabled` derive does. Would that mean it does something special that's not reproduced? And if so, could expanding the macro help seeing what that is?

Regarding the removal of itertools, everything you said sounds like a net-win to me, and I'd love to use the opportunity to stop the proliferation of `itertools` in its tracks - when it's already there, I will use it, if not, I will think twice.

benmkw:
The output now looks like:

```console
👌️ gitoxide.objects.cacheLimit                   ❗️If unset or 0, there is no object cache. overridden by 'GIX_OBJECT_CACHE_MEMORY'❗️
👌️ gitoxide.objects.replaceRefBase               ❗️overridden by 'GIT_REPLACE_REF_BASE'❗️
👌️ gitoxide.pathspec.glob                        ❗️pathspec wildcards don't match the slash character, then needing '**' to get past them. overridden by 'GIT_GLOB_PATHSPECS'❗️
👌️ gitoxide.pathspec.icase                       ❗️Compare string in a case-insensitive manner. overridden by 'GIT_ICASE_PATHSPECS'❗️
👌️ gitoxide.pathspec.inheritIgnoreCase           ❗️Inherit `core.ignoreCase` for defaults in pathspecs❗️
👌️ gitoxide.pathspec.literal                     ❗️Make the entire spec used verbatim, the only way to get ':()name' verbatim for instance. overridden by 'GIT_LITERAL_PATHSPECS'❗️
👌️ gitoxide.pathspec.noglob                      ❗️Enable literal matching for glob patterns, effectively disabling globbing. overridden by 'GIT_NOGLOB_PATHSPECS'❗️
👌️ gitoxide.ssh.commandWithoutShellFallback      ❗️is always executed without shell and treated as fallback. overridden by 'GIT_SSH'❗️
👌️ gitoxide.tracePacket                          ❗️overridden by 'GIT_TRACE_PACKET'❗️
👌️ gitoxide.user.emailFallback                   ❗️overridden by 'EMAIL'❗️
👌️ gitoxide.userAgent                            ❗️The user agent presented on the git protocol layer, serving as fallback for when no `http.userAgent` is set❗️
🕒 http.<url>.*                                                 planned ℹ definitely needed for correctness, testing against baseline is a must ℹ
🤔 http.cookieFile                                              not planned ℹ on demand ℹ
🤔 http.curloptResolve                                          not planned ℹ on demand ℹ
🤔 http.delegation                                              not planned ℹ on demand ℹ
🤔 http.emptyAuth                                               not planned ℹ on demand ℹ
👌️ http.extraHeader                                             ❗️fails on illformed UTF-8, without leniency❗️
✅ http.followRedirects                                         
👌️ http.lowSpeedLimit                                           ❗️fails on negative values❗️
```

(gitoxide is still green)

One issue could be that the rust format specifiers don't understand terminal escapes and thus calculate the length differently https://stackoverflow.com/a/69611481 so this was one reason why it was pulling in so many crates.

The question then is how to best replicate that behaviour in a simple way, if desired.

benmkw:
> The question then is how to best replicate that behaviour in a simple way, if desired.

- added a space to ? to make it take the same space as the emoji icons
- removed green and bold formatting from the config string, any fancy colors in the description can stay as they are just appended

```console
?  fetch.fsckObjects                                 : ❓
✅ fetch.negotiationAlgorithm                        : 
🤔 fetch.output                                      : not planned ℹ 'gix' might support it, but there is no intention on copying the 'git' CLI ℹ
🕒 fetch.parallel                                    : planned
🕒 fetch.prune                                       : planned
🕒 fetch.pruneTags                                   : planned
🕒 fetch.recurseSubmodules                           : planned ℹ Seems useful for cargo as well ℹ
❌ fetch.showForcedUpdates                           : not applicable: we don't support advices
🕒 fetch.unpackLimit                                 : planned
🕒 fetch.writeCommitGraph                            : planned
👌️ gitoxide.allow.protocolFromUser                   : ❗️overridden by 'GIT_PROTOCOL_FROM_USER'❗️
👌️ gitoxide.author.emailFallback                     : ❗️overridden by 'GIT_AUTHOR_EMAIL'❗️
👌️ gitoxide.author.nameFallback                      : ❗️overridden by 'GIT_AUTHOR_NAME'❗️
👌️ gitoxide.committer.emailFallback                  : ❗️overridden by 'GIT_COMMITTER_EMAIL'❗️
👌️ gitoxide.committer.nameFallback                   : ❗️overridden by 'GIT_COMMITTER_NAME'❗️
👌️ gitoxide.core.defaultPackCacheMemoryLimit         : ❗️If unset, we default to 96MB memory cap for the default 64 slot LRU cache for object deltas.❗️
👌️ gitoxide.core.externalCommandStderr               : ❗️overridden by 'GIX_EXTERNAL_COMMAND_STDERR
```

Byron:
Thanks for the great work - I am amazed how close both versions are with 14 dependencies less and overall, much less and simpler code.

Screenshot 2024-02-16 at 13 43 42
Screenshot 2024-02-16 at 13 43 57

It's quite amazing to see what `tabled` *actually* does in this case, and I might say that it's certainly fine to drop it. Colors aren't too important.

But what I find important is the linebreak it makes for the second column. If `terminal-size` was re-added, it should be possible to compute linebreaks by hand without too much effort, presumably.

Is this something you think is possible on top?

benmkw:
> But what I find important is the linebreak it makes for the second column. If terminal-size was re-added, it should be possible to compute linebreaks by hand without too much effort, presumably.

sure I understand, terminal size is fairly small and I would not have removed it for the sake of codesize so it would be no problem to add it back. 

The problem is the calculation of the displayed size of the usage section. I'll have to look into how to best do this/ if this is easily possible. 

I think the tabled solution was a bit overkill because in this case the usage strings are fixed and the calculation does not have to be 100% robust but it should of course be at least reasonable. 

Another way would be to basically record the size of the strings before adding escape codes and also remembering to number of emojis per string but yeah .. I'll have to see if there is a clean solution to that or if its easier to basically re-parse the escape codes and that way calculate the lenght. owo-colors does not seem to have a way to say "give me the underlying string without the terminal codes" as far as I could see which is a bit unfortunate for this use case (being tied to the API of a dependency is sometimes a real bummer), but maybe there is a way.

Byron:
Right, I see what you mean - trying to adjust the right-hand size and split it needs to consider terminal escape codes, which makes this a whole lot harder than it otherwise would be.

Maybe limiting this to just unicode will be good enough, so no ansicodes have to be dealt with. On the right-hand side this seems to be just the blinking and bold text.

There is an excellent crate to help with all of that: `anstyle`: https://github.com/rust-cli/anstyle - it has a [`parse`](https://github.com/rust-cli/anstyle/tree/main/crates/anstyle-parse) crate that can probably help with that.

But honestly, I am OK cutting back on that complexity as I am sure once simple line-breaks of simple unicode strings work, it will look good enough. I am not even a fan of the blinking which is now shown since I switched from Alacritty to Wezterm.

Byron (approved):
That's absolutely wonderful, so much better than before, and way more readable.

I really had to used `tabled` I guess just to not implement the wordwrap 😅.

PR #1294, opened 2024-02-16
remove itertools

benmkw:
follow up from https://github.com/Byron/gitoxide/pull/1293#issuecomment-1948023333 and https://github.com/Byron/gitoxide/pull/1293#issuecomment-1948047161

Removes around 9s of cpu compilation time at the exchange of some allocations for displaying mails (which should not be critical) and a for loop instead of a fold.

### DIFF
--- a/gitoxide-core/src/hours/util.rs
+++ b/gitoxide-core/src/hours/util.rs
@@ -55,10 +55,10 @@ where
             // estimate lower bound of capacity needed
             let (lower, _) = iter.size_hint();
             let mut result = String::with_capacity(sep.len() * lower);
-            write!(&mut result, "{first_elt}").unwrap();
+            write!(&mut result, "{first_elt}").expect("enough memory");
             iter.for_each(|elt| {
                 result.push_str(sep);
-                write!(&mut result, "{elt}").unwrap();
+                write!(&mut result, "{elt}").expect("enough memory");
             });
             result
         }
--- a/gitoxide-core/src/hours/util.rs
+++ b/gitoxide-core/src/hours/util.rs
@@ -42,6 +42,29 @@ impl<'a> From<&'a WorkByEmail> for WorkByPerson {
     }
 }
 
+fn join<I>(mut iter: I, sep: &str) -> String
+where
+    I: Iterator,
+    <I as Iterator>::Item: std::fmt::Display,
+{
+    use ::std::fmt::Write;
+
+    match iter.next() {
+        None => String::new(),
+        Some(first_elt) => {
+            // estimate lower bound of capacity needed
+            let (lower, _) = iter.size_hint();
+            let mut result = String::with_capacity(sep.len() * lower);
+            write!(&mut result, "{first_elt}").unwrap();
+            iter.for_each(|elt| {
+                result.push_str(sep);
+                write!(&mut result, "{elt}").unwrap();
+            });
+            result
+        }
+    }
+}
+
 impl WorkByPerson {
     pub fn write_to(
         &self,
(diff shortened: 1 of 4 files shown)


gitoxide-core/src/hours/util.rs
```diff
@@ -53,9 +52,23 @@ impl WorkByPerson {
...
-            "{} <{}>",
-            self.name.iter().join(", "),
-            self.email.iter().join(", ")
+            "{names} <{mails}>",
+            names = self
+                .name
+                .iter()
+                // BStr does not impl slice::Join
+                .map(|s| s.to_string())
+                .collect::<Vec<_>>()
```
NobodyXu, line 61: I think you could write an iterator for `BStr` to avoid this allocation (and the `.to_string()`)?

Given that gix uses `BStr` a lot, I think it might worth the trouble.

NobodyXu (changes requested):

gitoxide-core/src/hours/util.rs
benmkw: I was not yet sure what the precise missing bound was as to why BStr does not impl Join. Given that the results are probably short and are directly written to the user it seemed fine to convert them to Strings but that might be wrong (?).

NobodyXu: > I was not yet sure what the precise missing bound was as to why BStr does not impl Join.

`std::slice::join` only accept types that `impl Borrow<str>`, `BStr` doesn't implement it.

> Given that the results are probably short and are directly written to the user it seemed fine to convert them to Strings but that might be wrong (?).

True, though I still think achieving zero-allocation is better.

It probably won't take that much effect given that [`BStr` implements `std::fmt::Display`](https://docs.rs/bstr/latest/bstr/struct.BStr.html#impl-Display-for-BStr), so basically you just need to re-implement join from itertool.

benmkw: Yeah, I inlined the join impl which was fairly compact.

gitoxide-core/src/hours/util.rs
```diff
@@ -43,6 +42,29 @@ impl<'a> From<&'a WorkByEmail> for WorkByPerson {
     }
 }
 
+fn join<I>(mut iter: I, sep: &str) -> String
```
NobodyXu, line 45: IMO you can avoid the `String` creation by turning this into a struct, with `iter` and `sep` stored in it.

Then impl `fmt::Display` on it, and you will achieve zero-allocation joining.

NobodyXu (changes requested):

gitoxide-core/src/hours/util.rs
Byron: I think it's fine to allocate here, it's similar to what was available before with `std::slice::join()`.

Byron (approved):
Thanks a lot, that's great value!

gitoxide-core/src/hours/util.rs
benmkw: I took it from here https://github.com/rust-itertools/itertools/blob/762643f1be2217140a972745cf4d6ed69435f722/src/lib.rs#L2295-L2324

I also think its fine as is but one could certainly think of ways to improve it. The function could also be made non generic and just accept a slice as that would have less indirections in this case but I figured the way it is now it is easier to move it to utils in case it happens to be needed somewhere else.

Byron: Oh, if you find the time, maybe even in the other PR, could you add a note to say where it's from?
I'd like to keep track of any code that isn't specifically written for `gitoxide` to be sure I can comply with licensing terms.
Thanks in advance.

benmkw: yeah sure 👍

PR #1296, opened 2024-02-16
Copy over code from gix to generate completions for ein as well

sbatial:
One thing to note: It currently takes `gitoxide` instead of `ein` as the command.
I don't know why it does that.
The command name is taken from Args, and using the locally build version of gix does not do that (I figured it might be an issue with the manual build).

I'll try to investigate but I'd be glad about any hints.

If nothing comes up it would always be possible to just hardcode the name of the binary until the culprit is found.

### DIFF
--- a/src/porcelain/main.rs
+++ b/src/porcelain/main.rs
@@ -3,8 +3,8 @@ use std::sync::{
     Arc,
 };
 
-use anyhow::Result;
-use clap::Parser;
+use anyhow::{anyhow, Result};
+use clap::{CommandFactory, Parser};
 use gitoxide::shared::pretty::prepare_and_run;
 use gitoxide_core as core;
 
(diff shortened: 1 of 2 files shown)


Skgland:
> One thing to note: It currently takes `gitoxide` instead of `ein` as the command. I don't know why it does that. The command name is taken from Args, and using the locally build version of gix does not do that (I figured it might be an issue with the manual build).

`app.get_name()` appears to default to the default to the package name, adding a `#[command(name = "gix")]` to the definition of `App` might help, [`clap docs`](https://docs.rs/clap/latest/clap/_derive/_cookbook/git_derive/index.html)

sbatial:
@Skgland thanks! that should do the trick.

I tried to find where gix and ein differ in implementation but didn't spot that Args does indeed have a name set.

sbatial:
Not sure if I should add you as a co-author/how to do that properly, but at least according to my local build that did the trick!

Byron (approved):
Thanks a lot - a great addition!

PR #1300, opened 2024-02-17
Fix a typo in `gix::clone::PrepareFetch::new`, crate_opts -> create_opts

DianaNites:
Fix a documentation typo in `gix::clone::PrepareFetch::new` crate_opts -> create_opts


Byron (approved):
Thanks, good catch!

PR #1329, opened 2024-03-24
Remove dep reqwest from gix

NobodyXu:
Instead forward reqwest feature defined in gix-transport

### DIFF
--- a/gix-transport/Cargo.toml
+++ b/gix-transport/Cargo.toml
@@ -22,20 +22,42 @@ default = []
 ## If set, blocking implementations of the typical git transports become available in `crate::client`
 blocking-client = ["gix-packetline/blocking-io"]
 ## Implies `blocking-client`, and adds support for the http and https transports.
-http-client = ["base64", "gix-features/io-pipe", "blocking-client", "gix-credentials"]
+http-client = [
+    "base64",
+    "gix-features/io-pipe",
+    "blocking-client",
+    "gix-credentials",
+]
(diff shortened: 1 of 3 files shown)


gix-transport/Cargo.toml
```diff
@@ -22,20 +22,42 @@ default = []
 ## If set, blocking implementations of the typical git transports become available in `crate::client`
 blocking-client = ["gix-packetline/blocking-io"]
 ## Implies `blocking-client`, and adds support for the http and https transports.
-http-client = ["base64", "gix-features/io-pipe", "blocking-client", "gix-credentials"]
+http-client = [
+    "base64",
+    "gix-features/io-pipe",
+    "blocking-client",
+    "gix-credentials",
+]
```
NobodyXu, line 30: I'm using zed which does auto formatting of toml 😂

I can remove them if you don't like it, but I will be away for computer for the rest of the day

Byron: I really don't mind and believe that there is no use to 'fight' a formatter - over time, they will win :D.
The reason it's not autoformatted yet is probably because IntelliJ doesn't format TOML files for me.

I just hope Rust can align on one formatter at some point, maybe it did already, because 'formatter wars' is definitely nothing I want to be involved in ;).

Byron (approved):
Thanks a lot, both for suggesting this change and for implementing it.

I have added the curl-forwarding (just one feature, after all), and will merge once CI is green.

PR #1336, opened 2024-04-03
Add topological walk

Osse:
This pull request adds a new module `commit::topo` to `gix-traverse` to enable topological commit walk. See [this discussion](https://github.com/Byron/gitoxide/discussions/1266) for some context.

Since that discussion was last active I have added a bunch of tests in the same spirit as existing tests. I have also taken the liberty of rearranging some of the code in `gix-traverse` to make it easier to "slot in" my code.

One thing I have not implemented is exposing any of this to `gix` itself. I have started [sketching it](https://github.com/Osse/gitoxide/commit/6a746082f42eb1a63b0de2e2a5e5237e8c7217bb) but am not happy with it currently.

Any feedback is welcome!


gix-traverse/tests/traverse.rs
```diff
@@ -5,4 +5,5 @@ fn hex_to_id(hex: &str) -> gix_hash::ObjectId {
 }
 
 mod commit;
+mod topo;
```
Byron, line 8: Tests must follow the package structure, so it should be in `commit/topo`.

gix-traverse/Cargo.toml
```diff
@@ -20,5 +20,7 @@ gix-date = { version = "^0.8.5", path = "../gix-date" }
 gix-hashtable = { version = "^0.5.2", path = "../gix-hashtable" }
 gix-revwalk = { version = "^0.13.0", path = "../gix-revwalk" }
 gix-commitgraph = { version = "^0.24.2", path = "../gix-commitgraph" }
+gix-revision = { version = "^0.27.0", path = "../gix-revision" }
 smallvec = "1.10.0"
 thiserror = "1.0.32"
+flagset = "0.4.4"
```
Byron, line 26: This should most certainly be `bitflags` as this crate is already used in the dependency tree.

gix-traverse/src/commit/mod.rs
```diff
@@ -0,0 +1,56 @@
+use gix_object::FindExt;
+use smallvec::SmallVec;
+
+/// Simple ancestors traversal
+pub mod ancestors;
+pub use ancestors::{Ancestors, Sorting};
```
Byron, line 6: I think I will refactor the way `ancestors` is made to be more aligned with what's in `topo`, which would also avoid having to `pub use ` anything here for historical reasons.

There is other issues I have with `ancestors` mostly around unnecessary complexity that I'd like to resolve once and for all then.

Byron (commented):
Thanks a million for contributing true topological traversal!

I only had a quick look this time, less than 15 minutes, and really like what I see! There is a few minor comments, only, and the lint CI failure is something that would certainly also be going away with my refactor. However, please feel free to fix it earlier as I won't get to work on this immediately.

gix-traverse/Cargo.toml
Osse: Ah, yeah that would make more sense. I wasn't aware bitflags was in use so when I searched around it seemed like `flagset` was the new shiny thing to use. WIll convert to `bitflags`.

On that topic... The whole construct of using a bitfield for this is taken straight from Git, where there are many many more fields and the topological sort just adds a few. This thing *could* just be a `struct` with a couple of booleans if we wanted to. Another idea would be to use some sort of typestate thing where a commit changes state from been `Seen` to being `Explored`. which would very much make sense and mimic how the different "walks" happen.

Byron: Thanks for elaborating! I took another look at `flagset` and loved that it comes with only 410 lines of code. Otherwise, it can probably cover everything that `bitflags` is currently used for.

That's probably something for another PR, but I certainly wouldn't mind switching. If you consider doing that please note that `gix-index` is the place to look first, as it has the most elaborate use of `bitflags` that should be covered.

gix-traverse/tests/traverse.rs
Osse: Hmm... Can you clarify how would want this solved? [Currently](https://github.com/Byron/gitoxide/blob/main/gix-traverse/tests/commit/mod.rs) there is only a `commit/mod.rs` which contains `mod ancestors { ... }` inline. Should I split that out into `commit/ancestors.rs` and then add `commit/topo.rs` or just append `mod topo { ... }`?

Byron: Sorry for the confusion, don't worry about it - I will do it as part of my refactoring which might change things around even more. A lot is intuition by now and I just have to find a moment to focus on this PR to get it merged.

Byron:
### Review Tasks

* [ ] unify module structure even if it would be breaking
* [ ] add baseline comparison to assure it's the same as Git topo order

gix-traverse/src/commit/topo/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+        const Added = 0b100000;
+    }
+}
+
+/// Sorting to use for the topological walk
+#[derive(Clone, Copy, Debug, Default)]
+pub enum Sorting {
+    /// Show no parents before all of its children are shown, but otherwise show
+    /// commits in the commit timestamp order.
+    #[default]
```
Byron, line 68: A quick question: Since this is all about `topo`, should the default not be `TopoOrder`?

I also have trouble understanding how this `DateOrder` maps to Git. Is there a `git rev-list` equivalent?

Byron (approved):
Thanks again for this massive PR, I think it's great!

After some failed attempts to unify the package structure, I ended up with something that feels quite 'plumbing', but which at least is consistent with the current structure which sees verything as traversal (and not as Walk). I also considered to put both traversals/walks into `gix-revwalk`, and into `gix-revision` as well, but felt like `gix-traverse` is most 'meant' for these implementations.

I have also looks at the `gix::Repository::rev_walk()` implementation and think that it should be possible to use either one or the other implementation, based on a new `Sorting` enum that merges both `topo::Sorting` and `simple::Sorting` (previously `ancestors::Sorting`.

But doing so is left for another PR, and if you want, you can give the implementation a try.

gix-traverse/src/commit/topo/mod.rs
Osse: I don't have Git's code (with my debug modifications) in front of me. I can fill in details later, but from memory:

In Git there is an older sorting algorithm that that's used if gen numbers aren't available. If neither `--date-order` nor `--topo-order` is specified then _that_ algorithm is used, for some reason. So in Git the default is "neither of them". (this is the part I must double-check). I think the idea is that with time the old algorithm can be removed from Git entirely, but it hasn't happened yet. 

`DateOrder` maps to `git rev-list --date-order`. I have no strong feelings about which is the default for us; I imagine most callers would want to set the order explictly, anyway. Git's default is equivalent to the new algorithm with `--date-order` (again, I need to double check this).

Osse:
Did you call it `simple` based on my doc comment "Simple ancestors traversal"? I didn't mean to be rude 😛

gix-traverse/src/commit/topo/mod.rs
Byron: Wonderful, thanks so much for the explanation! The level of detail is perfectly fine, and I just added b8486f7c to add another test and incorporate the information.

It's amazing that Git still has this older algorithm, but then added more just by means of flags which don't seem to be mutually exclusive even - ` git rev-list --topo-order --date-order @` does *something*.

gix-traverse/src/commit/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+/// A commit walker that walks in topographical order, like `git rev-list
+/// --topo-order` or `--date-order` depending on the chosen [`topo::Sorting`].
+///
+/// Instantiate with [`topo::Builder`].
+pub struct Topo<Find, Predicate> {
+    commit_graph: Option<gix_commitgraph::Graph>,
+    find: Find,
+    predicate: Predicate,
+    indegrees: IdMap<i32>,
+    states: IdMap<topo::WalkFlags>,
```
Byron, line 32: Do you think that `states` and `indegrees` could be merged? I am asking as these seem to be insert-only, and will use considerable amounts of memory on huge graphs.

Byron (commented):
Hehe, indeed I may be been influenced by this, and the `Ancestors` iterator looked a lot simpler after seeing the state handling in the new topological one 😁.

Of course, both have their place and I'd use `Simple` when memory consumption must be kept in check and there is no particular need for more advanced ordering.

In a way, `Ancestors` was planned to be the 'one' iterator which handles a bunch of algorithms internally. I'd still think its possible to merge both in this crate, but for now I am going with 'merging happens in `gix`. I am writing this in case you only put everything in `topo` to have a clear slate and it's easier, and not because there is anything inherent to its operation that makes it unmergable.

With that said, if you want to integrate it in `gix`, I think that's absolutely possible, but it would definitely be easier if the algorithm(s) would be integrated in what formerly was called `Ancestors`. Both will work though, and I kind of like the separation as it makes it easier to implement new algorithms in future as well.

So I think this was a long-winded way of saying that `gix` should abstract over it to have a single `Sorting` parameter there that allows to implicitly choose any of the available implementations.

One issue there might be with that is that `ends` are not universally implemented yet, but it's something that can probably be done, one way or another, particularly once `specs` are implemented faithfully.

gix-traverse/src/commit/mod.rs
Osse: At first glance I don't see why not. But how much memory are we talking about? I'd guess the memory used by `IdMap<T1>` + `IdMap<T2>` wouldn't be that far off the amount used by `IdMap<(T1, T2)>`.

They are indeed insert-only. And that's just because in Git they are :stuck_out_tongue: There's a point where the indegree of a commit is set to zero. At that point I _think_ it can be removed from all the maps, but I'm not sure.

Osse: As I mentioned in the discussion a couple of months ago (damn, time flies) I did try to add my stuff to `Ancestors` but couldn't get it to work well. I'm all for it, still, but I just don't see it. Doing it in `gix` seems fine to me.

Byron: That's great to know (both :))!

This means if ever there is an issue with memory, there might be avenues to improve that. I assume Git also has multiple hashmaps for that? Or is it all in one there? * (if you recall, please ignore otherwise)*

PR #1337, opened 2024-04-05
[gix-config] Run parse::tests::section::size_of_events only on 64 bit architectures

blinxen:
The problem with `parse::tests::section::size_of_events` is that assumes that the test is always run on a 64 bit architectures. This PR simply disables the test on non 64 bit architectures. I am unsure if you want this kind of problem to be handled like this.  What would be your (@Byron) desired behavior in this case?

### DIFF
--- a/gix-config/src/parse/tests.rs
+++ b/gix-config/src/parse/tests.rs
@@ -5,6 +5,7 @@ mod section {
     use std::borrow::Cow;
 
     #[test]
+    #[cfg(target_pointer_width = "64")]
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot for the fix!

For background, there was a fuzzer test which genuinely managed to make the memory consumption go incredibly high, which led to monitoring the size of this data structure.
It should be enough to keep an eye on it on a single known platform, as it will be indicative to others as well, while being easier to maintain.

PR #1340, opened 2024-04-09
Autogenerate ci-packetline-blocking/src

### LINKED ISSUE #1310: building on windows fails due to symlinks in working tree
### Current behavior 😯

```
error: failed to load manifest for workspace member `C:\code\joe\gitoxide\gix-diff`

Caused by:
  failed to load manifest for dependency `gix-filter`

Caused by:
  failed to load manifest for dependency `gix-packetline-blocking`

Caused by:
  failed to parse manifest at `C:\code\joe\gitoxide\gix-packetline-blocking\Cargo.toml`

Caused by:
  can't find library `gix-packetline-blocking`, rename file to `src/lib.rs` or specify lib.path
```

### Expected behavior 🤔

I can compile

### Git behavior

_No response_

### Steps to reproduce 🕹

1. I have checked out the `byron/gitoxide` repository (using vscode)
2. I have run any `cargo` command anywhere in the repository

I suspect this is because the repo uses symlinks to reduce copying.

EliahKagan:
Fixes #1310

**This replaces the symbolic link** at `gix-packetline-blocking/src` with a copied source tree from `gix-packetline/src`, with headers prepended to state that they are copies and, for each file, what file it is a copy of (other than that header).

The headers are in the format suggested in https://github.com/Byron/gitoxide/issues/1310#issuecomment-2039154003, except that a blank line (i.e., two newlines rather than one) appears after each one. In files where there are already `//!` comments at the top, I am unsure if a blank line is correct, or if it should instead be omitted or a second line of just `//!` added to signify a paragraph break. Currently there is only one such file, `gix-packetline/src/lib.rs`.

**This also adds a shell script**, `etc/copy-packetline.sh`, which was used to perform the copy, and can be used at any time to update the copy when changes are made in `gix-packetline/src`. It can be invoked directly, or by running `just copy-packetline`. Both approaches work even if `gix-packeltine-blocking/src` has is absent or has been checked out as a regular file due to `core.symlinks` being `false` or symbolic links being otherwise unavailable.

However, the latter approach, even though at this time no Rust tool is used, will still print an error message that may mislead the person running it into thinking it has failed. It shows the same message as in https://github.com/Byron/gitoxide/issues/1310#issuecomment-2040299934. But this message actually *always* happens whenever a `just` command is run when the project structure is unacceptable to `cargo`. The cause is:

https://github.com/Byron/gitoxide/blob/68d1a29e8e497fa130e5733d1ecb6b7694a489ca/justfile#L193

It seems to me that it should be possible to keep that message from being printed for recipes that don't rely on it. But I think that ought be considered outside the scope of this PR.

**The script is used on CI to check that it has no effect.** For now I have added three jobs with a matrix strategy, but this is unnecessary to check that the script has been run if needed; for that we just need one. Instead, running it on three platforms checks that the script works on all of them (I don't have a macOS system). Because it does not test the full range of the script's functionality--as noted below, the script implements specific rules for when it is willing to make the copy--I also tested the script, I believe thoroughly, locally on Windows and Ubuntu.

I've kept all three jobs so if the script has to be changed during further work on this PR then they will be there for testing. I think all but the `ubuntu-latest` job can ultimately be removed.

**The script tries to refuse only when its effects could lead to data loss or confusion.** The rules it follows, for the target, are implemented and commented in the `check_target_dir` shell function. Basically:

- If nothing at all exists at the target location `gix-packetline-blocking/src` on disk, then always proceed. This is safe, and it provides a simple way to get to a state where the script will regenerate the files: deleting the entire subtree indicates a willingness for it to be replaced.
- If something exists at the target on disk and there is no merge in progress, then proceed so long as all changes at or under the target location are staged and no ignored files are present there.
- If something exists at the target location on disk and there is a merge in progress (whether due to a conflict or a deliberately uncommitted merge), then proceed so long as *there are no changes of any kind* an the target location (not even staged changes) and no ignored files are present there.

The script does not stage its own changes. This helps avoid data loss, makes it easy to keep changes from the script identifiable by staging and then running it, and makes the script suitable for use in a pre-commit hook if desired.

**The script decides between `\n` and `\r\n`** for the line endings of each file's prepended header by checking how the first line of the file being copied ends. I believe this is a better approach that analyzing the file as a whole, since if a file is to be inconsistent, it ought not to be made more inconsistent by having more separate sections with different line endings. Because various platforms, git configurations, and approaches to editing files affect line endings, I think there is no reasonable way to decide this besides to look at the files themselves. They are checked in a way that overcomes the strong tendency of tools in Cygwin-like environments (including MSYS2 and "Git Bash") to claim that `\r\n` is merely `\n`.

**The script performs a single pass**, adding the headers at the time the files are written, rather than copying the files and then modifying them. This has some advantages, which *might* justify retaining this approach over others that have been discussed, as detailed below.

**It may be that this shell script should be partly or wholly replaced with a tool coded in Rust.** I believe that was the idea in https://github.com/Byron/gitoxide/issues/1310#issuecomment-2040323536: that a shell script would check if it is safe and reasonable to copy the file and, if so, do so and then invoke a Rust tool to prepend the headers. The reason not to write the whole thing in Rust was that it can fail initially if there is no symlink or directory tree. I embarked on writing the script with the idea that I would write the whole thing as a shell script to get clear on exactly what it should do, then pare down the script, replacing the header-prepending code with a Rust tool.

As I wrote the script, I found that it was unexpectedly tricky, relative to the apparent simplicity of the task, to implement it as a sufficiently portable shell script that was robust to rare but plausible oddities. That made me think as much of it should be turned into Rust as possible. But I persisted in implementing the whole thing as a shell script initially, because I also noticed some advantages of it being implemented that way. I now suspect those advantages are sufficient to justify it remaining as a shell script in its entirety:

- Copying the files and then prepending their headers runs the risk of having copies without headers, that appear to be files that are reasonable to edit and that are easy to edit by accident instead of the correct ones, if the task is interrupted or fails. Failure is likely as long as running a separate Rust tool requires `cargo` to be able to run and see files where they're supposed to be everywhere in the workspace; it may fail due to conceptually unrelated breakages in any crate.
- Copying the files and then prepending their headers seems like it will introduce delays, and occasional minor annoyances, due to language servers (as used in editors and IDEs) triggering an analysis of the files as they are originally created, and then *again* when they are edited. The delay in between may be enough for this to happen, and slight experimentation suggests this is a real issue at least on Windows, but I do not really know for sure how big of a problem this would be. (In particular, this sort of thing could be even further ameliorated by avoiding removing an existing tree before updating it, and I did not go that far.)
- The script is very fast on CI, which runs the script directly rather than through `just`, because no Rust toolchain has to be installed.
- When one is making frequent edits inside `gix-packetline/src`, one might wish to perform the updates automatically using a pre-commit hook. Having them done by a script that doesn't depend on other state of the project benefits this, mainly in that one could leave the hook enabled all the time even when making other unrelated changes that don't require `cargo` to accept the state of the workspace.

I am inclined to find those factors persuasive when taken together, but I am not certain. In particular, this should be weighed against one of the benefits of having as much of this logic in Rust as possible: more gitoxide developers are likely to be able to work efficiently with the code, since gitoxide is a Rust project. (On the other hand, gitoxide does rely substantially on shell scripts as fixtures.) In addition, while this is the sort of thing one would think a shell script would be great for, it actually incurs some complexities due to limitations related to portability that I believe would be absent in Rust code. Ways the script is weird or cumbersome in service of portability are detailed below.

Although I did use CI checks during development (rebasing onto a branch of my fork in which CI runs on all branches), I've decided not to deliberately push each commit separately in this PR as I often do in GitPython. I'd be pleased to do something that like on request. The reasons I've refrained are that I don't know if it would be welcome here, with the more involved CI checks, and that I wonder if I may need to rebase this anyway, as I think this is a case where messages in the conventional commits style aren't called for, but I'm not sure.

---

Cygwin-like environments on Windows, including MSYS2 and the MSYS2-ish "Git Bash" environment, are weird, and some of the weirdness, as mentioned above and as commented in the script, constrains the approaches that can be taken to distinguish `\r\n` from mere `\n`, especially since the script should avoid using anything that may not be available on other systems that have nothing to do with Windows. This is covered in full in the big comment in `first_line_ends_crlf`, and thus not detailed here.

Bash offers a number of nice features beyond what are standardized for `sh` by POSIX. But many of them were introduced in Bash 4, which macOS does not and probably never will ship. This needs to work on macOS systems where no newer Bash has been installed; furthermore, most scripts in the project specify `/bin/bash` as the interpreter, and it could be confusing to change that to `/usr/bin/env bash` just for this. (I have tried to adhere to the style of other scripts.) Before major version 4, Bash lacks recursive globbing (`globstar`), the ability to have the rightmost command in a pipeline affect the calling environment (`lastpipe`), and the ability to check the exit status of a command run in process substitution (process substitution would otherwise be a workaround for not having `lastpipe`).

I assume Bash 3 or later, and the existence of command-line tools that are required by POSIX (in addition, I try to avoid using any that are, in practice, ever absent; for example, `file` is often not installed in Docker images). This is with one notable exception: I assume that `find`, which I'm using because I don't have recursive `**` globs, supports **the `-print0` action**; if it does not, then the script will fail (but signal the failure properly, due to `set -o pipefail`, which is supported even in Bash 3).

The contents of this repository are generally trusted, and I doubt anyone will deliberately create a file with a newline character in its name (which Windows can't tolerate, but other systems can) without good reason. But:

- There may be a reason to have such files, for the purpose of testing gitoxide (or its interaction with other tools such as `gpg` or `ssh`) in their presence.
- They can be created by accident.
- Assuming the full range of possible filenames cannot occur goes wrong more often than one tends to predict.

Of course, this can be made fully portable by using the `-exec` action. But this has a couple of problems:

- It makes lots of subprocess invocations (or incurs the further complexity of *also* implementing a loop to receive multiple paths from the `+` form).
- It is hard to portably stop the entire traversal on the first failure.
- I did it that way first and the script was more complicated, longer, and harder to read.

**Fortunately `-print0` seems in practice to be very portable.** GNU findutils, BusyBox `find` (as in minimal environments or Alpine Linux), and the `find` command in BSD systems, including macOS, all support it, and have supported it for a long time.

It may be that there is a much simpler and more elegant way to do this than I think, even in a shell script. Ordinarily, when writing an explicit loop in a shell script, one should consider alternatives (and I have considered some). However, before proceeding with an idea to simplify the script, the general problem that most Unix-style tools that conceptually operate on *text* do not properly distinguish between `\n` and `\r\n` in Cygwin-like environments should be considered. If the cumbersome state of the script ought to be ameliorated, then I suspect that replacing it with Rust may be the overall best approach. I am also curious if there is some way to do the whole thing in Rust yet let it work even if `cargo` frowns on the state of the workspace.

### DIFF
--- a/etc/copy-packetline.sh
+++ b/etc/copy-packetline.sh
@@ -51,7 +51,10 @@ function check_target_dir () {
     fi
   else
     # We can lose data if anything of value at the target is not in the index.
-    # (Even unstaged deletions, for we can forget what was and wasn't deleted.)
+    # (This includes unstaged deletions, for two reasons. One is that we could
+    # lose track of which files had been deleted. More importantly, replacing a
+    # staged symlink or regular file with an unstaged directory is shown by
+    # git-status as only a deletion, even if the directory is non-empty.)
     if target_dir_status | grep -q '^.[^ ]'; then
       fail 'target exists, with unstaged changes or ignored files'
     fi
--- a/etc/copy-packetline.sh
+++ b/etc/copy-packetline.sh
@@ -34,7 +34,7 @@ function merging () {
 }
 
 function target_dir_status () {
-  git status --short --ignored=traditional -- "$target_dir" ||
+  git status --porcelain --ignored=traditional -- "$target_dir" ||
     fail 'git-status failed'
 }
 
--- a/etc/copy-packetline.sh
+++ b/etc/copy-packetline.sh
@@ -2,9 +2,9 @@
 
 set -euC -o pipefail
 
-readonly source_dir='gix-packetline/src'
-readonly destination_parent_dir='gix-packetline-blocking'
-readonly destination_dir="$destination_parent_dir/src"
+readonly input_dir='gix-packetline/src'
+readonly output_parent_dir='gix-packetline-blocking'
+readonly output_dir="$output_parent_dir/src"
 
 function fail () {
   printf '%s: error: %s\n' "$0" "$1" >&2
@@ -14,7 +14,7 @@ function fail () {
 function chdir_toplevel () {
   local root_padded root
 
-  # Find the working tree's root. (Padding is for the trailing-newline case.)
+  # Find the working tree's root. (Padding covers the trailing-newline case.)
   root_padded="$(git rev-parse --show-toplevel && echo -n .)" ||
     fail 'git-rev-parse failed to find top-level dir'
   root="${root_padded%$'\n.'}"
@@ -25,7 +25,7 @@ function chdir_toplevel () {
 function merging () {
   local git_dir_padded git_dir
 
-  # Find the .git directory. (Padding is for the trailing-newline case.)
+  # Find the .git directory. (Padding covers the trailing-newline case.)
   git_dir_padded="$(git rev-parse --git-dir && echo -n .)" ||
     fail 'git-rev-parse failed to find git dir'
   git_dir="${git_dir_padded%$'\n.'}"
@@ -33,30 +33,30 @@ function merging () {
   test -e "$git_dir/MERGE_HEAD"
 }
 
-function target_dir_status () {
-  git status --porcelain --ignored=traditional -- "$destination_dir" ||
+function output_dir_status () {
+  git status --porcelain --ignored=traditional -- "$output_dir" ||
     fail 'git-status failed'
 }
 
-function check_destination_dir () {
-  if ! test -e "$destination_dir"; then
-    # The target does not exist on disk, so nothing will be lost. Proceed.
+function check_output_dir () {
+  if ! test -e "$output_dir"; then
+    # The destination does not exist on disk, so nothing will be lost. Proceed.
     return
   fi
 
   if merging; then
-    # In a merge, it would be confusing to replace anything at the target.
-    if target_dir_status | grep -q '^'; then
-      fail 'target exists, and a merge is in progress'
+    # In a merge, it would be confusing to replace anything at the destination.
+    if output_dir_status | grep -q '^'; then
+      fail 'output location exists, and a merge is in progress'
     fi
   else
-    # We can lose data if anything of value at the target is not in the index.
-    # (This includes unstaged deletions, for two reasons. One is that we could
-    # lose track of which files had been deleted. More importantly, replacing a
+    # We can lose data if anything of value at the destination is not in the
+    # index. (This includes unstaged deletions, for two reasons. We could lose
+    # track of which files had been deleted. More importantly, replacing a
     # staged symlink or regular file with an unstaged directory is shown by
     # git-status as only a deletion, even if the directory is non-empty.)
-    if target_dir_status | grep -q '^.[^ ]'; then
-      fail 'target exists, with unstaged changes or ignored files'
+    if output_dir_status | grep -q '^.[^ ]'; then
+      fail 'output location exists, with unstaged changes or ignored files'
     fi
   fi
 }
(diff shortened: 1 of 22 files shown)


etc/copy-packetline.sh
```diff
@@ -0,0 +1,153 @@
...
+}
+
+function make_header () {
+  local source_file endline
+
+  source_file="$1"
+  endline="$2"
+
+  # shellcheck disable=SC2016  # The backticks are intentionally literal.
+  printf '//! DO NOT EDIT - this is a copy of %s. Run `just copy-packetline` to update it.%s%s' \
```
Byron, line 90: I just realized that this crate gets published, and we'd not want actual docstrings for this, but just a comment.

Byron (commented):
Thanks a million for this improvement!

Please note that I just took a quick look and didn't read *any* of the PR description yet, which I will get to in a bit.

etc/copy-packetline.sh
EliahKagan: Is it okay for a normal `//` comment to precede a top-level `//!` documentation comment? (There is one file in that subtree, `lib.rs`, where that would happen.) If so, then this is very easy to change; if not, it should still be feasible to fix without too much trouble.

Byron: I think it's OK, and validated it.

EliahKagan: Great--I'll make that change shortly.

EliahKagan: I have changed the `//!` comments to `//` (modifying and rerunning the script).

EliahKagan:
Building on [d25e5d3](https://github.com/Byron/gitoxide/pull/1340/commits/d25e5d3781e0faca0b31bb6ea58d3057b3a5f16a), another option for names could be `input` and `output`.

Byron (approved):
Thanks a lot for this wonderful work, leading to a very impactful fix! Finally, we can use `cargo` on Windows again.

I only have made a couple of smaller adjustments. One change that I am not perfectly certain about is the removal of the `continue-on-error: true` line, which I thought was a left-over rather than intentional. If that's wrong, let's fix it before the merge.

> Although I did use CI checks during development (rebasing onto a branch of my fork in which CI runs on all branches), I've decided not to deliberately push each commit separately in this PR as I often do in GitPython. I'd be pleased to do something that like on request. The reasons I've refrained are that I don't know if it would be welcome here, with the more involved CI checks, and that I wonder if I may need to rebase this anyway, as I think this is a case where messages in the conventional commits style aren't called for, but I'm not sure.

I'd be happy to *keep* the current style of pushing once, as otherwise I am pretty sure CI would be blocked for a long time which could be an impediment.

> If the cumbersome state of the script ought to be ameliorated, then I suspect that replacing it with Rust may be the overall best approach. I am also curious if there is some way to do the whole thing in Rust yet let it work even if `cargo` frowns on the state of the workspace.

I understand the reasoning for the script, find it very elaborate and well made, and chose to consider it read-only. It's not that it's unmaintainable, it's just that it has one purpose that it fulfils so we can move on to greener pastures :).

Byron:
> Building on [d25e5d3](https://github.com/Byron/gitoxide/pull/1340/commits/d25e5d3781e0faca0b31bb6ea58d3057b3a5f16a), another option for names could be `input` and `output`.

I am done with the refactor and encourage you to bring in last minute changes, also to the names - the ones proposed are shorter, and they are also not names `target` :).

I plan to merge once CI is complete, but probably only in an hour or so.

EliahKagan:
Thanks, I'll push a change to rename the identifiers shortly.

Byron:
Thank you, much better! 

I also just now realise that the new CI job takes a mere 3 seconds, so running more of these isn't ever going to be an issue.
The only argument I can make to leave it as is, is that it's less noisy with just one job, and less confusing as well since it would otherwise be puzzling why this has to be tested on all platforms (if one knows the intention of the script). Thus, I think I can indeed merge in a moment.

EliahKagan:
> I only have made a couple of smaller adjustments. One change that I am not perfectly certain about is the removal of the `continue-on-error: true` line, which I thought was a left-over rather than intentional. If that's wrong, let's fix it before the merge.

Really I should never have used that, because it was applied to each job in the matrix for the purpose of ensuring that results from all platforms appeared during testing, rather than to specific jobs considered experimental. I should have used `fail-fast: false` instead. The effect of `fail-fast` makes no difference to a matrix strategy of only one job, but I've put that in so that, if the other jobs ever have to be enabled to test a change to the script, then all platforms' results will be produced and shown.

> I also just now realise that the new CI job takes a mere 3 seconds, so running more of these isn't ever going to be an issue. The only argument I can make to leave it as is, is that it's less noisy with just one job, and less confusing as well since it would otherwise be puzzling why this has to be tested on all platforms (if one knows the intention of the script). Thus, I think I can indeed merge in a moment.

I think it is less confusing with just one, yes. If the script will rarely be edited, then I think it's better to have just the ubuntu-latest job for the check.

PR #1341, opened 2024-04-10
Fix typos

szepeviktor:
- make typos config more strict and resilient
- check fixtures
- check dotfiles


gix-filter/tests/fixtures/baseline.sh
```diff
@@ -17,7 +17,7 @@ function repo_assertions() {
 
   git config filter.arrow.clean "$driver clean %f"
   git config filter.arrow.smudge "$driver smudge %f"
-  git config filter.arrow.requred true
+  git config filter.arrow.required true
```
Byron, line 20: That's a good one! It's also interesting that the test doesn't seem to care about it either way.

Byron (approved):
Thanks a lot! Great work - I didn't think the tool could be made to work, but here we are!

szepeviktor:
Glad to contribute.

PR #1342, opened 2024-04-12
Refine and strengthen diagnostics for problematic URLs

EliahKagan:
This adds more unit tests and journey tests for existing error messages about URLs that are often not usable as remotes because they cause problems when used with external commands. It also distinguishes a further subcategory of these URLs, adds both kinds of test cases for them, and ensures they are identified and that the associated error messages are accurate.

For background on what led me to some of the new `Url` methods, see [that comment from an intermediate stage](https://github.com/Byron/gitoxide/blob/f56ad390a5569d0129b7b16632991d18b9ddb4f7/gix-url/src/lib.rs#L183-L189). I had at that time thought the existing ambiguous method (and the new ambiguous method I added) should be removed, but now I think that may not be the case, though I've moved code over to using the new ones.

This may be ready, though I strongly suspect some revisions may be needed, stylistically, in naming, or otherwise.

Two things I had thought of while working on this but that seemed better to leave out of this PR, unless requested, were:

- Fuzz testing already touches on some of the kinds of inputs that are already handled properly, for which this PR only adds more unit and journey tests. I am unsure if fuzz testing should be extended to cover the other cases that this PR addresses, and if so, if there is anything this PR should do for that.
- As before, checking for strange paths is done separately from the gix-url facilities related to that. This doesn't seem like a problem, but I am curious if something ought to be documented to avoid regressions and/or to advise application developers. I haven't identified such an area though; the documentation seems to cover the issues.

The most important points and considerations are not those, but rather are presented outside this PR, such as in code comments/documentation and commit messages.

### DIFF
--- a/gix-transport/src/client/blocking_io/ssh/program_kind.rs
+++ b/gix-transport/src/client/blocking_io/ssh/program_kind.rs
@@ -2,6 +2,8 @@ use std::{ffi::OsStr, io::ErrorKind};
 
 use bstr::{BString, ByteSlice, ByteVec};
 
+use gix_url::ArgumentSafety::*;
+
 use crate::{
     client::{ssh, ssh::ProgramKind},
     Protocol,
@@ -60,23 +62,13 @@ impl ProgramKind {
                 }
             }
         };
-        let host_maybe_with_user_as_ssh_arg = match url.user() {
-            Some(user) => {
-                // FIXME: See the fixme comment on Url::user_argument_safe() about its return type.
-                if url.user_argument_safe() != Some(user) {
-                    return Err(ssh::invocation::Error::AmbiguousUserName { user: user.into() });
-                }
-                let host = url.host().expect("present in ssh urls");
-                format!("{user}@{host}")
-            }
-            None => {
-                let host = url
-                    .host_argument_safe()
-                    .ok_or_else(|| ssh::invocation::Error::AmbiguousHostName {
-                        host: url.host().expect("ssh host always set").into(),
-                    })?;
-                host.into()
-            }
+
+        let host_maybe_with_user_as_ssh_arg = match (url.user_as_argument(), url.host_as_argument()) {
+            (Usable(user), Usable(host)) => format!("{user}@{host}"),
+            (Absent, Usable(host)) => host.into(),
+            (Dangerous(user), _) => Err(ssh::invocation::Error::AmbiguousUserName { user: user.into() })?,
+            (_, Dangerous(host)) => Err(ssh::invocation::Error::AmbiguousHostName { host: host.into() })?,
+            (_, Absent) => panic!("BUG: host should always be present in SSH URLs"),
         };
 
         // Try to force ssh to yield English messages (for parsing later).
--- a/gix-transport/src/client/blocking_io/ssh/program_kind.rs
+++ b/gix-transport/src/client/blocking_io/ssh/program_kind.rs
@@ -72,9 +72,9 @@ impl ProgramKind {
             (_, Absent) => panic!("BUG: host should always be present in SSH URLs"),
         };
 
-        // Try to force ssh to yield English messages (for parsing later).
         Ok(prepare
             .arg(host_maybe_with_user_as_ssh_arg)
+            // Try to force ssh to yield English messages (for parsing later).
             .env("LANG", "C")
             .env("LC_ALL", "C"))
     }
--- a/gix-transport/src/client/blocking_io/ssh/program_kind.rs
+++ b/gix-transport/src/client/blocking_io/ssh/program_kind.rs
@@ -60,8 +60,12 @@ impl ProgramKind {
                 }
             }
         };
-        let host_as_ssh_arg = match url.user() {
+        let host_maybe_with_user_as_ssh_arg = match url.user() {
             Some(user) => {
+                // FIXME: See the fixme comment on Url::user_argument_safe() about its return type.
+                if url.user_argument_safe() != Some(user) {
+                    return Err(ssh::invocation::Error::AmbiguousUserName { user: user.into() });
+                }
                 let host = url.host().expect("present in ssh urls");
                 format!("{user}@{host}")
             }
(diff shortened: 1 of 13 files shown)


gix-transport/src/client/blocking_io/ssh/program_kind.rs
```diff
@@ -60,23 +62,21 @@ impl ProgramKind {
...
-            None => {
-                let host = url
-                    .host_argument_safe()
-                    .ok_or_else(|| ssh::invocation::Error::AmbiguousHostName {
-                        host: url.host().expect("ssh host always set").into(),
-                    })?;
-                host.into()
-            }
+
+        let host_maybe_with_user_as_ssh_arg = match (url.user_as_argument(), url.host_as_argument()) {
```
Byron, line 66: This is a great use of `match` and one of the main reasons Rust is so powerful! Never again can one miss handling a possible program state, as long as one remembers to match on them that is :).

Byron (approved):
This is a fantastic PR and it's like you have been doing Rust for a long time.

I love how you picked up on the commenting style to selectively incorporate changelog messages. On that note, I have a removed a method that was certainly mentioned in a commit, and I accept that the changelog will now be slightly misleading without rewriting history. It's a systematic shortcoming.

Further, the addition of `ArgumentSafety` is a great one which leads to nicer error messages than was possible before. The previous implementation didn't handle this very well as a the possible states weren't correctly represented.

I will now see to creating a release once CI is green.

EliahKagan:
Now that the advisory https://github.com/Byron/gitoxide/security/advisories/GHSA-98p4-xjmm-8mfh is published, I've updated this PR description to link to it and provide a small amount of additional information. In the near future, I'll likely edit it further to add more information, including about the approach taken, as well as to give it a clearer title.

I hope it's okay that I have also changed the "Affected versions" field of the advisory itself (the repository-local advisory, since as far as I know it is not yet in advisory databases) from `<= 0.41.3` to `< 0.42.0`. This makes clearer that there are no existing or forthcoming versions of the `gix-transport` crate lower than 0.42.0 that contain a fix. If this change is not wanted, it can of course be undone.

I am uncertain if there are any other crates--such as related to the [`gix`](https://github.com/Byron/gitoxide/releases/tag/gix-v0.62.0) CLI tool?--that should be listed as affected. If so, they could be added to the advisory. This can be done without replacing, [`gix-transport`](https://github.com/Byron/gitoxide/releases/tag/gix-transport-v0.42.0), which is the most important affected crate.

Byron:
> I hope it's okay that I have also changed the "Affected versions" field of the advisory itself (the repository-local advisory, since as far as I know it is not yet in advisory databases) from `<= 0.41.3` to `< 0.42.0`. This makes clearer that there are no existing or forthcoming versions of the `gix-transport` crate lower than 0.42.0 that contain a fix. If this change is not wanted, it can of course be undone.

Yes, thanks a lot for catching this. I somehow behaved like the bound of unpatched versions can't be changed.

> I am uncertain if there are any other crates--such as related to the [`gix`](https://github.com/Byron/gitoxide/releases/tag/gix-v0.62.0) CLI tool?--that should be listed as affected. If so, they could be added to the advisory. This can be done without replacing [`gix-transport`](https://github.com/Byron/gitoxide/releases/tag/gix-transport-v0.42.0), which is the most important affected crate: the advisory could list both.

Right, I wasn't aware that multiple items can be added. Instead, I added the affected items downstream as a 'Tip' right after the summary. I have corrected that by adding two more items accordingly and removed the 'Tip'.

EliahKagan:
I see the [RUSTSEC advisory](https://rustsec.org/advisories/RUSTSEC-2024-0335.html) is published! 🎉

I've updated the pull request description to include a link to that, but I still do plan to expand it further later, as noted.

EliahKagan:
It looks like I should have included [CWE-88](https://cwe.mitre.org/data/definitions/88.html) in the weaknesses list in https://github.com/Byron/gitoxide/security/advisories/GHSA-98p4-xjmm-8mfh. Its [Example 1](https://cwe.mitre.org/data/definitions/88.html#Demonstrative_Examples) clarifies that it encompasses this kind of vulnerability. That was also the (only) CWE listed in the earlier https://github.com/advisories/GHSA-rrjw-j4m2-mf34. I can edit the repository advisory to add [CWE-88](https://cwe.mitre.org/data/definitions/88.html), and it looks like RUSTSEC advisories don't list CWEs so nothing has to be added there.

But should [CWE-88](https://cwe.mitre.org/data/definitions/88.html) be added as a second CWE, or should it replace [CWE-77](https://cwe.mitre.org/data/definitions/77.html)? [CWE-77](https://cwe.mitre.org/data/definitions/77.html) is [often overused](https://cwe.mitre.org/data/definitions/77.html#Vulnerability_Mapping_Notes), but I think this is mostly because of uses where [CWE-78](https://cwe.mitre.org/data/definitions/78.html) should've been used instead. It seems to me that [CWE-77](https://cwe.mitre.org/data/definitions/77.html) still applies and *may* not be completely covered by [CWE-88](https://cwe.mitre.org/data/definitions/88.html), but I'm not sure.

Byron:
To me, as a layperson (strangely enough), and judging only by the title, they all seem pretty similar and I *feel* that they all somewhat apply. Since CWE assignments are probably useful for people to find (or better understand) an advisory, maybe having multiple slightly overlapping CWEs isn't the worst thing to do.

But I am definitely not sure either.

EliahKagan:
That sounds reasonable. I'll try to look a bit more into how, if at all, [CWE-77](https://cwe.mitre.org/data/definitions/77.html) and [CWE-88](https://cwe.mitre.org/data/definitions/88.html) tend to be used together, then I'll add [CWE-88](https://cwe.mitre.org/data/definitions/88.html) while also keeping [CWE-77](https://cwe.mitre.org/data/definitions/77.html) unless I've found a reason to do otherwise.

riking:
> Although ssh supports aliases for hosts, even if git-upload-pack could be made an alias, that is made difficult by the URL-encoding transformation.

Note that DNS Search Domains could turn that into a valid domain: `git-upload-pack.corp.example.com`

However, it will be difficult to place a malicious host in a machine's search domains, as the search domains typically have common control with the machine they are configured on.

PR #1349, opened 2024-04-21
fix: don't ignore packed ref deletion in non-default transaction mode

yuja:
Before, `Change::Delete` edits weren't propagated to packed refs if the mode was `DeletionsAndNonSymbolicUpdates`/`RemoveLooseSourceReference`.


Byron (approved):
Thanks a lot, I love the test!

Given how simple it is, I wonder how many more tests might be missing, despite my best attempts to put down a pole in every important spot.

I plan to rebase the PR onto `main` once CI is fixed there, and then merge.

PR #1356, opened 2024-05-06
Use v2 of taiki-e/install-action

### LINKED ISSUE #1355: CI install-action now fails on Windows, can't find .cargo/bin
### Current behavior 😯

Sometime in the last couple of weeks and probably in the last few days, the `Run taiki-e/install-action@v1` step in the `test-fast` job on `windows-latest` fails whenever it is run, showing:

```text
Run taiki-e/install-action@v1
  with:
    tool: nextest
  env:
    CARGO_TERM_COLOR: always
    CLICOLOR: 1
    CARGO_HOME: C:\Users\runneradmin/.cargo
    CARGO_INCREMENTAL: 0
    CACHE_ON_FAILURE: false
info: installing nextest@latest
info: downloading https://get.nexte.st/latest/windows-tar
tar: C\:\\Users\runneradmin/.cargo/bin: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
curl: ([2](https://github.com/EliahKagan/gitoxide/actions/runs/8969527945/job/24655063846#step:8:2)3) Failure writing output to destination
curl: (2[3](https://github.com/EliahKagan/gitoxide/actions/runs/8969527945/job/24655063846#step:8:3)) Failure writing output to destination
curl: (23) Failure writing output to destination
curl: (23) Failure writing output to destination
curl: (23) Failure writing output to destination
curl: (23) Failure writing output to destination
Error: Command failed: bash --noprofile --norc D:\a\_actions\taiki-e\install-action\v1/main.sh
```

I don't know specifically what is causing this, but experiments within my fork strongly suggest that it can be solved by upgrading `taiki-e/install-action` to a current version of the action.

I'm opening this issue rather than only a PR because there is more than one reasonable approach that could be taken to do this, and because I had needed to do some more experimentation, but I have opened #1356 to fix this.

### Expected behavior 🤔

The action should successfully complete, installing `nextest` as intended. This is the behavior on all three `test-fast`-tested platforms until very recently, and remains the behavior except on Windows.

### Git behavior

Not applicable, as this is a CI setup issue not identifiably related to any problem in gitoxide's code or test suite.

### Steps to reproduce 🕹

Because the job has not run in this upstream gitoxide repository since the problem began, and because CI is not configured to run on feature branches except in PRs, so far this has not occurred on this upstream repository's main branch due to the `push` trigger, but I have reproduced it in two ways in my fork as well as in this upstream repository with the `pull_request` trigger:

- [On a branch in my fork where I have modified the workflow](https://github.com/EliahKagan/gitoxide/actions/runs/8969527945/job/24655063846) to trigger even though it is a feature branch. Notice how, when [further modified to use `v2` rather than `v1`](https://github.com/EliahKagan/gitoxide/actions/runs/8969841238/job/24655075060) as the major version of the action, the problem goes away.
- [On the main branch of my fork, by rerunning it.](https://github.com/EliahKagan/gitoxide/actions/runs/8795078265/job/24655054842) Notice how multiple runs fail, so it is not a one-off problem, and how the first run, from when I synchronized the fork with the most recent push (as of this writing) to the upstream repository, succeeded.
- In this upstream repository, on [#1357](https://github.com/Byron/gitoxide/pull/1357), from the `pull_request` trigger, [in this failing job](https://github.com/Byron/gitoxide/actions/runs/8977522195/job/24656429375?pr=1357). Note that no corresponding failure happens in [#1356](https://github.com/Byron/gitoxide/pull/1356), even though [the job runs there](https://github.com/Byron/gitoxide/actions/runs/8977363125/job/24655965698?pr=1356), because [#1356](https://github.com/Byron/gitoxide/pull/1356) provides a fix for this issue.

EliahKagan:
Fixes #1355 

This switches from major version 1 to major version 2 of `taiki-e/install-action` to fix the new Windows failure described in #1355.

For consistency, and because it is preferable to use new versions over very old versions provided they are stable and there is no known reason to prefer the much older version, this changes it in both job definitions that use it, even though the problem only occurs in one generated job (the Windows job) from one of the job definitions (the other one does not generate a Windows job).

However, this is *not* a general GitHub Action upgrading PR. It is intended to make the minimal change, of those that seem both reasonable and non-confusing, to fix that issue and allow CI checks to pass again. Although I think it would be good to use later versions of some other actions, some of which are quite old, I think that automating such upgrades should be considered, and my inclination is to regard that decision beyond the scope of this PR. But see #1357.

As a demonstration that #1355 really does affect his upstream repository and thus needs fixing--since when I originally opened that I had only produced it in my fork--see [this failing job](https://github.com/Byron/gitoxide/actions/runs/8977522195/job/24656429375?pr=1357) in #1357. No corresponding failure happens here, even though [the job does run here](https://github.com/Byron/gitoxide/actions/runs/8977363125/job/24655965698?pr=1356), because this PR fixes the issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
       - run: |
           # it should never be a failure not to get the caches, as they can be regenerated.
           git lfs fetch && git lfs checkout || true
-      - uses: taiki-e/install-action@v1
+      - uses: taiki-e/install-action@v2
         with:
           tool: nextest
       - name: "Test (nextest)"
@@ -102,7 +102,7 @@ jobs:
         with:
           toolchain: stable
           targets: ${{ matrix.target }}
-      - uses: taiki-e/install-action@v1
+      - uses: taiki-e/install-action@v2
         with:
           tool: cross
       - name: "check"


Byron:
Thanks a lot for the fix, and for going through such lengths to verify it.

Issue #144, opened 2021-07-30
Dependabot frequency


Byron:
### Discussed in https://github.com/Byron/gitoxide/discussions/143

<div type='discussions-op-text'>

<sup>Originally posted by **edward-shen** July 30, 2021</sup>
While keeping up to date is a good thing, frequent updates create a lot of noise and can be annoying. Perhaps we can configure it to once a week, or once a month? 

We can take larger period between dependabot checks because we already run `cargo deny` in CI, so any urgent issues will be noticed quickly, where we manually update.

https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#scheduleinterval</div>

PR #1357, opened 2024-05-07
Use Dependabot version updates but just for GitHub Actions

EliahKagan:
A number of reusable actions, such as `actions/checkout`, are in old major versions, including versions that use old deprecated versions of Node.js, such as Node 16 (rather than the current Node 20).

Dependabot version updates were disabled for Rust crates in #143 and #144, with other means used to keep things up to date. This does not reverse that even partially, as it enables version updates only for GitHub Actions (which it has not been used on before).

Because most actions are specified with major version tags that get reset to the latest stable minor/patch version under that major version, the actual frequency of updates should not be anywhere near daily. Nonetheless, it may make sense to further refine this configuration, such as to check for updates less often as well as to group multiple updates into a single PR when doing so.

If this were merged before #1356 and then a Dependabot version update for `taiki-e/install-action` were merged, then #1356 would be superseded. (That is one of the reasons I opted to open #1355 rather than just describing the issue in #1356.) However, even if you know you will want to merge this, my recommendation is to merge #1356 first provided CI passed on it and it is otherwise acceptable, because it addresses a specific current breakage, while this is broader (though it should prevent future such breakages).


EliahKagan:
As a brief update on this, I'm testing what seems to me to be a reasonable use of grouped Dependabot version updates for GitHub Actions, and I expect to be able to add a commit to this PR for that in about ten minutes. That forthcoming second commit can then be kept if wanted, or dropped or reverted otherwise.

Byron (approved):
Thanks a lot, a great idea to automate these 'meta' dependency-updates.

Issue #1319, opened 2024-03-15
Oxidize Radicle-Git


Byron:
Here is the repository: https://app.radicle.xyz/nodes/seed.radicle.xyz/rad:z6cFWeWpnZNHh9rUW8phgA3b5yGt

### Required Features

Analysis TBD

### Notes

The repository contains a lot of extra-functionality that also isn't present in `git2`. This should probably be available from `gix` as long as it's not Radicle specific.

* Re-implement some basic features that were lacking in `git2` in `gix`
* Use `gix` where possible, slowly replacing `git2`
* Remove `git2` once there is nothing left.

Byron:
Thank you for clarifying that. From my experience, using `gix` in the frontend has advantages in terms of compatibility at the very least, but I also hear that recently `libgit2` really ramped up its contributions so these issues might even go away in the mid-term.

And I'd definitely love to *finally* come up with a fetch-API that is easy to use but not unnecessarily limiting, and can thus work for you as well. If that was the case, I think you could start tracking a more recent version which might ultimately pay off.

If you would link the latest code that uses the `Delegate` here, I should be able to see what can or can't be done with higher-level APIs and fix these. Sharing what it specifically was that prevented the adoption last time you tried would probably certainly be helpful to me as well.

Thanks again!

FintanH:
> Thank you for clarifying that. From my experience, using gix in the frontend has advantages in terms of compatibility at the very least, but I also hear that recently libgit2 really ramped up its contributions so these issues might even go away in the mid-term.

Compatibility in which sense? :) In my mind, the `radicle-fetch` code is quite isolated and the only conversion points are OIDs and refnames, which are already in place.

> And I'd definitely love to finally come up with a fetch-API that is easy to use but not unnecessarily limiting, and can thus work for you as well. If that was the case, I think you could start tracking a more recent version which might ultimately pay off.

Interesting. I can have a look again because, admittedly, it was a while that I looked and then got swept away by other tasks. 

> If you would link the latest code that uses the Delegate here, I should be able to see what can or can't be done with higher-level APIs and fix these. Sharing what it specifically was that prevented the adoption last time you tried would probably certainly be helpful to me as well.

I should have taken notes when I explored trying to use the updated version, but unfortunately my foresight is not 20/20 :sweat_smile: 

So the main entry point to using the gix fetch code are the following helpers:

[...]

PR #1363, opened 2024-05-07
tree -> index diff for status

Byron:
Based on #1350

----

### diff-correctness → `gix-status` → gix reset

----

Improve `gix status` to the point where it's suitable for use in `reset` functionality.
Leads to a proper worktree reset implementation, eventually leading to a high-level reset similar to how git supports it.

### Architecture

The reason this PR deals quite a bit with `gix status` is that for a safe implementation of `reset()` we need to be sure that the files we *would* want to touch don't don't carry modifications or are untracked files. In order to know what would need to be done, we have to diff the `current-index with target-index`. The set of files to touch can then be used to lookup information provided by `git-status`, like worktree modifications, index modifications, and untracked files, to know if we can proceed or not. Here is also where the reset-modes would affect the outcome, i.e. what to change and how.

This is a very modular approach which facilitates testing and understanding of what otherwise would be a very complex algorithm. Having a set of changes as output also allows to one day parallelize applying these changes.

This leaves us in a situation where the current `checkout()` implementation wants to become a fastpath for situations where the reset involves an empty tree as source (i.e. create everything and overwrite local changes).

### Extra Tasks

Out-of-band tasks that just should finally be done, with potential for great impact.

* [ ] support for [`hasconfig`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-codehasconfigremoteurlcode)  as part of `resolve_includes()` without actual lookahead.

### Tasks 

* [ ] diff index with index to learn what we would want to do in the worktree, or alternatively, 
      diff tree with index (with reverse-diff functionality to simulate diff of index with tree), for better performance as it 
      would avoid having to allocate a whole index even though we are only interested in a diff.
     - *Must include rename tracking*.
* [ ] how to make diff results available from status with all transformations applied, to allow user to obtain diffs of any kind? Compare to tree-tree-diff which has that functionality already.
* [ ] update `is_dirty()` and `Submodule::status()` to do full status.

### Status Enables

* `cargo package` and its use of complete status information.
* https://github.com/Byron/built/pull/1
* starship native dirty check (but needs diffstats of worktree-vs-index)
* `built` can get fully-functional is-dirty flag for 'describe()'

### Inbetween

* [ ] sort out #1319 

### Next PR: Reset

* [ ] `reset()` that checks if it's allowed to perform a worktree modification is allowed, or if an entry should be skipped. *That way we can postpone safety checks like --hard*

### Postponed

* [ ] a way to expand sparse dirs (but figure out if this is truly always necessary) - probably not, unless sparse dirs can be empty, but even then no expansion is needed
     - [ ] wire it up in `gix index entries` to optionally expand sparse entries
* [ ] `gix status` with implemented 'porcelain-v2` display mode

### Research
 
* Ignored files are considered expandable and can be overwritten on reset
* How to integrate submodules - probably easy to answer once `gix status` can deal a little better with submodules. Even though in this case a lot of submodule-related information is needed for a complete reset, probably only doable by a higher-level caller which orchestrates it.
* How to deal with various modes like `merge` and `keep`? How to control `refresh`? Maybe partial (only the files we touch), and full, to also update the files we don't touch as part of status? Maybe it's part of status if that is run before.
* figure out how this relates to the current `checkout()` method as technically that's a `reset --hard` with optional overwrite check. Could it be rolled into one, with pathspec support added? 
   - just keep them separate until it's clear that `reset()` performs just as well, which is unlikely as there is more overhead. But maybe it's not worth to maintain two versions over it. But if so, one should probably rename it.
* for `git status`: what about rename tracking? It's available for tree-diffs and quite complex on its own. Probably only needs HEAD-vs-index rename tracking. No, also can have worktree rename tracking, even though it's hard to imagine how this can be fast unless it's tightly integrated with untracked-files handling. This screams for a generalization of the tracking code though as the testing and implementation is complex, but *should* be generalisable.

### Re-learn

* `pathspecs` normalize themselves to turn from any kind of specification into repo-root relative patterns.
* attribute/ignore file sources are naturally relative to the root of the repo, which remains relative (i.e. can be `..` and that root will be always be used to open files like `../.gitignore`, which is useful for display to the user)


EliahKagan:
Regarding [91b6549](https://github.com/Byron/gitoxide/pull/1363/commits/91b65495f14cd58c8b178f359354749b39dc4008), is it required to use PowerShell? Otherwise, `shell: bash` can be used, even on Windows, which gives Git Bash, so anything that is due to PowerShell rather than a POSIX-compatible shell being used may be fixable that way.

Byron:
Thanks for the reminder! Yes, that would have worked too. Overall, I think using the `${{ env.var }}` syntax is more portable and should be preferred. There might be other cases where it's definitely better to change the shell though.

EliahKagan:
Yes, there is nothing wrong with `${{ env.var }}`. The only reason I mentioned `shell: bash` was this bit from the commit message:

> What's worse is that it doesn't seem to fail if something doesn't work as one
would expect.

Achieving the effect of `set -e` or `set -u` is less obvious and automatic in PowerShell than in POSIX-compatible shells, so I figured the ability to use `bash` might be relevant there.

Byron:
> Achieving the effect of `set -e` or `set -u` is less obvious and automatic in PowerShell than in POSIX-compatible shells, so I figured the ability to use `bash` might be relevant there.

And after writing the bit about powershell I also realized that I am so used to `set -euo pipefail` that maybe I forgot how `bash` would have done the same in that situation. But I am not sure how the defaults are set on CI, where one would hope a failing command indeed triggers failure by default.

EliahKagan:
> But I am not sure how the defaults are set on CI, where one would hope a failing command indeed triggers failure by default.

I had meant to replied to this, but it looks like I forgot. This is relevant to https://github.com/GitoxideLabs/gitoxide/pull/2340#issuecomment-3710416306, which reminded me of it, but I think other aspects related to the problem are more central there. So I figured I'd reply here, as I'd meant to. (I'll also reply there, about those aspects I think are more central to what we're discussing there.)

See also #1559, which fixed a bug related to fail-fast behavior inside a script step. (That shouln't be confused with #2342, which fixed a conceptually similar but worse bug that was not caused by script-specific semantics.)

#### The what

PowerShell does not generally treat an external command that fails with a nonzero exit status as an error. So while it can stop on errors, and it can be made to return the exit status of the last command it runs, as far as I know there is no simple foolproof way to *implicitly* cause intermediate external commands in a multi-command script to terminate the script (and indicate failure).

The way PowerShell is run by default on CI does not attempt to do that, both back when this PR was merged and today. For example, consider [this test workflow](https://github.com/EliahKagan/hello-world/blob/18fd8e20eb9b850ed42c2d59232af3e0ed0caec2/.github/workflows/failfast.yml):

```yaml
name: Test if PowerShell fails fast

on: [push, pull_request, workflow_dispatch]

jobs:
  powershell-failfast:
    runs-on: windows-latest

    steps:
      - name: Check out repository
        uses: actions/checkout@v4
      - name: External command reporting failure
        run: |
          echo a >t
          git add t
          echo b >>t
          git diff --exit-code -- t  # Should fail.
          echo $LASTEXITCODE
          git log --oneline -1  # Should succeed, if run.
      - name: Cmdlet reporting an error
        run: |
          cat foo
          echo done
```

As shown in [a test run](https://github.com/EliahKagan/hello-world/actions/runs/20727140513/job/59505379811#step:3:10):

1. The `git diff` command has an exit status of 1, but the first script step that contains it continues--and reports success, since the last command it runs has an exit status of 0.
2. In contrast, since `cat` in PowerShell on Windows is an alias for the `Get-Content` cmdlet, when it fails to open a file, this is an error closer to PowerShell. That fails the step that runs it, and does so immediately, not running the subsequent command.

### The why

This doesn't entirely answer the question of *why* PowerShell is not run by default on GitHub Actions runners (with the `pwsh` or `powershell` values of the `shell` key, or behavior on Windows that defaults to them based on availability in that order) with something like `-e`.

I think the reason is that PowerShell doesn't easily support doing this. My recollection is that it can be done for simple cases, but that this feature in PowerShell deviates from the behavior of POSIX `-e` in significant ways that developers would find unexpected if it were turned on automatically, relating to the distinction in PowerShell between an error and the result of a test.

Actually, POSIX-compatible shells [also have subtleties](https://mywiki.wooledge.org/BashFAQ/105) where the behavior of `-e` is unexpected a significant fraction fo the time. But my understanding is that making PowerShell approximate the behavior of `set -e` neither well-approximates the real behavior of `set -e` nor approximates the effect people tend to assume `set -e` has.

In practice, I don't usually see problems in GHA steps arising from unexpected mismatch between what `set -e` does and what one may intuitively think it does. In contrast, the absence of fail-fast behavior in PowerShell is very much unexpected, when PowerShell is used implicitly on CI for script steps that are conceptually "lists of commands to run in any way." Ironically, this seems not to cause problems in big complex PowerShell script steps, where the semantics of PowerShell are front and center.

PR #1334, opened 2024-04-03
Let nonstandard worktree fixtures work even if Git < 2.37.2

EliahKagan:
This reorders the operations in the [nonstandard worktree fixtures](https://github.com/Byron/gitoxide/blob/4bb8ef5fd05133b87eb1e79a0ac6e45eb66940a2/gix-dir/tests/fixtures/many.sh#L110-L134) to work even when Git is at a lower version than 2.37.2. This makes [`root_enters_directory_with_dot_git_in_reconfigured_worktree_tracked`](https://github.com/Byron/gitoxide/blob/4bb8ef5fd05133b87eb1e79a0ac6e45eb66940a2/gix-dir/tests/walk/mod.rs#L2885) pass on more systems, including common development systems.

When fixture scripts are run on a system whose Git version is lower than [2.37.2](https://github.com/git/git/blob/c2cbfbd2e28cbe27c194d62183b42f27a6a5bb87/Documentation/RelNotes/2.37.2.txt#L24-L26), including currently maintained downstream versions with backported security (but not most other) fixes on some popular OSes (e.g. [Ubuntu 22.04.4 LTS with 2.34.1-1ubuntu1.10](https://launchpad.net/ubuntu/jammy/+source/git/+changelog)), the `gix-dir` nonstandard worktree fixtures had silently failed to stage one of two intended files. They appeared to succeed, but created a repo with only the `seemingly-outside` file, and not the `inside` file.

This is due to the bug https://github.com/git/git/commit/27128996b8 fixed in Git. The nonstandard worktree fixtures had needed Git to support the first condition noted there, where `core.worktree` is the parent of the default worktree, in order to stage both files.

The change made here overcomes the problem by staging before rather than after moving `.git` and setting `core.worktree`. The `git add` command used to stage is adjusted to remain equivalent in its effect. (Committing can be, and is, still done last.)

CI hadn't suffered/shown the problem, since while [GHA ubuntu-latest runners are currently Ubuntu 22.04](https://github.com/actions/runner-images?tab=readme-ov-file#available-images), they have many newer software versions installed on them, [currently including Git 2.43.2](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#tools).

### The original test failure

[...]

PR #1365, opened 2024-05-08
Edit CI and docs to reflect generated archives not needing LFS

EliahKagan:
Since 35439de (#1176), and as mentioned in https://github.com/Byron/gitoxide/pull/1334#pullrequestreview-1975585626 and https://github.com/Byron/gitoxide/pull/1360#pullrequestreview-2043902532, generated archives that formerly required Git LFS are now included in full as ordinary blobs in the `gitoxide` repository. Therefore:

- The documentation changes that were part of [#1360](https://github.com/Byron/gitoxide/pull/1360) are incomplete, because the explanation of *why* it is enough to let CI test changed fixture scripts on `ubuntu-latest` still gives the old reason related to Git LFS.
- Documentation and CI workflows no longer need to use or mention Git LFS where the reason they had done so was related to fixtures.

The changes in this pull request address both those points.

LFS handling in `tests/tools/src/lib.rs` is intentionally retained, since there is no reason to remove LFS capabilities from the test tools (this could be revisited in the future if it is found to impose a significant maintenance burden) and LFS might potentially be used again in the future. Unlike in documentation and CI workflows, LFS capabilities in the test tools do not make working with the project any more complicated and do not make anything take longer.

I have also not made a change in `.gitattributes`. I'm unsure if any changes is needed or warranted there. The relevant line is:

https://github.com/Byron/gitoxide/blob/e791bc5da52a1237fb7cac230af583199162825d/.gitattributes#L1

I have permitted my editor to remove trailing whitespace, in files that I was editing anyway, since in all cases it appeared unintended. This deliberately does not remove trailing whitespace in files that would otherwise be untouched, nor does it do any other kind of cleanup.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,9 +79,6 @@ jobs:
       - name: "cargo check default features"
         if: startsWith(matrix.os, 'windows')
         run: cargo check --all --bins --examples
-      - run: |
-          # it should never be a failure not to get the caches, as they can be regenerated.
-          git lfs fetch && git lfs checkout || true
       - uses: taiki-e/install-action@v2
         with:
           tool: nextest
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,6 +1,6 @@
 # Development Guide
 
-## Practices 
+## Practices
 
  * **test-first development**
    * protect against regression and make implementing features easy.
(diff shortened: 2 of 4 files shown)


Byron:

> I have also not made a change in `.gitattributes`. I'm unsure if any changes is needed or warranted there. The relevant line is:

I'd be glad to keep this line as well as a reminder of what once was. The reason for ditching LFS was my inability to fairly pay for it - all I needed was 1GB per month of *bandwidth* more, but was only able to purchase 50GB at a time, which would also not carry over between months.

Skgland:
I don't believe this changes anything but GitHub is moving to [metered billing for Git LFS ](https://github.blog/changelog/2024-06-03-new-enterprise-accounts-have-metered-billing-for-git-lfs/)

Byron:
Thanks for sharing! It's great to see they are starting to make proper billing available so eventually this project can use that, too.

Just because `git-lfs` is perfect for this use-case - optional cache files that can accelerate test runs, but also are entirely optional in case `git lfs` isn't available on the client.

PR #1371, opened 2024-05-17
empty paths don't error in lenient mode

### LINKED ISSUE #1370: `core.excludesFile` config entry exists but has blank value causes error: is this considered a bug or expected behavior?
I encountered this error when building docs using `cargo doc` for my project. I did some digging. and managed to locate where the error originated: in `config::Cache::excludes_file()` it calls `self.trusted_file_path()` to get the path for the EXCLUDES_FILE:

https://github.com/Byron/gitoxide/blob/d3588ca4fe0364c88e42cdac24ceae548355d99d/gix/src/config/cache/access.rs#L208-L212

in `trusted_file_path()`, `self.resvoled.path_fillter(...)` returns a value of `Some("")` (it's actual type is `Path` not `&str` but you get the point) which cause `path.interpolate(ctx)` to return an `Err(interpolate::Error)`:

https://github.com/Byron/gitoxide/blob/d3588ca4fe0364c88e42cdac24ceae548355d99d/gix/src/config/cache/access.rs#L222-L232

--------

it turns out in my git config file, there exists an entry for `core.excludesfile`, but it's value is blank:

```text
[core]
	excludesfile =
```

I don't remember how it came to be, I think this might be a leftover by the SourceTree GUI app.

and I'm not sure if this is expected behavior or misbehavior, but apparently `libgit2` didn't treat this as an error case.

I guess another way to state question is, should `path_filter()` return `Some("")` or should it return `None` for an blank path?

Byron:
When using `gix::open_opts(path, options.strict_config(false))`, empty `core.excludesFile` values will not cause an error anymore.

Note that in strict mode, the behaviour is unchanged so invalid configuration can rather be fixed than ignored.

Fixes #1370 .


EliahKagan:
[The Windows test failure](https://github.com/Byron/gitoxide/actions/runs/9123648565/job/25086429126?pr=1371) is due to https://github.com/nextest-rs/nextest/issues/1493 (see also https://github.com/rust-lang/rustup/issues/3825). Rerunning the test should produce the same error (though I agree with doing so, just to be sure). It is not specific to the code in this PR, also occurring when formerly successful checks on the main branch are rerun, and locally. I have also verified, both locally and on CI, that it is also not related to any recent changes in `gix-macros`.

Until a proper fix is released, it should be possible to fix it by setting `RUSTUP_WINDOWS_PATH_ADD_BIN=1`, as in https://github.com/ruffle-rs/ruffle/pull/16342. I'll open a pull request here in gitoxide with that workaround shortly, so PRs such as this one can move forward until an upstream fix is available.

Byron:
Thanks a lot for your help here - incredible how fast you figured that out, I was still puzzled.

I'd be happy to wait until your PR is merged to fix this one.

Issue #1373, opened 2024-05-20
Nondeterministic macOS `is_symlink` assertion failure in `overwriting_files_and_lone_directories_works`


EliahKagan:
### Current behavior 😯

In `gix-worktree-state/tests/state/checkout.rs`, the test `overwriting_files_and_lone_directories_works` fails occasionally on the assertion:

https://github.com/Byron/gitoxide/blob/3c2174101ed35dcb9bdb4585b3245507b15efe59/gix-worktree-state/tests/state/checkout.rs#L238

Usually it passes, but occasionally it fails. This happens both on CI and in manual runs. As far as I know, the failure only happens on macOS. [Here's an example from CI in my fork](https://github.com/EliahKagan/gitoxide/actions/runs/9156200348/job/25170096086#step:8:2341):

```text
        FAIL [   0.422s] gix-worktree-state-tests::worktree state::checkout::overwriting_files_and_lone_directories_works

--- STDOUT:              gix-worktree-state-tests::worktree state::checkout::overwriting_files_and_lone_directories_works ---

running 1 test
test state::checkout::overwriting_files_and_lone_directories_works ... FAILED

failures:

failures:
    state::checkout::overwriting_files_and_lone_directories_works

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 10 filtered out; finished in 0.42s


--- STDERR:              gix-worktree-state-tests::worktree state::checkout::overwriting_files_and_lone_directories_works ---
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
thread 'state::checkout::overwriting_files_and_lone_directories_works' panicked at gix-worktree-state/tests/state/checkout.rs:238:9:
assertion `left == right` failed
  left: false
 right: true
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I don't know why this happens, but maybe it is somehow related to concurrency? I say this based only on https://github.com/Byron/gitoxide/issues/1354#issuecomment-2107303983, indicating that "the filesystem probe" for symlinks is "still not safe for concurrent operations."

### Expected behavior 🤔

The test should pass in all runs, rather than just most.

### Git behavior

I'm unsure if there is a corresponding Git behavior, because I am unsure if this is a test bug or a bug in the code under test. Above the line with the assertion, the test has this comment:

[...]

Issue #917, opened 2023-06-29
feature `sha1/asm` is enabled on windows


NobodyXu:
### Duplicates

- [X] I have searched the existing issues

### Current behavior 😯

Compiling on [windows failed](https://github.com/cargo-bins/cargo-binstall/actions/runs/5413274155/jobs/9838597776) failed with errors from `sha1-asm`, which is known in https://github.com/RustCrypto/asm-hashes/issues/17 and should be migrated by `gix-features` by disabling `sha1/asm` on windows:

<details>

```
warning: cl : Command line warning D9024 : unrecognized source file type 'src/x64.S', object file assumed
warning: cl : Command line warning D9027 : source file 'src/x64.S' ignored

error: failed to run custom build command for `sha1-asm v0.5.1`

Caused by:
  process didn't exit successfully: `D:\a\cargo-binstall\cargo-binstall\target\release\build\sha1-asm-898e65df03af5add\build-script-build` (exit code: 1)
  --- stdout
  TARGET = Some("x86_64-pc-windows-msvc")
  OPT_LEVEL = Some("z")
  HOST = Some("x86_64-pc-windows-msvc")
  cargo:rerun-if-env-changed=CC_x86_64-pc-windows-msvc
  CC_x86_64-pc-windows-msvc = None
  cargo:rerun-if-env-changed=CC_x86_64_pc_windows_msvc
  CC_x86_64_pc_windows_msvc = None
  cargo:rerun-if-env-changed=HOST_CC
  HOST_CC = None
  cargo:rerun-if-env-changed=CC
  CC = None
  cargo:rerun-if-env-changed=CFLAGS_x86_64-pc-windows-msvc
  CFLAGS_x86_64-pc-windows-msvc = None
  cargo:rerun-if-env-changed=CFLAGS_x86_64_pc_windows_msvc
  CFLAGS_x86_64_pc_windows_msvc = None
  cargo:rerun-if-env-changed=HOST_CFLAGS
  HOST_CFLAGS = None
  cargo:rerun-if-env-changed=CFLAGS
  CFLAGS = None
  cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
  CRATE_CC_NO_DEFAULTS = None
  CARGO_CFG_TARGET_FEATURE = Some("fxsr,sse,sse2")
  DEBUG = Some("false")
[...]

PR #1374, opened 2024-05-22
Various fixes

Byron:
### Tasks

* [x] fix https://github.com/Byron/gitoxide/issues/1373
* [x] fix https://github.com/Byron/gitoxide/issues/917
* [x] update dependencies


Byron:
@EliahKagan If you take a look at the CI failure on Windows you will see that the error there is due to a lock file that couldn't be obtained. I think this is akin to what Git does. The problem here is that it shouldn't even get to that point - it should fail when trying to read the reference, which I think it might actually do here but without issue. Unfortunately, I couldn't reproduce it in the VM, the test succeeds, but I will keep trying with added debug information.

Byron:
Turns out I was blind! CI is actually reproducing the correct issue, and the problem was that the error has handled too late.

Byron:
Alright, I think CI will pass ~~now~~, and once done, a new `gitoxide` release will be created.

EliahKagan:
When I run the tests locally on Windows, at [5f1c865](https://github.com/Byron/gitoxide/pull/1374/commits/5f1c865e025d116ae3aa293f23968bc16b5dfbc9), this still fails:

```text
        FAIL [   0.846s] gix-archive::archive from_tree::basic_usage_zip
```

The same thing seems to be happening [on CI](https://github.com/Byron/gitoxide/actions/runs/9189857631/job/25272808151?pr=1374) for it.

Byron:
Oh, that's a new failure probably due to Windows not really having symlinks. Maybe there is a way to support that better, but for now it's not to be expected to be a symlink on Windows.
It's interesting as in an older version of the `zip` crate, symlinks didn't work at all, so that's an improvement.

EliahKagan (approved):
LGTM.

I have installed on Windows from [e955770](https://github.com/Byron/gitoxide/pull/1374/commits/e955770c0b5c06a6c8518a06df4aa0cc3b506f16) with `cargo install --path .` and repeated the manual tests of the refs-related features that had not been fully implemented yet when I had tested them yesterday. Those are working now.

Once this passes CI, I think it can be merged. It seems to me that the pull request description should eventually be updated with the other relevant checklist items, but that this could be done later (which might even be preferable).

RivenSkaye:
Symlinks on Windows do exist, but currently they're locked behind one of two options:
1. Admin only to create and edit
2. Developer Options allow you to toggle making them free for all

The latter is also what you need to have a hassle-free `symlinks=native-strict` experience in MSYS2/cygwin envs. More info on [this 2016 windows developer blog article](https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/). For most intents and purposes, they're not reliably present

gix-fs/tests/stack/mod.rs
```diff
@@ -26,6 +27,287 @@ impl gix_fs::stack::Delegate for Record {
...
+    assert_eq!(
+        p("/").join("C:/"),
+        p("C:\\"),
+        "unix-absolute + win-drive = win-drive, strangely enough it changed the trailing slash to backslash, so better not have trailing slashes"
+    );
+    assert_eq!(p("/").join("C:\\"), p("C:\\"), "unix-absolute + win-drive = win-drive");
+    assert_eq!(
+        p("\\\\.").join("C:"),
+        p("C:"),
+        "device-namespace-unc + win-drive-relative = win-drive-relative - c: was supposed to be relative, but it's not acting like it."
```
EliahKagan, line 116: The `device-namespace-unc` part of the custom message suggests that this may have been meant to start with `p("\\\\.\\")` rather than `p("\\\\.")`. As currently written, I am not sure what the ideal behavior would be here, because I am not sure the trailing `\` of the prefix `\\?\` is strictly speaking functioning as a path separator. The Rust standard library [does sometimes treat `\\.\` and `\\.` differently](https://github.com/EliahKagan/pathjoin-unc) as the left side of a path join. However, both for the relative path `C:` and the absolute path `C:\` on the right, the result is the same.

However, for the same reason as in the above comment, that `C:` is relative should not be a reason to expect anything other than this behavior. It's less clear what it means to be *located at* `\\.` or `\\.\`. But I don't think it should be an exception to the usual effect of `C:` on the right side.

Furthermore, although `\\.\C:` and `\\.\C:\` are both valid paths, we *cannot* correctly form either one of them by concatenating `\\.` or `\\.\` with `C:`, and the reason is *because* a path `C:` is a relative path:

- `\\.\C:` is in the device namespace and ends in `C:` with no path separator, so that `C:` is a *device name*. Therefore `\\.\C:` refers to the `C:` drive as a device; the `C:` in `\\.\C:` is not a reference to the current directory, root directory, or *any* file or directory on the `C:` drive, but instead to the `C:` *drive* itself. (`\\.\C:` is an absolute path, but not to a file or directory.)
- `\\.\C:\`, due to the trailing `\`, is the root directory on the `C:` drive, and thus generally equivalent to `\\?\C:\` (though I don't know that all APIs that accept one will accept the other or that all APIs that accept both will necessarily treat them identically). But this is an absolute path to the root of the `C:` drive, which `C:` by itself does not designate because it is a relative path whose meaning depends on the current directory on the `C:` drive.

gix-fs/tests/stack/mod.rs
```diff
@@ -26,6 +27,287 @@ impl gix_fs::stack::Delegate for Record {
...
+    );
+    assert_eq!(p("relative").join("C:"), p("C:"), "relative + win-drive = win-drive");
+
+    assert_eq!(
+        p("/").join("\\\\localhost"),
+        p("\\localhost"),
+        "unix-absolute + win-absolute-unc = win-absolute-unc"
```
EliahKagan, line 123: These are not raw strings, so this is indicating that `/` joined with `\\localhost` gives `\localhost` with one leading `\`, and I don't understand why this is what happens. It also does not appear to be what this meant to assert, since it characterizes the effect of the join as `= win-absolute-unc`, but a path that begins with only one `\` is not a UNC path. So I don't know what's going on here.

EliahKagan (commented):
These comments about the semantics of joining Windows paths are mainly to identify areas of interest for forthcoming work.

I've posted them as review comments because that seemed like an easier way to identify context than posting a regular comment here or a separate discussion question.

Some potentially identify areas where the tests may be improved, but they do not represent worries about the correctness of the code under test that was introduced in this PR.

I have two main goals:

- In some cases, there are clarifications that can be made in the custom messages, which I plan to do soon in a small PR. *Edit: I've opened [#1528](https://github.com/Byron/gitoxide/pull/1528) and added links to the comments below to corresponding comments in that PR about the related changes.*

- When working on this, I recall that we had discussed the possibility that there was at least one area where the behavior of the standard library should be improved, and you suggested that I might wish to contribute an improvement. I recall being interested in doing so, and also, I think, that you may have been willing to help review or otherwise provide guidance with that. But unfortunately I don't remember *what* that was!

  It may be possible to figure it out from comments we exported from the temporary private fork in which most of the work that made it into this PR occurred. But those exports, even taken together, were not quite complete, and I have not yet managed to figure out which behavior of the standard library we thought was wrong. I'm hoping you may recall that. These comments include details about behaviors that I am confident at least on re-examination are correct but that may seem wrong, as well as about things that I am not sure are correct.

I also hope these comments may be useful to refer back to later when working on other parts of the code where these concepts are relevant. So if you identify any technical mistakes in the comments themselves, please let me know.

gix-fs/tests/stack/mod.rs
Byron: That's a mistake, as it indeed doesn't consider that "\\…" is not a raw string.
I don't know what an alternative comment would be, but it definitely isn't `= win-absolute-unc`.

Byron:
> * I recall being interested in doing so, and also, I think, that you may have been willing to help review or otherwise provide guidance with that. But unfortunately I don't remember _what_ that was!

I also have faint memory of that, but not the specific thing we thought could be fixed. Hopefully that will come back once these cases are improved and/or fixed up.

RivenSkaye:
Considering recent changes in std in the same domain, could you have been considering implementing/adding something along the lines of [`std::path::absolute`](https://doc.rust-lang.org/std/path/fn.absolute.html#windows-paths) perhaps? Came across that when [Av1an had a clash with stuff in other langs](https://github.com/master-of-zen/Av1an/issues/860) and proposed that as a fix for the issues with extended path length syntax caused by not all Windows APIs supporting the syntax

EliahKagan:
> Considering recent changes in std in the same domain, could you have been considering implementing/adding something along the lines of [`std::path::absolute`](https://doc.rust-lang.org/std/path/fn.absolute.html#windows-paths) perhaps?

Although that makes sense and would have been a good idea, I don't think so. If I remember correctly, I think some behavior of the standard library shown or commented on in those tests had seemed like it was wrong, at the time.

EliahKagan:
https://github.com/GitoxideLabs/gitoxide/pull/1374#issuecomment-2124574975 @RivenSkaye 

> Edit: [Current win32 docs](https://learn.microsoft.com/en-us/windows/win32/fileio/creating-symbolic-links) and the corresponding [page for the function in the headers](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createsymboliclinka) seem to indicate Developer Mode is now only required for UWP and makes no mention of perms anymore

It looks to me like that is a bug in the Microsoft documentation.

Over time, but more recently than that article was edited to mention UWP, I've attempted to use a limited user account (without extra privileges added), as well as with a UAC-unelevated administrator account, to create symbolic links on a few Windows 10 and Windows 11 systems, and including multiple versions of both operating systems, and on a Windows Server 2022 system. All of this was with non-UWP executables. Besides being compiled as traditional rather than UWP or WinRT executables, the executables were fairly diverse, and included:

- `cmd.exe` (via its `mklink` builtin).
- MSYS2 `ln.exe` with [`MSYS=winsymlinks:nativestrict`](https://www.msys2.org/docs/symlinks/#cygwin-symlink-modes) to cause symlink-creation operations via `msys-2.0.dll` to create real Windows symlinks ([#1444](https://github.com/GitoxideLabs/gitoxide/pull/1444)).
- `git.exe` (Git for Windows) with the `core.symlinks` configuration option set to `true` (`git.exe`, unlike `ln.exe`, [is a native Windows program](https://github.com/GitoxideLabs/gitoxide/pull/1727#pullrequestreview-2513107760)).

In these tests, I have never been able to create symlinks without enabling Developer Mode, performing UAC elevation, or turning UAC off. These experiments were after the change in the above-linked [`CreateSymbolicLinkA`](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createsymboliclinka) article. That change was made in https://github.com/MicrosoftDocs/sdk-api/pull/1284 with the commit message:

> Proposed change is based upon testing under MSIX on Windows 10 21H2.  I do not know whether developer mode is still required for UWP, so that should be checked prior to accepting this change.

I am unsure if "under MSIX" is a reference to operations that take place during an MSIX installation process, or if this is referring to the behavior of all software installed via MSIX, or if this referring to software built as a the MSIX project type in Visual Studio, or if "MSIX" here is meant to refer to all non-UWP (or all non-WinRT) software.

Although it seems unlikely, it could be that the above observations were due to the functions not being called with the necessary flags, or something peculiar about the builds (for example, some seem to have been built against `msvcrt.dll`, as happens in [MINGW64](https://www.msys2.org/docs/environments/#__tabbed_1_1) builds, even though the Microsoft recommendation for programs that are not part of the operating system is to build against a specific versioned MSVCRT or against the UCRT), or a difference in the behavior of `CreateSymbolicLinkW` and `CreateSymbolicLinkA` (more on that below). Furthermore, I did not keep detailed notes about exactly what builds of Windows itself the above observations were from, nor do I know with certainty that I didn't skip or miss something.

For those reasons, I have now also written and run [a small test program](https://github.com/EliahKagan/symlink) in C that tests out both the `CreateSymbolicLinkW` and `CreateSymbolicLinkA` functions, each for both file and directory symlinks, making sure to pass `SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE` in all calls. I did this on a Windows 11 system for which `WindowsBuildLabEx` is `22621.1.amd64fre.ni_release.220506-1250` and compiled it with Visual Studio 2022 17.12.3 and version 10.0.22621.0 of the Windows SDK. I can probably manage to run it on other versions and builds of Windows if a difference is suspected. [The code, and further details, can be seen in the repo.](https://github.com/EliahKagan/symlink) The results were the same: with Developer Mode turned off, I needed to elevate in order for the symlink creation attempts to succeed.

Although that shows that the documentation is at least partially wrong and incomplete, I regard the test using that program to be the most limited source of information I have about this, just because I have not built it against multiple SDK versions or run it on multiple Windows operating systems or versions of the same Windows operating system.

The other thing I have noticed, which I think is the most important of all, is that the change in the `CreateSymbolicLinkA` article mentioning UWP was never applied to the [`CreateSymbolicLinkW`](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createsymboliclinkw) article, which in the "meaning" column for `SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE` still says only:

> Specify this flag to allow creation of symbolic links when the process is not elevated. [Developer Mode](https://learn.microsoft.com/en-us/windows/uwp/get-started/enable-your-device-for-development) must first be enabled on the machine before this option will function.

But when the Windows API has pairs of "A" and "W" functions, the "A" function [is an ANSI wrapper](https://learn.microsoft.com/en-us/windows/win32/learnwin32/working-with-strings#unicode-and-ansi-functions) for the "W" function. An "A" function converts ANSI to Unicode for any strings referenced by input parameters, calls the corresponding "W" function, and then converts Unicode to ANSI for any strings referenced by output parameters or return values. (This is how they work on NT systems. In Windows 9x/Me with [MSLU](https://en.wikipedia.org/wiki/Microsoft_Layer_for_Unicode), it was the other way around, with "W" functions being Unicode wrappers for the "A" functions.)

So the API documentation does still say that Developer Mode is needed, in the documentation for the "W" version of the function. This is furthermore the primary version. Thus a revision only to the documentation of the "A" version seems unintentional in and of itself.

(Something I have noticed is that, when I search using the learn.microsoft.com interface--and sometimes even with general purpose search engines--I get the "A" versions of functions first. This seems to be an unfortunate result of the Microsoft website disambiguating by picking the name that sorts earlier alphabetically--and then search engines ranking partly based on links that go to the "A" versions that were found that way. When one of the unsuffixed macros, such as `CreateSymbolicLink`, is used, these days it expands to the "W" version--here, `CreateSymbolicLinkW`--in the most common configurations, i.e., "Unicode" builds. But the version of the documentation most likely to be consulted seems often to be the "A" version.)

RivenSkaye:
That is some A+ effort there. Good to know the docs and reality are out of sync on that front. I hope they'll be widely supported in the future. Perhaps a compile-time check or feature flag could be implemented at some point, but as-is it's not a safe default. Thanks for the poke!

Issue #1378, opened 2024-05-25
Installing `gitoxide@0.36.0` via `cargo install` not possible because the `zip` crate in the specified verision is yanked


kamulos:
### Current behavior 😯

As the title says: a `cargo install gitoxide --force` does not work because the `zip` crate is yanked.

This is the output:

```
Caused by:
  failed to select a version for the requirement `zip = "^1.3.1"`
  candidate versions found which didn't match: 2.1.0, 2.0.0, 1.1.4, ...
  location searched: crates.io index
  required by package `gix-archive v0.13.0`
      ... which satisfies dependency `gix-archive-for-configuration-only = "^0.13.0"` of package `gitoxide-core v0.38.0`
      ... which satisfies dependency `gitoxide-core = "^0.38.0"` of package `gitoxide v0.36.0`
  if you are looking for the prerelease package it needs to be specified explicitly
      zip = { version = "0.10.0-alpha.1" }
  perhaps a crate was updated and forgotten to be re-vendored?
```

Byron:
Thanks for reporting, a fix is in the works.

PR #1379, opened 2024-05-25
chore: version bump dependency zip-2.0.0

belt:
upstream yanked version `zip-1.3.1`

See https://github.com/Byron/gitoxide/issues/1378


Byron:

Since CI also checks installation which fails now, I will merge once the tests pass (which they now should). Then a patch release of `gix-archive` should do the trick.

PR #1380, opened 2024-05-26
feat: Dockerfile for faster CI/CD pipelines

belt:
Some pipeline require repository cloning. Provide an alternative to `git` within such workflows


README.md
```diff
@@ -62,6 +62,31 @@ Follow linked crate name for detailed status. Please note that all crates follow
 
 [semver]: https://semver.org
 
+### Pipeline Integration
+
+Some CI/CD pipelines leverage repository cloning. Below is a copy-paste-able example to build docker images for such workflows.
```
Byron, line 67: Could you provide more context? Which pipelines are you talking about? What's the general idea on how to get from images to using them in a pipeline?

Think blank slate, and I am good test for that as I wouldn't be able to use the docker images from this description.

README.md
```diff
@@ -62,6 +62,31 @@ Follow linked crate name for detailed status. Please note that all crates follow
...
+
+#### Pipeline Integration (recommended)
+
+Build an image without a target and then copy the binaries into your local image
```
Byron, line 71: I don't get that - could you describe that without assuming too much docker knowledge?

README.md
```diff
@@ -62,6 +62,31 @@ Follow linked crate name for detailed status. Please note that all crates follow
...
+
+```dockerfile
+COPY --from gitoxide:latest /bin/gix /usr/local/bin/
```
Byron, line 74: This works if the `gitoxide` image was already built, right? That wasn't clear to me and I thought maybe it pulls it from the network. Could that be clarified in the paragraph above as well?

README.md
```diff
@@ -62,6 +62,31 @@ Follow linked crate name for detailed status. Please note that all crates follow
...
+COPY --from gitoxide:latest /bin/ein /usr/local/bin/
+```
+
+#### Pipeline Integration (base image)
+
```
Byron, line 79: I think there should be text here that explains what it's doing and why - it probably all flows together with the paragraphs above, and it feels like these could be re-ordered so the user first builds an image, then uses the image in some way, and of course also gets some idea on how to use it in a pipeline. My preference would be to show how to use such image with GitHub actions if at all possible.

docker/Dockerfile.alpine
```diff
@@ -0,0 +1,69 @@
+ARG GITOXIDE_VERSION=0.36.0
```
Byron, line 1: Let's move these files into `etc/docker/`.

Byron (changes requested):
Thanks a lot! I love the idea of being able to use `gix clone` in pipelines.

I left some comments, and one more request: there should be a new job that validates the dockerfiles, probably by running them. That's probably quite costly but can hopefully be done within 15 minutes or so.

Finally, I wonder if two dockerfiles are needed, or if one could just endorse one base image - after all, both base images build a binary, the result should be the same. Or does one binary not run in most places? Or asked differently, is `alpine` not producing the more compatible binary, so `bookworm` is superseded?

Thanks again for your help.

README.md
belt: @Byron Do you have a docker hub account? If so, that should be the "official" image... and `docker` would automatically pull that image and copy the file.

belt:
> Thanks a lot! I love the idea of being able to use `gix clone` in pipelines.
> 
> I left some comments, and one more request: there should be a new job that validates the dockerfiles, probably by running them. That's probably quite costly but can hopefully be done within 15 minutes or so.
> 
> Finally, I wonder if two dockerfiles are needed, or if one could just endorse one base image - after all, both base images build a binary, the result should be the same. Or does one binary not run in most places? Or asked differently, is `alpine` not producing the more compatible binary, so `bookworm` is superseded?
> 
> Thanks again for your help.

Alpine is based off the `musl` system library. Debian is based off `libc` system library. The resulting binaries are different. Though, alpine does provide a `gcompat` library when it's _needed_. For what it's worth, Ubuntu (or any other distro for that matter) isn't guaranteed to be running the same `libc` version as Debian, so things might need recompilation for compatibility on those systems.

README.md
Byron: I am trying to say that I don't understand the paragraph, and that it should be adjusted to fit people who don't do that every day. Ideally, no prior knowledge about this usually works is assumed.

README.md
Byron: Let's leave dockerhub out of the picture for now as I have no plans to maintain an official image for now. This would mean the `gitoxide:latest` name assumes a local image, which probably means the description on how to build that should come first.

Byron:
> Alpine is based off the `musl` system library. Debian is based off `libc` system library. The resulting binaries are different. Though, alpine does provide a `gcompat` library when it's _needed_.

Wouldn't this mean that the alpine version could run anywhere where there is a recent-enough kernel? That seems to mean it supersedes the debian version which only runs on libc compatible distros, which seems less compatible.
If that's true, who would want a libc version of the binary? Somebody saving every bit of bandwidth as it's probably a little smaller?

belt:
> > Alpine is based off the `musl` system library. Debian is based off `libc` system library. The resulting binaries are different. Though, alpine does provide a `gcompat` library when it's _needed_.
> 
> Wouldn't this mean that the alpine version could run anywhere where there is a recent-enough kernel? That seems to mean it supersedes the debian version which only runs on libc compatible distros, which seems less compatible. If that's true, who would want a libc version of the binary? Somebody saving every bit of bandwidth as it's probably a little smaller?

Correct. Trade-offs. Do you want me to call that out explicitly or just provide `musl` support?

Byron:
Alright, let's go with just alpine then, MUSL seems to be the most compatible while keeping things simple.

Would you be able to setup a CI step to build the docker image, to assure it keeps working? I probably won't run it locally very much beyond initial testing.

Byron:
Thanks a lot! With the improved documentation I was able to understand what's intended, and with that it could be pushed over the finishing line.

Please note that I removed the CI integration as it was to slow, taking 1h to build with a budget of only 30 minutes - however, I noted that the dockerfile might stop working eventually so users are encouraged to fix it by PR.

PR #1271, opened 2024-02-03
Remove Rustix special-cases

Byron:
Fixes #1269

> [!WARNING]
> Blocked on https://github.com/bytecodealliance/rustix/pull/999#issuecomment-1925203175

### Tasks

* [x] update Rustix
* [ ] use `StatExt` trait

PR #1383, opened 2024-05-26
hurd: fix accessing st_[mc]time

sthibaul:
GNU/Hurd uses a st_[mc]tim timespec, like aix


Byron (approved):
Thank you.

I wish I could get rid of this code, but it's blocked by #1271 .

Issue #575, opened 2022-11-02
IOS Support -and more generally any rust target + headache-free cross compiling VIA using native rust libs


colbyn:
### Summary 💡

Add a feature to use native rust libraries for cases where optimized C libs become a huge headache or refuse to build, such as when cross compiling to IOS. 

### Motivation 🔦

In #122 someone remarked

> Just a heads-up, we decided on using zlib-ng over a native rust implementation due to the large performance benefits; you can see the discussion in #47. 

The problem with `zlib-ng` is that it becomes problematic when cross-compiling to e.g. IOS, whereas native rust code tends to be headache free.

Byron:
Thanks for the suggestion!

In case it matters, one can get a (probably close to) Rust-only build with `cargo build --no-default-features --features small`, or the respective features it sets in the `git-features` crate. As far as I know there is no way to know if a crate uses a C compiler in its build scripts or does anything else non-Rusty. Maybe I am missing something though.

colbyn:
@Byron the current issue I have is compiling `libz-sys`, 

<details>
<summary>Here is the log if you want to see it</summary>
<pre>error: failed to run custom build command for `libz-sys v1.1.8`

Caused by:
  process didn't exit successfully: `/Users/colbyn/Projects/subscript-root/target/release/build/libz-sys-f3ff5c4f94c5a6e0/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-env-changed=LIBZ_SYS_STATIC
  cargo:rerun-if-changed=build.rs
  CMAKE_TOOLCHAIN_FILE_aarch64-apple-ios = None
  CMAKE_TOOLCHAIN_FILE_aarch64_apple_ios = None
  TARGET_CMAKE_TOOLCHAIN_FILE = None
  CMAKE_TOOLCHAIN_FILE = None
  CMAKE_GENERATOR_aarch64-apple-ios = None
  CMAKE_GENERATOR_aarch64_apple_ios = None
  TARGET_CMAKE_GENERATOR = None
  CMAKE_GENERATOR = None
  running: "xcrun" "--show-sdk-path" "--sdk" "iphoneos"
  exit status: 0
  running: "xcrun" "--show-sdk-path" "--sdk" "iphoneos"
  exit status: 0
  CMAKE_PREFIX_PATH_aarch64-apple-ios = None
  CMAKE_PREFIX_PATH_aarch64_apple_ios = None
  TARGET_CMAKE_PREFIX_PATH = None
[...]

PR #1384, opened 2024-05-27
Refactor: Make it easier to compile gitoxide as dynlib

NobodyXu:
And also optimize compilation time: By making `plumbing` and `porcelain` as modules the `lib.rs`, they can be compiled after the rmeta for the dependencies are generated.

For the `uni.rs` which uses both `plumbing` and `porcelain`, this would avoid compiling these two modules twice.

This PR would also help @holzschu 's [use case](https://github.com/Byron/gitoxide/issues/575#issuecomment-2132390429).

With this PR merged and published to https://crates.io , you can compile gitoxide as a dynlib using:

`Cargo.toml`:

```rust
[profile]
name = "gitoxide-dynlib"
publish = false

[lib]
crate-type = ["cdylib"]
bench = false

[dependencies.anyhow]
version = "1.0.86"

[dependencies.gitoxide]
version = "0.36.0"  # Or you could use git = "https://github.com/Byron/gitoxide"
features = ["max-pure"]
```

in `src/lib.rs`:

```rust
use anyhow::Result;

#[no_mangle]
pub extern "C" fn gix_main() -> Result<()> {
    gitoxide::plumbing::main()
}

#[no_mangle]
pub extern "C" fn ein_main() -> Result<()> {
    gitoxide::porcelain::main()
}
```

Then you would have a C dynlib, with two C API in it: `gix_main` and `ein_main`

### DIFF
--- a/src/ein.rs
+++ b/src/ein.rs
@@ -1,9 +1,7 @@
 #![deny(rust_2018_idioms, unsafe_code)]
 
-mod porcelain;
-
 fn main() -> anyhow::Result<()> {
-    porcelain::main()
+    crate::porcelain::main()
 }
 
 #[cfg(not(feature = "pretty-cli"))]
--- a/src/ein.rs
+++ b/src/ein.rs
@@ -1,7 +1,7 @@
 #![deny(rust_2018_idioms, unsafe_code)]
 
 fn main() -> anyhow::Result<()> {
-    crate::porcelain::main()
+    gitoxide::porcelain::main()
 }
 
 #[cfg(not(feature = "pretty-cli"))]
--- a/src/gix.rs
+++ b/src/gix.rs
@@ -2,7 +2,7 @@
 
 #[cfg(feature = "pretty-cli")]
 fn main() -> anyhow::Result<()> {
-    crate::plumbing::main()
+    gitoxide::plumbing::main()
 }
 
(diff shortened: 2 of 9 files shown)


NobodyXu:
@Byron It seems like the CI is failing because of use of `unsafe` in gitoxide conflicts with `forbid(unsafe)`

Byron:
Thanks a lot for your help!

Regarding CI, if this happens, I usually put the lib-attribute to `deny`. Will take a look later if you don't fix it before me.

Lastly:

```
use anyhow::Result;

#[no_mangle]
pub extern "C" fn gix_main() -> Result<()> {
    crate::plumbing::main()
}

#[no_mangle]
pub extern "C" fn ein_main() -> Result<()> {
    crate::porcelain::main()
}
```

Should that not be `gitixoide::porcelain::main()`?

NobodyXu:
Thanks, I've fixed the compilation error.

> Should that not be `gitixoide::porcelain::main()`?

Turns out that I also uses the wrong path in this PR, update the description and the PR.

Byron (approved):
Looks great now, thank you!

NobodyXu:
Just found another compilation error.

And I've run `cargo build --release --timings`:

image

Moving these module plumbing and porcelain to `lib.rs` enables these two modules to compile in parallel to gitoxide-core

PR #1386, opened 2024-05-28
fix: ignore `.gitignore` files that are directories

Byron:
This happens in real-life, and Git itself doesn't budge.

Related to https://github.com/gitbutlerapp/gitbutler/issues/3876

### DIFF
--- a/gix-fs/src/symlink.rs
+++ b/gix-fs/src/symlink.rs
@@ -59,7 +59,7 @@ pub fn create(original: &Path, link: &Path) -> io::Result<()> {
 pub fn is_collision_error(err: &std::io::Error) -> bool {
     // TODO: use ::IsDirectory as well when stabilized instead of raw_os_error(), and ::FileSystemLoop respectively
     err.kind() == AlreadyExists
-            || err.raw_os_error() == Some(21)
+            || err.raw_os_error() == Some(if cfg!(windows) { 5 } else { 21 })
             || err.raw_os_error() == Some(62) // no-follow on symlnk on mac-os
             || err.raw_os_error() == Some(40) // no-follow on symlnk on ubuntu
 }
--- a/gix-glob/src/search/pattern.rs
+++ b/gix-glob/src/search/pattern.rs
@@ -48,14 +48,27 @@ fn read_in_full_ignore_missing(path: &Path, follow_symlinks: bool, buf: &mut Vec
     };
     Ok(match file {
         Ok(mut file) => {
-            file.read_to_end(buf)?;
-            true
+            if let Err(err) = file.read_to_end(buf) {
+                if io_err_is_dir(&err) {
+                    false
+                } else {
+                    return Err(err);
+                }
+            } else {
+                true
+            }
         }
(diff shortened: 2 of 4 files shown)


gix-fs/src/symlink.rs
```diff
@@ -59,7 +59,7 @@ pub fn create(original: &Path, link: &Path) -> io::Result<()> {
 pub fn is_collision_error(err: &std::io::Error) -> bool {
     // TODO: use ::IsDirectory as well when stabilized instead of raw_os_error(), and ::FileSystemLoop respectively
     err.kind() == AlreadyExists
-            || err.raw_os_error() == Some(21)
+            || err.raw_os_error() == Some(if cfg!(windows) { 5 } else { 21 })
```
Skgland, line 62: This function here appears to be under `#[cfg(not(windows))]` making `cfg!(windows)` always false.
The `#[cfg(windows)]` below should probably be changed instead.

Byron: That's a great catch, thank you!

I pushed the fix [into this standing PR](https://github.com/Byron/gitoxide/pull/1368).

EliahKagan:
Should something be done so that new binary installations of `gitoxide` (i.e., pre-built `gix` and `ein` binaries) use the new dependencies?

When installed from source using `cargo install gitoxide`, or with `cargo install --path .` from a clone of the `gitoxide` repository, commands like `gix status` no longer fail due to the presence of a `.gitignore` directory. They behave as expected, no longer being tripped up by a the presence of such a directory:

```text
% git ls-files
.gitignore/a
% gix status
 12:07:41 status done 1.0 files in 0.01s (183.0 files/s)

head -> index isn't implemented yet
% echo "$?"
0
```

But when installing with `cargo quickinstall gitoxide` or `cargo binstall gitoxide`, this still happens:

```text
% git ls-files
.gitignore/a
% gix status
Error: Failed to update the excludes stack to see if a path is excluded
% echo "$?"
1
```

In both cases, this is with `gitoxide` 0.36.0.

It may be that nothing should be done. If the main affected subcommand is `status`, then the behavior of `gix` could perhaps wait until #1368 is done and merged, anyway. On the other hand, `gix` commands having two different behaviors even at the same version of `gitoxide` installed, due to non-intended differences in dependency versions, could be confusing. Then again, directories named `.gitignore` are uncommon. I'm not sure.

Byron:
Thanks for making me aware of this.

I think the only way to assure such updates become available in `cargo quickinstall` is to create a dummy `gitoxide` patch release to incorporate them and I think for important (i.e. security related) fixes this should certainly be done one way or another.

For this one, I think it's OK to leave it as is and do nothing.

NobodyXu:
> I think the only way to assure such updates become available in `cargo quickinstall` is to create a dummy `gitoxide` patch release to incorporate them

I could delete the cargo-quickinstall release for gitoxide, that should trigger a rebuild later.

EliahKagan:
> This happens in real-life, and Git itself doesn't budge.

While `git` permits this on all systems, on Windows it gives a warning that crates the appearance that it does not. I've opened https://github.com/git-for-windows/git/issues/5068 for this.

PR #1402, opened 2024-06-16
gix-url fix

Byron:
Reproduce a URL parsing overflow when using the latest version of the `url` crate.

This one probably needs to be fixed in `idna`.

Ideally, this would be solved by adding clusterfuzz support to the `idna` crate, for which `gitoxide` is acting like a proxy now.

### Tasks

* [x] wait for [the upstream PR](https://github.com/servo/rust-url/pull/940) to be merged
* [x] wait for new release.
* [x] rerun CI with updated version of `idna` crate.


Byron:
@DaveLak This is a wonderful example of `gitoxide` being a proxy-fuzz-target for the upstream `url` crate. Maybe this crate would be a great place to start fuzzing an incredibly high-value Rust crate - [nearly 200m downloads](https://crates.io/crates/url) and something that naturally has to deal with untrusted input all the time.

That would also help `gitoxide` as it might not be a proxy anymore, and if `clusterfuzz` would start fuzzing the `url` crate (and related crates), those would get much more fuzzing time to increase the chances of finding issues in the parser.

DaveLak:
@Byron , thanks for the ping and the insight! This is very interesting to me!

I'm planning on taking my first shot at fuzzing contributions on this repo in the second half of August, so I may very well start here!

PR #1407, opened 2024-06-19
Add rela_path utility to gix::status::index_worktree::iter::Item

jsimonrichard:
Add a small utility so that rela_path is easily accessible from `gix::status::index_worktree::iter::Item`.


Byron (approved):
Thanks for that, it's very useful! 

*I think I remember wanting that too, but don't recall why it wasn't implemented in the end.*

PR #1425, opened 2024-06-26
Always fall back to creating file symlinks on Windows

### LINKED ISSUE #1420: Checkout fails when Windows symlinks have strangely named targets
### Current behavior 😯

Windows permits symbolic links to specify targets with unusual names, such as invalid filenames like `???` and reserved legacy DOS device names like `CON`. Cloning a repository on Windows with `gix clone` that contains such a symlink fails, due to the error that occurs when attempting to look up metadata for such a target, which is done to figure out if a file symlink or directory symlink should be created.

`gix clone` has supported creating dangling symlinks on Windows since 31d02a8 (#1363), which fixed #1354 by special-casing `NotFound` errors and falling back to creating a file symlink:

https://github.com/Byron/gitoxide/blob/8d89b865c84d1fb153d93343d1ce4e1d64e53541/gix-fs/src/symlink.rs#L44-L48

Most cases of dangling symlinks are covered by `NotFound` errors. This includes cases where there is no target file at the destination, where there is no destination (e.g., the target path has nonexistent intermediate components), and possibly some other cases where errors prevent the target from being visible. However, this does not cover all errors. When the target is an invalid or reserved name, checkout fails.

Furthermore, it is not only those specific symlinks that are refused; checkout as a whole fails, and `gix clone` leaves no cloned directory behind. This can be seen when cloning a repository with a symlink to `???`:

```text
C:\Users\ek\src> gix clone git@github.com:EliahKagan/symlink-to-qmarks.git
 18:04:17 indexing done 7.0 objects in 0.00s (30.0k objects/s)
 18:04:17 decompressing done 1.5KB in 0.00s (6.3MB/s)
 18:04:17     Resolving done 7.0 objects in 0.05s (136.0 objects/s)
 18:04:17      Decoding done 1.5KB in 0.05s (29.5KB/s)
 18:04:17 writing index file done 1.3KB in 0.00s (8.6MB/s)
 18:04:17  create index file done 7.0 objects in 0.06s (115.0 objects/s)
 18:04:17          read pack done 1.1KB in 0.10s (10.7KB/s)
Error: IO error while writing blob or reading file metadata or changing filetype

Caused by:
    The filename, directory name, or volume label syntax is incorrect. (os error 123)
C:\Users\ek\src> ls symlink-to-qmarks
Get-ChildItem: Cannot find path 'C:\Users\ek\src\symlink-to-qmarks' because it does not exist.
```

And when cloning a repository with a symlink to `CON`:

```text
C:\Users\ek\src> gix clone git@github.com:EliahKagan/symlink-to-con.git
 18:09:42 indexing done 7.0 objects in 0.00s (24.6k objects/s)
 18:09:42 decompressing done 1.6KB in 0.00s (5.6MB/s)
 18:09:42     Resolving done 7.0 objects in 0.05s (136.0 objects/s)
 18:09:42      Decoding done 1.6KB in 0.05s (32.2KB/s)
 18:09:42 writing index file done 1.3KB in 0.00s (8.3MB/s)
 18:09:42  create index file done 7.0 objects in 0.06s (116.0 objects/s)
 18:09:42          read pack done 1.2KB in 0.06s (18.5KB/s)
Error: IO error while writing blob or reading file metadata or changing filetype

Caused by:
    The parameter is incorrect. (os error 87)
C:\Users\ek\src> ls symlink-to-con
Get-ChildItem: Cannot find path 'C:\Users\ek\src\symlink-to-con' because it does not exist.
```

More detailed reproduction is presented below.

This is only relevant when `gix` operates with a configuration that allows it to create symlinks. When it detects that it cannot create any symlinks at all, or when it is invoked with `-c core.symlinks=false`, it creates regular files instead of symlinks, so the behavior described here does not occur. Note that `gix` will usually attempt to create symlinks, though, and as of this writing this bug can be observed even when `core.symlinks` is set to `false` globally, due to #1353.

#### Option 1: Create the symlinks

The strange symlink, which is typically a dangling symlink (see below for details), should probably just be created. This is simple, should be the simplest kind of fix to implement, and would make `gix` behave more closely to Git. As detailed in the next section on Git behavior, Git creates these without complaint.

It should be feasible to implement this by using an expression like `orig_abs.is_dir()` in place of the entire existing `match` logic, because [`Path::is_dir`](https://doc.rust-lang.org/std/path/struct.Path.html#method.is_dir) performs an `fs::metadata` check, as we are currently doing explicitly, and then converts *all* errors to `false`, rather than just `NotFound` as we are currently doing. **I intend to open a PR for this shortly.**

(Such a change would leave the behavior invariant with respect to what kind of symlink is created, in cases where a symlink is already being created. Since `fs::metadata` rather than `fs::symlink_metadata` is used, this behavior is to attempt to fully dereference the target if possible. If the target is a symlink whose ultimate target does not exist, or cannot be reached—such as due to the wrong kind of symlink, i.e. file vs. directory, appearing [somewhere in the chain](https://github.com/EliahKagan/is-dir-symlink?tab=readme-ov-file#results-of-the-experiment)—then that is treated as an error. So a symlink to a directory symlink that dangles is currently created as a file symlink. This is not necessarily the best possible behavior, but I believe it should be considered independent of this bug and, if it is to be changed, then that can be done separately. Note that Git will sometimes do this as well, and always does so [in the straightforward scenario](https://gist.github.com/EliahKagan/590972953431a604cc574ced3e3fb789#file-experiment-3-txt-L55-L68) that the ultimate target directory is not also being created.)

#### Option 2: Refuse with a specific error and check out other files

However, if for some reason it is intentional that symlink creation be refused, then `gix` should:

- Report the condition specifically so that the user can understand what failed.
- Check out other files, unless there is a clear design reason why that shouldn't be done.

I think there is probably no justification for refusing to create symlinks, even unusual ones. I recommend Option 1.

#### A note on the semantics of such symlinks

Symlinks whose targets are invalid filenames or reserved device names are, *for the most part*, special cases of dangling symlinks. It seems to me that the behavior of creating dangling symlinks, and making them file symlinks on Windows where the type is unknown because the target does not exist, should apply here.

The following two symlink demonstrations are done in `cmd.exe`. Usually PowerShell should be preferred, but `cmd.exe` is easier to use for symlink demonstrations because its `mklink` builtin creates symlinks and its `dir` command shows their exact symlink type (file or directory).

It is intuitive that a symlink to an invalid path like `???` is dangling, since no file named `???` exists. This shows that I can create the symlink, and attempting to create the target file by writing through the symlink fails:

```text
C:\Users\ek\tmp>dir
 Volume in drive C is OS
 Volume Serial Number is B203-10FB

 Directory of C:\Users\ek\tmp

06/25/2024  06:24 PM    <DIR>          .
06/25/2024  06:24 PM    <DIR>          ..
               0 File(s)              0 bytes
               2 Dir(s)  84,951,973,888 bytes free

C:\Users\ek\tmp>mklink myqmarks ???
symbolic link created for myqmarks <<===>> ???

C:\Users\ek\tmp>dir
 Volume in drive C is OS
 Volume Serial Number is B203-10FB

 Directory of C:\Users\ek\tmp

06/25/2024  06:24 PM    <DIR>          .
06/25/2024  06:24 PM    <DIR>          ..
06/25/2024  06:24 PM    <SYMLINK>      myqmarks [???]
               1 File(s)              0 bytes
               2 Dir(s)  84,951,842,816 bytes free

C:\Users\ek\tmp>echo hello >myqmarks
The filename, directory name, or volume label syntax is incorrect.
```

It is less intuitive that a symlink to a reserved DOS device name like `CON` would be dangling, since in most file access APIs, such names, except in `\\?\` paths, are treated as though they exist in every directory. So isn't a symlink to `CON` a non-dangling symlink to the `CON` device?

Although I would not be confident that no application or API treats it that way, the answer is generally no--instead, it refers to a file named `CON` in the current directory on the actual filesystem, which typically does not exist, but which can be created by writing through the symlink:

```text
C:\Users\ek\tmp>del myqmarks

C:\Users\ek\tmp>echo hello >CON
hello

C:\Users\ek\tmp>dir
 Volume in drive C is OS
 Volume Serial Number is B203-10FB

 Directory of C:\Users\ek\tmp

06/25/2024  06:32 PM    <DIR>          .
06/25/2024  06:32 PM    <DIR>          ..
               0 File(s)              0 bytes
               2 Dir(s)  84,944,207,872 bytes free

C:\Users\ek\tmp>mklink mycon CON
symbolic link created for mycon <<===>> CON

C:\Users\ek\tmp>dir
 Volume in drive C is OS
 Volume Serial Number is B203-10FB

 Directory of C:\Users\ek\tmp

06/25/2024  06:33 PM    <DIR>          .
06/25/2024  06:33 PM    <DIR>          ..
06/25/2024  06:33 PM    <SYMLINK>      mycon [CON]
               1 File(s)              0 bytes
               2 Dir(s)  84,944,207,872 bytes free

C:\Users\ek\tmp>echo hello >mycon

C:\Users\ek\tmp>dir
 Volume in drive C is OS
 Volume Serial Number is B203-10FB

 Directory of C:\Users\ek\tmp

06/25/2024  06:33 PM    <DIR>          .
06/25/2024  06:33 PM    <DIR>          ..
06/25/2024  06:33 PM                 8 CON
06/25/2024  06:33 PM    <SYMLINK>      mycon [CON]
               2 File(s)              8 bytes
               2 Dir(s)  84,944,207,872 bytes free

C:\Users\ek\tmp>type mycon
hello

C:\Users\ek\tmp>type \\?\C:\Users\ek\tmp\CON
hello

C:\Users\ek\tmp>del \\?\C:\Users\ek\tmp\CON

C:\Users\ek\tmp>dir
 Volume in drive C is OS
 Volume Serial Number is B203-10FB

 Directory of C:\Users\ek\tmp

06/25/2024  06:33 PM    <DIR>          .
06/25/2024  06:33 PM    <DIR>          ..
06/25/2024  06:33 PM    <SYMLINK>      mycon [CON]
               1 File(s)              0 bytes
               2 Dir(s)  84,944,076,800 bytes free
```

### Git behavior

Git creates the symlinks to the strangely named targets, without complaint. Since these are dangling symlinks, Git creates them as file symlinks. I tested with:

```text
C:\Users\ek\src> git version
git version 2.45.2.windows.1
```

Cloning the repository containing a symlink to `???`:

```text
C:\Users\ek\src> git clone git@github.com:EliahKagan/symlink-to-qmarks.git
Cloning into 'symlink-to-qmarks'...
remote: Enumerating objects: 7, done.
remote: Counting objects: 100% (7/7), done.
remote: Compressing objects: 100% (5/5), done.
remote: Total 7 (delta 0), reused 7 (delta 0), pack-reused 0
Receiving objects: 100% (7/7), done.
C:\Users\ek\src> cmd /c dir symlink-to-qmarks
 Volume in drive C is OS
 Volume Serial Number is B203-10FB

 Directory of C:\Users\ek\src\symlink-to-qmarks

06/25/2024  07:18 PM    <DIR>          .
06/25/2024  07:18 PM    <DIR>          ..
06/25/2024  07:18 PM               617 COPYING
06/25/2024  07:18 PM               298 README.md
06/25/2024  07:18 PM    <SYMLINK>      symlink [???]
               3 File(s)            915 bytes
               2 Dir(s)  83,582,005,248 bytes free
```

Cloning the repository containing a symlink to `CON`:

```text
C:\Users\ek\src> git clone git@github.com:EliahKagan/symlink-to-con.git
Cloning into 'symlink-to-con'...
remote: Enumerating objects: 7, done.
remote: Counting objects: 100% (7/7), done.
remote: Compressing objects: 100% (5/5), done.
remote: Total 7 (delta 0), reused 7 (delta 0), pack-reused 0
Receiving objects: 100% (7/7), done.
C:\Users\ek\src> cmd /c dir symlink-to-con
 Volume in drive C is OS
 Volume Serial Number is B203-10FB

 Directory of C:\Users\ek\src\symlink-to-con

06/25/2024  07:20 PM    <DIR>          .
06/25/2024  07:20 PM    <DIR>          ..
06/25/2024  07:20 PM               617 COPYING
06/25/2024  07:20 PM               441 README.md
06/25/2024  07:20 PM    <SYMLINK>      symlink [CON]
               3 File(s)          1,058 bytes
               2 Dir(s)  83,575,894,016 bytes free
```

I showed the output of the `dir` builtin of `cmd.exe` because it distinguishes file and directory symlinks, thereby verifying that Git creates them as file symlinks. If they had been directory symlinks, `<SYMLINKD>` would have been shown instead of `<SYMLINK>`.

The details of how those repositories were prepared are included below.

#### 1. Create test repositories or examine the ones I made

Either create a repository with a symlink to `???` and a repository with a symlink to `CON`, or use the preexisting test repositories I created, [`symlink-to-qmarks`](https://github.com/EliahKagan/symlink-to-qmarks) and [`symlink-to-con`](https://github.com/EliahKagan/symlink-to-con). To verify that both cases are broken, these need to be two separate repositories, because the current behavior is to fail the checkout entirely and leave no cloned files.

I created both repositories on an Ubuntu system (and WSL is definitely sufficient), though it could be done in native Windows. Note that `ln -s` may not be sufficient to create symlinks on native Windows, even though `ln` exists in Git Bash; this may create copies instead.

A repository like `symlink-to-qmarks` can be created in a Unix-like system like this:

```sh
git init symlink-to-qmarks
cd symlink-to-qmarks
ln -s '???' symlink  # Create the dangling symlink.
git add .
git commit -m 'Initial commit'
```

A repository like `symlink-to-con` can be created in a Unix-like system analogously:

```sh
git init symlink-to-con
cd symlink-to-con
ln -s CON symlink  # Create the dangling symlink.
git add .
git commit -m 'Initial commit'
```

They can then be uploaded to remotes and used to test `gix clone`. I made them roughly like that, but I also added more commits with things like readme files, so it would be clear what they were.

All subsequent steps should be done on Windows. I will use SSH URLs here because the output with `gix --trace` is shorter and more readable, but  I have also tested this with HTTPS URLs to verify that the problem is not specific to any transport (though it would be very strange if it were, especially since the specific cause of the bug is known, as detailed above).

#### 2. Optionally, clone the repositories with `git clone` to verify Git behavior

This is detailed in "Git behavior" above, including specific commands to clone the repositories and to inspect the checked out contents.

This shows that, with Git, the clone succeeds, and the specific dangling symlinks (with strangely named targets) are created.

Before proceeding to test with `gix`, remove the created directories (or, alternatively, do all subsequent steps in a different directory) to avoid clashes confusing the experiment. They can be removed in PowerShell with `rm -r -fo`:

```powershell
rm -r -fo symlink-to-qmarks
rm -r -fo symlink-to-con
```

#### 3. Clone the repositories with `gix clone` to observe the bug

Output without `--trace` is shown above in "Current behavior" demonstrating that:

- `gix clone git@github.com:EliahKagan/symlink-to-qmarks.git` fails and reports that the failure is "Caused by":

  ```text
  The filename, directory name, or volume label syntax is incorrect. (os error 123)
  ```

- `gix clone git@github.com:EliahKagan/symlink-to-con.git` fails and reports that the failure is "Caused by":

  ```text
  The parameter is incorrect. (os error 87)
  ```

Inspection reveals that `symlink-to-qmarks` and `symlink-to-con` directories do not exist after the operation. Thus the clone as a whole failed, rather than just refusing to check out specific files.

#### 4. Optionally, clone the repositories with `gix --trace clone`

When passing `--trace`, the output is very long for HTTPS URLs, which is why SSH URLs have been used throughout, so this output is reasonably short and can be compared directly to non-`--trace` output (and to Git behavior).

Cloning the repository with a symlink to `???` with `--trace` looks like:

```text
C:\Users\ek\src> gix --trace clone git@github.com:EliahKagan/symlink-to-qmarks.git
 20:03:05 indexing done 7.0 objects in 0.00s (25.4k objects/s)
 20:03:05 decompressing done 1.5KB in 0.00s (5.3MB/s)
 20:03:05     Resolving done 7.0 objects in 0.05s (136.0 objects/s)
 20:03:05      Decoding done 1.5KB in 0.05s (29.5KB/s)
 20:03:05 writing index file done 1.3KB in 0.00s (6.9MB/s)
 20:03:05  create index file done 7.0 objects in 0.06s (126.0 objects/s)
 20:03:05          read pack done 1.1KB in 0.07s (16.2KB/s)
 20:03:05            tracing INFO     run [ 1.09s | 3.73% / 100.00% ]
 20:03:05            tracing INFO     ┝━ open_from_paths() [ 135ms | 7.60% / 12.34% ]
 20:03:05            tracing INFO     │  ┝━ gix_path::git::install_config_path() [ 51.4ms | 4.71% ]
 20:03:05            tracing DEBUG    │  │  ┕━ 🐛 [debug]: invoking git for installation config path | cmd: "git.exe" "config" "-l" "--show-origin"
 20:03:05            tracing DEBUG    │  ┝━ 🐛 [debug]: invoking git to get system prefix/exec path | cmd: "git.exe" "--exec-path"
 20:03:05            tracing INFO     │  ┕━ gix_odb::Store::at() [ 326µs | 0.03% ]
 20:03:05            tracing INFO     ┝━ remote::Connection::ref_map() [ 735ms | 0.00% / 67.32% ]
 20:03:05            tracing INFO     │  ┕━ remote::Connection::fetch_refs() [ 735ms | 0.01% / 67.32% ]
 20:03:05            tracing DEBUG    │     ┝━ gix_protocol::handshake() [ 647ms | 59.26% ] service: UploadPack | extra_parameters: []
 20:03:05            tracing DEBUG    │     │  ┕━ 🐛 [debug]: gix_transport::SpawnProcessOnDemand | command: "C:\\Windows\\System32\\OpenSSH\\ssh.exe" "-o" "SendEnv=GIT_PROTOCOL" "git@github.com" "git-upload-pack" "\'EliahKagan/symlink-to-qmarks.git\'"
 20:03:05            tracing DEBUG    │     ┕━ gix_protocol::ls_refs() [ 87.9ms | 8.05% ] capabilities: Capabilities { data: "agent=git/github-74f58a100f14\nls-refs=unborn\nfetch=shallow wait-for-done filter\nserver-option\nobject-format=sha1", value_sep: 10 }
 20:03:05            tracing INFO     ┝━ fetch::Prepare::receive() [ 179ms | 0.28% / 16.41% ]
 20:03:05            tracing DEBUG    │  ┝━ negotiate [ 87.6ms | 0.09% / 8.03% ] protocol_version: 2
 20:03:05            tracing DEBUG    │  │  ┝━ mark_complete_and_common_ref [ 1.91ms | 0.14% / 0.18% ] mappings: 2
 20:03:05            tracing INFO     │  │  │  ┝━ mark_all_refs [ 333µs | 0.03% ]
 20:03:05            tracing DEBUG    │  │  │  ┝━ mark_alternate_refs [ 200ns | 0.00% ] num_odb: 0
 20:03:05            tracing INFO     │  │  │  ┝━ mark known_common [ 1.50µs | 0.00% ]
 20:03:05            tracing DEBUG    │  │  │  ┕━ mark tips [ 300ns | 0.00% ] num_tips: 0
 20:03:05            tracing DEBUG    │  │  ┕━ negotiate round [ 84.8ms | 7.76% ] round: 1
 20:03:05            tracing INFO     │  ┝━ gix_pack::Bundle::write_to_directory() [ 68.5ms | 6.28% ]
 20:03:05            tracing DEBUG    │  ┕━ update_refs() [ 20.0ms | 0.61% / 1.83% ] mappings: 2
 20:03:05            tracing DEBUG    │     ┕━ apply [ 13.3ms | 1.22% ] edits: 2
 20:03:05            tracing INFO     ┕━ gix::clone::PrepareCheckout::main_worktree() [ 2.16ms | 0.20% / 0.20% ]
 20:03:05            tracing INFO        ┕━ gix_index::State::from_tree() [ 15.3µs | 0.00% ]
Error: IO error while writing blob or reading file metadata or changing filetype

Caused by:
    The filename, directory name, or volume label syntax is incorrect. (os error 123)
```

Cloning the repository with a symlink to `CON` with `--trace` looks like:

```text
C:\Users\ek\src> gix --trace clone git@github.com:EliahKagan/symlink-to-con.git
 20:04:08 indexing done 7.0 objects in 0.00s (12.6k objects/s)
 20:04:08 decompressing done 1.6KB in 0.00s (2.9MB/s)
 20:04:08     Resolving done 7.0 objects in 0.05s (137.0 objects/s)
 20:04:08      Decoding done 1.6KB in 0.05s (32.5KB/s)
 20:04:08 writing index file done 1.3KB in 0.00s (7.5MB/s)
 20:04:08  create index file done 7.0 objects in 0.06s (109.0 objects/s)
 20:04:08          read pack done 1.2KB in 0.07s (17.4KB/s)
 20:04:08            tracing INFO     run [ 1.20s | 4.57% / 100.00% ]
 20:04:08            tracing INFO     ┝━ open_from_paths() [ 143ms | 7.48% / 11.95% ]
 20:04:08            tracing INFO     │  ┝━ gix_path::git::install_config_path() [ 53.1ms | 4.44% ]
 20:04:08            tracing DEBUG    │  │  ┕━ 🐛 [debug]: invoking git for installation config path | cmd: "git.exe" "config" "-l" "--show-origin"
 20:04:08            tracing DEBUG    │  ┝━ 🐛 [debug]: invoking git to get system prefix/exec path | cmd: "git.exe" "--exec-path"
 20:04:08            tracing INFO     │  ┕━ gix_odb::Store::at() [ 326µs | 0.03% ]
 20:04:08            tracing INFO     ┝━ remote::Connection::ref_map() [ 792ms | 0.00% / 66.26% ]
 20:04:08            tracing INFO     │  ┕━ remote::Connection::fetch_refs() [ 792ms | 0.01% / 66.26% ]
 20:04:08            tracing DEBUG    │     ┝━ gix_protocol::handshake() [ 686ms | 57.39% ] service: UploadPack | extra_parameters: []
 20:04:08            tracing DEBUG    │     │  ┕━ 🐛 [debug]: gix_transport::SpawnProcessOnDemand | command: "C:\\Windows\\System32\\OpenSSH\\ssh.exe" "-o" "SendEnv=GIT_PROTOCOL" "git@github.com" "git-upload-pack" "\'EliahKagan/symlink-to-con.git\'"
 20:04:08            tracing DEBUG    │     ┕━ gix_protocol::ls_refs() [ 106ms | 8.86% ] capabilities: Capabilities { data: "agent=git/github-74f58a100f14\nls-refs=unborn\nfetch=shallow wait-for-done filter\nserver-option\nobject-format=sha1", value_sep: 10 }
 20:04:08            tracing INFO     ┝━ fetch::Prepare::receive() [ 204ms | 0.03% / 17.07% ]
 20:04:08            tracing DEBUG    │  ┝━ negotiate [ 91.5ms | 0.06% / 7.65% ] protocol_version: 2
 20:04:08            tracing DEBUG    │  │  ┝━ mark_complete_and_common_ref [ 1.11ms | 0.06% / 0.09% ] mappings: 2
 20:04:08            tracing INFO     │  │  │  ┝━ mark_all_refs [ 342µs | 0.03% ]
 20:04:08            tracing DEBUG    │  │  │  ┝━ mark_alternate_refs [ 200ns | 0.00% ] num_odb: 0
 20:04:08            tracing INFO     │  │  │  ┝━ mark known_common [ 2.00µs | 0.00% ]
 20:04:08            tracing DEBUG    │  │  │  ┕━ mark tips [ 200ns | 0.00% ] num_tips: 0
 20:04:08            tracing DEBUG    │  │  ┕━ negotiate round [ 89.7ms | 7.50% ] round: 1
 20:04:08            tracing INFO     │  ┝━ gix_pack::Bundle::write_to_directory() [ 68.1ms | 5.69% ]
 20:04:08            tracing DEBUG    │  ┕━ update_refs() [ 44.2ms | 1.32% / 3.69% ] mappings: 2
 20:04:08            tracing DEBUG    │     ┕━ apply [ 28.4ms | 2.38% ] edits: 2
 20:04:08            tracing INFO     ┕━ gix::clone::PrepareCheckout::main_worktree() [ 1.81ms | 0.15% / 0.15% ]
 20:04:08            tracing INFO        ┕━ gix_index::State::from_tree() [ 16.4µs | 0.00% ]
Error: IO error while writing blob or reading file metadata or changing filetype

Caused by:
    The parameter is incorrect. (os error 87)
```

#### Solution

My forthcoming PR will have tests that fail before the bugfix and pass afterwards. I have already written those tests, verified that they fail with no change to the code under test, written the fix, and verify that the tests pass with it. My pull request is not ready yet because:

- [ ] I want to see if there are any symlink target related errors that are currently misinterpreted as *collisions* and, if so, to report that and to verify that this fixes that as well. I had originally meant to finish researching that and include it as part of this issue, but, if present, such errors are really a separate bug, due to the additional wrong behavior of being reported as collisions, and due to causing only individual symlinks not to be created rather than an entire clone and checkout to fail. *Edit: I have reported this as [#1421](https://github.com/Byron/gitoxide/issues/1421).*
- [ ] I need to generate some new archives for new fixtures, which I do on a separate machine from the Windows system where I am developing the fix (so all committed archives are generated on Unix-like systems).
- [ ] It looks like there are no tests that directory symlinks are *ever* created on Windows. I'll look more into that and, if it is the case, I may open a new issue for that and/or include such a test in my PR. The reason this may be appropriate to cover at the same time is that it would facilitate verifying that my solution does not cause file symlinks to be created even in situations where directory symlinks are already intended to be created. *Edit: I have reported this as [#1422](https://github.com/Byron/gitoxide/issues/1422).*

But when that is ready, running its own tests against the current version of the code will also work as verification, as would running the new fixtures manually to produce test repositories and then testing with those repositories.

### LINKED ISSUE #1421: PermissionDenied checking Windows symlink target is misinterpreted as collision
#### Background

Checking out files into a working tree skips any individual files that already exist, reporting them correctly as collisions. This happens when overwriting is prohibited, as is the case in `gix clone`. For example, if a separate application creates files inside the working tree `gix clone` is checking out, and such a file would be overwritten if checked out, `gix clone` declines to check out that specific file and reports it as a collision. This behavior is correct and should not be changed.

Like other operating systems that support symbolic links, Windows permits a symbolic link to point to a location that is inaccessible to the user creating the symbolic link, including to a location where, due to permissions, traversal to the location will fail due to permissions. The user will not be able to deference that symbolic link other than to inspect the target path, not even to check it's target's existence and metadata, but the symlink itself is no problem.

Unlike other operating systems that support symbolic links, Windows symbolic links come in two varieties: file and directory symbolic links. For this reason, gitoxide attempts to check the metadata of the symlink's target, to decide which type of symlink to create.

#### The problem

The problem is that, when attempting to check the metadata of the symlink's target fails with a `PermissionDenied` error, this is wrongly misinterpreted as a collision, even though there is no collision because the error happens in reading what the index specifies is the symlink's target, not when creating a new file. A wrong or unusual target of a symlink is never a real collision, because checkout does not create or mutate anything at the target.

This is what it looks like to use `gix clone` to clone a repository with such a symlink:

```text
C:\Users\ek\src> gix clone git@github.com:EliahKagan/symlink-to-inaccessible.git
 20:53:39 indexing done 19.0 objects in 0.00s (60.9k objects/s)
 20:53:39 decompressing done 3.6KB in 0.00s (11.2MB/s)
 20:53:39     Resolving done 19.0 objects in 0.05s (377.0 objects/s)
 20:53:39      Decoding done 3.6KB in 0.05s (70.9KB/s)
 20:53:39 writing index file done 1.6KB in 0.00s (7.4MB/s)
 20:53:39  create index file done 19.0 objects in 0.06s (343.0 objects/s)
 20:53:39          read pack done 2.5KB in 0.09s (27.1KB/s)
 20:53:39           checkout done 3.0 files in 0.00s (2.4k files/s)
 20:53:39            writing done 1.4KB in 0.00s (1.1MB/s)
HEAD:refs/remotes/origin/HEAD (implicit)
        77d5afef8ef5ff6f3af10ab900cd1a08e18fc1cc HEAD symref-target:refs/heads/main -> refs/remotes/origin/HEAD [new]
+refs/heads/*:refs/remotes/origin/*
        77d5afef8ef5ff6f3af10ab900cd1a08e18fc1cc refs/heads/main -> refs/remotes/origin/main [new]
symlink: collision (PermissionDenied)
Error: One or more errors occurred - checkout is incomplete: encountered 1 collision(s)
```

The repository is created, but due to the symlink wrongly being detected as a collision, no symlink is created there:

```text
C:\Users\ek\src> ls symlink-to-inaccessible

    Directory: C:\Users\ek\src\symlink-to-inaccessible

d----           6/25/2024  8:53 PM                .git
-a---           6/25/2024  8:53 PM            617 COPYING
-a---           6/25/2024  8:53 PM            826 README.md
```

#### Relationship to #1420

This is related to #1420. As there, the error occurs when attempting to access metadata on the symlink's target, which is only done on Windows, so this bug does not affect non-Windows systems, and which is only done when actually attempting to create symbolic links, so this bug does not affect uses of `gix clone` where `core.symlinks` has been forced to `false` (though, as there, #1353 affects this).

Depending on how #1420 is fixed, the fix may automatically take care of this as well, by making it so no `PermissionDenied` error can ever be propagated upward from a position where it doesn't represent a real collision. I had originally planned to report this as part of #1420. However, on reflection I decided these are separate bugs, because:

- Other ways of fixing that bug would not necessarily fix this, and vice versa.
- This bug has the additional wrong behavior of reporting a non-collision condition as a collision.
- This bug has the mitigating factor that it does not cause an entire clone to be canceled and the partially checked out working tree to be deleted (though the reason it doesn't do that is that it is wrongly treated as a collision).
- This bug is less straightforward to write a regression test for in the test suite. It can definitely be done, though. Whether it needs to be done, and whether it needs to be done before a change that fixes the bug is accepted, might depend on how this bug is fixed. If it is fixed together with [#1420](https://github.com/Byron/gitoxide/issues/1420) then the tests for that *might* be considered enough. Otherwise, it probably needs its own tests.

### Expected behavior 🤔

Symbolic links to locations to which the application cannot traverse, including when the error is `PermissionDenied`, should be created, unless a specific rationale is known for declining to do so. Such a rationale would have to overcome the following factors:

- Windows, like other operating systems, generally allows this.
- Git creates them (as detailed below).
- Dangling symlinks are created—since https://github.com/Byron/gitoxide/commit/31d02a89ee15a0df463a5b5c34864497f8adfae1 ([#1363](https://github.com/Byron/gitoxide/pull/1363)), which fixed [#1354](https://github.com/Byron/gitoxide/issues/1354)—and this is not appreciably different.
- Dangling symlinks are created, and they can subsequently *become* this scenario if the target is created with restrictive permissions.
- *Non-dangling* symlinks are created, and they can become this scenario if permissions are changed, or if the referred-to files are removed (making the symlinks dangle) and later replaced with files with restrictive permissions.
- This is not fundamentally different from a symlink that *sometimes* dangles because the volume it is on is disconnected, not mounted, or otherwise unavailable.
- Allowing this, as Git does, seems considerably simpler than not allowing it but correcting the misreported collision.

As for dangling symlinks, these can be created as file symlinks. As in dangling symlinks, this is imperfect, but not inherently different from the situation of dangling symlinks where the target may later created as a directory. In addition, there may be existing conditions where a file or directory target is hidden from the process rather than access being denied explicitly, in which case that is already treated as dangling.

If for some reason that is not to be done, another aspect of the expected behavior is that these are not collisions and should not be reported as such.

### Git behavior

Git creates the symlinks. It creates them as file symlinks, even when the inaccessible targets are directories, since it cannot know they are directories and it falls back to creating file symlinks when the target type cannot be discerned:

```text
C:\Users\ek\src> rm -r -fo symlink-to-inaccessible
C:\Users\ek\src> git clone git@github.com:EliahKagan/symlink-to-inaccessible.git
Cloning into 'symlink-to-inaccessible'...
remote: Enumerating objects: 19, done.
remote: Counting objects: 100% (19/19), done.
remote: Compressing objects: 100% (9/9), done.
remote: Total 19 (delta 0), reused 19 (delta 0), pack-reused 0
Receiving objects: 100% (19/19), done.
C:\Users\ek\src> cmd /c dir symlink-to-inaccessible
 Volume in drive C is OS
 Volume Serial Number is B203-10FB

 Directory of C:\Users\ek\src\symlink-to-inaccessible

06/25/2024  09:25 PM    <DIR>          .
06/25/2024  09:25 PM    <DIR>          ..
06/25/2024  09:25 PM               617 COPYING
06/25/2024  09:25 PM               826 README.md
06/25/2024  09:25 PM    <SYMLINK>      symlink [\Program Files\WindowsApps\MutableBackup]
               3 File(s)          1,443 bytes
               2 Dir(s)  84,863,729,664 bytes free
```

The referred-to `MutableBackup` entry is actually a directory, but the `git` process has no way of knowing that because it is not allowed to access metadata on that item or list the contents of its parent directory. So Git has created a file symlink.

This can be seen because the `dir` builtin in `cmd.exe` shows `<SYMLINK>`, whereas for a directory symlink it would show `<SYMLINKD>`.

Details on that particular path and why I used it for testing are included below.

#### 1. Create a repository with a symlink into an inaccessible location

Whether such a location suitable for testing already exists or has to be created may vary by system.

On current desktop versions of Windows, the Microsoft Store keeps its Store applications in the `WindowsApps` subdirectory of the `Program Files` directory, which has restrictive permissions such that even administrators cannot list the directory's contents or access information on specific known subdirectory, at least if UAC is enabled and elevation has not been performed. (On server versions of Windows, the Microsoft Store is absent.)

The specific paths vary across systems. I decided to use the drive-relative path to a known subdirectory of `WindowsApps`. Although I actually performed more steps because I tried some other locations and ways of specifying the path--and although it is fairly straightforward to create such a repository in a native Windows system--I created [the repository](https://github.com/EliahKagan/symlink-to-inaccessible) on an Ubuntu system (WSL is sufficient) in a manner *similar to* this, which is sufficient:

```sh
git init symlink-to-inaccessible
cd symlink-to-inaccessible
ln -s '\Program Files\WindowsApps\MutableBackup' symlink
git add .
git commit -m 'Initial commit'
```

This is a symlink to the `MutableBackup` subdirectory of `WindowsApps`, which users cannot typically access, if the repository is cloned on the system drive (or otherwise on the same drive that contains a `Program Files` directory that contains `WindowsApps`, in the rare case that this is different).

I made the test repository this way for convenience, so that the steps to verify this bug are simple in most cases, once the repository is already available. This can alternatively be tested by creating a new directory tree and setting restrictive ACL permissions.

The repository is available [on GitHub](https://github.com/EliahKagan/symlink-to-inaccessible). The following steps assume it is used and will have to be modified otherwise. The output shown reflects the presence of additional files in the repository, such as `README.md`, which are present mainly as documentation, though they also help verify that other files, besides the symlink, are still checked out in `gix clone`.

#### 2. Optionally, clone the repo with `git clone` to verify Git behavior

The `git clone` behavior, including the specific command, is shown above in the "Git behavior" section.

The created `symlink-to-inacessible` directory should be deleted before testing again with `gix clone` (if it is to be run in the same location), which in PowerShell can be done with:

```powershell
rm -r -fo symlink-to-inaccessible
```

#### 3. Clone the repo with `gix clone` to observe the bug

This is shown above in "Current behavior," showing the error:

```text
symlink: collision (PermissionDenied)
Error: One or more errors occurred - checkout is incomplete: encountered 1 collision(s)
```

After checking the output, and that the working tree exists with other files besides the symlink that is wrongly taken to be a collision, the created directory should be removed again before running further tests, to avoid clashes.

#### 4. Optionally, clone with `--trace` for more output

This is mainly to help verify that nothing else unusual is happening. It can also be compared to the output in #1420.

```text
C:\Users\ek\src> gix --trace clone git@github.com:EliahKagan/symlink-to-inaccessible.git
 20:54:22 indexing done 19.0 objects in 0.00s (55.1k objects/s)
 20:54:22 decompressing done 3.6KB in 0.00s (10.0MB/s)
 20:54:22     Resolving done 19.0 objects in 0.05s (375.0 objects/s)
 20:54:22      Decoding done 3.6KB in 0.05s (70.5KB/s)
 20:54:22 writing index file done 1.6KB in 0.00s (6.7MB/s)
 20:54:22  create index file done 19.0 objects in 0.05s (347.0 objects/s)
 20:54:22          read pack done 2.5KB in 0.06s (41.0KB/s)
 20:54:22           checkout done 3.0 files in 0.00s (2.5k files/s)
 20:54:22            writing done 1.4KB in 0.00s (1.2MB/s)
 20:54:22            tracing INFO     run [ 1.12s | 2.43% / 100.00% ]
 20:54:22            tracing INFO     ┝━ open_from_paths() [ 147ms | 8.21% / 13.11% ]
 20:54:22            tracing INFO     │  ┝━ gix_path::git::install_config_path() [ 54.4ms | 4.86% ]
 20:54:22            tracing DEBUG    │  │  ┕━ 🐛 [debug]: invoking git for installation config path | cmd: "git.exe" "config" "-l" "--show-origin"
 20:54:22            tracing DEBUG    │  ┝━ 🐛 [debug]: invoking git to get system prefix/exec path | cmd: "git.exe" "--exec-path"
 20:54:22            tracing INFO     │  ┕━ gix_odb::Store::at() [ 343µs | 0.03% ]
 20:54:22            tracing INFO     ┝━ remote::Connection::ref_map() [ 767ms | 0.00% / 68.59% ]
 20:54:22            tracing INFO     │  ┕━ remote::Connection::fetch_refs() [ 767ms | 0.01% / 68.58% ]
 20:54:22            tracing DEBUG    │     ┝━ gix_protocol::handshake() [ 678ms | 60.61% ] service: UploadPack | extra_parameters: []
 20:54:22            tracing DEBUG    │     │  ┕━ 🐛 [debug]: gix_transport::SpawnProcessOnDemand | command: "C:\\Windows\\System32\\OpenSSH\\ssh.exe" "-o" "SendEnv=GIT_PROTOCOL" "git@github.com" "git-upload-pack" "\'EliahKagan/symlink-to-inaccessible.git\'"
 20:54:22            tracing DEBUG    │     ┕━ gix_protocol::ls_refs() [ 89.1ms | 7.97% ] capabilities: Capabilities { data: "agent=git/github-74f58a100f14\nls-refs=unborn\nfetch=shallow wait-for-done filter\nserver-option\nobject-format=sha1", value_sep: 10 }
 20:54:22            tracing INFO     ┝━ fetch::Prepare::receive() [ 175ms | 0.30% / 15.62% ]
 20:54:22            tracing DEBUG    │  ┝━ negotiate [ 87.8ms | 0.07% / 7.85% ] protocol_version: 2
 20:54:22            tracing DEBUG    │  │  ┝━ mark_complete_and_common_ref [ 1.10ms | 0.07% / 0.10% ] mappings: 2
 20:54:22            tracing INFO     │  │  │  ┝━ mark_all_refs [ 328µs | 0.03% ]
 20:54:22            tracing DEBUG    │  │  │  ┝━ mark_alternate_refs [ 300ns | 0.00% ] num_odb: 0
 20:54:22            tracing INFO     │  │  │  ┝━ mark known_common [ 1.80µs | 0.00% ]
 20:54:22            tracing DEBUG    │  │  │  ┕━ mark tips [ 300ns | 0.00% ] num_tips: 0
 20:54:22            tracing DEBUG    │  │  ┕━ negotiate round [ 86.0ms | 7.69% ] round: 1
 20:54:22            tracing INFO     │  ┝━ gix_pack::Bundle::write_to_directory() [ 63.3ms | 5.66% ]
 20:54:22            tracing DEBUG    │  ┕━ update_refs() [ 20.3ms | 0.66% / 1.81% ] mappings: 2
 20:54:22            tracing DEBUG    │     ┕━ apply [ 12.9ms | 1.16% ] edits: 2
 20:54:22            tracing INFO     ┕━ gix::clone::PrepareCheckout::main_worktree() [ 2.84ms | 0.19% / 0.25% ]
 20:54:22            tracing INFO        ┝━ gix_index::State::from_tree() [ 14.7µs | 0.00% ]
 20:54:22            tracing ERROR       ┝━ 🚨 [error]: symlink: collided (PermissionDenied)
 20:54:22            tracing DEBUG       ┕━ gix_index::File::write() [ 720µs | 0.06% / 0.06% ] path: "symlink-to-inaccessible\\.git\\index"
 20:54:22            tracing DEBUG          ┕━ gix_index::File::write_to() [ 11.9µs | 0.00% / 0.00% ] skip_hash: false
 20:54:22            tracing INFO              ┕━ gix_index::State::write() [ 7.70µs | 0.00% ]
HEAD:refs/remotes/origin/HEAD (implicit)
        77d5afef8ef5ff6f3af10ab900cd1a08e18fc1cc HEAD symref-target:refs/heads/main -> refs/remotes/origin/HEAD [new]
+refs/heads/*:refs/remotes/origin/*
        77d5afef8ef5ff6f3af10ab900cd1a08e18fc1cc refs/heads/main -> refs/remotes/origin/main [new]
symlink: collision (PermissionDenied)
Error: One or more errors occurred - checkout is incomplete: encountered 1 collision(s)
```

### LINKED ISSUE #1422: Test suite does not assert directory symlink creation
### Current behavior 😯

On Windows, which has separate file and directory symlinks, gitoxide falls back to creating file symlinks when the target is not found, since 31d02a8 (#1363), to support dangling symlinks (#1354). I wish to create file symlinks in even more situations, to fix #1420 and #1421. But this runs the risk of accidentally converting situations where directory symlinks are currently being created into situations where file symlinks are created, if I make a mistake.

On a feature branch which should never be merged, I [tried deliberately breaking symlink creation](https://github.com/EliahKagan/gitoxide/commit/cf9e727c4ad66bddcebf1ead8a92d7f8b919a331) to only ever create file symlinks on Windows. [Enabling CI](https://github.com/EliahKagan/gitoxide/commit/a7dea6bbc23f0abec560e7273d67a841cfd2609b) on [that branch](https://github.com/EliahKagan/gitoxide/commits/ciall-no-dir-symlinks-experiment/) reveals that [this breaking causes no existing tests to fail](https://github.com/EliahKagan/gitoxide/actions/runs/9672723831/job/26685515780). I have also produced this effect locally.

Ordinarily I would just open a PR that adds a test. I intend to open a PR that fixes #1420 and hopefully also #1421, and to include a test that a symlink to a directory is created as a directory symlink on Windows. I plan to include it there so that it validates that my fix there does not break this functionality. I am also opening an issue for this, for two reasons:

1. If, for simplicity, the new test is not wanted in that PR, then it can be removed and the need for it will still not be forgotten.
2. It is not obvious how this test should be done, or how thoroughly it should be done. Rather than have that PR be the only place to discuss it, I'm opening this issue so that possible concerns with the approach I plan to take can be easily raised.

On that second point, the approach I am thinking of taking is actually not to check the created symlink type at all, but instead to write an operating-system agnostic test that will fail on Windows if the wrong type of symlink is created, testing the actual functionality of the symlink.

[Some experiments I have done](https://github.com/EliahKagan/is-dir-symlink?tab=readme-ov-file#results-of-the-experiment) reveal that obtaining metadata with [`std::fs::metadata`](https://doc.rust-lang.org/std/fs/fn.metadata.html) will fail to traverse through the wrong type of symlink to access metadata on the target. (This is not for lack of trying. Unlike [`symlink_metadata`](https://doc.rust-lang.org/std/fs/fn.symlink_metadata.html), which is `lstat`-like, the `metadata` function is `stat`-like and attempts to obtain metadata of the ultimate target of the symlink.) Therefore, a test can simply attempt to do that, and also assert that the target is a directory; together, that should provide both validation and insight into what goes wrong if there is a regression.

The advantage of that approach is that it tests the desired behavior that symlinks to directories are, when possible, created in a way that actually works, on all systems, including Windows where it happens to be easy to get that wrong.

The disadvantage is that I am not sure the *in*ability to traverse through the wrong type of symlink to obtain metadata is a guaranteed part of the behavior of the `metadata` function in the Rust standard library.

If this approach is inadequate, then it could either be complimented or supplanted by another test that uses Windows-specific functions to check the actual type.

### Expected behavior 🤔

I think at least one test in the test suite should fail if `gix clone` on Windows always creates symlinks as file symbolic links and never as directory symbolic links even if the target is a directory that already exists.

To be clear, I don't believe gitoxide currently has that or any bug like it. This issue is solely about what the test suite checks for. The expected behavior I'm describing is the behavior of helping to keep me from creating such a bug when changing the code that decides what type of symlink to create on Windows.

### Git behavior

I have not researched if, or how, Git's own test suite checks for this.

Whether or not the approach I suggested above and plan to implement soon is used, I don't anticipate any large difficulties extending the test suite to cover this. However, if I encounter problems, I may research that for comparison. I can also look into it if there is interest, separate from any need.

### Steps to reproduce 🕹

I recommend just checking the linked experimental branch and test results above.

However, because I intend to eventually delete that branch (after this issue is closed, but the issue will still *exist*), here's a diff that introduces the breakage that this issue is about the test suite not yet failing on:

```diff
diff --git a/gix-fs/src/symlink.rs b/gix-fs/src/symlink.rs
index ce639c48f..a7571c0c6 100644
--- a/gix-fs/src/symlink.rs
+++ b/gix-fs/src/symlink.rs
@@ -33,24 +33,12 @@ pub fn remove(path: &Path) -> io::Result<()> {

 /// Create a new symlink at `link` which points to `original`.
 ///
-/// Note that if a symlink target (the `original`) isn't present on disk, it's assumed to be a
-/// file, creating a dangling file symlink. This is similar to a dangling symlink on Unix,
-/// which doesn't have to care about the target type though.
+/// This temporarily broken implementation is to test the test suite.
+/// In real life, we should sometimes create directory symlinks!
+/// This change should not be merged and should not appear in any releases!!
 #[cfg(windows)]
 pub fn create(original: &Path, link: &Path) -> io::Result<()> {
-    use std::os::windows::fs::{symlink_dir, symlink_file};
-    // TODO: figure out if links to links count as files or whatever they point at
-    let orig_abs = link.parent().expect("dir for link").join(original);
-    let is_dir = match std::fs::metadata(orig_abs) {
-        Ok(m) => m.is_dir(),
-        Err(err) if err.kind() == io::ErrorKind::NotFound => false,
-        Err(err) => return Err(err),
-    };
-    if is_dir {
-        symlink_dir(original, link)
-    } else {
-        symlink_file(original, link)
-    }
+    std::os::windows::fs::symlink_file(original, link)
 }

 /// Return true if `err` indicates that a file collision happened, i.e. a symlink couldn't be created as the `link`
```

I emphasize that this diff is included for posterity and not as a request that anyone use it.

EliahKagan:
Fixes #1420
Fixes #1421
Fixes #1422

#### Changes

This adds tests that symlinks can specify targets with unusual names that, on Windows, are invalid or reserved (#1420), which I verified locally to fail until changes are made to the code under test; as well as a regression test that symlinks to directories on Windows are created as directory symlinks (#1422; see [verification](https://github.com/EliahKagan/gitoxide/actions/runs/9673788215/job/26688333583) that the test is able to fail).

This also modifies `gix-fs/src/symlink.rs`, using the technique proposed in #1420 to convert all errors accessing metadata to guess the proper symlink type, and not just `NotFound` errors, into `false`. This falls back to creating file symlinks in more situations, solving #1420 and #1421.

This does not contain new tests specifically for #1421. Since the same technique fixed both, I am unsure if that is needed at this time. There may also be various other error conditions that would have prevented checkout or been misinterpreted as collisions that this now addresses by falling back to creating file symlinks, as Git does.

#### Comparison to Git

There is one subtlety, though not a new one, in comparison to the Git behavior. On Windows, Git repeatedly rechecks if symbolic links that it created as file symlinks should be replaced with directory symlinks. This allows Git to check out the best type of symlink in more situations where ordering would otherwise matter, though it still will check out a file symlink when the target is an existing but dangling directory symlink (thus it is still not intuitive), and I think it may have performance implications. Even if it does not cause performance problems, I think it may need to be implemented very carefully if done for gitoxide, since the facilities in the Rust standard library access a large amount of metadata with multiple Windows API calls plus further fallback calls. (The algorithm Git for Windows uses for this is worst-case quadratic in the number of files involved. If only files from the index were involved, then it would be possible to implement a linear-work alternative by using topological sort. But the files involved need not all be from the index.)

However, that difference between Git and gitoxide in the treatment of symlinks was already present. This leaves that difference unchanged, and merely converts various situations that were previously errors into successful creation of file symlinks.

#### Manual testing

I have repeated the manual testing described in #1420 and #1421 after installing the version from this feature branch with `cargo install --path .`, further verifying the fixes.

With the repository that has a symlink to `???`:

```text
C:\Users\ek\src> gix clone git@github.com:EliahKagan/symlink-to-qmarks.git
 01:28:31 indexing done 7.0 objects in 0.00s (24.3K objects/s)
 01:28:31 decompressing done 1.5KB in 0.00s (5.0MB/s)
 01:28:31     Resolving done 7.0 objects in 0.05s (137.0 objects/s)
 01:28:31      Decoding done 1.5KB in 0.05s (29.6KB/s)
 01:28:31 writing index file done 1.3KB in 0.00s (5.1MB/s)
 01:28:31  create index file done 7.0 objects in 0.06s (110.0 objects/s)
 01:28:31          read pack done 1.1KB in 0.06s (16.9KB/s)
 01:28:31           checkout done 3.0 files in 0.00s (1.8K files/s)
 01:28:31            writing done 918B in 0.00s (549.7KB/s)
HEAD:refs/remotes/origin/HEAD (implicit)
        1f0a60ad33050fe90560d80b752f49d6b2121b06 HEAD symref-target:refs/heads/main -> refs/remotes/origin/HEAD [new]
+refs/heads/*:refs/remotes/origin/*
        1f0a60ad33050fe90560d80b752f49d6b2121b06 refs/heads/main -> refs/remotes/origin/main [new]
C:\Users\ek\src> cmd /c dir symlink-to-qmarks
 Volume in drive C is OS
 Volume Serial Number is B203-10FB

 Directory of C:\Users\ek\src\symlink-to-qmarks

06/26/2024  01:28 AM    <DIR>          .
06/26/2024  01:28 AM    <DIR>          ..
06/26/2024  01:28 AM    <DIR>          .git
06/26/2024  01:28 AM               617 COPYING
06/26/2024  01:28 AM               298 README.md
06/26/2024  01:28 AM    <SYMLINK>      symlink [???]
               3 File(s)            915 bytes
               3 Dir(s)  88,286,535,680 bytes free
```

With the repository that has a symlink to `CON`:

```text
C:\Users\ek\src> gix clone git@github.com:EliahKagan/symlink-to-con.git
 01:29:00 indexing done 7.0 objects in 0.00s (21.1K objects/s)
 01:29:00 decompressing done 1.6KB in 0.00s (4.7MB/s)
 01:29:00     Resolving done 7.0 objects in 0.05s (137.0 objects/s)
 01:29:00      Decoding done 1.6KB in 0.05s (32.3KB/s)
 01:29:00 writing index file done 1.3KB in 0.00s (4.3MB/s)
 01:29:00  create index file done 7.0 objects in 0.06s (111.0 objects/s)
 01:29:00          read pack done 1.2KB in 0.07s (17.3KB/s)
 01:29:00           checkout done 3.0 files in 0.00s (1.8K files/s)
 01:29:00            writing done 1.1KB in 0.00s (646.6KB/s)
HEAD:refs/remotes/origin/HEAD (implicit)
        fe64a5bc3362f7657fbb9e5da495101b5d815fe5 HEAD symref-target:refs/heads/main -> refs/remotes/origin/HEAD [new]
+refs/heads/*:refs/remotes/origin/*
        fe64a5bc3362f7657fbb9e5da495101b5d815fe5 refs/heads/main -> refs/remotes/origin/main [new]
C:\Users\ek\src> cmd /c dir symlink-to-con
 Volume in drive C is OS
 Volume Serial Number is B203-10FB

 Directory of C:\Users\ek\src\symlink-to-con

06/26/2024  01:29 AM    <DIR>          .
06/26/2024  01:29 AM    <DIR>          ..
06/26/2024  01:29 AM    <DIR>          .git
06/26/2024  01:29 AM               617 COPYING
06/26/2024  01:29 AM               441 README.md
06/26/2024  01:29 AM    <SYMLINK>      symlink [CON]
               3 File(s)          1,058 bytes
               3 Dir(s)  88,287,436,800 bytes free
```

With the repository that has a symlink to a file that, on my system, cannot even have its metadata accessed due to giving a permission error:

```text
C:\Users\ek\src> gix clone git@github.com:EliahKagan/symlink-to-inaccessible.git
 01:29:19 indexing done 19.0 objects in 0.00s (43.0K objects/s)
 01:29:19 decompressing done 3.6KB in 0.00s (7.7MB/s)
 01:29:19     Resolving done 19.0 objects in 0.05s (373.0 objects/s)
 01:29:19      Decoding done 3.6KB in 0.05s (70.1KB/s)
 01:29:19 writing index file done 1.6KB in 0.00s (5.4MB/s)
 01:29:19  create index file done 19.0 objects in 0.06s (305.0 objects/s)
 01:29:19          read pack done 2.5KB in 0.06s (40.2KB/s)
 01:29:19           checkout done 3.0 files in 0.00s (1.6K files/s)
 01:29:19            writing done 1.5KB in 0.00s (774.6KB/s)
HEAD:refs/remotes/origin/HEAD (implicit)
        77d5afef8ef5ff6f3af10ab900cd1a08e18fc1cc HEAD symref-target:refs/heads/main -> refs/remotes/origin/HEAD [new]
+refs/heads/*:refs/remotes/origin/*
        77d5afef8ef5ff6f3af10ab900cd1a08e18fc1cc refs/heads/main -> refs/remotes/origin/main [new]
C:\Users\ek\src> cmd /c dir symlink-to-inaccessible
 Volume in drive C is OS
 Volume Serial Number is B203-10FB

 Directory of C:\Users\ek\src\symlink-to-inaccessible

06/26/2024  01:29 AM    <DIR>          .
06/26/2024  01:29 AM    <DIR>          ..
06/26/2024  01:29 AM    <DIR>          .git
06/26/2024  01:29 AM               617 COPYING
06/26/2024  01:29 AM               826 README.md
06/26/2024  01:29 AM    <SYMLINK>      symlink [\Program Files\WindowsApps\MutableBackup]
               3 File(s)          1,443 bytes
               3 Dir(s)  88,286,195,712 bytes free
```

In addition, I have further tested that the changes do not break the creation of directory symlinks where that was already working, by cloning [this test repository](https://github.com/EliahKagan/has-symlink-to-directory). Unlike the above results, this is the same as before the changes (because it was already working before the changes):

```text
C:\Users\ek\src> gix clone git@github.com:EliahKagan/has-symlink-to-directory.git
 01:27:49 indexing done 9.0 objects in 0.00s (18.0K objects/s)
 01:27:49 decompressing done 1.6KB in 0.00s (2.9MB/s)
 01:27:49     Resolving done 9.0 objects in 0.05s (175.0 objects/s)
 01:27:49      Decoding done 1.6KB in 0.05s (31.4KB/s)
 01:27:49 writing index file done 1.3KB in 0.00s (5.5MB/s)
 01:27:49  create index file done 9.0 objects in 0.07s (127.0 objects/s)
 01:27:49          read pack done 1.2KB in 0.12s (9.6KB/s)
 01:27:49           checkout done 4.0 files in 0.00s (1.6K files/s)
 01:27:49            writing done 949B in 0.00s (380.7KB/s)
HEAD:refs/remotes/origin/HEAD (implicit)
        0e838aaff928701c050ec77dc46a66d16aa75464 HEAD symref-target:refs/heads/main -> refs/remotes/origin/HEAD [new]
+refs/heads/*:refs/remotes/origin/*
        0e838aaff928701c050ec77dc46a66d16aa75464 refs/heads/main -> refs/remotes/origin/main [new]
C:\Users\ek\src> cmd /c dir has-symlink-to-directory
 Volume in drive C is OS
 Volume Serial Number is B203-10FB

 Directory of C:\Users\ek\src\has-symlink-to-directory

06/26/2024  01:27 AM    <DIR>          .
06/26/2024  01:27 AM    <DIR>          ..
06/26/2024  01:27 AM    <DIR>          .git
06/26/2024  01:27 AM    <SYMLINKD>     a [b]
06/26/2024  01:27 AM    <DIR>          b
06/26/2024  01:27 AM               617 COPYING
06/26/2024  01:27 AM               331 README.md
               2 File(s)            948 bytes
               5 Dir(s)  88,283,258,880 bytes free
```

#### Other considerations

I have checked [on another branch](https://github.com/EliahKagan/gitoxide/commits/ciall-strange-symlink-targets/), as well as locally, that tests pass [even without generated archives](https://github.com/EliahKagan/gitoxide/actions/runs/9674155178/job/26689272585).

I have proceeded conservatively with respect to the shell script heredoc style issues raised in #1423, leaving the seemingly unnecessary use in place in `<<-` even where in existing tests that are directly relevant to the changes here, while using `<<` in the new tests. I'd be pleased to change this in either direction; but that could probably be done separately and later.


gix-worktree-state/tests/state/checkout.rs
```diff
@@ -291,6 +319,54 @@ fn dangling_symlinks_can_be_created() -> crate::Result {
     Ok(())
 }
 
+#[test]
+fn dangling_symlink_to_windows_invalid_target_can_be_created() -> crate::Result {
```
Byron, line 323: Could you combine the two test fixtures and the two tests itself into one?

They seem to be copy-paste with exactly one part changed (the target of the symlink), and in that case I think it's better to consolidate.

The fixture can reasonably create two dangling links at once as well.

If you prefer to keep the fixtures separate to make them more descriptive, I am up for that too, but truly believe the Rust side should be consolidated.

You will find the following useful.

```Rust
for (fixture, expected_name) in [("fixture1", "CON"), ("fixture2", "???")] {

}
```

Byron (changes requested):
Thanks a lot for the fix and all the research that went into it.

Your diligence (among other traits) I look up to and attempt to apply similar standards in my own work. My primary school teacher always said one should reach for the impossible to reach the possible :D.

With that said, I have a small request regarding deduplication of tests, but with that done I am looking forward to merging.

gix-worktree-state/tests/state/checkout.rs
EliahKagan: Yes, absolutely. Unfortunately I have to go away from my computer for a while right now, but I plan to do this when I get back.

I can see how it is more important to consolidate the Rust code of the tests than the fixtures, since among other things the Rust code of the tests is significantly longer and more complicated. I'll try to figure out which of the two approaches is better--in terms of whether the fixtures should also be consolidated--when I return.

The tests are based very heavily on the `dangling_symlinks_can_be_created` test, and maybe all three could even be combined into one. Then again, these tests, `dangling_symlink_to_windows_invalid_target_can_be_created` and `dangling_symlink_to_windows_reserved_target_can_be_created`, are somewhat more conceptually related to each other than they are to that test. If you have a preference as to whether all three could reasonably be combined, you can let me know. Otherwise, I will also try to figure that out when I come back.

The `symlinks_to_directories_are_usable` test is also somewhat similar, though less so, because it has an additional assertion--the most important of the assertions in it--that the metadata are accessible through the symlink.

> You will find the following useful.

Yes, this pattern appears throughout the test suite and I should really have just used it originally based on that.

I am curious if there is a reasonably uncomplicated way to parameterize test functions, whether through the testing framework, or by using the genericity features of the Rust language itself, so that they are run as independent tests with independent statuses, yet still without code duplication. (Don't worry; that curiosity will not prevent me from using this pattern, which is a decisive improvement compared to what I currently have written, and which can always be improved on further if such a thing could/should ultimately be done.)

Byron: >  If you have a preference as to whether all three could reasonably be combined, you can let me know. Otherwise, I will also try to figure that out when I come back.

I am also very cautious about trying to deduplicate too much - if it's low-hanging, then it should be done, but to my mind there is no need to consolidate more unless one feels it's a problem.

> The `symlinks_to_directories_are_usable` test is also somewhat similar, though less so, because it has an additional assertion--the most important of the assertions in it--that the metadata are accessible through the symlink.

I thought it's specific enough to not try to consolidate. In the test-suite, there are a few examples for complex test-programs with lots of parameters effectively, and despite having good use, they are also hard to understand. It's always a trade-off, but generally I don't spend much brainpower on deduplication, but do it if it's very obvious like here.

> I am curious if there is a reasonably uncomplicated way to parameterize test functions, whether through the testing framework, or by using the genericity features of the Rust language itself, so that they are run as independent tests with independent statuses, yet still without code duplication. (Don't worry; that curiosity will not prevent me from using this pattern, which is a decisive improvement compared to what I currently have written, and which can always be improved on further if such a thing could/should ultimately be done.)

I think the answer is 'yes', but with the right framework. There are property-test crates out there that provide proc-macros to annotate functions, even though I have not used them.

Once somebody introduced a new test-runner that allows to procedurally create test-objects with closures, but ultimately I removed it as it was harder to deal with while some of the tests aren't passing. [This is the file I am talking about, it's history should reveal everything](
https://github.com/Byron/gitoxide/blob/3ef3bc20a1b90799e5ac26858f898bc7a7c96901/gix-url/tests/baseline.rs#L13).

Custom test-runners are not worth it here, but certain utility crates might be, I just didn't use them yet.
Personally, I prefer the simple loop if there isn't too much data, and baseline programs if there is more. Direct Rust code, no magic.

EliahKagan: I ended up consolidating all three dangling symlink tests rather than just the two new ones, though I'd be pleased to consolidate just the two new ones if you prefer that.

The reason I ended up deciding to consolidate all three is that it shows the similarities between what the are testing. Keeping them fully separate had obscured that somewhat; it seems to me that consolidating two of them but not the other might obscure the relationship much more.

Part of this reflects my belief that #1420 should be understood primarily as being about cases of dangling symlinks that were not covered before (per the "A note on the semantics of such symlinks" subsection there). However, that is was actually less of a factor in my thinking than the value that it seems to me comes from seeing exactly what varies across all three of those tests:

https://github.com/Byron/gitoxide/blob/87422ab34922131e3243d5d7db7910a3f68978a4/gix-worktree-state/tests/state/checkout.rs#L306-L310

However, to reiterate, I am fully willing to split the original test and the two new ones, so that only the two new ones are deduplicated.

I have of course, and as mentioned above, *not* attempted to deduplicate anything to do with the new `symlinks_to_directories_are_usable` test, which remains separate from all others, new and old.

Byron (approved):
Great work, thanks a lot!

PR #1430, opened 2024-06-27
gate some deps under features

klensy:
Each commit gates some dep under corresponding features.
Tested locally, but let's see CI.


Byron:
Great work, thanks a lot!

I wonder how you found these - it's usually not at all straightforward to do this unless there is tooling support.

klensy:
Some was catched by cargo udeps, some - manually.

PR #1433, opened 2024-06-29
feat: add `mailmap::Snapshot::iter()`

emilazy:
Allow iterating through entries without allocation.

### DIFF
--- a/gitoxide-core/src/repository/mailmap.rs
+++ b/gitoxide-core/src/repository/mailmap.rs
@@ -45,10 +45,7 @@ pub fn entries(
     }
 
     #[cfg(feature = "serde")]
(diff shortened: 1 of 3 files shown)


Byron (approved):
Thanks a lot!

I have added a baseline test prior to the change, and started using the iterator in a place that previously had to allocate.

When green, I will merge and create a new point release. Thanks also for using using `feat:` to indicate the new feature, this will go straight into the changelog.

Byron:
And the new release is available here: https://github.com/Byron/gitoxide/releases/tag/gix-mailmap-v0.23.4

emilazy:
Thank you so much for the quick release as always! I should have grepped for uses of `.entries()` :)

PR #1435, opened 2024-06-30
Use `<<` rather than `<<-` heredoc operator

EliahKagan:
This makes the change from `<<-` to `<<` discussed in #1423, changing to `<<` because the distinctive behavior with `<<-` of stripping consistent leading tabs was not relevant to any of the here documents in this project where it was being used.

See the commit messages for further details. In particular, the second commit, which regenerates archives, involves a choice related to an ongoing situation where a couple of committed archives have for some reason not be used, and you might prefer that a different choice be made there.

PR #1436, opened 2024-06-30
Remove execution-specific code in sourcing-only .sh files

EliahKagan:
Some of the `.sh` script files are only meant ever to be sourced and not to be executed. These are already correctly not marked executable, but some of them contained code that would only be proper for scripts that are actually run, and not for scripts that are only sourced:

- One contained a `#!` line, which this removes because it falsely indicated an intent to be executed, and because it was not effective even in principle when sourcing the script.

- Two contained `set -eu` commands affecting the behavior of the running interpreter, which were redundant because the script for which they are helpers already runs such a command. If they are later sourced into other scripts (a use for which they are not intended and would probably require further modification), those other scripts might not have these `set` commands, but in that case those commands' behavior would probably not be wanted.

This also adds "Must be sourced into the main journey test" comments at the top of `.sh` files to which they apply, mirroring the presence of such comments in other such files. Besides consistency, this has the advantage of making clear what all the scripts that are meant only for sourcing are being used for.

---

I was originally going to include this in #1435, but it doesn't overlap with the changes there, is not essential to them, probably makes sense to evaluate separately, and could overcomplicate #1435 given the other subtlety raised there relating to which archives that PR should regenerate. So I've opened this separate PR instead.

### DIFF
--- a/tests/helpers.sh
+++ b/tests/helpers.sh
@@ -1,3 +1,4 @@
+# Must be sourced into the main journey test
 
 function set-static-git-environment() {
   set -a
@@ -66,4 +67,3 @@ function launch-git-daemon() {
     done
     trap 'kill $daemon_pid' EXIT
 }
-
--- a/tests/journey/ein.sh
+++ b/tests/journey/ein.sh
@@ -1,5 +1,4 @@
 # Must be sourced into the main journey test
-set -eu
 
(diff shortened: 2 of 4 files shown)


Byron (approved):
Thanks a lot for cleaning this up!

Issue #1442, opened 2024-07-04
many_different_states fails on Windows with GIX_TEST_IGNORE_ARCHIVES=1


EliahKagan:
### Current behavior 😯

Starting after #1441, which fixed #1440, there is a new failing test on Windows with `GIX_TEST_IGNORE_ARCHIVES=1`, in addition to those that had failed before (#1358).

The new failing test is:

```text
        FAIL [   9.690s] gix-diff-tests::diff tree::changes::to_obtain_tree::many_different_states
```

This is one of the tests using the fixture the changes in #1441 applied to.

#### The recent change seems otherwise good

Only Windows is affected, all tests still pass on Windows when `GIX_TEST_IGNORE_ARCHIVES=1` is not used, and no unexpected untracked files are created on any system (at least on Windows, Ubuntu, or macOS).

This is to say that #1441 does appear to have fully fixed #1440, but unfortunately has also produced a test suite regression, on Windows only, for `tree::changes::to_obtain_tree::many_different_states` in `gix-diff-tests::diff`, when the use of pre-generated archives is suppressed with `GIX_TEST_IGNORE_ARCHIVES=1`.

#### Full run output

Full output for this test run, as well as the full error summary (which otherwise lists the same tests as #1358), can be seen in [**this new gist**](https://gist.github.com/EliahKagan/cf7f0063d67c08a67747036bb000f8b9).

#### Details on the failed test

Cleaning ignored files again, and then entering the `gix-diff/tests/tree` directory and running tests from there with `GIX_TEST_IGNORE_ARCHIVES=1` and `RUST_BACKTRACE=full`, shows some more information about the failing test. The full output can, if wanted, be seen in [this other gist](https://gist.github.com/EliahKagan/d374302372a3a43ae2f4bfff2eb3fce4). The other `gix-diff` tests still pass as expected, and the full traceback for the failing test is unremarkable. The interesting part is the assertion message itself:

```text
ek@Glub MINGW64 ~/source/repos/gitoxide/gix-diff/tests/tree (main)
        FAIL [   7.550s] gix-diff-tests::diff tree::changes::to_obtain_tree::many_different_states

--- STDOUT:              gix-diff-tests::diff tree::changes::to_obtain_tree::many_different_states ---

running 1 test
test tree::changes::to_obtain_tree::many_different_states ... FAILED

failures:

failures:
    tree::changes::to_obtain_tree::many_different_states

[...]

Issue #1358, opened 2024-05-07
14 tests fail on Windows with GIX_TEST_IGNORE_ARCHIVES=1


EliahKagan:
### Current behavior 😯

Running tests on Windows in a Git Bash environment (similar to the environment in which they run in Windows on CI, where `bash` is Git Bash), all tests are able to pass normally. However, this apparently relies on the use of generated archives. When the tests are run with  the environment variable `GIX_TEST_IGNORE_ARCHIVES` set to `1` rather than unset, 16 tests fail. The failing tests, as quoted from the end of the test run, are:

```text
     Summary [ 875.860s] 2244 tests run: 2228 passed (30 slow, 17 leaky), 16 failed, 9 skipped
        FAIL [   8.804s] gix-glob::glob pattern::matching::compare_baseline_with_ours
        FAIL [   3.411s] gix-pathspec::pathspec parse::baseline
        FAIL [   0.015s] gix-pathspec::pathspec parse::valid::glob_negations_are_always_literal
        FAIL [   0.017s] gix-pathspec::pathspec parse::valid::whitespace_in_pathspec
        FAIL [   2.050s] gix-pathspec::pathspec search::files
        FAIL [   4.685s] gix-pathspec::pathspec search::prefixes_are_always_case_sensitive
        FAIL [ 180.967s] gix-ref-tests::refs packed::iter::performance
        FAIL [   5.502s] gix-submodule::submodule file::baseline::common_values_and_names_by_path
        FAIL [  25.197s] gix-submodule::submodule file::is_active_platform::pathspecs_matter_even_if_they_do_not_match
        FAIL [  17.391s] gix-submodule::submodule file::is_active_platform::submodules_with_active_config_are_considered_active_or_inactive
        FAIL [  11.325s] gix-submodule::submodule file::is_active_platform::submodules_with_active_config_override_pathspecs
        FAIL [  31.321s] gix-submodule::submodule file::is_active_platform::without_any_additional_settings_all_are_inactive_if_they_have_a_url
        FAIL [  37.674s] gix-submodule::submodule file::is_active_platform::without_submodule_in_index
        FAIL [  37.356s] gix::gix object::tree::diff::track_rewrites::copies_in_entire_tree_by_similarity_with_limit
        FAIL [   0.271s] gix::gix revision::spec::from_bytes::regex::find_youngest_matching_commit::regex_matches
[...]

PR #1444, opened 2024-07-05
windows test robsustness

### LINKED ISSUE #1443: 9 tests rely on commands like `ln -s` making copies instead of symlinks on Windows
#### Background

As detailed in #1442, `ln -s` in Git Bash (or other MSYS2 environments) in Windows does not ordinarily attempt to create symlinks, but instead copies files. Its default behavior of copying files is independent of whether the current user has the ability to create symlinks, and also independent of the `core.symlinks` Git configuration variable. The behavior is also not directly related to Git's behavior of creating regular files when `core.symlinks` is not set to `true`, because when Git creates a regular file instead of a symlink, that file still represents a symlink in the index and contains its target path, rather than being a *copy* of the target.

`ln -s` is used in [about 14](https://github.com/search?q=repo%3AByron%2Fgitoxide%20%22ln%20-s%22%20path%3A*.sh&type=code) of gitoxide's fixture scripts.

Although `ln -s` does not typically attempt to create symlinks, it will attempt to do so--and succeed if the user has the ability to create them without UAC elevation--if the default behavior of MSYS2's emulated `symlink` system call is customized in the `MSYS` environment variable. As noted in https://github.com/git-for-windows/git/issues/3122#issuecomment-800915597 and #1442, one way to do this is to set `MSYS=winsymlinks:nativestrict`.

One should be careful about doing this automatically, because the user may already have it set; like [the `CYGWIN` variable](https://cygwin.com/cygwin-ug-net/using-cygwinenv.html) that it is based on, it supports a number of options, not all of which relate to symlinks, which can be set together in its value by separating them with spaces (or tabs). Therefore, if in the future gitoxide test fixtures automatically enable `winsymlinks:nativestrict`, they should probably do so in a way that preserves unrelated options in `$MSYS` if present. However, such subtleties can usually be ignored in manual testing on a system where `MSYS` is known to be unset.

#### Test failures with `MSYS=winsymlinks:nativestrict`

Originally my goal was to find out if any tests documented as failing in #1358 were failing due to `ln -s` or other Unix-style commands (e.g., `tar` to unpack an archive with symlinks) not really creating symlinks, such that they would pass with `MSYS=winsymlinks:nativestrict`. It turns out that there was exactly one such test; see https://github.com/Byron/gitoxide/issues/1358#issuecomment-2209859265.

However, the bigger effect is that there are nine new *failures*! This is to say that some tests have been depending on `ln -s`, or other commands that would ordinarily create symlinks, *not* attempting to do so.

#### Failures when using pre-generated archives

Using pre-generated archives by *not* defining `GIX_TEST_IGNORE_ARCHIVES`, I ran:

```sh
MSYS=winsymlinks:nativestrict cargo nextest run --all --no-fail-fast
```

This produced 9 failures:

```text
     Summary [ 360.891s] 2363 tests run: 2354 passed (12 slow, 1 leaky), 9 failed, 9 skipped
        FAIL [   0.695s] gix-archive::archive from_tree::basic_usage_internal
        FAIL [   0.706s] gix-archive::archive from_tree::basic_usage_tar
        FAIL [   0.032s] gix-archive::archive from_tree::basic_usage_zip
        FAIL [   0.020s] gix-status-tests::worktree stack::paths_not_going_through_symlink_directories_are_ok_and_point_to_correct_item
        FAIL [   0.024s] gix-status-tests::worktree status::index_as_worktree::modified
        FAIL [   0.031s] gix-status-tests::worktree status::index_as_worktree::refresh
        FAIL [   0.029s] gix-status-tests::worktree status::index_as_worktree::unchanged
        FAIL [   0.061s] gix-worktree-state-tests::worktree state::checkout::allow_or_disallow_symlinks
        FAIL [   3.897s] gix-worktree-stream::stream from_tree::will_provide_all_information_and_respect_export_ignore
error: test run failed
```

See [**this gist**](https://gist.github.com/EliahKagan/a31d37c71b56755c11ff7e32f015c722) for the full test output of this run. Other than the information in this issue description, that gist contains the most important information relevant to this issue. The next subsection covers the effect of setting both `MSYS` and `GIX_TEST_IGNORE_ARCHIVES`, but for the purpose of showing that their effects are *mostly* independent, so it may be of less interest.

#### Failures when ignoring pre-generated archives

I independently tried it with:

```sh
MSYS=winsymlinks:nativestrict GIX_TEST_IGNORE_ARCHIVES=1 cargo nextest run --all --no-fail-fast
```

This gave 24 failures:

```text
     Summary [ 891.930s] 2363 tests run: 2339 passed (22 slow, 2 leaky), 24 failed, 9 skipped
        FAIL [   0.948s] gix-archive::archive from_tree::basic_usage_internal
        FAIL [   1.089s] gix-archive::archive from_tree::basic_usage_tar
        FAIL [   0.998s] gix-archive::archive from_tree::basic_usage_zip
        FAIL [   7.922s] gix-glob::glob pattern::matching::compare_baseline_with_ours
        FAIL [   8.021s] gix-pathspec::pathspec parse::baseline
        FAIL [   0.018s] gix-pathspec::pathspec parse::valid::glob_negations_are_always_literal
        FAIL [   0.018s] gix-pathspec::pathspec parse::valid::whitespace_in_pathspec
        FAIL [   1.988s] gix-pathspec::pathspec search::files
        FAIL [   4.424s] gix-pathspec::pathspec search::prefixes_are_always_case_sensitive
        FAIL [ 180.755s] gix-ref-tests::refs packed::iter::performance
        FAIL [   0.436s] gix-status-tests::worktree stack::paths_not_going_through_symlink_directories_are_ok_and_point_to_correct_item
        FAIL [   0.792s] gix-status-tests::worktree status::index_as_worktree::modified
        FAIL [   0.027s] gix-status-tests::worktree status::index_as_worktree::refresh
        FAIL [   0.909s] gix-status-tests::worktree status::index_as_worktree::unchanged
        FAIL [   5.599s] gix-submodule::submodule file::baseline::common_values_and_names_by_path
        FAIL [  17.483s] gix-submodule::submodule file::is_active_platform::pathspecs_matter_even_if_they_do_not_match
        FAIL [  30.376s] gix-submodule::submodule file::is_active_platform::submodules_with_active_config_are_considered_active_or_inactive
        FAIL [  38.142s] gix-submodule::submodule file::is_active_platform::submodules_with_active_config_override_pathspecs
        FAIL [  22.849s] gix-submodule::submodule file::is_active_platform::without_any_additional_settings_all_are_inactive_if_they_have_a_url
        FAIL [   5.158s] gix-submodule::submodule file::is_active_platform::without_submodule_in_index
        FAIL [   0.804s] gix-worktree-state-tests::worktree state::checkout::allow_or_disallow_symlinks
        FAIL [   1.531s] gix-worktree-stream::stream from_tree::will_provide_all_information_and_respect_export_ignore
        FAIL [   0.218s] gix::gix revision::spec::from_bytes::regex::find_youngest_matching_commit::regex_matches
        FAIL [   0.117s] gix::gix revision::spec::from_bytes::regex::with_known_revision::contained_string_matches_in_unanchored_regex_and_disambiguates_automatically
error: test run failed
```

But most of those failures are those that occur without setting the `MSYS` environment variable and were reported in #1358. The new failures are the tests that failed above without `GIX_TEST_IGNORE_ARCHIVES`.

See [this other gist](https://gist.github.com/EliahKagan/937680282966a85deee882c9f8585e0b) for the full test output from this run without using pre-generated archives.

#### Not necessarily specific to `ln -s`

These results are not necessarily specific to `ln -s`. This does not necessarily produce the same effect as setting the environment variable only for `ln -s` calls, because the MSYS behavior it is affecting is not specifically the behavior of `ln -s`, but rather the behavior of any executable that links to `msys-2.0.dll` and makes use of the POSIX `symlink` call that it emulates. Some more details on this distinction are given in #1442.

Note that fully native executables, including `git`, do not link to `msys-2.0.dll`, so whether they attempt to create symlinks is unaffected by the `MSYS` environment variable. In the MSYS2 documentation, this is correlated with the "C library" column in [this table](https://www.msys2.org/docs/environments/#overview).

### Expected behavior 🤔

When `ln -s` creates symbolic links, that should not cause any tests to fail on any system. If other POSIX commands such as `tar` are also involved in this, the same applies: when such a command ordinarily would create symbolic links but may not always do so on Windows, its ability to do so on Windows should not break the tests.

If special-casing is in place to work around the difficulty creating symlinks on Windows by treating their presence as unexpected, then this disparity with other platforms should be eliminated when possible. If not possible, the special-casing should be made more robust, so it does not assume the absence of features that may be present in reasonable configurations on non-broken systems.

If commands such as `ln -s` are used in test fixtures, or invoked by other means in testing, in any situations where they are really not intended to create symbolic links, then either a different command should be used, or the necessary configuration change should be made, such as removing symlink-related options from the value of `$MSYS` when present. My guess, however, is that calls to `ln -s` in gitoxide's test suite are not deliberately intended to copy a file rather than create a symlink.

### Git behavior

Not applicable.

### Steps to reproduce 🕹

Test commands are all shown above. I ran `gix clean -xde` between tests.

The Windows 10 system I tested on has "developer mode" enabled, which permits me to create symlinks without UAC elevation; this is important to the results in that it demonstrates that the problem is not the inability of `ln -s` to create symlinks when it tries, but rather that these nine failures seem to be due to it *actually* creating them.

Byron:
Allow windows tests to work even if symlinks are available.

Fixes #1443.

### Tasks

* [x] fix tests
* [x] actually enable `MSYS=winsymlinks:nativestrict` to make these tests pass

### DIFF
--- a/tests/tools/src/lib.rs
+++ b/tests/tools/src/lib.rs
@@ -590,6 +590,8 @@ fn configure_command<'a>(
     script_result_directory: &Path,
 ) -> &'a mut std::process::Command {
     let never_path = if cfg!(windows) { "-" } else { ":" };
+    let mut msys_for_git_bash_on_windows = std::env::var("MSYS").unwrap_or_default();
+    msys_for_git_bash_on_windows.push_str(" winsymlinks:nativestrict");
     cmd.args(args)
         .stdout(std::process::Stdio::piped())
         .stderr(std::process::Stdio::piped())
@@ -597,6 +599,7 @@ fn configure_command<'a>(
         .env_remove("GIT_DIR")
         .env_remove("GIT_ASKPASS")
         .env_remove("SSH_ASKPASS")
+        .env("MSYS", msys_for_git_bash_on_windows)
         .env("GIT_CONFIG_SYSTEM", never_path)
         .env("GIT_CONFIG_GLOBAL", never_path)
         .env("GIT_TERMINAL_PROMPT", "false")
--- a/tests/tools/src/lib.rs
+++ b/tests/tools/src/lib.rs
@@ -754,7 +754,6 @@ fn extract_archive(
         }
         #[cfg(not(feature = "xz"))]
         {
-            use std::io::Read;
             input_archive.read_to_end(&mut buf)?;
         }
         buf
--- a/gix-archive/tests/archive.rs
+++ b/gix-archive/tests/archive.rs
@@ -17,7 +17,7 @@ mod from_tree {
     #[test]
     fn basic_usage_internal() -> gix_testtools::Result {
         basic_usage(gix_archive::Format::InternalTransientNonPersistable, |buf| {
-            assert_eq!(buf.len(), if cfg!(windows) { 565 } else { 551 });
+            assert_eq!(buf.len(), 551);
 
             let mut stream = gix_worktree_stream::Stream::from_read(std::io::Cursor::new(buf));
             let mut paths_and_modes = Vec::new();
@@ -27,11 +27,7 @@ mod from_tree {
                 entry.read_to_end(&mut buf).expect("stream can always be read");
             }
 
-            let expected_link_mode = if cfg!(windows) {
-                EntryKind::Blob
-            } else {
-                EntryKind::Link
-            };
+            let expected_link_mode = EntryKind::Link;
             let expected_exe_mode = if cfg!(windows) {
                 EntryKind::Blob
             } else {
@@ -53,11 +49,7 @@ mod from_tree {
                     (
                         "symlink-to-a".into(),
                         expected_link_mode,
-                        hex_to_id(if cfg!(windows) {
-                            "45b983be36b73c0788dc9cbcb76cbb80fc7bb057"
-                        } else {
-                            "2e65efe2a145dda7ee51d1741299f848e5bf752e"
-                        })
+                        hex_to_id("2e65efe2a145dda7ee51d1741299f848e5bf752e")
                     ),
                     (
                         "dir/b".into(),
@@ -119,34 +111,20 @@ mod from_tree {
                     header.mode()?,
                 ));
             }
-            let expected_symlink_type = if cfg!(windows) {
-                EntryType::Regular
-            } else {
-                EntryType::Symlink
-            };
+            let expected_symlink_type = EntryType::Symlink;
             let expected_exe_mode = if cfg!(windows) { 420 } else { 493 };
             assert_eq!(
                 out,
                 [
                     ("prefix/.gitattributes", EntryType::Regular, 56, 420),
                     ("prefix/a", EntryType::Regular, 3, 420),
-                    (
-                        "prefix/symlink-to-a",
-                        expected_symlink_type,
-                        if cfg!(windows) { 3 } else { 0 },
-                        420
-                    ),
+                    ("prefix/symlink-to-a", expected_symlink_type, 0, 420),
                     ("prefix/dir/b", EntryType::Regular, 3, 420),
                     ("prefix/dir/subdir/exe", EntryType::Regular, 0, expected_exe_mode),
                     ("prefix/extra-file", EntryType::Regular, 21, 420),
                     ("prefix/extra-exe", EntryType::Regular, 0, expected_exe_mode),
                     ("prefix/extra-dir-empty", EntryType::Directory, 0, 420),
-                    (
-                        "prefix/extra-dir/symlink-to-extra",
-                        expected_symlink_type,
-                        if cfg!(windows) { 21 } else { 0 },
-                        420
-                    )
+                    ("prefix/extra-dir/symlink-to-extra", expected_symlink_type, 0, 420)
                 ]
                 .into_iter()
                 .map(|(path, b, c, d)| (bstr::BStr::new(path).to_owned(), b, c, d))
@@ -183,7 +161,7 @@ mod from_tree {
             },
             |buf| {
                 assert!(
-                    buf.len() < 1270,
+                    buf.len() < 1280,
                     "much bigger than uncompressed for some reason (565): {} < 1270",
                     buf.len()
                 );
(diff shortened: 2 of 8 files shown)


EliahKagan:
In addition to fixing #1443, I think this would also fix #1442. The new `many_different_states` failure covered in #1442 went away, as expected, when I ran the tests with `SYS=winsymlinks:nativestrict`. (That test is not among the "Failures when ignoring pre-generated archives" in #1443.)

Byron:
Now that I have done this, I am pretty sure the other issues that happen with archives ignored are very similar, so should be more than fixable. I spent some time today to actually be able to have a half-decent dev-cycle on Windows so I can iterate on these and do minor code edits without ripping out my hair, so I am definitely considering to finally solve #1358 .

tests/tools/src/lib.rs
```diff
@@ -590,13 +590,16 @@ fn configure_command<'a>(
     script_result_directory: &Path,
 ) -> &'a mut std::process::Command {
     let never_path = if cfg!(windows) { "-" } else { ":" };
+    let mut msys_for_git_bash_on_windows = std::env::var("MSYS").unwrap_or_default();
+    msys_for_git_bash_on_windows.push_str(" winsymlinks:nativestrict");
```
EliahKagan, line 594: I decided to check to make sure this behaves correctly even when another `winsymlinks` [option](https://cygwin.com/cygwin-ug-net/using-cygwinenv.html) is already listed in `$MSYS`.

**This code is correct!** 😸 I say that based on this experiment:

```text
ek@Glub MINGW64 ~/tmp
$ touch target

ek@Glub MINGW64 ~/tmp
$ MSYS=winsymlinks:lnk ln -s target lnk-symlink

ek@Glub MINGW64 ~/tmp
$ MSYS=winsymlinks:nativestrict ln -s target nativestrict-symlink

ek@Glub MINGW64 ~/tmp
$ MSYS='winsymlinks:lnk winsymlinks:nativestrict' ln -s target lnk-nativestrict-symlink

ek@Glub MINGW64 ~/tmp
$ MSYS='winsymlinks:nativestrict winsymlinks:lnk' ln -s target nativestrict-lnk-symlink

ek@Glub MINGW64 ~/tmp
$ cmd //c dir
 Volume in drive C is OS
 Volume Serial Number is B203-10FB

 Directory of C:\Users\ek\tmp

07/06/2024  07:22 PM    <DIR>          .
07/06/2024  07:22 PM    <DIR>          ..
07/06/2024  07:21 PM    <SYMLINK>      lnk-nativestrict-symlink [target]
07/06/2024  07:21 PM               525 lnk-symlink.lnk
07/06/2024  07:22 PM               525 nativestrict-lnk-symlink.lnk
07/06/2024  07:21 PM    <SYMLINK>      nativestrict-symlink [target]
07/06/2024  07:19 PM                 0 target
               5 File(s)          1,050 bytes
               2 Dir(s)  157,514,899,456 bytes free
```

The later option overrides the earlier one, as one would expect. This doesn't seem to be documented for `CYGWIN` (in Cygwin) or `MSYS`, but I think it's reasonable to rely on. The alternative of removing prior `winsymlinks` options seems much less good to me, because that effect could be unintuitive when inspecting the environment and could also create the false impression that other options such as `error_start` would be removed, even though they would not be.

Byron: Thanks a lot for trying it out! Just this morning I was thinking about this, wondering if it's working as expected :).

PR #1449, opened 2024-07-10
Mention cargo-diet and sort in README

cruessler:
This PR mentions 2 issues I encountered when I ran `just test check-size` for the first time.

- I did not have `cargo-diet` installed, so I thought it might be a good idea to mention it.
- I did not have `LC_ALL` set in my environment which made one of the tests fail with the following diff:

    ```
    < ./a-non-bare-repo-with-extension.git/.git
    4a4
    > ./a-non-bare-repo-with-extension.git/.git
    8d7
    < ./no-origin/.git
    9a9
    > ./no-origin/.git
    11d10
    < ./origin-and-fork/.git
    12a12
    > ./origin-and-fork/.git
    14,15c14,15
    < ./special-origin/.git
    < ./special-origin/a
    \ No newline at end of file
    ---
    > ./special-origin/a
    > ./special-origin/.git
    \ No newline at end of file
    ```

    Notice that dotfiles and non-dotfiles were sorted in a way that did not match the snapshot. One could also individually set `LC_ALL=C` for all or just the failing invocations of `sort` in shell scripts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,7 +324,7 @@ Project non-goals can change over time as we learn more, and they can be challen
 
 If what you have seen so far sparked your interest to contribute, then let us say: We are happy to have you and help you to get started.
 
-We recommend running `just test check-size` during the development process to assure CI is green before pushing. Make sure to have [cargo-diet][cargo-diet] installed before you run `just check-size`. When running tests, make sure to have `LC_ALL=C` set in order to make `sort` which is used in some of the tests behave as expected.
+We recommend running `just test` during the development process to assure CI is green before pushing.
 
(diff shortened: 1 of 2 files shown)


Byron (commented):
Thanks a lot for the fly-by fixes!

Now that I see it, I think even though `cargo diet` can be mentioned, the part that says `running` should just be `just test`, `check-size` I haven't run myself in a long time, it's very occasional  only.

With `LC_ALL`, I think it would be better to at least put it into the top-level journey test, I wasn't aware that the sort-order is dependent on the locale.

Does that make sense?

cruessler:
That makes a lot of sense! I’ve updated the PR accordingly.

Byron (approved):
Perfect, thanks a lot :)!

PR #1454, opened 2024-07-13
use portable-atomic for AtomicU64 where needed

Fabian-Gruenbichler:
AtomicU64 isn't available in libstd on all targets. portable-atomic fills the gap, and can be nicely cfg-guarded to only be used on such targets.

without this patch, gix-status (and thus a lot of other gix crates, and cargo) cannot be built on Debian armel.

technically, AtomicUsize is also not guaranteed to exist for all targets, and a similar patch could be written to support those as well if desired.

### DIFF
--- a/gix-status/Cargo.toml
+++ b/gix-status/Cargo.toml
@@ -34,6 +34,7 @@ gix-diff = { version = "^0.44.0", path = "../gix-diff", default-features = false
 thiserror = "1.0.26"
 filetime = "0.2.15"
 bstr = { version = "1.3.0", default-features = false }
+portable-atomic = "1"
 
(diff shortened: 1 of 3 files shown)


Byron (approved):
Thanks a lot for the fix and the detailed explanation - I will do my best to remember that and use portable atomics for `AtomicU64` types in future.

Byron:
> are there any limitations regarding architecture support we should be aware off? I see you are already running some CI for windows and (limited?) for 32-bit..

I wouldn't go below 16-bit, but that's probably not a target for Debian either :).

Otherwise, the goal would be to be as compatible as Cargo is, at least, so any fixes you make definitely should end up here eventually.

Thank a lot for your help with this.

PR #1455, opened 2024-07-14
Pull in portable-atomic only when needed

joshtriplett:
Use cargo conditional dependencies to only depend on portable-atomic on
platforms without 64-bit atomics.

Centralize the conditionals for `AtomicU64` import in one place, to avoid
duplicating them.


Byron (approved):
That's wonderful and the better template going forward, as I expect some more incompatibilities to be lurking in other plumbing crates as well (they might just not yet be used by `cargo`).

Fabian-Gruenbichler:
thanks, that is indeed cleaner! :)

PR #1456, opened 2024-07-17
Don't assume program files folder locations

EliahKagan:
In #1419, `gix-path` facilities got the ability to check Program Files directories as alternative locations for when Git for Windows is installed but the `git` binary isn't in the `PATH`. However, it hard-codes the paths, which are not correct for all Windows systems.

The most common scenario in which the currently hard-coded paths are not correct is a 32-bit Windows system, for which the 32-bit Program Files directory is the only Program Files directory and is (it least in all ordinary setups) just called `Program Files`. It is also technically possible, though in practice very uncommon, for the system drive not to be `C:` and, in extremely weird setups, *maybe* even for Program Files directories to have different names.

This obtains them through three specially named environment variables, which is one of the reasonably reliable and documented means that works for all combinations of process architecture (i.e. build target) and architecture of the system, as well as if they are somehow in unusual places. Because locating these directories carries potentially surprising subtleties, I've included extensive tests, including an end-to-end style test that uses a combination of other techniques to obtain them that is actually more reliable, but that would require new production dependencies, or nontrivial unsafe code in production, to do outside the tests.

I don't think CI covers the cases most likely to go wrong here, but I have tested this on a 32-bit x86 Windows 10 system, on a 64-bit x64 Windows 10 system with both 64-bit and 32-bit builds, and an ARM64 (AArch64) Windows 11 system with 64-bit ARM, 64-bit x64, and 32-bit x86 builds, both manually and by running all tests for `gix-path`, including the new tests I added.


gix-path/src/env/git/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+    let rules = [
+        (varname_64bit, suffix_64),
+        (varname_x86, suffix_32),
+        (varname_current, suffix_current),
+    ];
+
+    let mut locations = vec![];
+
+    for (name, suffix) in rules {
+        let Some(pf) = var_os_func(name) else { continue };
```
Byron, line 58: let-else blocks help to flatten trees.

gix-path/src/env/git/tests.rs
```diff
@@ -0,0 +1,403 @@
...
+
+    #[derive(Clone, Copy, Debug)]
+    enum PlatformArchitecture {
+        Is32on32,
+        Is32on64,
+        Is64on64,
+    }
+
+    impl PlatformArchitecture {
+        fn current() -> WindowsResult<Self> {
```
Byron, line 182: This was the only implementation of a trait, probably just a left-over, so I removed it.

Byron (approved):
Thanks a lot for this incredible improvement - things get complex, and it's nice that most of that complexity is with the tests which nicely illustrate the paths that are produced.

It's unfortunate that I made the mistake not creating a new commit after the module changes, so my edits aren't easily visible. However, I made comments with highlights.

gix-path/src/env/git/tests.rs
EliahKagan: Thanks for catching this! I believe this was actually due to my inexperience with Rust. If I recall correctly, for some reason I was not sure if I could write an `impl` block for an `enum` itself (even though there is no reason for there to be such a limitation) wrote this with the intention to simplify it if I could, then forgot.

EliahKagan:
> It's unfortunate that I made the mistake not creating a new commit after the module changes, so my edits aren't easily visible. However, I made comments with highlights.

Thanks! It is a big improvement splitting the module into its implementation and tests. I guess the correct organization for this code was obvious, but for whatever reason I didn't see it.

Regarding the change being in just one commit, another commit for that could be rebased in if desired, but I think it should not be necessary.

PR #1462, opened 2024-07-22
various fixes

### LINKED ISSUE #1458: `gix clean -xde` deletes whole repo if `.gitignore` lists `*` or `/`
### Current behavior 😯

`gix clean -xde` will delete the entire top-level repository it is operating on, including tracked files and the `.git` directory itself--thus the whole local history--if a `.gitignore` file contains `*` or `/`. This seems to happen because the repository itself is identified as an untracked nested repository.

#### Illustration with a real-world example

One approach to writing `.gitignore` is to list `*` followed by `!` exclusions. In such a repository, running `gix clean -xde` deletes the entire contents of the repository directory, causing inconvenience and the loss of any unpushed data. For example, on the [`cargo-update`](https://crates.io/crates/cargo-update) repository:

```text
ek@Glub:~/src$ git clone https://github.com/nabijaczleweli/cargo-update.git
Cloning into 'cargo-update'...
remote: Enumerating objects: 328251, done.
remote: Counting objects: 100% (83271/83271), done.
remote: Compressing objects: 100% (1630/1630), done.
remote: Total 328251 (delta 81523), reused 83220 (delta 81477), pack-reused 244980
Receiving objects: 100% (328251/328251), 110.53 MiB | 30.62 MiB/s, done.
Resolving deltas: 100% (320324/320324), done.
ek@Glub:~/src$ cd cargo-update
ek@Glub:~/src/cargo-update (master=)$ cat .gitignore
*
!.gitignore
!.travis.yml
!gh_rsa.enc
!appveyor.yml
!LICENSE
!Cargo.toml
!rustfmt.toml
!build.rs
!cargo-install-update-manifest.rc
!cargo-install-update.exe.manifest
!*.sublime-project
!*.md
!.github
!.github/**
!src
!src/**
!man
!man/**
!tests
!tests/**
!test-data
!test-data/**
ek@Glub:~/src/cargo-update (master=)$ gix clean -xdn
WOULD remove / ( )

WARNING: would remove repositories hidden inside ignored directories - use --skip-hidden-repositories to skip
ek@Glub:~/src/cargo-update (master=)$ gix clean -xde
removing / ( )
Error: Invalid argument (os error 22)
ek@Glub:~/src/cargo-update[1]$ ls -al
total 8
drwxr-xr-x  2 ek ek 4096 Jul 21 15:11 .
drwxr-xr-x 20 ek ek 4096 Jul 21 15:10 ..
```

That is on Ubuntu 22.04 LTS.

#### Windows is affected when outside the directory

Although Windows superficially appears unaffected because open files and directories cannot usually be deleted, the entire repository directory will still be deleted if one runs `gix -r cargo-update clean -xde` from the parent directory. (This `-r` is a `gix` option and should not be confused with `gix clean -r`.) So really all systems are affected, though to different degrees.

```text
C:\Users\ek\src\cargo-update [master ≡]> gix clean -xde
removing / (🗑️)
Error: The process cannot access the file because it is being used by another process. (os error 32)
```

```text
C:\Users\ek\src\cargo-update [master ≡]> cd ..
C:\Users\ek\src> gix -r cargo-update clean -xdn
WOULD remove / (🗑️)

WARNING: would remove repositories hidden inside ignored directories - use --skip-hidden-repositories to skip
C:\Users\ek\src> gix -r cargo-update clean -xde
removing / (🗑️)
C:\Users\ek\src> cd cargo-update
Set-Location: Cannot find path 'C:\Users\ek\src\cargo-update' because it does not exist.
```

#### The `/` means the repository

The `/` is in each case referring to the top-level directory of the repository. It fortunately does not refer to the actual root of the filesystem. Likewise:

- As far as I have been able to find, this does not ever delete *upwards*. For example, paths with `..` components in `.gitignore` do not cause files outside of the repository to be deleted.
- It also does not seem possible to have `.gitignore` patterns that cause only *some* of the contents of `.git` to be deleted. Although this may seem like cold comfort, it is actually a major mitigating factor, because partial deletion of `.git` would not necessarily be noticed and could result in situations that could be much harder to recover from, since local refs could be silently removed, recreated with different referents, and then force-pushed to a remote without awareness that they were replacing preexisting conceptually unrelated tags or branches.

However, all local branches and their history (except the tip of any branches checked out in separate worktrees and unmodified), the index, any stashes, and other objects available through the reflog, can all be deleted.

#### Some relevant code

It looks like, in traversals, paths that are not conceptually part of the repository are pruned, except that when a `.git` directory would be pruned for this reason but also matches a `.gitignore` entry, then it is instead retained and given the `Ignored` status:

https://github.com/Byron/gitoxide/blob/15f1cf76764221d14afa66d03a6528b19b9c30c9/gix-dir/src/entry.rs#L43-L49

But this is not necessarily to say that this aspect of the classification has to be changed. It may be suitable for most repositories found during traversal, just not for the top-level working tree or `.git` directory, nor for any submodules or their `.git` files.

### Expected behavior 🤔

One never expects cleaning to remove tracked files or cause data loss in the repository's local history, much less loss of the whole history.

Although the behavior described here was not intended, and exposition may not be required to *demonstrate* that this is a bug, I've nonetheless detailed what I think is the expected behavior below. This is because I think it may be useful identifying current or future cases of the bug, some of which are less obvious than others.

A `*` entry in a `.gitignore` has real-world uses (as shown in the example of the `cargo-update` repository) and should be taken as a pattern that matches all files, which subsequent `!` exclusions can then make exceptions to. A `/` might likewise be used deliberately, though I'm not sure I have seen it outside of testing. Furthermore, the ability to delete *nested* untracked repositories is a deliberate feature of `gix clean` when passed some combinations of options, and a very useful feature at that. However:

1. The top-level directory of a repository, i.e., the entire repository, should not be deleted in a clean. This applies to the case of running `gix -r ... clean -xde`. This would apply even if the directory were actually empty due to a nonstandard worktree. (This `-r` is a `gix` option and should not be confused with `-r` as a `gix clean` option as presented below in case 4.)
2. Tracked files should not be deleted or modified in a clean, regardless of whether they have any (staged or unstaged) changes. This applies to the unexpected deletions other than those of `.git`.
3. Neither the current repository's `.git` directory nor anything inside it should ever be deleted in a clean, even if intentionally cleaning ignored files and even if intentionally including ignored subdirectories that are themselves repositories. Although the current repository's `.git` directory is "ignored" in the sense that commands like `git add .` do not stage anything from it, this behavior is separate from, and supersedes, the effect of `.gitignore`. The `.git` directory is effectively a *void* in the working tree.
4. As a special case of 3, even if `.git` is specified explicitly in `.gitignore` and `-r` is included in the options passed to `gix clean`, the `.git` directory should not be deleted. Currently adding `-r` will make this happen, covered in "Steps to reproduce" below. (This `-r` is a `gix clean` option and should not be confused with `-r` as  a `gix` option as presented above in case 1.)
5. When the current repository is a submodule, is is expected to have a `.git` file instead of a `.git` directory. This likewise should not be deleted when running `gix clean` in the submodule, irrespective of the options to `gix clean` or the contents of any `.gitignore` file. I have not tested this case yet.
6. When the current repository *has* submodules--that is, directories corresponding to entries in `.gitmodules`--those submodules should not be deleted by `gix clean`, since submodule directories are likewise *voids* in the superproject's working tree, rather than being ignored due to `.gitignore`. I have not tested this case yet either.

Cases 2 and 3 are the most important, because they are the most common and they are the most likely to cause data loss, especially case 3.

Although I have not tested cases 5 and 6 (yet), I mention them because it seems like incorrect behavior for them might be easy to bring about by accident when fixing this bug for the other cases.

#### Notes on `-p` and `-r`

One possible implementation approach comes to mind for a fix that I want to recommend *against*. While `gix clean` recognizes precious files, I recommend against allowing any of the above to occur even when `-p` is passed. I think regarding `.git` as typically ineligible for deletion by automatically considering it a precious directory would still be far from strong enough protection. It also doesn't really fit conceptually: I believe that precious files are conceptually those that should usually not be deleted due to their status as being important for reasons independent of source control.

In addition to the above, the effect of `-x` and `-d` on actual nested repositories, especially if they are to continue to delete them under some conditions even in the absence of `-r`, should be documented explicitly, including in the output of `gix help clean`. But that could be done separately from the fix for this bug.

#### No one-to-one comparison...

There is no *exact* comparison to `git clean` behavior, because `gix clean` deliberately deletes ignored nested repositories when `-x` and `-d` are passed (provided that `-e` is passed to allow it do anything at all). It furthermore seems to do so intentionally even without `-r`, though as examined below in *Steps to reproduce*, perhaps this is unintentional in the absence of `-r`.

More broadly, `gix clean` is not intended to behave exactly the same as `git clean`, as detailed in #1308.

#### ...but `git` behavior is relevant

However, it's true that `git clean` does set strong expectations for what kinds of deletions are within the ambit of cleaning, and no way of using `git clean` produces this effect.

For example, running `git clean -xdf` in the `cargo-update` repository used as an example above does not delete any tracked files or the `.git` directory.

### Steps to reproduce 🕹

To reproduce this, one can carry out the example shown above with the non-toy `cargo-update` repository, which confirms the practical significance.

#### Simplified reproducer

One can alternatively run the following commands to reproduce it with a simple repository. These and all commands shown in this section were tested in Ubuntu 22.04 LTS.

```sh
git init ignore-star
cd ignore-star
echo '*' >.gitignore
echo '!.gitignore' >>.gitignore
cat .gitignore
git add .
git status
gix clean -xdn
gix clean -xde
ls -al
```

With full output:

```text
ek@Glub:~/src$ git init ignore-star
Initialized empty Git repository in /home/ek/src/ignore-star/.git/
ek@Glub:~/src$ cd ignore-star
ek@Glub:~/src/ignore-star (main #)$ echo '*' >.gitignore
ek@Glub:~/src/ignore-star (main #)$ echo '!.gitignore' >>.gitignore
ek@Glub:~/src/ignore-star (main #%)$ cat .gitignore
*
!.gitignore
ek@Glub:~/src/ignore-star (main #%)$ git add .
ek@Glub:~/src/ignore-star (main +)$ git status
On branch main

No commits yet

Changes to be committed:
  (use "git rm --cached <file>..." to unstage)
        new file:   .gitignore

ek@Glub:~/src/ignore-star (main +)$ gix clean -xdn
WOULD remove / (🗑️)

WARNING: would remove repositories hidden inside ignored directories - use --skip-hidden-repositories to skip
ek@Glub:~/src/ignore-star (main +)$ gix clean -xde
removing / (🗑️)
Error: Invalid argument (os error 22)
ek@Glub:~/src/ignore-star[1]$ ls -al
total 8
drwxr-xr-x  2 ek ek 4096 Jul 21 19:14 .
drwxr-xr-x 20 ek ek 4096 Jul 21 19:14 ..
```

#### Demonstration that this relates to nested repository handling

Some of the above commands are not necessary to confirm the bug but illustrate relevant aspects of it. For example, consider this part of the output of the dry-run `gix clean -xdn` command run before doing the real clean:

```text
WARNING: would remove repositories hidden inside ignored directories - use --skip-hidden-repositories to skip
```

This strongly suggests that the problem relates to the way code in `gix::dirwalk` identifies nested repositories.

#### Variation with `/`

The above commands can be repeated with a `/` entry instead of `*` to confirm that it also happens with that. This works both with and without the second line, `!.gitignore`, since `git` itself treats a `/` entry in `.gitignore` not to cover `.gitignore`. Actually I am not sure what a `/` in `.gitignore` is supposed to do.

#### Variation with tracked files that are not special

Although the real-world example with `cargo-update`, as well as the minimal example where `.gitignore` lists `*` and `!.gitingore`, illustrate that files excluded from being ignored by matching a `!` pattern that comes after `*` are not spared, here's a minimal example focused on that:

```sh
git init ignore-tracked
cd ignore-tracked
echo $'*\n!.gitignore\n!a' >.gitignore
cat .gitignore
touch a
git add .
git status
gix clean -xdn
gix clean -xde
ls -al
```

That produces:

```text
ek@Glub:~/src$ git init ignore-tracked
Initialized empty Git repository in /home/ek/src/ignore-tracked/.git/
ek@Glub:~/src$ cd ignore-tracked
ek@Glub:~/src/ignore-tracked (main #)$ echo $'*\n!.gitignore\n!a' >.gitignore
ek@Glub:~/src/ignore-tracked (main #%)$ cat .gitignore
*
!.gitignore
!a
ek@Glub:~/src/ignore-tracked (main #%)$ touch a
ek@Glub:~/src/ignore-tracked (main #%)$ git add .
ek@Glub:~/src/ignore-tracked (main +)$ git status
On branch main

No commits yet

Changes to be committed:
  (use "git rm --cached <file>..." to unstage)
        new file:   .gitignore
        new file:   a

ek@Glub:~/src/ignore-tracked (main +)$ gix clean -xdn
WOULD remove / (🗑️)

WARNING: would remove repositories hidden inside ignored directories - use --skip-hidden-repositories to skip
ek@Glub:~/src/ignore-tracked (main +)$ gix clean -xde
removing / (🗑️)
Error: Invalid argument (os error 22)
ek@Glub:~/src/ignore-tracked[1]$ ls -al
total 8
drwxr-xr-x  2 ek ek 4096 Jul 21 22:05 .
drwxr-xr-x 20 ek ek 4096 Jul 21 22:04 ..
```

The key seems to be that the top-level directory is taken to match the entry `*`, causing its contents all to be deleted even if some of them match `!` exclusions.

#### Variation listing `.git` and passing `-r`

Making the `.gitignore` file contain `.git` does not cause `gix clean -xde` to delete `.git`, but it does cause `gix clean -xdre` to delete `.git`. (Note that this `-r` is an option to `gix clean` and should not be confused with the `-r` option to `gix` before a subcommand, which specifies a repository to operate on.) This can be seen by running the commands:

```sh
git init ignore-dotgit
cd ignore-dotgit
echo '.git' >.gitignore
cat .gitignore
git add .
gix clean -xdn
gix clean -xde
gix clean -xdrn
gix clean -xdre
ls -al
```

Here's what that looks like:

```text
ek@Glub:~/src$ git init ignore-dotgit
Initialized empty Git repository in /home/ek/src/ignore-dotgit/.git/
ek@Glub:~/src$ cd ignore-dotgit
ek@Glub:~/src/ignore-dotgit (main #)$ echo '.git' >.gitignore
ek@Glub:~/src/ignore-dotgit (main #%)$ cat .gitignore
.git
ek@Glub:~/src/ignore-dotgit (main #%)$ git add .
ek@Glub:~/src/ignore-dotgit (main +)$ gix clean -xdn
Nothing to clean (Skipped 1 repository - show with -r)
ek@Glub:~/src/ignore-dotgit (main +)$ gix clean -xde
ek@Glub:~/src/ignore-dotgit (main +)$ gix clean -xdrn
WOULD remove repository .git/ (🗑️)
ek@Glub:~/src/ignore-dotgit (main +)$ gix clean -xdre
removing repository .git/ (🗑️)
ek@Glub:~/src/ignore-dotgit$ ls -al
total 12
drwxr-xr-x  2 ek ek 4096 Jul 21 19:51 .
drwxr-xr-x 20 ek ek 4096 Jul 21 19:50 ..
-rw-r--r--  1 ek ek    5 Jul 21 19:50 .gitignore
```

#### When should `-r` affect what happens?

I believe this should not happen even with `-r`.

In addition, combined with the above results, this raises the question of whether `gix clean -xde` without `-r` is actually intended to delete any actually nested repositories:

- If so, then, as noted above, I think this should be documented.
- If not, then that this happens--that, for example, `gix clean -xde` without `-r` is currently a good command to delete both build output and generated archives when run in `gitoxide`'s own repository--might be considered part of this bug, or its own separate related bug.

#### Other variations with `.git` and `-r`

Listing `/.git` has the same effect as `.git` at least when the current directory is the top-level directory of the repository.

Listing paths *under* `.git` in `.gitignore`, such as with the line `.git/config` or `/.git/config`, fortunately has no effect. It does not appear that a `.gitignore` entry can cause `gix clean` with any combination of options to attempt to delete only some files inside `.git`.

### LINKED ISSUE #1460: The kstring integration in gix-attributes is unsound
### Current behavior 😯

gix-attributes (in [state::ValueRef](https://github.com/Byron/gitoxide/blob/main/gix-attributes/src/state.rs#L19-L27)) unsafely creates a `&str` from a `&[u8]` containing non-UTF8 data, with the justification that so long as nothing reads the &str and relies on it being UTF-8 in the &str, there is no UB:

```rust
// SAFETY: our API makes accessing that value as `str` impossible, so illformed UTF8 is never exposed as such.
```

The problem is that the non-UTF8 `str` **is** exposed to outside code: first to the `kstring` crate itself, which requires UTF-8 in its documentation and may have UB as a consequence of this, but also to `serde`, where it propagates to e.g. `serde_json`, `serde_yaml`, etc., where the same problems occur.

As far as I know, this is not sound, and either is or can cause UB down the line in these places that can view the `&str`.

### Expected behavior 🤔

I think gix-attributes should probably use a `Vec<u8>` or `smallvec` or similar, at least until `kstring` can support bytes.

Absolute worst case, one could at least add an `unsafe` feature, so that code that opts out of `unsafe` can get the slower `Vec<u8>`-based implementation which is guaranteed to be sound.

### Git behavior

N/A

### Steps to reproduce 🕹

N/A

Byron:
### Tasks

* [x] fix #1460
* [x] baseline for `/` to assure it doesn't match all in root (or anywhere for that matter)
* [x] fix #1458
* [x] "I would not consider this to disprove that, but I have found that when the current repository is a submodule with a .git file and its own .gitignore lists .git, deleting the .git file in a submodule does not require -r."

### DIFF
--- a/gix-dir/src/walk/classify.rs
+++ b/gix-dir/src/walk/classify.rs
@@ -169,6 +169,7 @@ pub fn path(
                         .map(|platform| platform.excluded_kind())
                 })
                 .map_err(Error::ExcludesAccess)?
+                .filter(|_| filename_start_idx > 0)
             {
                 out.status = entry::Status::Ignored(excluded);
             }
@@ -256,6 +257,7 @@ pub fn path(
     if let Some(excluded) = ctx
         .excludes
         .as_mut()
+        .filter(|_| !rela_path.is_empty())
         .map_or(Ok(None), |stack| {
             stack
                 .at_entry(rela_path.as_bstr(), is_dir, ctx.objects)
--- a/gix-dir/tests/walk/mod.rs
+++ b/gix-dir/tests/walk/mod.rs
@@ -4276,3 +4276,159 @@ fn type_mismatch_ignore_case_clash_file_is_dir() {
          If there was no special handling for this, it would have found the file (`d` in the index, icase), which would have been wrong."
     );
 }
+
+#[test]
+fn top_level_slash_with_negations() -> crate::Result {
+    for repo_name in ["slash-in-root-and-negated", "star-in-root-and-negated"] {
+        let root = fixture(repo_name);
+        let ((out, _root), entries) = collect(&root, None, |keep, ctx| walk(&root, ctx, options_emit_all(), keep));
+        assert_eq!(
+            out,
+            walk::Outcome {
+                read_dir_calls: 0,
+                returned_entries: entries.len(),
+                seen_entries: 1,
+            }
+        );
+        assert_eq!(
+            entries,
+            &[entry("", Ignored(Expendable), Directory)],
+            "This is wrong - the root can never be listed"
+        );
+
+        let ((out, _root), entries) = collect(&root, None, |keep, ctx| {
+            walk(
+                &root,
+                ctx,
+                walk::Options {
+                    for_deletion: Some(ForDeletionMode::FindRepositoriesInIgnoredDirectories),
+                    emit_tracked: false,
+                    ..options_emit_all()
+                },
+                keep,
+            )
+        });
+        assert_eq!(
+            out,
+            walk::Outcome {
+                read_dir_calls: 2,
+                returned_entries: entries.len(),
+                seen_entries: 5,
+            }
+        );
+        assert_eq!(
+            entries,
+            &[
+                entry_nokind(".git", Ignored(Expendable))
+                    .with_property(DotGit)
+                    .with_match(Always),
+                entry("file", Ignored(Expendable), File)
+            ],
+            "This is still wrong, but consistent within what it should do.\
+            Top-level `.git` should always be Pruned, even if ignored.\
+            Except for `file` which should be untracked."
+        );
+    }
+    Ok(())
+}
+
+#[test]
+fn subdir_slash_with_negations() -> crate::Result {
+    for repo_name in ["slash-in-subdir-and-negated", "star-in-subdir-and-negated"] {
+        let root = fixture(repo_name);
+        let ((out, _root), entries) = collect(&root, None, |keep, ctx| walk(&root, ctx, options_emit_all(), keep));
+        assert_eq!(
+            out,
+            walk::Outcome {
+                read_dir_calls: 3,
+                returned_entries: entries.len(),
+                seen_entries: 5,
+            }
+        );
+        assert_eq!(
+            entries,
+            &[
+                entry_nokind(".git", Pruned).with_property(DotGit).with_match(Always),
+                entry("sub/.github/workflow.yml", Tracked, File),
+                entry("sub/.gitignore", Tracked, File),
+                entry("sub/file", Untracked, File),
+                entry("sub/readme.md", Tracked, File),
+            ],
+            "subdirectory matches work as expected, also with a `/` which has no bearing."
+        );
+
+        let ((out, _root), entries) = collect(&root, None, |keep, ctx| {
+            walk(
+                &root,
+                ctx,
+                walk::Options {
+                    for_deletion: Some(ForDeletionMode::FindRepositoriesInIgnoredDirectories),
+                    emit_tracked: false,
+                    ..options_emit_all()
+                },
+                keep,
+            )
+        });
+        assert_eq!(
+            out,
+            walk::Outcome {
+                read_dir_calls: 3,
+                returned_entries: entries.len(),
+                seen_entries: 5,
+            }
+        );
+        assert_eq!(
+            entries,
+            &[
+                entry_nokind(".git", Pruned).with_property(DotGit).with_match(Always),
+                entry("sub/file", Untracked, File)
+            ],
+            "This is expected, and the `.git` top-level is pruned."
+        );
+    }
+    Ok(())
+}
+
+#[test]
+fn one_ignored_submodule() -> crate::Result {
+    let root = fixture("one-ignored-submodule");
+    let ((out, _root), entries) = collect(&root, None, |keep, ctx| walk(&root, ctx, options_emit_all(), keep));
+    assert_eq!(
+        out,
+        walk::Outcome {
+            read_dir_calls: 1,
+            returned_entries: entries.len(),
+            seen_entries: 5,
+        }
+    );
+    assert_eq!(
+        entries,
+        &[
+            entry_nokind(".git", Pruned).with_property(DotGit).with_match(Always),
+            entry(".gitignore", Untracked, File),
+            entry(".gitmodules", Tracked, File),
+            entry("empty", Tracked, File),
+            entry("submodule", Tracked, Repository),
+        ],
+        "when traversing the worktree root, this is correct, the submodule doesn't count as ignored"
+    );
+
+    let troot = root.join("submodule");
+    let ((out, _root), entries) = collect(&root, Some(&troot), |keep, ctx| {
+        walk(&root, ctx, options_emit_all(), keep)
+    });
+    assert_eq!(
+        out,
+        walk::Outcome {
+            read_dir_calls: 0,
+            returned_entries: entries.len(),
+            seen_entries: 1
+        }
+    );
+    assert_eq!(
+        entries,
+        &[entryps("submodule", Tracked, Repository, Verbatim)],
+        "The submodule is simply tracked, it doesn't count as ignored"
+    );
+    Ok(())
+}
--- a/gix-dir/tests/walk/mod.rs
+++ b/gix-dir/tests/walk/mod.rs
@@ -4285,15 +4285,21 @@ fn top_level_slash_with_negations() -> crate::Result {
         assert_eq!(
             out,
             walk::Outcome {
-                read_dir_calls: 0,
+                read_dir_calls: 2,
                 returned_entries: entries.len(),
-                seen_entries: 1,
+                seen_entries: 5,
             }
         );
         assert_eq!(
             entries,
-            &[entry("", Ignored(Expendable), Directory)],
-            "This is wrong - the root can never be listed"
+            &[
+                entry_nokind(".git", Pruned).with_property(DotGit).with_match(Always),
+                entry(".github/workflow.yml", Tracked, File),
+                entry(".gitignore", Tracked, File),
+                entry("file", Untracked, File),
+                entry("readme.md", Tracked, File),
+            ],
+            "the top-level is never considered ignored"
         );
 
         let ((out, _root), entries) = collect(&root, None, |keep, ctx| {
@@ -4319,14 +4325,10 @@ fn top_level_slash_with_negations() -> crate::Result {
         assert_eq!(
             entries,
             &[
-                entry_nokind(".git", Ignored(Expendable))
-                    .with_property(DotGit)
-                    .with_match(Always),
-                entry("file", Ignored(Expendable), File)
+                entry_nokind(".git", Pruned).with_property(DotGit).with_match(Always),
+                entry("file", Untracked, File)
             ],
-            "This is still wrong, but consistent within what it should do.\
-            Top-level `.git` should always be Pruned, even if ignored.\
-            Except for `file` which should be untracked."
+            "And the negated file is correctly detected as untracked"
         );
     }
     Ok(())
--- a/gix-attributes/src/lib.rs
+++ b/gix-attributes/src/lib.rs
@@ -6,7 +6,8 @@
     doc = ::document_features::document_features!()
 )]
 #![cfg_attr(all(doc, feature = "document-features"), feature(doc_cfg, doc_auto_cfg))]
-#![deny(missing_docs, rust_2018_idioms, unsafe_code)]
+#![deny(missing_docs, rust_2018_idioms)]
+#![forbid(unsafe_code)]
 
 pub use gix_glob as glob;
 use kstring::{KString, KStringRef};
--- a/gix-attributes/src/name.rs
+++ b/gix-attributes/src/name.rs
@@ -1,7 +1,6 @@
+use crate::{Name, NameRef};
 use bstr::{BStr, BString, ByteSlice};
 use kstring::KStringRef;
-
-use crate::{Name, NameRef};
 
 impl<'a> NameRef<'a> {
     /// Turn this ref into its owned counterpart.
--- a/gix-attributes/src/parse.rs
+++ b/gix-attributes/src/parse.rs
@@ -1,9 +1,8 @@
 use std::borrow::Cow;
 
+use crate::{name, AssignmentRef, Name, NameRef, StateRef};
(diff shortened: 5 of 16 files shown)


gix-dir/src/walk/classify.rs
```diff
@@ -169,6 +169,7 @@ pub fn path(
                         .map(|platform| platform.excluded_kind())
                 })
                 .map_err(Error::ExcludesAccess)?
+                .filter(|_| filename_start_idx > 0)
```
Byron, line 172: This line and the line below constitute the fix, it's quite neat.

The idea is that previously, the classifier didn't distinguish between the worktree root (and the contained repository) and sub-repositories. For the latter, the current classification is very much intentional, but the root should never ever be counted as ignored, and thus removable.

The commit [with the fix](https://github.com/Byron/gitoxide/pull/1462/commits/1e92d1ed3f23e9de84203252d64e701e2c7fb941) also shows how the classification changes due to the fix, which is all it takes to prevent `gix clean` to go off the rail.

When looking for weaknesses, I also recommend using the pathspec, as it is quite integrated to the algorithm to have some (possibly unexpected) influence.

gix-dir/tests/walk/mod.rs
```diff
@@ -4276,3 +4276,161 @@ fn type_mismatch_ignore_case_clash_file_is_dir() {
...
+            walk::Outcome {
+                read_dir_calls: 2,
+                returned_entries: entries.len(),
+                seen_entries: 5,
+            }
+        );
+        assert_eq!(
+            entries,
+            &[
+                entry_nokind(".git", Pruned).with_property(DotGit).with_match(Always),
```
Byron, line 4296: Before the fix, this entry was ignored which was part of the problem. With the fix, it turned into a pruned `DotGit` entry, which naturally won't be touched by `gix clean`.

Byron (commented):
It took a while to write the tests, but once they were available homing on a fix was quite straightforward and felt nice, despite the complexity of the algorithm.

EliahKagan (commented):
I've checked manually on Windows 10 and Ubuntu 22.04 LTS, on both systems installing with `cargo install --path .`, that this fixes the cases covered in #1458. If I find other cases I'll open a new issue. Thanks!

EliahKagan:
Regarding [the failure of the `pure-rust-build` job](https://github.com/Byron/gitoxide/actions/runs/10063920534/job/27819753640) on CI from the merge commit, [it passed in my fork](https://github.com/EliahKagan/gitoxide/actions/runs/10064016018/job/27820056478), so maybe the failure would go away if the job is rerun.

EliahKagan:
> maybe the failure would go away if the job is rerun.

I'm pleased to see it [went away](https://github.com/Byron/gitoxide/actions/runs/10063920534/job/27820567423) (and the subsequent commits, related to releasing, are also green).

> If I find other cases I'll open a new issue.

I've opened #1464.

PR #1470, opened 2024-07-26
improvements to `gix clean`

Byron:
Related to #1469

### Tasks

* [x] reproduce issue with test
* [x] see if a fix is easy

Issue #1469, opened 2024-07-26
`gix clean -xde` deletes the repo's own hidden nested worktrees, but they are not really hidden


EliahKagan:
### Current behavior 😯

#1465 fixed #1464 by checking if a git repository encountered during traversal is a worktree of the current repository and declining to delete it even if `-p` is passed, and also even if `-x` and `-d` are passed to delete ignored directories. This protects the repository's own worktrees, even if nested, provided that they are encountered during traversal rather than being present below a directory that is found during traversal and determined to be ignored.

As described in https://github.com/Byron/gitoxide/issues/1464#issuecomment-2249475179, `gix clean` continues to require `--skip-hidden-repositories` to avoid deleting nested repositories that require full traversal to discover, and this is important for performance. This behavior holds even when those nested repositories are worktrees of the current repository.

But worktrees of the current repository, regardless of their location, should never require much work to discover. This is because the repository knows where its worktrees are. For example, a non-bare repository with `git worktree` managed worktrees, if it is not a submodule, has a `.git` directory with a `worktrees` subdirectory that contains information about its worktrees.

Whether a directory *contains* any descendants that are worktrees of the current repository should likewise be efficient to determine and likewise does not require full traversal.

#### Broad implementation idea

I suspect that either there is a simpler and more elegant approach, or maybe I am just expressing this idea poorly and it can be implemented simply? I am not sure. I present this more to clarify why I believe this can be achieved efficiently than to argue for a particular algorithm.

At the beginning of traversal, before actually beginning the walk:

1. Examine the repository `.git/worktrees` directory or equivalent and make a list of its worktrees.
2. Remove any from the list that physically do not exist on disk. This requires checking if they exist.
3. Normalize their paths relative to the root of the incipient traversal, i.e., so they are the same as they will be seen to be in traversal.
4. Remove any from the list that are not inside the repository. Preferably, remove any that are not nested inside the working tree that `gix clean` is operating on. If feasible, remove any that are outside the part of it that `gix clean` is operating on, if that is not the whole thing.
5. Optionally, as a performance optimization (if such an optimization is justified), fall back to the old implementation if the list is empty. At this point, the list will usually be empty, because usually there are no nested worktrees.
6. Store those paths, or maybe explicitly store each intermediate subdirectory path, in such a way that allows recursive deletion of a complete subdirectory that contains them to be avoided.

For elaboration on that last pre-step:

[...]

PR #1471, opened 2024-07-26
Improve help text about hidden repos

EliahKagan:
*Note: This PR ~~is either mostly or completely~~ was in substantial part superseded by #1470 as discussed in #1469, and has been narrowed in scope. The following description does not reflect that.*

#### Changes to the warning message

In the first commit, this unconditionally augments the warning suggesting to use `--skip-hidden-repositories` so that it mentions worktrees as well. It shows that slightly longer message even in a repository that has no worktrees. This has the benefit of simplicity, is the requested change in https://github.com/Byron/gitoxide/issues/1469#issuecomment-2252163765 if I understand properly, and makes sense as an initial mitigation for #1469.

The message is not much longer, and if that level of detail is acceptable, it might even be okay to have it like this longer term, though I suspect it might not be necessary given other forthcoming changes.

I considered parenthesizing "and worktrees" but it looked to like "repositories (and worktrees)" could be misread as "repositories (and their worktrees)" to mean the same thing as "repositories."

I did not make a corresponding change in the other similarly worded warning message suggesting to use `--find-untracked-repositories`, because there does not appear to be an analogous situation with untracked directories where a worktree of the current repository would be removed. I did some testing to check this.

#### Changes to the help text

It seemed like the right time to also propose changes to the help text for `gix clean` options directly relevant to the deletion of nested repositories, since I was already planning to do that for `-r`. In the second commit, this does that, at least an an initial wording that might be possible to improve on, and also applies an analogous change--but with more details, since this is help text--to the change in the warning message.

I made them separate commits so it's easy to take only one (I imagine it would be the first) if not all these changes are wanted at this time.

#### Commit messages

Even though this only adjusts messages, from [this guideline](https://github.com/Byron/gitoxide/blob/main/DEVELOPMENT.md#commit-messages) it seems like the commits should have `fix:` prefixes, so I did that. I'm not sure if that's right. This PR does not carry any behavioral change other than the text of messages that are displayed.

### DIFF
--- a/gitoxide-core/src/repository/clean.rs
+++ b/gitoxide-core/src/repository/clean.rs
@@ -325,7 +325,7 @@ pub(crate) mod function {
                     wrote_nl = true;
                     writeln!(
                         err,
-                        "WARNING: would remove repositories and worktrees hidden inside ignored directories - use --skip-hidden-repositories to skip{}",
+                        "WARNING: would remove repositories hidden inside ignored directories - use --skip-hidden-repositories to skip{}",
                         wrap_in_parens(msg.take().unwrap_or_default())
                     )?;
                 }
--- a/gitoxide-core/src/repository/clean.rs
+++ b/gitoxide-core/src/repository/clean.rs
@@ -325,7 +325,7 @@ pub(crate) mod function {
                     wrote_nl = true;
                     writeln!(
                         err,
-                        "WARNING: would remove repositories hidden inside ignored directories - use --skip-hidden-repositories to skip{}",
+                        "WARNING: would remove repositories and worktrees hidden inside ignored directories - use --skip-hidden-repositories to skip{}",
                         wrap_in_parens(msg.take().unwrap_or_default())
                     )?;
                 }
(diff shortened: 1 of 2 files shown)


Byron (approved):
Thanks a lot! This can be merged once the `(and worktrees)` portions have been removed.

EliahKagan:
Sounds good, I'll have a commit that removes the mention of worktrees in a couple of minutes.

EliahKagan:
I've done that in c27b167. Because of the way it was done, all three of the commits here are prefixed `fix:`. Given the effect on the automatically generated release notes, maybe this is not justified. Perhaps a rebase should be done to edit the messages or to squash the commits into one. Or maybe it's fine as is.

Byron:
I think it's fair to use `fix:` three times if adding three messages is intended or at least makes sense to you. It's definitely your choice on how to design this.

EliahKagan:
> I think it's fair to use `fix:` three times if adding three messages is intended or at least makes sense to you. It's definitely your choice on how to design this.

Each one was something I intended at the time given what had come before, but none of these messages are really ideal now -- they basically all have to be read together.

On the other hand, they do make clear what the progression was in relation to the changes from #1470.

Maybe the best thing to do would be if I were to rebase this to edit the message in the first commit, 7a4d239, to remove `fix:`. I think the other two make sense together and they avoid claiming a change across versions that is actually absent. Would you be okay with me doing that?

Byron:
Thanks again for all the great work!

PR #1474, opened 2024-07-28
ag/jiff

BurntSushi:
- **gix-date: make `time` a private dependency of `gix-date`**
- **gix-date: bump MSRV to Rust 1.70**
- **gix-{archive,date}: switch from `time` to `jiff`**

### DIFF
--- a/gix-date/src/parse.rs
+++ b/gix-date/src/parse.rs
@@ -14,7 +14,7 @@ pub enum Error {
 pub(crate) mod function {
     use std::{str::FromStr, time::SystemTime};
 
-    use time::{format_description::well_known, Date, OffsetDateTime};
+    use jiff::{civil::Date, fmt::rfc2822, tz::TimeZone, Zoned};
 
     use crate::{
         parse::{relative, Error},
@@ -32,27 +32,27 @@ pub(crate) mod function {
             return Ok(Time::new(42, 1800));
         }
 
-        Ok(if let Ok(val) = Date::parse(input, SHORT.0) {
-            let val = val.with_hms(0, 0, 0).expect("date is in range").assume_utc();
-            Time::new(val.unix_timestamp(), val.offset().whole_seconds())
-        } else if let Ok(val) = OffsetDateTime::parse(input, &well_known::Rfc2822) {
-            Time::new(val.unix_timestamp(), val.offset().whole_seconds())
-        } else if let Ok(val) = OffsetDateTime::parse(input, ISO8601.0) {
-            Time::new(val.unix_timestamp(), val.offset().whole_seconds())
-        } else if let Ok(val) = OffsetDateTime::parse(input, ISO8601_STRICT.0) {
-            Time::new(val.unix_timestamp(), val.offset().whole_seconds())
-        } else if let Ok(val) = OffsetDateTime::parse(input, GITOXIDE.0) {
-            Time::new(val.unix_timestamp(), val.offset().whole_seconds())
-        } else if let Ok(val) = OffsetDateTime::parse(input, DEFAULT.0) {
-            Time::new(val.unix_timestamp(), val.offset().whole_seconds())
+        Ok(if let Ok(val) = Date::strptime(SHORT.0, input) {
+            let val = val.to_zoned(TimeZone::UTC).expect("date is in range");
+            Time::new(val.timestamp().as_second(), val.offset().seconds())
+        } else if let Ok(val) = rfc2822_relaxed(input) {
+            Time::new(val.timestamp().as_second(), val.offset().seconds())
+        } else if let Ok(val) = strptime_relaxed(ISO8601.0, input) {
+            Time::new(val.timestamp().as_second(), val.offset().seconds())
+        } else if let Ok(val) = strptime_relaxed(ISO8601_STRICT.0, input) {
+            Time::new(val.timestamp().as_second(), val.offset().seconds())
+        } else if let Ok(val) = strptime_relaxed(GITOXIDE.0, input) {
+            Time::new(val.timestamp().as_second(), val.offset().seconds())
+        } else if let Ok(val) = strptime_relaxed(DEFAULT.0, input) {
+            Time::new(val.timestamp().as_second(), val.offset().seconds())
         } else if let Ok(val) = SecondsSinceUnixEpoch::from_str(input) {
             // Format::Unix
             Time::new(val, 0)
         } else if let Some(val) = parse_raw(input) {
             // Format::Raw
             val
-        } else if let Some(time) = relative::parse(input, now).transpose()? {
-            Time::new(time.unix_timestamp(), time.offset().whole_seconds())
+        } else if let Some(val) = relative::parse(input, now).transpose()? {
+            Time::new(val.timestamp().as_second(), val.offset().seconds())
         } else {
             return Err(Error::InvalidDateString { input: input.into() });
         })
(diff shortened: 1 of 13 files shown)


Byron:
Thanks so much 🙏!

> Hopefully this PR isn't too forward, but I noticed [you mentioned switching to `jiff`](https://github.com/Byron/gitoxide/issues/471), so I figured I'd take a crack at it. The transition went pretty smoothly, although I did need to [add support for skipping weekday checking in RFC 2822 style datetimes](https://github.com/BurntSushi/jiff/pull/53). So this ended up being a good exercise for Jiff, even if you don't take the PR.

When I saw the email this morning I just thought: Wow, as if he knew I wanted that :D! Thus I am very glad the person best suitable for the job performed the transition, and that it tought `jiff` a new thing as well *(a feature I wasn't even aware of relying on)*.

> This doesn't completely remove `time` from the dependency tree though, since `prodash` depends on it. However, I do have a patch for `prodash` that makes the swap, but wanted to wait for the OK before sending it.

Your patch will be warmly welcomed, I have had plans to migrate off `time` for the local-time handling for a long time now. Even though my initial thoughts revolved around fixing it in `time`, I am also glad not to have to do it and to get `jiff` from a very trusted and highly appreciated author.

> Plus, this PR gets rid of some `unsafe` usage. :-)

❤️

> Main caveat: merging this will require bumping the MSRV to at least Rust 1.70, which is Jiff's MSRV. (I do intend for Jiff's MSRV to be about as conservative as `regex`'s MSRV, which is usually around N-9, or one year of Rust releases.)

Actually, it's not a caveat anymore as I effectively follow `helix`, which follows `firefox` AFAIK. They are at 1.74 by now, so going to 1.70 will be fine.

Let me make the remaining changes to this PR so I can merge it.

Byron:
One more note on `jiff`: What I like about `time` is its flexible parsing, and as `git` essentially comes with its own equally flexible (and probably buggy) date parser, that was always something I hoped to get from the time-library I use. Thus far, I didn't try to achieve feature-parity with this, but when that's happening I'd ping you to see if or what of that should be contributed to `jiff` itself.

gix-date/src/time/format.rs
```diff
@@ -1,37 +1,22 @@
...
 /// E.g. `Thu, 8 Aug 2022 12:45:06 +0800`. This is output by `git log --pretty=%aD`.
-pub const GIT_RFC2822: CustomFormat = CustomFormat(format_description!(
-    "[weekday repr:short], \
-     [day padding:none] \
-     [month repr:short] \
-     [year] \
-     [hour]:[minute]:[second] \
-     [offset_hour sign:mandatory][offset_minute]"
-));
+pub const GIT_RFC2822: CustomFormat = CustomFormat("%a, %-d %b %Y %H:%M:%S %z");
```
Byron, line 13: This crate finally comes full circle :): initially `time` would have a format string like this, then it moved to `format_description!` with more verbose declarations and optimized efficiency. Now we go back to simple format strings which are probably just the right choice here.

gix-date/src/time/init.rs
```diff
@@ -27,28 +23,14 @@ impl Time {
...
-        let now = time::OffsetDateTime::now_utc();
-        let seconds = now.sub(std::time::SystemTime::UNIX_EPOCH).whole_seconds();
-        // TODO: make this work without cfg(unsound_local_offset), see
-        //       https://github.com/time-rs/time/issues/293#issuecomment-909158529
-        let offset = time::UtcOffset::local_offset_at(now)
-            .map(time::UtcOffset::whole_seconds)
-            .unwrap_or(0);
+        let zdt = jiff::Zoned::now();
+        let seconds = zdt.timestamp().as_second();
+        let offset = zdt.offset().seconds();
```
Byron, line 33: Oh wow, this cannot fail!
But… how can that be as so much can go wrong even when using the system TZ database?
The answer must be that internally it ignores failures and falls back to UTC automatically (by not having a TZ database if it could not be read). If that's the case, I wonder if there is a way to explicitly initialize the TZ database - `gix` (CLI) might be interested in that in case it performs operations for which displaying or using (i.e. when creating commits) the correct local time is paramount.

gix-date/tests/time/parse.rs
```diff
@@ -173,12 +173,19 @@ mod relative {
         let two_weeks_ago = gix_date::parse("2 weeks ago", Some(now)).unwrap();
         assert_eq!(Sign::Plus, two_weeks_ago.sign);
         assert_eq!(0, two_weeks_ago.offset);
-        let expected = OffsetDateTime::from(now).saturating_sub(Duration::weeks(2));
-        // account for the loss of precision when creating `Time` with seconds
-        let expected = expected.replace_nanosecond(0).unwrap();
+        let expected = Zoned::try_from(now)
+            .unwrap()
+            // account for the loss of precision when creating `Time` with seconds
+            .round(
```
Byron, line 179: I like this!

src/plumbing/main.rs
```diff
@@ -54,11 +54,6 @@ pub mod async_util {
 
 pub fn main() -> Result<()> {
     let args: Args = Args::parse_from(gix::env::args_os());
-    #[allow(unsafe_code)]
-    unsafe {
-        // SAFETY: we don't manipulate the environment from any thread
-        time::util::local_offset::set_soundness(time::util::local_offset::Soundness::Unsound);
```
Byron, line 60: I love this!

gix-date/src/parse.rs
```diff
@@ -83,52 +83,79 @@ pub(crate) mod function {
...
-    pub(crate) fn parse(input: &str, now: Option<SystemTime>) -> Option<Result<OffsetDateTime, Error>> {
-        parse_inner(input).map(|offset| {
-            let offset = std::time::Duration::from_secs(offset.whole_seconds().try_into()?);
+    pub(crate) fn parse(input: &str, now: Option<SystemTime>) -> Option<Result<Zoned, Error>> {
+        parse_inner(input).map(|result| {
+            let span = result?;
+            // This was an error case in a previous version of this code, where
+            // it would fail when converting from a negative signed integer
+            // to an unsigned integer. This preserves that failure case even
+            // though the code below handles it okay.
```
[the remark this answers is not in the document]
BurntSushi: Yeah I noticed that because I was intensely curious as to why there was a `catch_unwind` here deep inside datetime parsing. And yeah indeed, I believe [`time`'s panic comes from its `From<SystemTime>` impl](https://github.com/time-rs/time/blob/aca5c3652aff6fc7df3fbd8591c3e324d3a6ffb2/time/src/offset_date_time.rs#L1565-L1566), which in turn calls [`OffsetDateTime`'s `Add` impl that specifically panics](https://github.com/time-rs/time/blob/aca5c3652aff6fc7df3fbd8591c3e324d3a6ffb2/time/src/offset_date_time.rs#L1362-L1368). The `Add` impl panicking makes sense, since that's consistent with what `std` does too. But the `From` impl should probably be a `TryFrom` impl. Since it's a `From` impl, `time`'s behavior here is pretty much locked into its API. With that said, I think you probably could have avoided the `catch_unwind` here by first converting `now` to a `time::OffsetDateTime` and then doing your arithmetic, since `time` exposes checked operations on `time::OffsetDateTime`. With Jiff, you can't mess this up because it doesn't provide a `From<SystemTime>` impl. Only `TryFrom<SystemTime>`.

gix-date/src/parse.rs
```diff
@@ -83,52 +83,79 @@ pub(crate) mod function {
...
-                        .into()
-                })
-                .map_err(|_| Error::RelativeTimeConversion)
+                let ts = Timestamp::try_from(now).map_err(|_| Error::RelativeTimeConversion)?;
+                // N.B. This matches the behavior of this code when it was
+                // written with `time`, but we might consider using the system
+                // time zone here. If we did, then it would implement "1 day
+                // ago" correctly, even when it crosses DST transitions. Since
+                // we're in the UTC time zone here, which has no DST, 1 day is
+                // in practice always 24 hours. ---AG
```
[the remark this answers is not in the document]
BurntSushi: Yeah I definitely don't have the context to be fully confident in such a change here. It depends on your test coverage I think. But the idea here is that you'd do `ts.to_zoned(TimeZone::system())`, and then `zdt.checked_sub(span)` would become DST aware according to the end user's configured time zone.

gix-date/src/parse.rs
```diff
@@ -83,52 +83,79 @@ pub(crate) mod function {
...
-            "hour" => 60 * 60,
-            "day" => 24 * 60 * 60,
-            "week" => 7 * 24 * 60 * 60,
+        let result = match period {
+            "second" => Span::new().try_seconds(units),
+            "minute" => Span::new().try_minutes(units),
+            "hour" => Span::new().try_hours(units),
+            "day" => Span::new().try_days(units),
+            "week" => Span::new().try_weeks(units),
             // TODO months & years? YES
```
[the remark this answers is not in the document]
BurntSushi: Yup. It would just be `Span::new().try_months` and then `Span::new().try_years`. Then everything else should just work.

gix-date/src/time/format.rs
BurntSushi: Hah! Although Jiff's `strtime` APIs are quite fast. At least in Jiff's benchmarks, they're faster than `time`'s `format_description!` parsers.

But yeah I don't really like the `strtime` APIs myself, but they are handy and it seems like almost everyone is familiar with them.

gix-date/src/time/init.rs
BurntSushi: You're correct that Jiff will automatically fall back to UTC. This is actually what `time` does as well, courtesy of `libc`. Try, for example, `sudo rm /etc/localtime` and then `date` on your system. Things don't blow up. They just fall back to UTC. Hah.

In terms of fallible APIs, I think it makes sense for `TimeZone::try_system` to exist: https://github.com/BurntSushi/jiff/issues/65

For the tzdb itself, I'd have to think about that.

Note though that even today, Jiff will emit copious log messages when it fails to find a tzdb or the system configured time zone. So if "end user should enable debug logs for gix" is a workflow that exists, then you _should_ already get pretty detailed information when Jiff fails.

BurntSushi (commented):
Hiya! Sorry about the late reply. I finally had a chance to pop my stack enough that I could swing back around here. Awesome to see the PR already merged! w00t!

> One more note on jiff: What I like about time is its flexible parsing, and as git essentially comes with its own equally flexible (and probably buggy) date parser, that was always something I hoped to get from the time-library I use. Thus far, I didn't try to achieve feature-parity with this, but when that's happening I'd ping you to see if or what of that should be contributed to jiff itself.

@sharkdp also mentioned wanting something like this. Basically a, "just let me throw a datetime at your library and have it try a whole bunch of different formats to parse it." Although maybe that's different than what you're asking for, since you want to specifically target what `git` supports. IDK if it should be a goal for Jiff to "support what `git` does" specifically though. I think the main problem with this sort of thing is that it could be very difficult to offer it without it having footguns. On the other hand, it's probably better for Jiff to do it if the alternative is ~everyone else just rolling their own ad hoc version of it.

Byron:
Thanks for your responses, much appreciated!
As I can't comment on comments on a closed PR apparently, I will just do it here by quoting you.

> Yeah I definitely don't have the context to be fully confident in such a change here. It depends on your test coverage I think. But the idea here is that you'd do ts.to_zoned(TimeZone::system()), and then zdt.checked_sub(span) would become DST aware according to the end user's configured time zone.

My stance here is that it's good to be more correct than Git if that's an option, especially if that would be less surprising. Test-coverage around this probably isn't great, but if you know what correctness would mean, you'd probably put down a test to validate it. Having that would also make it possible to see if Git is also correct, just for fun - I wouldn't be completely surprised if Git would have managed to roll its own time-handling *and* handle all the edge-cases.

> Hah! Although Jiff's strtime APIs are quite fast. At least in Jiff's benchmarks, they're faster than time's format_description! parsers.

I should have known just by looking at the author name :D!

> Note though that even today, Jiff will emit copious log messages when it fails to find a tzdb or the system configured time zone. So if "end user should enable debug logs for gix" is a workflow that exists, then you should already get pretty detailed information when Jiff fails.

Ah, that's great! If `tracing` automatically picks up `log` messages, then these would already be visible today in `gix` (CLI). But just in case, I added a note which hopefully will also be enough to remember.

And finally, here is a quote of mine:

> And as gix-date is Google-OSS-fuzzed, I will be curious if it will find new ways of crashing parsing.

The truth is that two bugs weren't reproducible anymore and closed automatically since the introduction of `jiff`, without creating any new ones. Truly impressive, you beat the machine!

BurntSushi:
> > And as gix-date is Google-OSS-fuzzed, I will be curious if it will find new ways of crashing parsing.
> 
> The truth is that two bugs weren't reproducible anymore and closed automatically since the introduction of `jiff`, without creating any new ones. Truly impressive, you beat the machine!

Wow.

PR #1475, opened 2024-07-29
Fix 32-bit Windows releases so they are the correct architecture

### LINKED ISSUE #1472: The 32-bit Windows releases are 64-bit
### Current behavior 😯

On the GitHub release pages where binaries can be downloaded, the four archives named to indicate that they are `i686-pc-windows-msvc` builds are actually all `x86_64-pc-windows-msvc` builds.

I don't believe this is specific to recent versions. But in [the current latest version](https://github.com/Byron/gitoxide/releases/tag/v0.37.0) the affected archives are:

- [gitoxide-lean-v0.37.0-i686-pc-windows-msvc.zip](https://github.com/Byron/gitoxide/releases/download/v0.37.0/gitoxide-lean-v0.37.0-i686-pc-windows-msvc.zip)
- [gitoxide-max-pure-v0.37.0-i686-pc-windows-msvc.zip](https://github.com/Byron/gitoxide/releases/download/v0.37.0/gitoxide-max-pure-v0.37.0-i686-pc-windows-msvc.zip)
- [gitoxide-max-v0.37.0-i686-pc-windows-msvc.zip](https://github.com/Byron/gitoxide/releases/download/v0.37.0/gitoxide-max-v0.37.0-i686-pc-windows-msvc.zip)
- [gitoxide-small-v0.37.0-i686-pc-windows-msvc.zip](https://github.com/Byron/gitoxide/releases/download/v0.37.0/gitoxide-small-v0.37.0-i686-pc-windows-msvc.zip)

Downloading these releases from a releases page or using `gh release download`, or attempting to automatically download and install them using `cargo binstall`, obtains 64-bit executables instead.

This problem appears specific to the 32-bit Windows releases. Other releases all appear to be of the claimed architecture.

#### The cause?

The release workflow contains `win32-msvc` jobs that are [meant to](https://github.com/Byron/gitoxide/blob/55cffe4ecf78bfea76059b742fafaebe964a725f/.github/workflows/release.yml#L108-L111) build and produce archives of 32-bit Windows executables. But this is producing archives of 64-bit executables instead.

I'm not sure exactly what is going on. It seems to me from [the build output](https://github.com/Byron/gitoxide/actions/runs/10064652615/job/27822039710) that those jobs' behavior may be accidentally equivalent in effect to that of the `win-msvc` jobs--that is, they may be building the wrong target. But I may just be reading the build output wrong.

However, separately from the compilation itself, it looks like there is a bug in the "Build archive" step, where it accounts for the target that was built on Unix-like systems but does not account for it on Windows:

https://github.com/Byron/gitoxide/blob/55cffe4ecf78bfea76059b742fafaebe964a725f/.github/workflows/release.yml#L203-L211

That shows that, on Windows, the files being copied for inclusion in the archive are specified as:

```sh
target/release/${{ env.EXE_NAME }}.exe target/release/gix.exe
```

While, in contrast, on other operating systems, they are specified as:

```sh
target/${{ matrix.target }}/release/${{ env.EXE_NAME }} target/${{ matrix.target }}/release/gix
```

If that is not the full cause, or the needed fix is not immediately apparent when you look at it, then I would be pleased to investigate further and try to fix it.

I am not sure of the best way is to experiment with this, but it looks feasible. In my fork, I could remove the "Upload release archive" step check the output in some other way. Or if necessary I could leave it in and publish "releases" from my fork while testing.

### Expected behavior 🤔

The releases for Windows marked `i686` should target that architecture rather than `x86_64`.

(If some builds with particular feature configuration cannot be produced, then they can simply be omitted. But I suspect they all can be built. In particular, https://github.com/rust-lang/libz-sys/issues/197 does not seem to affect local cross-compilation from a 64-bit Windows machine.)

### Git behavior

32-bit builds of Git for Windows offered for download are 32-bit, targeting i686.

### Steps to reproduce 🕹

This can be verified manually, by attempting to install a 32-bit build with `cargo binstall` and examining the result, or semi-automatically by downloading all archives of a release and checking which of them match the archive names.

#### Demonstration with `cargo binstall`

This is from a 32-bit Windows 10 system:

```text
PS C:\Users\ek> cargo binstall gitoxide
 INFO resolve: Resolving package: 'gitoxide'
 WARN The package gitoxide v0.37.0 (i686-pc-windows-msvc) has been downloaded from github.com
 INFO This will install the following binaries:
 INFO   - ein.exe (ein.exe -> C:\Users\ek\.cargo\bin\ein.exe)
 INFO   - gix.exe (gix.exe -> C:\Users\ek\.cargo\bin\gix.exe)
Do you wish to continue? yes/[no]
? y
 INFO Installing binaries...
 INFO Done in 26.3840891s
PS C:\Users\ek> gix
ResourceUnavailable: Program 'gix.exe' failed to run: An error occurred trying to start process 'C:\Users\ek\.cargo\bin\gix.exe' with working directory 'C:\Users\ek'. The specified executable is not a valid application for this OS platform.At line:1 char:1
+ gix
+ ~~~.
PS C:\Users\ek> ein
ResourceUnavailable: Program 'ein.exe' failed to run: An error occurred trying to start process 'C:\Users\ek\.cargo\bin\ein.exe' with working directory 'C:\Users\ek'. The specified executable is not a valid application for this OS platform.At line:1 char:1
+ ein
+ ~~~.
PS C:\Users\ek> &'C:\Program Files\Git\usr\bin\file' 'C:\Users\ek\.cargo\bin\gix.exe'
C:\Users\ek\.cargo\bin\gix.exe: PE32+ executable (console) x86-64, for MS Windows
PS C:\Users\ek> &'C:\Program Files\Git\usr\bin\file' 'C:\Users\ek\.cargo\bin\ein.exe'
C:\Users\ek\.cargo\bin\ein.exe: PE32+ executable (console) x86-64, for MS Windows
```

#### Demonstration checking binaries in all archives of a release

Downloading all archives of a release can be done with the [`gh`](https://cli.github.com/) command. I did the following in PowerShell on a Windows system with 7-Zip installed (providing the `7z` command) and with the MSYS2 `file` command made available in the `PATH`:

```powershell
mkdir gitoxide-v0.37.0-release
cd gitoxide-v0.37.0-release
gh release -R byron/gitoxide download v0.37.0
gci | %{ 7z x $_ }
gci *.tar | %{ 7z x $_ }
file */gix* */ein*
file */gix* */ein* | sls -raw i686
```

This confirmed that all `i686-pc-windows-msvc` archives were affected, with their executables being `x86_64` rather than `i686`, and that no other archives were affected. [The full transcript is in this gist.](https://gist.github.com/EliahKagan/5d2b5a84451c11ccec1b10ba6f2454c6) The most useful part is the output of the last command, which is:

```text
gitoxide-lean-v0.37.0-i686-pc-windows-msvc/gix.exe:        PE32+ executable (console) x86-64, for MS Windows, 5 sections
gitoxide-max-pure-v0.37.0-i686-pc-windows-msvc/gix.exe:    PE32+ executable (console) x86-64, for MS Windows, 5 sections
gitoxide-max-v0.37.0-i686-pc-windows-msvc/gix.exe:         PE32+ executable (console) x86-64, for MS Windows, 5 sections
gitoxide-small-v0.37.0-i686-pc-windows-msvc/gix.exe:       PE32+ executable (console) x86-64, for MS Windows, 5 sections
gitoxide-lean-v0.37.0-i686-pc-windows-msvc/ein.exe:        PE32+ executable (console) x86-64, for MS Windows, 5 sections
gitoxide-max-pure-v0.37.0-i686-pc-windows-msvc/ein.exe:    PE32+ executable (console) x86-64, for MS Windows, 5 sections
gitoxide-max-v0.37.0-i686-pc-windows-msvc/ein.exe:         PE32+ executable (console) x86-64, for MS Windows, 5 sections
gitoxide-small-v0.37.0-i686-pc-windows-msvc/ein.exe:       PE32+ executable (console) x86-64, for MS Windows, 5 sections
```

EliahKagan:
This fixes #1472 so that the 32-bit Windows releases contain 32-bit binaries rather than 64-bit binaries, makes it so pushing a version tag automatically triggers the release workflow, and somewhat streamlines and clarifies the workflow.

Further improvement is possible, including by incorporating some of the other recent changes from the ripgrep release workflow on which it was based. This is something I would be interested to do separately but that I think is not naturally within the scope of these changes.

At least in principle, the workflow can also at this point be extended easily for more build targets, so long as they are otherwise able to build and so long as either *(a)* they can be built with cross-compilation using `cross` on a Linux-based host or *(b)* they do not require `cross`. However, if adding more Linux-based targets, it would be a good idea to generalize the `strip` logic, or the resulting binaries would have debug symbols even though they would be optimized builds.

There are some more details in the commit messages. This includes some approaches I tried but did not keep, when they were entangled with the other changes. This information about the journey may be useful because the actual workflow runs I did for testing are not currently public. As discussed in #1472, I used a private reupload to avoid notifying people in their feeds of something the might misunderstand as an actual prerelease of gitoxide. However, I'd be pleased to republish from my public fork with distinctive tag names (e.g., with `do-not-use` in them), provide you access to the private repository, make the private repository at least temporarily public, or any combination of those things.

For verification, and to demonstrate that the bug does not occur when the release made by the revised workflow is used, I ran these commands on Windows, which correspond to those in "Demonstration checking binaries in all archives of a release" in #1472, except that they will not currently work for anyone but me because they refer to a private repository to which no one else currently has access:

```powershell
mkdir tmp
cd tmp
gh release -R EliahKagan/private-gitoxide download v0.38.0-alpha.5
gci | %{ 7z x $_ }
gci *.tar | %{ 7z x $_ }
file */gix* */ein*
file */gix* */ein* | sls -raw i686
```

The last four commands are the same. The full output is in [this second file in the related gist](https://gist.github.com/EliahKagan/5d2b5a84451c11ccec1b10ba6f2454c6#file-proposed-txt). The output just of the last command is:

```text
gitoxide-lean-v0.38.0-alpha.5-i686-pc-windows-msvc/gix.exe:        PE32 executable (console) Intel 80386, for MS Windows, 4 sections
gitoxide-max-pure-v0.38.0-alpha.5-i686-pc-windows-msvc/gix.exe:    PE32 executable (console) Intel 80386, for MS Windows, 4 sections
gitoxide-max-v0.38.0-alpha.5-i686-pc-windows-msvc/gix.exe:         PE32 executable (console) Intel 80386, for MS Windows, 4 sections
gitoxide-small-v0.38.0-alpha.5-i686-pc-windows-msvc/gix.exe:       PE32 executable (console) Intel 80386, for MS Windows, 4 sections
gitoxide-lean-v0.38.0-alpha.5-i686-pc-windows-msvc/ein.exe:        PE32 executable (console) Intel 80386, for MS Windows, 4 sections
gitoxide-max-pure-v0.38.0-alpha.5-i686-pc-windows-msvc/ein.exe:    PE32 executable (console) Intel 80386, for MS Windows, 4 sections
gitoxide-max-v0.38.0-alpha.5-i686-pc-windows-msvc/ein.exe:         PE32 executable (console) Intel 80386, for MS Windows, 4 sections
gitoxide-small-v0.38.0-alpha.5-i686-pc-windows-msvc/ein.exe:       PE32 executable (console) Intel 80386, for MS Windows, 4 sections
```

Note that these are now 32-bit executables. The other Windows builds remain 64-bit.

*Edit:* I have also manually tested the binaries from the `gitoxide-max-v0.38.0-alpha.5-i686-pc-windows-msvc.zip` test release to ensure that they run on a 32-bit Windows system and that I can use the `gix` executable to clone the `gitoxide` repository via SSH.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -75,32 +75,26 @@ jobs:
           - build: linux
             os: ubuntu-latest
             rust: stable
-            cargo: cargo
             target: x86_64-unknown-linux-musl
           - build: linux-arm
             os: ubuntu-latest
             rust: nightly
-            cargo: cross
             target: arm-unknown-linux-gnueabihf
           - build: macos
             os: macos-latest
             rust: stable
-            cargo: cargo
             target: x86_64-apple-darwin
           - build: win-msvc
             os: windows-latest
             rust: nightly
-            cargo: cargo
             target: x86_64-pc-windows-msvc
           - build: win-gnu
             os: windows-latest
             rust: nightly-x86_64-gnu
-            cargo: cargo
             target: x86_64-pc-windows-gnu
           - build: win32-msvc
             os: windows-latest
             rust: nightly
-            cargo: cargo
             target: i686-pc-windows-msvc
         # on linux we build with musl which causes trouble with open-ssl. For now, just build max-pure there
         # even though we could also build with `--features max-control,http-client-reqwest,gitoxide-core-blocking-client,gix-features/fast-sha1` for fast hashing.
@@ -145,7 +139,7 @@ jobs:
           targets: ${{ matrix.target }}
 
       - name: Use Cross
-        if: matrix.cargo == 'cross'
+        if: matrix.os == 'ubuntu-latest'
         run: |
           cargo install cross
           echo 'CARGO=cross' >> "$GITHUB_ENV"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -75,26 +75,32 @@ jobs:
           - build: linux
             os: ubuntu-latest
             rust: stable
+            cargo: cargo
             target: x86_64-unknown-linux-musl
           - build: linux-arm
             os: ubuntu-latest
             rust: nightly
+            cargo: cross
             target: arm-unknown-linux-gnueabihf
           - build: macos
             os: macos-latest
             rust: stable
+            cargo: cargo
             target: x86_64-apple-darwin
           - build: win-msvc
             os: windows-latest
             rust: nightly
+            cargo: cargo
             target: x86_64-pc-windows-msvc
           - build: win-gnu
             os: windows-latest
             rust: nightly-x86_64-gnu
+            cargo: cargo
             target: x86_64-pc-windows-gnu
           - build: win32-msvc
             os: windows-latest
             rust: nightly
+            cargo: cargo
             target: i686-pc-windows-msvc
         # on linux we build with musl which causes trouble with open-ssl. For now, just build max-pure there
         # even though we could also build with `--features max-control,http-client-reqwest,gitoxide-core-blocking-client,gix-features/fast-sha1` for fast hashing.
@@ -116,14 +122,10 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     env:
-      # For some builds, we use cross to test on 32-bit, and maybe later big-endian, systems.
-      CARGO: cargo
-      # Whether CARGO is `cargo` or `cross`, TARGET_FLAGS will be set to `--target matrix.target`.
+      CARGO: ${{ matrix.cargo }}
       TARGET_FLAGS: --target=${{ matrix.target }}
-      # Whether CARGO is `cargo` or `cross`, TARGET_DIR will include matrix.target.
       TARGET_DIR: ./target/${{ matrix.target }}
-      # Emit backtraces on panics.
-      RUST_BACKTRACE: 1
+      RUST_BACKTRACE: 1  # Emit backtraces on panics.
 
     steps:
       - name: Checkout repository
@@ -142,11 +144,9 @@ jobs:
           toolchain: ${{ matrix.rust }}
           targets: ${{ matrix.target }}
 
-      - name: Use Cross
-        if: matrix.os == 'ubuntu-latest'
-        run: |
-          cargo install cross
-          echo 'CARGO=cross' >> "$GITHUB_ENV"
+      - name: Install Cross
+        if: matrix.cargo == 'cross'
+        run: cargo install cross
 
       - name: Show command used for Cargo
         run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -54,7 +54,7 @@ jobs:
         run: echo '${{ steps.release.outputs.upload_url }}' > artifacts/release-upload-url
 
       - name: Save version number to artifact
-        run: echo '${{ env.VERSION }}' > artifacts/release-version
+        run: echo "$VERSION" > artifacts/release-version
 
       - name: Upload artifacts
         uses: actions/upload-artifact@v4
@@ -152,9 +152,9 @@ jobs:
 
       - name: Show command used for Cargo
         run: |
-          echo 'cargo command is: ${{ env.CARGO }}'
-          echo 'target flag is: ${{ env.TARGET_FLAGS }}'
-          echo 'target dir is: ${{ env.TARGET_DIR }}'
+          echo "cargo command is: $CARGO"
+          echo "target flag is: $TARGET_FLAGS"
+          echo "target dir is: $TARGET_DIR"
 
       - name: Get release download URL
         uses: actions/download-artifact@v4
@@ -168,11 +168,12 @@ jobs:
           echo "VERSION=$(< artifacts/release-version)" >> "$GITHUB_ENV"
 
       - name: Build release binary
-        run: ${{ env.CARGO }} build --verbose --release ${{ env.TARGET_FLAGS }} --no-default-features --features ${{ matrix.feature }}
+        run: |
+          "$CARGO" build --verbose --release "$TARGET_FLAGS" --no-default-features --features ${{ matrix.feature }}
 
       - name: Strip release binary (linux and macos)
         if: matrix.build == 'linux' || matrix.build == 'macos'
-        run: strip 'target/${{ matrix.target }}/release/ein' 'target/${{ matrix.target }}/release/gix'
+        run: strip -- "$TARGET_DIR"/release/{ein,gix}
 
       - name: Strip release binary (arm)
         if: matrix.build == 'linux-arm'
@@ -192,13 +193,13 @@ jobs:
           cp {README.md,LICENSE-*,CHANGELOG.md} "$staging/"
 
           if [ '${{ matrix.os }}' = 'windows-latest' ]; then
-            file 'target/${{ matrix.target }}/release/ein.exe' 'target/${{ matrix.target }}/release/gix.exe'
-            cp 'target/${{ matrix.target }}/release/ein.exe' 'target/${{ matrix.target }}/release/gix.exe' "$staging/"
+            file -- "$TARGET_DIR"/release/{ein,gix}.exe
+            cp -- "$TARGET_DIR"/release/{ein,gix}.exe "$staging/"
             7z a "$staging.zip" "$staging"
             echo "ASSET=$staging.zip" >> "$GITHUB_ENV"
           else
-            file 'target/${{ matrix.target }}/release/ein' 'target/${{ matrix.target }}/release/gix'
-            cp 'target/${{ matrix.target }}/release/ein' 'target/${{ matrix.target }}/release/gix' "$staging/"
+            file -- "$TARGET_DIR"/release/{ein,gix}
+            cp -- "$TARGET_DIR"/release/{ein,gix} "$staging/"
             tar czf "$staging.tar.gz" "$staging"
             echo "ASSET=$staging.tar.gz" >> "$GITHUB_ENV"
           fi
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -145,6 +145,9 @@ jobs:
         run: |
           cargo install cross
           echo 'CARGO=cross' >> "$GITHUB_ENV"
+
+      - name: Set target variables
+        run: |
           echo 'TARGET_FLAGS=--target ${{ matrix.target }}' >> "$GITHUB_ENV"
           echo 'TARGET_DIR=./target/${{ matrix.target }}' >> "$GITHUB_ENV"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -167,21 +157,17 @@ jobs:
...
       - name: Set release upload URL and release version
-        shell: bash
         run: |
-          release_upload_url="$(cat artifacts/release-upload-url)"
-          echo "RELEASE_UPLOAD_URL=$release_upload_url" >> $GITHUB_ENV
-          echo "release upload url: $RELEASE_UPLOAD_URL"
-          release_version="$(cat artifacts/release-version)"
-          echo "RELEASE_VERSION=$release_version" >> $GITHUB_ENV
-          echo "release version: $RELEASE_VERSION"
+          echo "UPLOAD_URL=$(< artifacts/release-upload-url)" >> "$GITHUB_ENV"
```
Byron, line 161: I didn't know this syntax existed! Initializing variables from file content, neat!

Byron (approved):
Thanks a lot for digging in for a fix, it's much appreciated, particularly because I really don't like working on this 😅.

> Further improvement is possible, including by incorporating some of the other recent changes from the ripgrep release workflow on which it was based. This is something I would be interested to do separately but that I think is not naturally within the scope of these changes.

I can only encourage to keep the flow going if you think it makes sense now, and if it interests you. Otherwise, it's probably just as good to keep it stable now and hope it doesn't break for a long time.

> At least in principle, the workflow can also at this point be extended easily for more build targets, so long as they are otherwise able to build and so long as either _(a)_ they can be built with cross-compilation using `cross` on a Linux-based host or _(b)_ they do not require `cross`. However, if adding more Linux-based targets, it would be a good idea to generalize the `strip` logic, or the resulting binaries would have debug symbols even though they would be optimized builds.

By now, it's possible to specify `strip = true` in the `Cargo.toml`  manifest directly, so `cargo` will perform the stripping of release binaries. Thus far I refrained from doing that because when profiling release builds, symbols are a necessity. However, maybe it's possible to define a specific `profile` configuration that performs no stripping but otherwise is equal to the release build which does perform stripping. If that's possible, I think one could just use `cargo` to strip symbols which should be cross-platform.
But when looking at the workflow, I could imagine that within `cross` it wouldn't work after all as it would have to be able to find strangely named platform-dependent `strip`-binaries.

And I a particularly looking forward to seeing the auto-release work again. Now that I see the regex change I wonder how I could ever think that `v.*` should work at all, oh, right, I must have thought it's full regex, even though it's only globs.

In any case, I will be looking forward to seeing this workflow running automatically the next time a release happens :).

.github/workflows/release.yml
EliahKagan: `$(< artifacts/release-upload-url)` actually has the same effect as `$(cat artifacts/release-upload-url)`, just without using `cat` or any other subprocess--the shell opens and reads the file. The Bash reference manual [documents it](https://www.gnu.org/software/bash/manual/bash.html#Command-Substitution) as a special form of command substitution, though its semantics entail that no command is run. A copy of [the manual for bash 3.2](https://www.bashcookbook.com/bashinfo/source/bash-3.2/doc/bashref.pdf) confirms that it was present back then too, and thus safe to use even in `bash` on macOS.

Normally, an unquoted `<` at the start of a command is just input redirection, the same as any other command. But when it is inside `$(` `)` and the entire "command" is a path, `bash` treats it specially. This `bash` feature is not standardized. However, to fix the way environment variables are set and used, I had already made all script steps use `bash` on all operating systems, so it is definitely available. (I wouldn't use it in a command passed to `docker run`, though, since the shell that uses is often `sh`.)

From a technical perspective, replacing `cat` with `<` is independent of the other change here, which was to remove the `echo` commands that tried to print the effect of setting the environment variable. These commands never worked, because appending to the file that `$GITHUB_ENV` expands to is setting environment variables for subsequent steps, not the current step, not even subsequent commands of the script run in the current step. Since in this case the variables didn't seem to require special attention and also are already available for inspection in the workflow run, I just removed them.

Removing those other `echo` commands allowed the initial assignment be eliminated while still avoiding duplication, and thus allowed the variables to be set with one command each. I felt that changing `$(cat ...)` to `$(< ...)` was part of keeping that new combined command as short and simple as it reasonably could be, which is why I did it at the same time.

Skgland:
> By now, it's possible to specify `strip = true` in the `Cargo.toml` manifest directly, so `cargo` will perform the stripping of release binaries. Thus far I refrained from doing that because when profiling release builds, symbols are a necessity. However, maybe it's possible to define a specific `profile` configuration that performs no stripping but otherwise is equal to the release build which does perform stripping. If that's possible, I think one could just use `cargo` to strip symbols which should be cross-platform. But when looking at the workflow, I could imagine that within `cross` it wouldn't work after all as it would have to be able to find strangely named platform-dependent `strip`-binaries.

That sounds like you are looking for a custom profile inheriting from release and overriding strip <https://doc.rust-lang.org/cargo/reference/profiles.html#custom-profiles>

Byron:
Wow, thanks, that's pretty much spot-on! I will leave it to @EliahKagan to implement these (e.g. add a `profile` profile), as I am afraid that the strip option doesn't work if it has to invoke external programs. But maybe the compiler simply emits no names for functions then instead of stripping them out in an extra step. But definitely something to validate I suppose, despite learning towards trusting the `rustc` implementation always works.

Skgland:
The strip setting is documented at <https://doc.rust-lang.org/cargo/reference/profiles.html#strip> and corresponds to the `-C strip=val` compiler flag documented at <https://doc.rust-lang.org/rustc/codegen-options/index.html#strip>.
The way that is worded sounds to me like rustc instructs the linker to strip the info while linking rather than invoking a separate strip tool. Though based on the [tracking issue for `-c strip=val`](https://github.com/rust-lang/rust/issues/72110) mac OS is special as the linker doesn't support stripping while linking.

Byron:
Thanks for sharing your research - it truly looks like this would work on all platforms. MacOS being special shouldn't be a problem as no crosscompilation is going on there. Also, I'd think they will be able to handle differences between MacOS versions better than us in a bash script.

Great, maybe @EliahKagan can use the momentum and adjust the publishing so that stripping is left to `cargo` itself. I'd think that I can use a release-profile named `profiling` to get an unstripped release binary for when a profiling run is desired.

PR #1483, opened 2024-08-01
Implement object checkout in PrepareCheckout

Fernthedev:
In continuation of #1403, I implemented `with_object_id` to allow checking out specific commits such as is the case in submodules. 

This currently does not include any test cases as of writing this. [However, I did make sure this worked from a high level in my project](https://github.com/QuestPackageManager/QPM.CLI/blob/acf6561f16fd700d012295e2bdf735221dfc8399/src/utils/git.rs#L258-L260)

Though this does not properly update the `HEAD` and may not clone properly. Hoping I can fix it by using `util::update_head` and refactoring it to use `oid`, though I leave that decision up to you if you have a better idea.

Apologies for the lack of tests as well.

Tasks:
- [x]  Checkout tree associated with the `oid` of the revision used.
- [ ] Update the `HEAD` with `util::update_head` to use the correct detached state.
- [ ] Add tests ensuring this works as intended
-   [ ] Shallow clone with specific revision, then checkout.
-   [ ] Standard clone while checking out specific revision

### DIFF
--- a/gix/src/clone/checkout.rs
+++ b/gix/src/clone/checkout.rs
@@ -60,6 +60,20 @@ pub mod main_worktree {
 
     /// Modification
     impl PrepareCheckout {
+        /// Set the `name` of the reference to check out, instead of the remote `HEAD`.
+        /// If `None`, the `HEAD` will be used, which is the default.
+        ///
+        /// Note that `name` should be a partial name like `main` or `feat/one`, but can be a full ref name.
+        /// If a branch on the remote matches, it will automatically be retrieved even without a refspec.
+        /// It can also be a commit id.
+        pub fn with_ref_name<'a, Name, E>(mut self, ref_name: Option<Name>) -> Result<PrepareCheckout, E>
+        where
+            Name: TryInto<&'a gix_ref::PartialNameRef, Error = E>,
+        {
+            self.ref_name = ref_name.map(TryInto::try_into).transpose()?.map(ToOwned::to_owned);
+            Ok(self)
+        }
+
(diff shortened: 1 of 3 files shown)


gix/src/clone/checkout.rs
```diff
@@ -60,6 +61,54 @@ pub mod main_worktree {
 
     /// Modification
     impl PrepareCheckout {
+        /// Set the `name` of the reference to check out, instead of the remote `HEAD`.
+        /// If `None`, the `HEAD` will be used, which is the default.
+        ///
+        /// Note that `name` should be a partial name like `main` or `feat/one`, but can be a full ref name.
+        /// If a branch on the remote matches, it will automatically be retrieved even without a refspec.
+        pub fn with_ref_name<'a, Name, E>(self, ref_name: Option<Name>) -> Result<PrepareCheckout, Error>
```
Byron, line 69: These three methods can be a single one if a `revspec` is used, see [rev_parse()](https://docs.rs/gix/latest/gix/struct.Repository.html#method.rev_parse) for details.

Byron (commented):
Thanks for contributing!

Maybe you could add a task list to communicate what you think is missing to make this PR reviewable.

Fernthedev:
Alright, I added some tasks as requested though the extent of my knowledge with this process is limited so there might be more things I'm not aware of.

Byron:
Thanks for giving this a shot! As the new year is looming, I am closing PRs that can't move and that I can't take over. Thus, sadly, this one is among them but at the same time I encourage to reopen it feels right.

PR #1479, opened 2024-07-30
Add ARM64 (AArch64) binary release archives for macOS and Windows

EliahKagan:
As mentioned in https://github.com/Byron/gitoxide/issues/1478#issuecomment-2257584083, this PR does some of the tasks for #1478. I believe this actually is a complete fix for that bug as described. But I suggest keeping the #1478 open and treating that issue more broadly, to encompass both this and other near-future changes. I've opened this PR now so that the core confusion can be avoided in the next release even if there are delays in doing the rest and even if the next release is sooner than anticipated.

Although I suggest scoping this to just those tasks, I'm willing to broaden this PR to include building universal binaries and just keep this PR open for a while until I can get that done. (But there may be some delays, even if nothing goes wrong, because I'm out of private runner minutes. That's probably a blessing in disguise, since this may not be the best way to work on this, even aside from that limitation. But it does mean I'll have to find another way to approach it.)

Details on the produced archives, analogous to [the information there](https://github.com/Byron/gitoxide/pull/1475#issue-2434501361), can be see [**in this gist**](https://gist.github.com/EliahKagan/91b16863df6873a1e0d6099725bff146), which may make even the most important part more readable than it can be here, but here's that part:

```text
C:\Users\ek\tmp> file */gix* */ein*
gitoxide-lean-v0.38.0-alpha.6-aarch64-apple-darwin/gix:            Mach-O 64-bit arm64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|PIE|HAS_TLV_DESCRIPTORS>
gitoxide-lean-v0.38.0-alpha.6-aarch64-pc-windows-msvc/gix.exe:     PE32+ executable (console) Aarch64, for MS Windows, 5 sections
gitoxide-lean-v0.38.0-alpha.6-i686-pc-windows-msvc/gix.exe:        PE32 executable (console) Intel 80386, for MS Windows, 4 sections
gitoxide-lean-v0.38.0-alpha.6-x86_64-apple-darwin/gix:             Mach-O 64-bit x86_64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|PIE|HAS_TLV_DESCRIPTORS>
gitoxide-lean-v0.38.0-alpha.6-x86_64-pc-windows-gnu/gix.exe:       PE32+ executable (console) x86-64 (stripped to external PDB), for MS Windows, 10 sections
[...]

Issue #1477, opened 2024-07-30
Unclear intent regarding optimizations for some profiles


EliahKagan:
### Current behavior 😯

There are a few conditions related to build profiles for `gitoxide` defined in the top-level `Cargo.toml` that seem like they can be addressed when—as planned in the comment discussion in #1475—splitting the `release` profile into separate `release` and `prof` (or `profile` or `profiling`) profiles.

1. [ ] Debug symbols - as discussed.
2. [ ] Unclear intent for LTO - `#lto = "fat"`.
3. [ ] Old plan for optimizations in debug builds.

#### 1. Debug symbols

Currently the `release` profile leaves debug symbols in because they are useful for profiling, but this creates the need to strip them in a separate step when they are not wanted, such as when building binaries to publish in releases on GitHub.

This is as discussed in comments in #1475, with a suggested plan in https://github.com/Byron/gitoxide/pull/1475#issuecomment-2256168338.

#### 2. Unclear intent for LTO

It is not clear if [link-time optimization](https://doc.rust-lang.org/cargo/reference/profiles.html#lto), at higher than the default very limited level, should be done. It's possible this might even differ between `release` and `prof` profiles. (This is the main sub-issue that may need examination and discussion, and the main reason I opened this issue.)

Early on, in 07859cb, various performance-related configuration [was set](https://github.com/Byron/gitoxide/commit/07859cb6031ccc1938f66481ec1dd88cf05185b8#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542) for the `release` profile. This included the most aggressive link-time optimization available—short of [LTO across different languages](https://doc.rust-lang.org/rustc/linker-plugin-lto.html)—which was still in place about a year ago:

https://github.com/Byron/gitoxide/blob/71efcbba112376b4acaf37d662cdb38d369462be/Cargo.toml#L185

Then in bd32e39 (#893) it was commented out, with the effect of changing the `lto` setting from `fat` to [the default](https://doc.rust-lang.org/cargo/reference/profiles.html#release) of `false`, which somewhat unintuitively still performs link-time optimization, just only within the crate itself.

https://github.com/Byron/gitoxide/blob/1d37bf6a773d56eea9003aa626ced413e8e0eaa3/Cargo.toml#L209

[...]

Issue #1242, opened 2024-01-09
Provide more pre-built binaries


NobodyXu:
It would be great to provide:
 - `aarch64-apple-darwin` builds
 - `universal-apple-darwin` builds
 - provide *-musl builds for all *-gnu builds
 - provide a new type of builds which uses `zlib-ng` for max perf and linked with it statically, uses `rustls` to avoid external `openssl` dep.

_Originally posted by @NobodyXu in https://github.com/Byron/gitoxide/issues/1239#issuecomment-1881026218_

mattyclarkson:
`aarch64-linux-{musl,gnu}` builds, if possible.

vitaly-zdanevich:
And no amd64 linux bin?

EliahKagan:
> And no amd64 linux bin?

Some such builds are published, so I'm not sure what you're referring to, but I can think of three things.

#### They are labeled x86_64

In Rust target names, AMD64 targets are labeled `x86_64`. So the archives built by [`release.yml`](https://github.com/GitoxideLabs/gitoxide/blob/main/.github/workflows/release.yml), and attached to GitHub [releases](https://github.com/GitoxideLabs/gitoxide/releases) such as [v0.44.0](https://github.com/GitoxideLabs/gitoxide/releases/tag/v0.44.0), include such builds, but searching for "amd64" does not find them. For example, that most recent `gitoxide` release has:

- [gitoxide-max-pure-v0.44.0-x86_64-unknown-linux-gnu.tar.gz](https://github.com/GitoxideLabs/gitoxide/releases/download/v0.44.0/gitoxide-max-pure-v0.44.0-x86_64-unknown-linux-gnu.tar.gz)
- [gitoxide-max-pure-v0.44.0-x86_64-unknown-linux-musl.tar.gz](https://github.com/GitoxideLabs/gitoxide/releases/download/v0.44.0/gitoxide-max-pure-v0.44.0-x86_64-unknown-linux-musl.tar.gz)

#### For now, they are max-pure only

Those are `max-pure` builds. It may be that you are looking for builds with other (combinations of) features, such as `max`, `small`, and `lean`, as are built and published for various other operating systems.

[...]

PR #1486, opened 2024-08-02
Modernize release CI workflow and include universal binaries and checksums

### LINKED ISSUE #1478: Prebuilt binaries obtained on ARM64 macOS are x86-64
Tasks:

- [x] Refactor so target architectures are always immediately clear ([#1479](https://github.com/Byron/gitoxide/pull/1479)).
- [x] Build `aarch64-apple-darwin` and make release archives for it ([#1479](https://github.com/Byron/gitoxide/pull/1479)).
- [x] Make release archives with universal binaries ([#1486](https://github.com/Byron/gitoxide/pull/1486)).
- [x] Build `aarch64-pc-windows-msvc` and make release archives for it ([#1479](https://github.com/Byron/gitoxide/pull/1479)).
- [x] Research if any analogous situation may apply on Linux-based systems ([#1478 (comment)](https://github.com/Byron/gitoxide/issues/1478#issuecomment-2259583839)).

### Current behavior 😯

The release workflow builds binaries for `x86_64-apple-darwin` but not for `aarch64-apple-darwin`.

An ARM64 (AArch64) macOS system is capable of running x86-64 macOS binaries by emulation, and will do so automatically, so it is easy to run the x86-64 builds on an ARM64 Mac. But in view of how performance is a key benefit of `gitoxide`, users would most often not prefer to do this, since the instruction sets are not compatible, so emulation imposes a significant overhead. Most importantly, users would most often not want to *inadvertently* run an x86-64 build on ARM64.

Unfortunately, this is likely to happen when using `cargo binstall`:

- When using it directly, it prompts for confirmation, but users may miss the different architecture in the target name.
- When invoking it indirectly, as through `cargo install-update`, it does not prompt for confirmation.

It is really the indirect case where I believe users likely to miss that this is installing an x86-64 binary on their ARM64 system. This is because:

- It may be one of many upgrades as when passing `-a`/`--all`.
- *This happens, without any warning, even when upgrading from an ARM64 build.*

This can be fixed for future releases by adding `aarch64-apple-darwin` jobs to the release workflow. That has further benefits, since even users who are not being inadvertently misled into using the wrong build will have another way, besides `cargo quickinstall`, to install same-architecture binaries.

`cargo quickinstall` does succeed at obtaining a native `aarch64-apple-darwin` build, since there is such a build among the `cargo quickinstall` releases. Current versions of `cargo binstall` are [capable of installing those](https://github.com/cargo-bins/cargo-quickinstall?tab=readme-ov-file#relationship-to-cargo-binstall). But that does not happen here with `cargo binstall`, since `cargo binstall` apparently prefers the `gitoxide` releases even though their architecture is not as good of a match.

Unless something gets in the way, I will try to do this in my next PR that modifies the workflow, and I may fix by itself or with limited other changes rather than combining it with fixes for #1477.

#### An analogous situation exists for Windows

The same problem happens on an ARM64 Windows system, which also supports x86-64 executables by emulation. On such a system, one would ordinarily wish to run a `aarch64-pc-windows-msvc` build, but `cargo binstall` provides a `x86_64-pc-windows-msvc` build instead. I expect that this is less often encountered in practice so far, in part because I don't think ARM64 Windows systems are are common yet as ARM64 macOS systems.

Fortunately, it should be feasible to add both at the same time. Unlike on the Ubuntu runner where the tools to perform cross-compilation without `cross` would have to be installed explicitly (such as via `sudo apt install ...`) and further steps might be needed to install some libraries or to enable `pkg-config` to find them, on the macOS and Windows runners both targets have the Rust-nonspecific tools for targeting both architectures already.

#### Linux-based systems

Something like this may happen for Linux-based systems. I haven't investigated that at this time, because I wouldn't be immediately fixing it for those systems, as #1477 should probably be fixed before adding new Linux targets.

#### Other systems

At this time, we don't make *any* binary builds for targets outside macOS, Windows, and Linux-based systems. So users of such systems are not likely to be affected.

### Expected behavior 🤔

Since macOS is increasingly run on ARM64 processors, one may expect `aarch64-apple-darwin` binaries to be available. But the reason I wanted to open this issue and ensure this is not forgotten even if I have trouble adding the target is two other expectations that I think users hold more strongly and implicitly:

- Getting a binary that is slower because it is for a different architecture that requires emulation should not happen without clear indication.
- Switching the architecture of the binary should not happen automatically when upgrading.
- The recommended binary installation method in the readme (which is `cargo binstall`) should deliver a build of the correct architecture.
- If `cargo binstall` is otherwise capable of delivering a build of the correct architecture by using the `cargo-quickinstall` releases, then nothing about the binaries `gitoxide` provides as its own releases should *prevent* that from happening.

To an extent the reason these expectations are not satisfied is due to aspects of the design of some other projects that could perhaps also be improved. For example, `cargo install-update` could be changed to warn or even prompt when it would upgrade to a different architecture, or `cargo binstall` could check for better architecture matches in the `cargo-quickinstall` releases even if an installable match has already been found.

However, for `gitoxide`, it should be feasible and maybe even easy to fix this by also shipping ARM64 macOS releases.

**Regarding Windows:** In this section, I didn't add anything about expectations for Windows ARM64 targets, since as noted above, I don't think the expectations about that are as strong, though I do believe we should provide such builds.

**Regarding Linux-based systems:** Users may be less likely to assume suitable builds are available, because most feature builds are not currently available (also something that #1477 will work toward).

### Git behavior

I think this is mostly not applicable, because as far as I know, the upstream Git project does not build official binaries. However:

#### Downstream `git` builds for macOS

macOS does have `git` (or maybe the Apple developer tools install it, I'm not sure). That binary carries both x86-64 and ARM64 code:

```text
~> git version
git version 2.39.3 (Apple Git-146)
~> type git
git is /usr/bin/git
~> file (command -v git)
/usr/bin/git: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64e:Mach-O 64-bit executable arm64e]
/usr/bin/git (for architecture x86_64): Mach-O 64-bit executable x86_64
/usr/bin/git (for architecture arm64e): Mach-O 64-bit executable arm64e
```

(The ARM64 code, at least on that particular macOS 14.5 system, is `arm64e` rather than `arm64`, but I *think* that does not have significant performance implications, and that, even though there are many different ARM architectures, `arm64` vs. `arm64e` is an ABI difference rather than being different architectures.)

Although this is a relevant comparison, I mainly present the above to show that the `file` command on macOS is capable of revealing when a binary contains code of both architectures, so it's clear that its output, as shown below, really does demonstrate the issue.

#### Git for Windows

Git for Windows does not yet provide ARM64 builds, so the x86-64 version of Git for Windows really is the best available target for an ARM64 Windows system (unless one is willing to use experimental builds, which in any case I don't think are provided as binaries).

This is one of the current benefits of `gitoxide` and one of the reasons I think it's useful to make it as easy as possible to install ARM64 binaries on Windows. While `gitoxide` has performance benefits for some of its functionality that overlaps with that of `git`, such as [for cloning](https://github.com/Byron/gitoxide/discussions/349), the performance benefit is far greater on an ARM64 Windows system where Git for Windows requires emulation but `gitoxide` does not. (Of course, Git for Windows native ARM64 builds for Windows [will be coming](https://github.com/git-for-windows/git/issues/3107); this is not likely to be a long-standing difference.)

### Steps to reproduce 🕹

I used a macOS 14.5 system running on an ARM64 (AArch64) processor, and nothing was set up in a way that would itself lead to an x86-64 build being seen as preferred:

```text
~> sw_vers -productVersion
14.5
~> uname -m
arm64
~> rustup toolchain list
stable-aarch64-apple-darwin (default)
~> rustup target list --installed
aarch64-apple-darwin
```

Installing with `cargo binstall` gave x86-64, not ARM64, binaries:

```text
~> cargo binstall gitoxide
 INFO resolve: Resolving package: 'gitoxide'
 WARN The package gitoxide v0.37.0 (x86_64-apple-darwin) has been downloaded from github.com
 INFO This will install the following binaries:
 INFO   - ein (ein -> /Users/ek/.cargo/bin/ein)
 INFO   - gix (gix -> /Users/ek/.cargo/bin/gix)
Do you wish to continue? yes/[no]
? yes
 INFO Installing binaries...
 INFO Done in 14.699131666s
~> file (command -v gix ein)
/Users/ek/.cargo/bin/gix: Mach-O 64-bit executable x86_64
/Users/ek/.cargo/bin/ein: Mach-O 64-bit executable x86_64
```

(This was in `fish`. To do this in `zsh`, `bash`, or another Bourne-style shell, I would use a `$` before the opening parenthesis of the command substitution.)

I uninstalled, then installed the *previous* version from source, then upgraded with `cargo install-upgrade`, showing that the result, due to its use of `cargo-binstall`, was to upgrade to a x86-64 build:

```text
~> cargo uninstall gitoxide
    Removing /Users/ek/.cargo/bin/ein
    Removing /Users/ek/.cargo/bin/gix
~> cargo install gitoxide@0.36.0 --locked --quiet
~> cargo install-update --all
    Polling registry 'https://index.crates.io/'.....

Package             Installed  Latest   Needs update
gitoxide            v0.36.0    v0.37.0  Yes
cargo-binstall      v1.8.0     v1.8.0   No
cargo-nextest       v0.9.72    v0.9.72  No
cargo-quickinstall  v0.2.10    v0.2.10  No
cargo-update        v13.4.0    v13.4.0  No

Updating gitoxide
 INFO resolve: Resolving package: 'gitoxide'
 WARN The package gitoxide v0.37.0 (x86_64-apple-darwin) has been downloaded from github.com
 INFO This will install the following binaries:
 INFO   - ein (ein -> /Users/ek/.cargo/bin/ein)
 INFO   - gix (gix -> /Users/ek/.cargo/bin/gix)
 INFO Installing binaries...
 INFO Done in 3.101160291s


Updated 1 package.
Overall updated 1 package: gitoxide.
~> file (command -v gix ein)
/Users/ek/.cargo/bin/gix: Mach-O 64-bit executable x86_64
/Users/ek/.cargo/bin/ein: Mach-O 64-bit executable x86_64
```

I uninstalled again, and installed the current version from source for contrast, showing that gives ARM64:

```text
~> cargo uninstall gitoxide
    Removing /Users/ek/.cargo/bin/ein
    Removing /Users/ek/.cargo/bin/gix
~> cargo install gitoxide --quiet
~> file (command -v gix ein)
/Users/ek/.cargo/bin/gix: Mach-O 64-bit executable arm64
/Users/ek/.cargo/bin/ein: Mach-O 64-bit executable arm64
```

And again, showing that `cargo quickinstall` gives ARM64 as well:

```text
~> cargo quickinstall gitoxide
Calling `cargo-binstall` to do the install
 INFO resolve: Resolving package: 'gitoxide'
 INFO resolve: Verified signature for package 'gitoxide-0.37.0-aarch64-apple-darwin': timestamp:1722141643      file:gitoxide-0.37.0-aarch64-apple-darwin.tar.gz        hashed
 WARN The package gitoxide v0.37.0 (aarch64-apple-darwin) has been downloaded from third-party source QuickInstall
 INFO This will install the following binaries:
 INFO   - ein (ein -> /Users/ek/.cargo/bin/ein)
 INFO   - gix (gix -> /Users/ek/.cargo/bin/gix)
 INFO Installing binaries...
 INFO Done in 2.783621666s
~> file (command -v gix ein)
/Users/ek/.cargo/bin/gix: Mach-O 64-bit executable arm64
/Users/ek/.cargo/bin/ein: Mach-O 64-bit executable arm64
```

**Windows:** On an ARM64 Windows 11 system with `rustup` installed via https://win.rustup.rs/aarch64 ([the Rust site](https://www.rust-lang.org/tools/install) and [the rustup site](https://rustup.rs/) don't yet link to it, but [it's official](https://github.com/rust-lang/rustup/issues/3406#issuecomment-2179792050)), and with `cargo-binstall` installed [with this technique](https://github.com/cargo-bins/cargo-binstall?tab=readme-ov-file#windows), installing `gitoxide` using `cargo binstall` installs the x86-64 version:

```text
PS C:\Users\parnassus> cargo uninstall gitoxide
    Removing C:\Users\parnassus\.cargo\bin\ein.exe
    Removing C:\Users\parnassus\.cargo\bin\gix.exe
PS C:\Users\parnassus> cargo binstall gitoxide
 INFO resolve: Resolving package: 'gitoxide'
 WARN The package gitoxide v0.37.0 (x86_64-pc-windows-msvc) has been downloaded from github.com
 INFO This will install the following binaries:
 INFO   - ein.exe (ein.exe -> C:\Users\parnassus\.cargo\bin\ein.exe)
 INFO   - gix.exe (gix.exe -> C:\Users\parnassus\.cargo\bin\gix.exe)
Do you wish to continue? yes/[no]
? yes
 INFO Installing binaries...
 INFO Done in 13.2124373s
PS C:\Users\parnassus> &'C:\Program Files\Git\usr\bin\file.exe' (gcm gix).path
C:\Users\parnassus\.cargo\bin\gix.exe: PE32+ executable (console) x86-64, for MS Windows, 5 sections
PS C:\Users\parnassus> &'C:\Program Files\Git\usr\bin\file.exe' (gcm ein).path
C:\Users\parnassus\.cargo\bin\ein.exe: PE32+ executable (console) x86-64, for MS Windows, 5 sections
PS C:\Users\parnassus> cargo uninstall gitoxide
    Removing C:\Users\parnassus\.cargo\bin\ein.exe
    Removing C:\Users\parnassus\.cargo\bin\gix.exe
```

### LINKED ISSUE #1484: Release workflow relies on deprecated `set-output` feature
### Current behavior 😯

The `release.yml` workflow uses the unmaintained [upload-release-asset](https://github.com/actions/upload-release-asset) action to attach archive files to just-made releases. This action sets step output by writing text to stdout that contains the string `::set-output`, which GitHub Actions treats specially.

This old technique is deprecated. [GitHub hopes eventually to remove it](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/), due to the security problems entailed by setting output when displaying untrusted data that may contain the string `::set-output`, though they have deferred doing so since [this and related features remain in somewhat wide use](https://github.blog/changelog/2023-07-24-github-actions-update-on-save-state-and-set-output-commands/). If and when it is eventually removed, workflows that use it, even indirectly as in `release.yml`, will break.

Currently, this accounts for many of the warnings issued when the release workflow runs. This is from the most recent run as of this writing, which was prior to the recent changes in #1475 and #1479, but those changes do not affect this behavior. The warnings about this can be seen [on a summary page](https://github.com/Byron/gitoxide/actions/runs/10064652615) for any run of the workflow, and say:

> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

The other warnings are also due to the use of the unmaintained  [upload-release-asset](https://github.com/actions/upload-release-asset) action. They are about a mismatch between the version of Node.js that the action was written for, and the version it is run on. It is a Node 12 action, but this Node version has been unsupported for a long time, and GitHub no longer provides it on GHA runners, both for security and general maintenance reasons. As a result, it runs on Node 16.

> The following actions uses node12 which is deprecated and will be forced to run on node16: actions/upload-release-asset@v1.0.2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

It has been doing so for a while and works, so it is unlikely to break due to running on Node 16. But Node 16 is *also* now unsupported for a long time, and GitHub shall soon no longer provide it; it will run automatically on Node 20 sometime soon. This is detailed in [GitHub Actions: Transitioning from Node 16 to Node 20](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/). My guess is that this will not break it, but it is hard to be completely sure. So this is a second reason to replace uses of that action.

### Expected behavior 🤔

There is not really a strong expectation, but ideally there would be nothing in the release workflow that could plausibly cause breakage in the foreseeable future, and accordingly no warnings about such conditions.

#### Git doesn't have corresponding workflows

To the best my of knowledge, Git does not use GitHub Actions for releasing, and neither does Git for Windows. However, they do use GitHub Actions for other purposes. It does not appear that their workflows issue such warnings, and a quick glance at the actions listed as dependencies on the Insights tab seems to confirm this. I didn't carefully check, though, since the relevance of workflows that don't point the way to what gitoxide could use is rather slight.

#### But ripgrep does

In contrast, the [ripgrep release workflow](https://github.com/BurntSushi/ripgrep/blob/master/.github/workflows/release.yml) is very relevant, because the gitoxide release workflow is [based on](https://github.com/Byron/gitoxide/blob/12313f2720bb509cb8fa5d7033560823beafb91c/.github/workflows/release.yml#L1-L2) it (https://github.com/Byron/gitoxide/pull/1239#pullrequestreview-1809363846, https://github.com/Byron/gitoxide/issues/1472#issuecomment-2254244666).

Over time, the ripgrep workflow has been revised significantly, and one of the changes has been to eliminate the use of that action, as well as some others, by switching to using the [`gh` command](https://cli.github.com/) instead, which is preinstalled on the runners.

This was done in https://github.com/BurntSushi/ripgrep/commit/c9584b035b19244e370a50fd872a3ae2039e2931 (https://github.com/BurntSushi/ripgrep/pull/2554), and for basically the same reason (https://github.com/BurntSushi/ripgrep/pull/2360), though gitoxide already does not use the [create-release](https://github.com/actions/create-release) mentioned there. gitoxide uses the [release-action](https://github.com/ncipollo/release-action) action instead since 43ce9f5, which is fine to use, though also natural to replace by `gh` at the same time.

Using `gh` for everything applicable would simplify things significantly (though, as noted below, I would probably be doing it together with other changes that add complexity elsewhere). Only the version, and not the URL, would have to be passed between dependent jobs. It would also be easier to understand the relationship between the operations on the GitHub release if they are all done in similar ways, so using the `gh` command for some of them but not others would likely be less clear then using it for all of them.

I plan to include this change in the same PR that adds builds of universal binaries, for the currently one remaining task in #1478. They make sense to do at the same time because, as planned in https://github.com/Byron/gitoxide/issues/1478#issuecomment-2257584083, this involves adding at least one more kind of operation on the release: downloading the aarch64 and x86_64 builds from the GitHub release in order to extract their binaries and use them to make the Universal 2 binary. Downloading assets from a release is also easy to do with `gh`.

A possible alternative solution is to use a different action in place of [upload-release-asset](https://github.com/actions/upload-release-asset) for the same functionality. I don't think there is a drop-in replacement, but [its readme](https://github.com/actions/upload-release-asset?tab=readme-ov-file) suggests [action-gh-release](https://github.com/softprops/action-gh-release). This is a much broader action, rather than being specific to uploading assets. This would probably be okay, but it's not the solution I favor, and I mention it only for completeness or in case there are disadvantages of using `gh` here that I am not aware of.

### Steps to reproduce 🕹

Go to any recent run of any job of the release workflow and click Summary near the upper left of the page to examine the summary for the entire workflow run. Then scroll down to see all warnings from all jobs in the workflow. (They are much easier to spot there than when examining individual workflows, where it would usually be necessary to expand specific steps.)

Alternatively, just follow the links above to the most recent one.

EliahKagan:
Closes #1478
Closes #1484

Major changes:

1. [x] Validate version against `Cargo.toml` unless specially named.
2. [x] Create release as a draft and publish after assets are all attached.
3. [x] Use `gh` for all release operations, eliminating deprecated actions ([#1484](https://github.com/Byron/gitoxide/issues/1484)).
4. [x] Make archives with macOS universal binaries ([#1478](https://github.com/Byron/gitoxide/issues/1478) task 3; [#1242](https://github.com/Byron/gitoxide/issues/1242) task 2).
5. [x] Generate and include SHA256 checksums.
6. [x] Document and refactor to avoid confusion.
7. [ ] Make the workflow more readily testable?

#### First… what it looks like

- [Here's an example release.](https://github.com/EliahKagan/gitoxide/releases/tag/v0.38.0-alpha.1-DO-NOT-USE) (This attempts not to change anything related to release *notes*, and I deliberately have not edited the example release in any way.)
- [Here's its associated workflow run.](https://github.com/EliahKagan/gitoxide/actions/runs/10211156907) Logs are available for each job. (Full information should be accessible to any logged-in user. Warnings from any job would be visible on that summary page as well, but there are no warnings anymore.)

To avoid confusion with real releases of gitoxide that people should actually use, I plan to delete that release eventually, but given its distinctive version "number," I'm not too worried and I'd be pleased to leave it up for at least as long as this PR is open and, if useful, possibly a bit longer.

If one wishes to do so, one can download all the files from that test release using [`gh`](https://cli.github.com/):

```sh
mkdir tmp
cd tmp
gh release -R EliahKagan/gitoxide download v0.38.0-alpha.1-DO-NOT-USE
```

Each of the above checkboxed points ("tasks") is elaborated below, including the last one which I've written as unchecked because I'm not sure the changes here really achieve it.

#### 1. Validate version against `Cargo.toml` unless specially named

This is inspired by a check performed in the ripgrep workflow, though the behavior here is different to permit `TEST-*` and `*-DO-NOT-USE` and messages that report the nature of the likely mistake. Also, while I think this may not have been the case when ripgrep added this check, now the `ubuntu-latest` runner has the `yq` command, which accepts TOML input and therefore allows extracting the version from `Cargo.toml` to be done robustly.

#### 2. Create release as a draft and publish after assets are all attached

This creates a draft release, which is only visible to users with push permissions on the repository. If any of steps (of any job) fail, it remains a draft release and the problem can be investigated. If they all succeed, then the last step of the last job marks it non-draft, publishing it.

This is directly adapted from the ripgrep workflow. The difference is that marking the release non-draft seems to be done manually there, rather than happening automatically. (For further discussion of the options related to this, see "Make the workflow more readily testable?" below.)

#### 3. Use `gh` for all release operations, eliminating deprecated actions

This fixes #1484 by using `gh` instead of several actions related to operating on releases, including the `upload-release-asset` action, but also others that are not deprecated. This is for the reasons presented there, and as mentioned there, this also uses a (single) job output rather than GHA artifacts for passing metadata (now it is just a version) to subsequent jobs, thereby eliminating the use of actions related to GHA assets as well.

Although there are minor differences, and also *more* uses of `gh` because it is also used in the changes presented below, this is perhaps the area where the greatest benefit has coming from incorporating changes that appeared in the ripgrep workflow.

#### 4. Make archives with macOS universal binaries

As recommended in https://github.com/Byron/gitoxide/issues/1478#issuecomment-2257408006, this uses `lipo` to create archives with macOS [Universal 2](https://en.wikipedia.org/wiki/Universal_binary#Universal_2) binaries. There is one such archive for each feature.

It makes them by downloading the `aarch64` and `x86_64` archives for the same feature from the draft release and making a repack in which the `ein` and `gix` binaries are each combinations of both architectures' binaries.

Unfortunately, I did not find a way to make the dependencies between jobs as granular as I had hoped in https://github.com/Byron/gitoxide/issues/1478#issuecomment-2257584083. I wanted to make them depend only on non-universal macOS and only on the ones for the same feature. It doesn't look like there is currently a supported way to do this with GitHub Actions using `needs`, and as far as I can tell it would not be made easier by using artifacts rather than the release itself to hold the assets.

It should be feasible to do it, or at least get part of the way there, using other techniques. In particular, the macOS jobs could be split off into a separate job definition, and substantial code duplication--which if present in this workflow would almost certainly lead to bugs, I think--could be avoided with a [reusable workflow or composite action](https://docs.github.com/en/actions/using-workflows/avoiding-duplication).

I decided not to try to do anything like that at this time. Instead, I just made the jobs that generate builds with universal binaries depend on all the jobs that generate the other builds. Unfortunately this means there is a substantial delay between when the architecture-specific builds are attached to the release and when the universal builds made from them are attached. There are two ways this could matter, though *currently* they are not very significant:

- If the release goes back to being crated as a non-draft, i.e., published before everything is attached, then the delay could create the appearance to users looking for them that universal builds are not available.
- If `fail-fast: false` is set in the `build-release` job definition so that jobs from that matrix are no longer cancelled automatically when others fail, then having the universal builds depend on the aarch64 and x86_64 builds could cause the universal builds not to be built when otherwise they would be able to.

#### 5. Generate and include SHA256 checksums

The ripgrep workflow does this and I figured it would be a good idea. But this differs in a few ways:

- Rather than publish with a separate file for each checksum, this initially creates the checksums that way and even attaches them to the draft release, but then once ready to publish, replaces them with a single `hashes.sha256` files that lists all of them. I wasn't sure what filename to use; the name `hashes.sha256` is inspired by that file in [PowerShell release assets](https://github.com/PowerShell/PowerShell/releases/tag/v7.4.4). If anything goes wrong in any step of any job up to that point, then the individual checksum files are retained in the unpublished draft.
- Checksums are generated in binary mode explicitly. This will probably make no difference to the value, but it leaves no doubt. Also, it causes `shasum` to include the leading `*` before each filename in its output, attesting that the checksums were computed while treating the files that way.
- `shasum` is used on all platforms, even Windows, because `certutil` produces output in a different format that, to the best of my knowledge, no common tools will automatically use. (The `shasum` command actually exists on the Windows GHA runners because Git for Windows ships it; it's just in a subdirectory that is not always in the `PATH`.) An example of a Rust project that uses this traditional Unix format for all archives including `.zip` archives for Windows targets is [`uv`](https://github.com/astral-sh/uv/releases/tag/0.2.33) (though I did not use anything from its release workflow).

Although I think this approach is reasonable or I wouldn't be doing it, I also chose it because of its proximity to other approaches: this should not be too hard to change into any of the following if preferred:

- Keep the individual `.sha256` files and still include `hashes.sha256`.
- Keep the individual `.sha256` files and don't include `hashes.sha256`.
- Never attach individual `.sha256` files even while the release is a draft.
- Use `gpg` to sign the `hashes.sha256` file (would need another secret if automatic).

#### 6. Document and refactor to avoid confusion

Overall the comments are actually sparer. But I placed one at the top of each job definition identifying what it does. I think that creating the release as a draft is also more intuitive, which does not mean we necessarily should keep it that way, but does mean that less confusion is likely when reading the workflow as long as it is the case.

The refactoring largely consists of changes to the way non-literal values get into script steps:

- Sometimes, this is simply avoided by splitting steps up so that conditional logic is in `if:` keys guarding the steps.
- For data from environment variables, access them with parameter expansion (`$VAR`) rather than by interpolating into the script with a template (`${{ env.VAR }}`).
- For data not from environment variables, put them in environment variables scoped to the step in a step-level `env:`. (When non-sensitive and frequently used, put them in the job-level `env:`.)

In other scenarios--not this one--the latter two points would be necessary for security, because untrusted data can contain arbitrary characters including quotation marks. In this case, that is not a worry (except by accident), because none of the values are ever untrusted or otherwise especially unpredictable. So this is largely a stylistic matter. But:

- It seems that the effect, in this case, is to increase readability. That is subjective.
- Previous versions of this workflow had bugs that arose due to confusion about the semantics of access to `env`, such as the lifetime of variables inserted by appending to `$GITHUB_ENV`. I think this helps avoid that too.
- Although secret redaction in GitHub Actions logs works just as well either way, having it be explicitly clear reading each step whether it can perform a privileged operation may be helpful to humans. Currently the only secret is `github.GITHUB_TOKEN`, but it does have write permissions.

Relating to the above, I had previously removed debug printing that didn't work because it attempted to print environment variables that were not set. I gave the workflow another look to see if maybe some of this information would be useful after all, and the ripgrep workflow helped too in what it prints. I brought some of that back, in subsequent steps so that the variables are set before being printed.

#### 7. Make the workflow more readily testable?

Having this make releases initially as drafts seems beneficial in and of itself, in that the release can get into its final state before being made public, and in that the failure to produce or attach assets may sometimes mean the release should not be published without fixing something.

But the bigger reason I did it this way was that it enabled me to test this in my (public) fork, without creating a large number of releases that could potentially confuse people especially if they appeared in their GitHub "what's new" type feeds. (Though I did also name my versions with the string `DO-NOT-USE`, so most likely no one will be misled.) Only pretty near the end did I add the commit that enabled the last step of the new `publish-release` job.

But there are other possible approaches and some considerations the current approach may not adequately satisfy:

- Maybe publishing should always be done manually.
- Maybe it should be made public immediately like it was before.
- Maybe considerations related to `cargo-smart-release` have bearing on this?
- Maybe considerations related to manual editing of release notes have bearing on this?
- Maybe it should work like this but have a way to keep it as a draft.
- Maybe it should work like this but (also?) have a way to make it publish immediately (i.e. never be a draft).

Customization is pretty easy to add when a workflow is triggered by the `workflow_dispatch` event (here's [an example](https://github.com/EliahKagan/hello-world/blob/d75d73384035b48590dc1285237ab4423e707e11/.github/workflows/reverse-shell.yml#L5-L14)). It is not always quite so easy in what is the more important case here of being triggered in push. But, for example, this could look at [repository variables](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/variables) to determine some of its behavior. It could also have different behavior based on factors like whether it is running in a fork, though I would want that only to be a matter of different defaults so that testing changes to it in a fork could still be done robustly.

Of course, it can also be changed over time as preferences about how it should work become clear or change. It will also be changing in order to simplify things related to stripping debug symbols (#1477) and to publish releases for more targets or for some existing targets with more features (#1242). This PR satisfies the second point of #1242 (and #1479 satisfied the first), but otherwise this PR does not attempt to cover any of that.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -292,23 +292,20 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-  # This checks the draft release on GitHub and publishes it. It does not upload to crates.io.
-  publish-release:
-    name: publish-release
+  check-release:
+    name: check-release
 
     runs-on: ubuntu-latest
 
     needs: [ create-release, build-release, build-macos-universal2-release ]
-
-    env:
-      REPOSITORY: ${{ github.repository }}
-      VERSION: ${{ needs.create-release.outputs.version }}
 
     steps:
       - name: Discover assets
         run: |
           gh release --repo="$REPOSITORY" view "$VERSION" --json assets --jq '.assets.[].name' > assets.txt
         env:
+          REPOSITORY: ${{ github.repository }}
+          VERSION: ${{ needs.create-release.outputs.version }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Show all asset names
@@ -329,8 +326,34 @@ jobs:
           diff -- <(mask aarch64) <(mask universal)
           diff -- <(mask x86_64) <(mask universal)
 
+  publish-release:
+    name: publish-release
+
+    runs-on: ubuntu-latest
+
+    needs: [ create-release, check-release ]
+
+    env:
+      REPOSITORY: ${{ github.repository }}
+      VERSION: ${{ needs.create-release.outputs.version }}
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    steps:
+      - name: Retrieve individual checksums
+        run: gh release --repo="$REPOSITORY" download "$VERSION" --pattern='gitoxide-*.sha256'
+
+      - name: Concatenate checksums into one file
+        run: cat gitoxide-*.sha256 > hashes.sha256
+
+      - name: Upload the combined checksum file
+        run: gh release --repo="$REPOSITORY" upload "$VERSION" hashes.sha256
+
+      - name: Discard the individual checksum files
+        run: |
+          for sumfile in gitoxide-*.sha256; do
+            gh release --repo="$REPOSITORY" delete-asset "$VERSION" "$sumfile" --yes
+          done
+
       - name: Publish the release
         if: false
         run: gh release --repo="$REPOSITORY" edit "$VERSION" --draft=false
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -347,7 +347,6 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Publish the release
-        if: false
         run: gh release --repo="$REPOSITORY" edit "$VERSION" --draft=false
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -88,11 +88,7 @@ jobs:
           - x86_64-pc-windows-gnu
           - i686-pc-windows-msvc
           - aarch64-pc-windows-msvc
-        feature: &features
-          - small
-          - lean
-          - max
-          - max-pure
+        feature: [ small, lean, max, max-pure ]
         include:
           - target: x86_64-unknown-linux-musl
             os: ubuntu-latest
@@ -239,7 +235,7 @@ jobs:
 
     strategy:
       matrix:
-        feature: *features
+        feature: [ small, lean, max, max-pure ]
 
     steps:
       - name: Placeholder
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -247,17 +247,28 @@ jobs:
       - name: Obtain single-architecture releases
         run: |
           gh release download "$VERSION" --pattern="$(name aarch64).tar.gz" --pattern="$(name x86_64).tar.gz"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Unpack single-architecture releases
+        run: |
           tar xf "$(name aarch64).tar.gz"
           tar xf "$(name x86_64).tar.gz"
 
-      - name: Repack into Universal-2 release
+      - name: Pre-populate directory for archive
         run: |
           cp -R -- "$(name aarch64)" "$(name universal)"
           rm -- "$(name universal)"/{ein,gix}
+
+      - name: Create Universal 2 binaries
+        run: |
           for bin in ein gix; do
             lipo -create "$(name aarch64)/$bin" "$(name x86_64)/$bin" -output "$(name universal)/$bin"
+            file "$(name universal)/$bin"
           done
-          file "$(name universal)"/{ein,gix}
+
+      - name: Build archive
+        run: |
           tar czf "$(name universal).tar.gz" "$(name universal)"
           echo "ASSET=$(name universal).tar.gz" >> "$GITHUB_ENV"
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,11 +12,6 @@ on:
     tags:
       - 'v*'
 
-env:
-  RUST_BACKTRACE: 1
-  CARGO_TERM_COLOR: always
-  CLICOLOR: 1
-
 defaults:
   run:
     shell: bash
@@ -27,10 +22,13 @@ jobs:
   # from building the release so that we only create the release once.
   create-release:
     name: create-release
+
     runs-on: ubuntu-latest
+
 #    env:
 #      # Set to force version number, e.g., when no tag exists.
 #      VERSION: TEST-0.0.0
+
     steps:
       - name: Create artifacts directory
         run: mkdir artifacts
@@ -133,10 +131,12 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     env:
-      CARGO: cargo  # On Linux, this will be changed to `cross` later.
+      CARGO: cargo  # On Linux, this will be changed to `cross` in a later step.
       TARGET_FLAGS: --target=${{ matrix.target }}
       TARGET_DIR: ./target/${{ matrix.target }}
-      RUST_BACKTRACE: 1  # Emit backtraces on panics.
+      RUST_BACKTRACE: '1'  # Emit backtraces on panics.
+      CARGO_TERM_COLOR: always
+      CLICOLOR: '1'
 
     steps:
       - name: Checkout repository
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -62,9 +62,6 @@ jobs:
             ;;
           esac
 
-      - name: Create artifacts directory
-        run: mkdir artifacts
-
       - name: Create GitHub release
         id: release
         uses: ncipollo/release-action@v1
@@ -76,6 +73,9 @@ jobs:
           omitBody: true
           omitPrereleaseDuringUpdate: true
           token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Create artifacts directory
+        run: mkdir artifacts
 
       - name: Save release upload URL to artifact
         run: echo '${{ steps.release.outputs.upload_url }}' > artifacts/release-upload-url
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -174,11 +174,13 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Install packages (Ubuntu)
-        # Because openssl doesn't work on musl by default, we resort to max-pure. And that won't need any dependency, so we can skip this.continue-on-error
+        # Because openssl doesn't work on musl by default, we resort to max-pure.
+        # And that won't need any dependency, so we can skip this or use `continue-on-error`.
         # Once we want to support better zlib performance, we might have to re-add it.
         if: matrix.os == 'ubuntu-latest-disabled'
         run: |
-          sudo apt-get update && sudo apt-get install -y --no-install-recommends xz-utils liblz4-tool musl-tools
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends xz-utils liblz4-tool musl-tools
 
       - name: Install Rust
         uses: dtolnay/rust-toolchain@master
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,7 +35,10 @@ jobs:
 
       - name: Get the release version from the tag
         if: env.VERSION == ''
-        run: echo 'VERSION=${{ github.ref_name }}' >> "$GITHUB_ENV"
+        run: echo "VERSION=$VERSION" >> "$GITHUB_ENV"
+        env:
+          VERSION: ${{ github.ref_name }}
+
 
       - name: Validate version against Cargo.toml
         run: |
@@ -78,7 +81,9 @@ jobs:
         run: mkdir artifacts
 
       - name: Save release upload URL to artifact
-        run: echo '${{ steps.release.outputs.upload_url }}' > artifacts/release-upload-url
+        run: echo "$URL" > artifacts/release-upload-url
+        env:
+          URL: ${{ steps.release.outputs.upload_url }}
 
       - name: Save version number to artifact
         run: echo "$VERSION" > artifacts/release-version
@@ -156,8 +161,10 @@ jobs:
 
     env:
       CARGO: cargo  # On Linux, this will be changed to `cross` in a later step.
+      TARGET: ${{ matrix.target }}
       TARGET_FLAGS: --target=${{ matrix.target }}
       TARGET_DIR: target/${{ matrix.target }}
+      FEATURE: ${{ matrix.feature }}
       RUST_BACKTRACE: '1'  # Emit backtraces on panics.
       CARGO_TERM_COLOR: always
       CLICOLOR: '1'
@@ -204,7 +211,7 @@ jobs:
 
       - name: Build release binary
         run: |
-          "$CARGO" build --verbose --release "$TARGET_FLAGS" --no-default-features --features ${{ matrix.feature }}
+          "$CARGO" build --verbose --release "$TARGET_FLAGS" --no-default-features --features "$FEATURE"
 
       - name: Strip release binary (x86-64 Linux, and all macOS)
         if: matrix.target == 'x86_64-unknown-linux-musl' || matrix.os == 'macos-latest'
@@ -222,12 +229,12 @@ jobs:
 
       - name: Build archive
         run: |
-          staging='gitoxide-${{ matrix.feature }}-${{ env.VERSION }}-${{ matrix.target }}'
+          staging="gitoxide-$FEATURE-$VERSION-$TARGET"
           mkdir -p -- "$staging"
 
           cp {README.md,LICENSE-*,CHANGELOG.md} "$staging/"
 
-          if [ '${{ matrix.os }}' = 'windows-latest' ]; then
+          if [ "$OS" = 'windows-latest' ]; then
             file -- "$TARGET_DIR"/release/{ein,gix}.exe
             cp -- "$TARGET_DIR"/release/{ein,gix}.exe "$staging/"
             7z a "$staging.zip" "$staging"
@@ -238,6 +245,8 @@ jobs:
             tar czf "$staging.tar.gz" "$staging"
             echo "ASSET=$staging.tar.gz" >> "$GITHUB_ENV"
           fi
+        env:
+          OS: ${{ matrix.os }}
 
       - name: Upload release archive
         uses: actions/upload-release-asset@v1.0.2
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,4 +1,4 @@
-# This is largely adapted from past and recent versions of the ripgrep release workflow.
+# This is largely adapted from the ripgrep release workflow.
 # https://github.com/BurntSushi/ripgrep/blob/master/.github/workflows/release.yml
 
 name: release
@@ -65,33 +65,12 @@ jobs:
           esac
 
       - name: Create GitHub release
-        id: release
-        uses: ncipollo/release-action@v1
-        with:
-          tag: ${{ env.VERSION }}
-          name: ${{ env.VERSION }}
-          allowUpdates: true
-          draft: true
-          omitBody: true
-          omitPrereleaseDuringUpdate: true
-          token: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Create artifacts directory
-        run: mkdir artifacts
-
-      - name: Save release upload URL to artifact
-        run: echo "$URL" > artifacts/release-upload-url
+        run: gh release create "$VERSION" --title="$VERSION" --draft
         env:
-          URL: ${{ steps.release.outputs.upload_url }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Save version number to artifact
-        run: echo "$VERSION" > artifacts/release-version
-
-      - name: Upload artifacts
-        uses: actions/upload-artifact@v4
-        with:
-          name: artifacts
-          path: artifacts
+    outputs:
+      version: ${{ env.VERSION }}
 
   build-release:
     name: build-release
@@ -199,17 +178,6 @@ jobs:
           echo "target flag is: $TARGET_FLAGS"
           echo "target dir is: $TARGET_DIR"
 
-      - name: Get release download URL
-        uses: actions/download-artifact@v4
-        with:
-          name: artifacts
-          path: artifacts
-
-      - name: Set release upload URL and release version
-        run: |
-          echo "UPLOAD_URL=$(< artifacts/release-upload-url)" >> "$GITHUB_ENV"
-          echo "VERSION=$(< artifacts/release-version)" >> "$GITHUB_ENV"
-
       - name: Build release binary
         run: |
           "$CARGO" build --verbose --release "$TARGET_FLAGS" --no-default-features --features "$FEATURE"
@@ -228,33 +196,36 @@ jobs:
             /target/arm-unknown-linux-gnueabihf/release/ein \
             /target/arm-unknown-linux-gnueabihf/release/gix
 
-      - name: Build archive
-        run: |
-          staging="gitoxide-$FEATURE-$VERSION-$TARGET"
-          mkdir -p -- "$staging"
-
-          cp {README.md,LICENSE-*,CHANGELOG.md} "$staging/"
-
-          if [ "$OS" = 'windows-latest' ]; then
-            file -- "$TARGET_DIR"/release/{ein,gix}.exe
-            cp -- "$TARGET_DIR"/release/{ein,gix}.exe "$staging/"
-            7z a "$staging.zip" "$staging"
-            echo "ASSET=$staging.zip" >> "$GITHUB_ENV"
-          else
-            file -- "$TARGET_DIR"/release/{ein,gix}
-            cp -- "$TARGET_DIR"/release/{ein,gix} "$staging/"
-            tar czf "$staging.tar.gz" "$staging"
-            echo "ASSET=$staging.tar.gz" >> "$GITHUB_ENV"
-          fi
+      - name: Determine version
+        run: echo "VERSION=$VERSION" >> "$GITHUB_ENV"
         env:
-          OS: ${{ matrix.os }}
+          VERSION: ${{ needs.create-release.outputs.version }}
+
+      - name: Determine archive basename
+        run: echo "ARCHIVE=gitoxide-$FEATURE-$VERSION-$TARGET" >> "$GITHUB_ENV"
+
+      - name: Pre-populate directory for archive
+        run: |
+          mkdir -- "$ARCHIVE"
+          cp {README.md,LICENSE-*,CHANGELOG.md} "$ARCHIVE/"
+
+      - name: Build archive (Windows)
+        if: matrix.os == 'windows-latest'
+        run: |
+          file -- "$TARGET_DIR"/release/{ein,gix}.exe
+          cp -- "$TARGET_DIR"/release/{ein,gix}.exe "$ARCHIVE/"
+          7z a "$ARCHIVE.zip" "$ARCHIVE"
+          echo "ASSET=$ARCHIVE.zip" >> "$GITHUB_ENV"
+
+      - name: Build archive (Unix)
+        if: matrix.os != 'windows-latest'
+        run: |
+          file -- "$TARGET_DIR"/release/{ein,gix}
+          cp -- "$TARGET_DIR"/release/{ein,gix} "$ARCHIVE/"
+          tar czf "$ARCHIVE.tar.gz" "$ARCHIVE"
+          echo "ASSET=$ARCHIVE.tar.gz" >> "$GITHUB_ENV"
 
       - name: Upload release archive
-        uses: actions/upload-release-asset@v1.0.2
+        run: gh release upload "$VERSION" "$ASSET"
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        with:
-          upload_url: ${{ env.UPLOAD_URL }}
-          asset_path: ${{ env.ASSET }}
-          asset_name: ${{ env.ASSET }}
-          asset_content_type: application/octet-stream
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,4 +1,4 @@
-# This is largely adapted from the ripgrep release workflow.
+# Much of this workflow is adapted from the ripgrep release workflow.
 # https://github.com/BurntSushi/ripgrep/blob/master/.github/workflows/release.yml
 
 name: release
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,6 +46,7 @@ jobs:
           tag: ${{ env.VERSION }}
           name: ${{ env.VERSION }}
           allowUpdates: true
+          draft: true
           omitBody: true
           omitPrereleaseDuringUpdate: true
           token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,9 +17,7 @@ defaults:
     shell: bash
 
 jobs:
-  # The create-release job runs purely to initialize the GitHub release itself,
-  # and names the release after the version tag that was pushed. It's separate
-  # from building the release so that we only create the release once.
+  # Create a draft release, initially with no binary assets attached.
   create-release:
     runs-on: ubuntu-latest
 
@@ -70,6 +68,7 @@ jobs:
     outputs:
       version: ${{ env.VERSION }}
 
+  # Build for a particular feature and target, and attach an archive for it.
   build-release:
     needs: [ create-release ]
 
@@ -223,6 +222,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+  # Add a macOS universal binary archive for a feature using its built aarch64 and x86_64 assets.
   build-macos-universal2-release:
     runs-on: macos-latest
 
@@ -286,6 +286,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+  # Check for some problems, consolidate checksum files into one, and mark the release non-draft.
   publish-release:
     runs-on: ubuntu-latest
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -134,14 +134,15 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     env:
-      CARGO: cargo  # On Linux, this will be changed to `cross` in a later step.
-      TARGET: ${{ matrix.target }}
-      TARGET_FLAGS: --target=${{ matrix.target }}
-      TARGET_DIR: target/${{ matrix.target }}
-      FEATURE: ${{ matrix.feature }}
       RUST_BACKTRACE: '1'  # Emit backtraces on panics.
       CARGO_TERM_COLOR: always
       CLICOLOR: '1'
+      CARGO: cargo  # On Linux, this will be changed to `cross` in a later step.
+      FEATURE: ${{ matrix.feature }}
+      VERSION: ${{ needs.create-release.outputs.version }}
+      TARGET: ${{ matrix.target }}
+      TARGET_FLAGS: --target=${{ matrix.target }}
+      TARGET_DIR: target/${{ matrix.target }}
 
     steps:
       - name: Checkout repository
@@ -192,11 +193,6 @@ jobs:
             /target/arm-unknown-linux-gnueabihf/release/ein \
             /target/arm-unknown-linux-gnueabihf/release/gix
 
-      - name: Determine version
-        run: echo "VERSION=$VERSION" >> "$GITHUB_ENV"
-        env:
-          VERSION: ${{ needs.create-release.outputs.version }}
-
       - name: Determine archive basename
         run: echo "ARCHIVE=gitoxide-$FEATURE-$VERSION-$TARGET" >> "$GITHUB_ENV"
 
@@ -226,8 +222,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-  build-macos-universal-binary-release:
-    name: build-macos-universal-binary-release
+  build-macos-universal2-release:
+    name: build-macos-universal2-release
 
     runs-on: macos-latest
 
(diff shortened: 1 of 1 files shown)


Byron:
Thank you so, so much for tackling this!

Here is my thoughts and comments related to the various points.

#### 1. Validate version against Cargo.toml unless specially named

Neat, appreciated!

#### 2. Create release as a draft and publish after assets are all attached

That is a major improvement, thank you!

#### 3. Use gh for all release operations, eliminating deprecated actions

That is great to hear, `gh` has been very stable and could hardly be more trusted to keep working like for some time to come, greatly improving the reliability of the workflow.

#### 4. Make archives with macOS universal binaries

This is fantastic and a great achievement, thank you!

> It should be feasible to do it, or at least get part of the way there, using other techniques. In particular, the macOS jobs could be split off into a separate job definition, and substantial code duplication--which if present in this workflow would almost certainly lead to bugs, I think--could be avoided with a [reusable workflow or composite action](https://docs.github.com/en/actions/using-workflows/avoiding-duplication).

I had touched points with these over at GitButler and didn't particularly like them for the added level of indirection. But then again, I also don't like YAML based things so I am biased. However, if you think the complexity is worth it and feel like it should be done, I'd absolutely go with it if it manages to avoid code duplication. Personally, I am fine with the trade-offs as is, as I assume builds work most of the time now.

#### 5. Generate and include SHA256 checksums

Thank you, I am absolutely happy with the choices made.

Something I'd find helpful is if the filename would reveal information on either how it was created or how users can validate the hash. I suppose the suffix already does it, but I wonder if naming it `shasum-hashes.sha256` would help, maybe along with a ready-made `shasum-hashes.md` to learn more on how these should be used and most importantly, validated. Those who know could already do it, but those who don't probably wouldn't and maybe the UX around this can be improved trivially.

**If you agree, this could be a follow-up PR.**

#### 6. Document and refactor to avoid confusion

I just read through the new file start to end (without using the diff-view), and can confirm that it makes a clean and much simpler impression, with all steps clearly marked. Great work!

There was the `ubuntu-latest-disabled` comparison that explained itself with a comment, but since I read it after the fact I was initially confused about the purpose. My expectation would have been that it's commented out, but that comes with its own issues so probably doing it like this is overall better. My initially confusion is due to not being used to it, and I am happy to learn and adapt.

#### 7. Make the workflow more readily testable?


* Maybe publishing should always be done manually.
    - I am glad I don't have to :).
* Maybe it should be made public immediately like it was before.
    - It seems objectively better to start it out as draft.
* Maybe considerations related to cargo-smart-release have bearing on this?
    - I am glad it has nothing to do with it.
* Maybe considerations related to manual editing of release notes have bearing on this?
    - Once `gix` and `gitoxide` go v1.0, I think the approach to top-level changelogs will be revised. In theory though, this can be done as part of `CHANGELOG.md` and manual edits to it (which are fully supported), so maybe pretty changelogs work already. They will naturally end up in the release.
* Maybe it should work like this but have a way to keep it as a draft. Maybe it should work like this but (also?) have a way to make it publish immediately (i.e. never be a draft).
    - I don't see an application for this yet, probably that's an indicator that it's not (yet) needed.

> Of course, it can also be changed over time as preferences about how it should work become clear or change.

I agree, let's go with that.

---

💜Lovely, thanks again 🙏

EliahKagan:
> Thank you so, so much for tackling this!

No problem!

> #### 4. Make archives with macOS universal binaries
> This is fantastic and a great achievement, thank you!

This is thanks to @NobodyXu for suggesting it and explaining how to use use `lipo` in https://github.com/Byron/gitoxide/issues/1478#issuecomment-2257408006!

> I had touched points with these over at GitButler and didn't particularly like them for the added level of indirection. But then again, I also don't like YAML based things so I am biased.

There are ways to bring in static type checking and even a different syntax. I don't know how much that would help or if it would be worthwhile. I'll try and open a discussion question about it soon.

> However, if you think the complexity is worth it and feel like it should be done, I'd absolutely go with it if it manages to avoid code duplication.

At this point, I think I would only want to do it if it does not increase complexity. Even then, my main reluctance is that putting code that really is tightly coupled to this workflow in another file could make it harder to understand what is going on, even if the combined length and complexity of both files were to decrease.

> Something I'd find helpful is if the filename would reveal information on either how it was created or how users can validate the hash. I suppose the suffix already does it, but I wonder if naming it `shasum-hashes.sha256` would help, maybe along with a ready-made `shasum-hashes.md` to learn more on how these should be used and most importantly, validated. Those who know could already do it, but those who don't probably wouldn't and maybe the UX around this can be improved trivially.

One problem is that I don't think I actually know how people should validate the hash, other than that they should use whatever program their system has for validating SHA256 checksums provided in a file formatted in this traditional way. I think most Unix-like systems have such a command, but the best command to do so may differ across systems. The available command for this is not always `shasum`, though when it is not, I think it is usually `sha256sum`.

The workflow uses the `shasum` command because it is present on all GitHub-hosted GHA runners. (In ripgrep, the workflow used `certutil` on Windows to generate the hashes in a different format. But it used `shasum` rather than `sha256sum` probably because, of the two, only `shasum` is shipped with macOS.) `shasum` is a `perl` script that may be absent if `perl` is not installed, and that I think may also be absent if `perl` is installed but in a minimal way. It may also be in an unusual location even if it is installed, as is the case on the Windows runners (where the workflow hard-codes the path).

One might think that `perl` would always be present, since some `git` commands use it. This is, after all, why a full installation of Git for Windows has it (in contrast, [MinGit](https://github.com/git-for-windows/git/wiki/MinGit) does not provide `perl`). But there are systems on which this is not the case, but where one might reasonably use or even *develop* `gitoxide`. In particular, my Alpine Linux 3.17 system has `git` but not `perl`, and all but one of the `gitoxide` unit tests pass. It does not have `shasum`, but it does have `sha256sum`. Its `sha256sum` is provided by BusyBox. GNU/Linux systems have `sha256sum` from GNU coreutils. The BusyBox `sha256sum` and GNU coreutils `sha256sum` differ in some ways; for example, both support `-c`, but the BusyBox implementation does not recognize long options so does not accept `--check`.

Most Linux-based systems that intentionally support command-line use will therefore have `sha256sum` one way or another. Furthermore, Git for Windows *also* ships `sha256sum`, since its Git Bash environment includes a port of GNU coreutils. (In light of this, I should very possibly have used that command to *generate* the hashes on Windows rather than hard-coding `/usr/bin/core_perl/shasum`. I'll try to look into which is better next time I make changes to the workflow.) It may mainly be macOS where `shasum` is more likely to be present than `sha256sum`.

> **If you agree, this could be a follow-up PR.**

The above is mostly a problem for naming the file in a way that makes clearer how it should be used. It is not necessarily a problem for giving instructions, which could simply say something like, "On most systems, you can use `shasum -c` or `sha256sum -c` to check the hashes," and maybe also something about Git Bash.

However, I am not sure it makes sense for a file with those instructions to be among the release files. It seems like this belongs in the top-level readme, or a separate top-level file giving or reiterating installation instructions. The disadvantage of this is that if the way the hashes are provided changes over time, then there could be skew between the version of the documentation that a user is reading and the version that would apply to the download the user is doing.

Some projects repeat the hashes of downloadable archives in the release notes and give instructions. If that is done, then this makes sense there. Or the release notes could mention the file of hashes and say how to check them, or link to information about how to check them. But I don't know if either of these things is worthwhile.

There is another consideration that may be important. Providing a single file of hashes may be confusing to users who have not often checked hashes, because running `shasum -c` or `sha256sum -c` (or other compatible commands, if any) on it [will report the absence of any files that were not downloaded](https://gist.github.com/EliahKagan/87cc990f4920d0de61da2324d1165242). Maybe this is the reason some projects provide separate `.sha256` files for each archive.

Whatever is done for this, the `publish-release` job can shortened based on it. Right now, [its check for feature consistency](https://github.com/Byron/gitoxide/blob/3604a3b519d76db367084ca49d395f6ec4094cf7/.github/workflows/release.yml#L300-L325) is written separately from [its consolidation of the checksums](https://github.com/Byron/gitoxide/blob/3604a3b519d76db367084ca49d395f6ec4094cf7/.github/workflows/release.yml#L327-L347). I did this deliberately to avoid committing to a particular approach, but with the idea of eventually combining shared work.

- If the approach of doing them independently is desired then it can be reinforced by splitting the job into multiple, probably just two, jobs. That may be better than just leaving things as they are. The clean up step in the middle is a strong signal that this one job is doing too much. (That step is actually not needed, but it makes the job easier to reason about because the reader need not verify that files with `.txt` suffixes are never used after that point.) But using fewer jobs also makes things run slightly faster.
- If we keep the current approach of replacing the individual checksum files with one file, then we don't need to download the asset list as JSON and then operate on it, because we have one checksum per archive and we can just check for feature consistency in the names of those downloaded files.
- If we don't keep that approach, then the code that replaces the individual checksum files with one file can be removed, and the rest left in place.

> #### 6. Document and refactor to avoid confusion
> I just read through the new file start to end (without using the diff-view), and can confirm that it makes a clean and much simpler impression, with all steps clearly marked. Great work!

Thanks--this is tricky to be sure of, and your appraisal increases my confidence that the stylistic changes have helped, or at least not hurt.

> There was the `ubuntu-latest-disabled` comparison that explained itself with a comment, but since I read it after the fact I was initially confused about the purpose. My expectation would have been that it's commented out, but that comes with its own issues so probably doing it like this is overall better. My initially confusion is due to not being used to it, and I am happy to learn and adapt.

This was actually there before, with the same `ubuntu-latest-disabled` check (which was added in 28d2cf5). The only change I made to that step was to try and improve its readability in f7ee365.

I didn't do anything further with it in this PR, because it is something I hope to remove altogether soon, when working on #1477 and adding more Linux targets and features in the spirit of #1242. I don't think the packages listed there would be sufficient to do everything we want to do with musl. I'm also not sure *any* such command could be sufficient, since `cross` is being used for all Linux builds, and headers and build tools need to be inside or otherwise accessible from the Docker containers it uses.

I tested `max` locally in an Alpine Linux 3.17 x86-64 system (since Alpine uses musl instead of glibc) to investigate [what those comments](https://github.com/Byron/gitoxide/blob/3604a3b519d76db367084ca49d395f6ec4094cf7/.github/workflows/release.yml#L148-L150) as well as [the related ones on the feature exclusions](https://github.com/Byron/gitoxide/blob/3604a3b519d76db367084ca49d395f6ec4094cf7/.github/workflows/release.yml#L113-L115) seem to indicate are the blockers:

- I was able to produce what I think are the problems referenced there when attempting to use OpenSSL, by building with the `openssl` and `openssl-dev` packages installed and no LibreSSL-related packages installed. [The linker reported](https://gist.github.com/EliahKagan/ba1563931c12a37fd2aa318e2bea48b4#file-openssl-build-txt-L256-L269) `cannot find -lssl` and `cannot find -lcrypto`. Maybe there are other packages that would rectify this while still using OpenSSL with musl, though the `libssl3` and `ssl_client` packages were also (already) installed.

- But when I uninstalled `openssl` and `openssl-dev` and installed `libressl-dev`--which also pulled in `libressl3.6-libcrypto`, `libressl3.6-libssl`, and `libressl3.6-libtls` as dependencies--I was able to build with no errors. Both 0.37.0 and the current tip of the main branch build with no problem.

  On main, I ran all tests with `cargo nextest run --all --no-fail-fast`. Without `bash`, there were over 100 failures, since some of the tests make use of it, but this is similar to what happens on a FreeBSD system without `bash` and is not related to the available libraries. With `bash` installed, only one test failed on Alpine Linux, [`from_path_no_follow`](https://gist.github.com/EliahKagan/86f3770f185fa66c40652b5196d750cd#file-output-txt-L2795). This happens both before and after the changes in [#1488](https://github.com/Byron/gitoxide/pull/1488) and is due to the BusyBox `touch` command not accepting that syntax (I think it doesn't accept nanoseconds).

  On Ubuntu 22.04 LTS, running `busybox touch -d "2446-05-10 22:38:55.111111111" future` produces the same `invalid date` error, so the problem relates to BusyBox vs. GNU coreutils but does not relate to musl vs glibc. The `touch` invocation is in a fixture whose generated archive is explicitly `.gitignore`d and that is therefore run on any system. The code under test is not itself invoking `touch` and does not itself seem to have any problems.

- As requested in [#1242](https://github.com/Byron/gitoxide/issues/1242), a third option would be to use neither library by having gitoxide use rustls instead.

In view of that, it seems to me that it may be possible to make and publish musl builds for all features that we are building other targets for, without changing any code of `gitoxide` or even its dependencies, by installing LibreSSL libraries and headers in the environment `cross` uses. I would want check if such builds work on a variety of systems, though. If not, then it may be necessary to change some configuration to link LibreSSL statically.

If this problem can be solved soon--even if not by the ultimately solution which is probably rustls--then I think further changes to the step with the `ubuntu-latest-disabled` comparison could wait until then.

NobodyXu:
For the signature part, I'd recommend to use minisign as cargo-binstall can use it to verify identity of the publisher.

The schema I recommend is to generate a new key-pair during publishing, write the private key into `Cargo.toml`, and then publish it.

For the private key, it would be used to sign the binary artifacts, and then be destroyed.

In cargo-binstall release PR, we achieve this by setting up another key-pair, use it to encrypt the temporary private key, and upload it as artifacts.

Once all package building/signature is done, we delete the private key.

Byron:
Thanks for the suggestion, @NobodyXu , I like the idea of huddling around a standard that is already supported by a tool tasked with installing software, it's a great match.

There is certainly complexities involved due to the key handling, but I'd hope it's exclusively during artifact production in this workflow. Otherwise, `cargo smart-release` would have to learn it as it is tasked with publishing, and it seems like a lot can go wrong especially *if* a private key for a published version has to be re-used for artifact creation. Admittedly, I don't claim I understand the workflow and I didn't look it up.

So all in all, maybe aligning around how influential projects do it might be more suitable as first step.

----

> There are ways to bring in static type checking and even a different syntax. I don't know how much that would help or if it would be worthwhile. I'll try and open a discussion question about it soon.

The IDE can certainly provide more support, but I don't know if this is anything that can be improved by changing the file. I suppose it's more about which plugins are installed. Probably I am missing something, apologies.


> If this problem can be solved soon--even if not by the ultimately solution which is probably rustls--then I think further changes to the step with the `ubuntu-latest-disabled` comparison could wait until then.

It's exciting to hear that this could go away and the MUSL builds could work with the `max` setting, and I am looking forward to whichever changes come first. If `ubuntu-latest-disabled` goes away in the process, even better.

NobodyXu:
> There is certainly complexities involved due to the key handling, but I'd hope it's exclusively during artifact production in this workflow. 

It depends on how you generate the minisign key-pair.

If you generate it once and stored the private key within repo secrets and write the public key to `Cargo.toml` and commit it, then yeah, it would only affect artifact production.

If you generate one per release like cargo-binstall does, then you'd have to change `cargo smart-release`:

It would have to generate a key-pair, write the public key into Cargo.toml before publishing, and the private key would be encrypted and uploaded to action artifact for signing (removed after building is completed).

We did this because crates.io is guaranteed to be immutable, whereas github release can be modified at anytime.

By generating a unique public key for each release and putting it into `Cargo.toml`, even if someone breaks one of the release, other is still fine.

Byron:
Thanks for elaborating. I think for one, I think minisign is an upgrade to just providing hashes as it can potentially validate authorship, hashes can only validate the file one was supposed to download was actually downloaded.

Having a per-release key that is thrown away seems to invalidate the idea of proving authorship, as it's an ever-changing public key that anyone could have created. Maybe this is also where I am wrong or misunderstand things.

NobodyXu:
> Having a per-release key that is thrown away seems to invalidate the idea of proving authorship, as it's an ever-changing public key that anyone could have created. 

The idea is to minimize the risk of malicious takeover.

By having a key for each release, stored on crates.io, we guarantee that even if attacker takeover the account/repository, they can't modify the existing release artifacts without getting noticed.

Sure they can create new release, but that create a clear timeline on when that malicious takeover happens, user can simply ignore all releases come after that.

Byron:
Ah, right, and `crates.io` stores the hashes it has seen so `cargo-binstall` can double-check it from there.
This also sounds like any hash could be used to do the same, as along as a trusted third party stores what it has seen, but using cryptographic keys seems like it's more secure and versatile.

NobodyXu:
> Ah, right, and crates.io stores the hashes it has seen so cargo-binstall can double-check it from there.

Yes we already check that in cargo-binstall

> This also sounds like any hash could be used to do the same, as along as a trusted third party stores what it has seen, but using cryptographic keys seems like it's more secure and versatile.

We currently only support crypto asymmetric keys in the `Cargo.toml`.

Storing the key in github release makes no sense at all, since it can be overwritten by malicious takeover.

We also have plans to support more algorithms

cargo-bins/cargo-binstall#1

EliahKagan:
> > There are ways to bring in static type checking and even a different syntax. I don't know how much that would help or if it would be worthwhile. I'll try and open a discussion question about it soon.
> 
> The IDE can certainly provide more support, but I don't know if this is anything that can be improved by changing the file. I suppose it's more about which plugins are installed. Probably I am missing something, apologies.

Yes, for VS Code there is the [GitHub Actions extension](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-github-actions). But I'm suggesting something else: generating them from a DSL written in a statically typed language. This may be a hard sell because, as far as I can tell, currently no one has made anything where Rust would be that language. I've opened #1512 where I have described this in more detail.

> > If this problem can be solved soon--even if not by the ultimately solution which is probably rustls--then I think further changes to the step with the `ubuntu-latest-disabled` comparison could wait until then.
> 
> It's exciting to hear that this could go away and the MUSL builds could work with the `max` setting, and I am looking forward to whichever changes come first. If `ubuntu-latest-disabled` goes away in the process, even better.

Due to #1493, I am not sure this can be solved the way I had hoped. Since the step with `ubuntu-latest-disabled` does not likely resemble anything that would be used when this is fixed (because we are using `cross` for all Linux builds), I am inclined to go ahead and remove it soon anyway, possibly when making changes discussed in #1498.

Byron:
> Regarding signing releases as discussed above, is there anything I can or should do to help out with that?

I'd leave it as is, but go back to that once `gitoxide` is stable or plans to become stable. Then supply-chain safety would have an even higher value.

The BoM already interests me a lot though, and if that interests you as well maybe that could be investigated for the purpose of adding it to GitHub releases.

PR #1488, opened 2024-08-03
reproduce and fix date fuzz error

Byron:
Reproduce a fuzz failure.

Fortunately it's just an incorrect expectation.

### Tasks

* [x] reproduce
* [x] fix

### DIFF
--- a/gix-date/src/parse.rs
+++ b/gix-date/src/parse.rs
@@ -33,7 +33,9 @@ pub(crate) mod function {
         }
 
         Ok(if let Ok(val) = Date::strptime(SHORT.0, input) {
-            let val = val.to_zoned(TimeZone::UTC).expect("date is in range");
+            let val = val
+                .to_zoned(TimeZone::UTC)
+                .map_err(|_| Error::InvalidDateString { input: input.into() })?;
             Time::new(val.timestamp().as_second(), val.offset().seconds())
         } else if let Ok(val) = rfc2822_relaxed(input) {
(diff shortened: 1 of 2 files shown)


Byron:
I thought you'd find this interesting, @BurntSushi , as this is an expectation failure.

```
date is in range: Error { inner: ErrorInner { kind: Adhoc(error converting datetime -009999-01-01T00:00:00 to instant in time zone UTC), cause: Some(Error { inner: ErrorInner { kind: Range(RangeError(Negative { what: "second", given: 0, min: -377705023201, max: 253402207200 })), cause: None } }) } }
thread 'time::parse::fuzz::invalid_but_does_not_cause_panic' panicked at gix-date/src/parse.rs:36:51:
```

Probably that's just a mistake and the fix is as easy as it looks, but I'd be glad if you would let me know. Thank you.

BurntSushi:
Yup that is indeed an expected error! Nice find fuzzer. I think I used `expect()` here because that's what it did previously with `time`, but indeed, this should be an error. (`time` I believe supports a bigger range than Jiff does.)

For why this happens, Jiff guarantees that every `Timestamp` can be converted to a `civil::DateTime`, but sacrifices the opposite direction at the boundaries. The error cases in [this example](https://docs.rs/jiff/latest/jiff/civil/struct.DateTime.html#example-errors) might help make this clearer. (The example uses `DateTime::MAX` instead of `DateTime::MIN`, but it's the same idea.)

PR #1495, opened 2024-08-07
Split `release` profile into `release` and `release-opt`

### LINKED ISSUE #1477: Unclear intent regarding optimizations for some profiles

EliahKagan:
This makes the kind of changes discussed in comments in #1475 and further in #1477. This fixes #1477, though some other fix might be preferred, so this may benefit from being changed, or further changes could be made afterward. See https://github.com/Byron/gitoxide/issues/1477#issuecomment-2257499363 and https://github.com/Byron/gitoxide/issues/1477#issuecomment-2274445001.

In `Cargo.toml`, this adds thin LTO to the `release` profile, and creates a separate more aggressively optimized `release-opt` profile with fat LTO and max 1 codegen unit. The effect on build times seems in both cases modest, but you might disagree. For more on that, see [this gist where I timed the builds](https://gist.github.com/EliahKagan/52cedc99a8544505525706351948664a) and https://github.com/Byron/gitoxide/issues/1477#issuecomment-2274445001 (which, among other related topics, discusses it). This also removes some old commented-out code, as also discussed in #1477.

In `release.yml`, this takes the profile to use from an environment variable (some occurrences of `release` had referred to that profile while others had not), does some related other refactoring, and most significantly changes which profile is used for the GitHub releases from `release` to the new `release-opt` profile.

The results [seem okay](https://gist.github.com/EliahKagan/a8a59e0f6fb4435dcbd9555b3c8b8680), and in particular the Linux executables, which without configuring stripping in `Cargo.toml` might require more explicit logic as new targets are added, show as stripped. (The `file` implementation I am using does not seem to report the absence of symbols in macOS binaries; that affected prior runs' results as well.) That output is on assets downloaded from a draft release produced in [this workflow run](https://github.com/EliahKagan/gitoxide/actions/runs/10293234759).

In case you decide to merge this without many changes but hope for an even better approach later, I've [included a comment](https://github.com/Byron/gitoxide/pull/1495/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R216-R217) in `Cargo.toml` stating that the `release-opt` profile is not guaranteed to stick around.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -139,6 +139,7 @@ jobs:
       TARGET: ${{ matrix.target }}
       TARGET_FLAGS: --target=${{ matrix.target }}
       TARGET_DIR: target/${{ matrix.target }}
+      PROFILE: release-opt
 
     steps:
       - name: Checkout repository
@@ -171,23 +172,9 @@ jobs:
           echo "target flag is: $TARGET_FLAGS"
           echo "target dir is: $TARGET_DIR"
 
-      - name: Build release binary
+      - name: Build release binary (with extra optimizations)
         run: |
-          "$CARGO" build --verbose --release "$TARGET_FLAGS" --no-default-features --features "$FEATURE"
-
-      - name: Strip release binary (x86-64 Linux, and all macOS)
-        if: matrix.target == 'x86_64-unknown-linux-musl' || matrix.os == 'macos-latest'
-        run: strip "$TARGET_DIR"/release/{ein,gix}
-
-      - name: Strip release binary (ARM Linux)
-        if: matrix.target == 'arm-unknown-linux-gnueabihf'
-        run: |
-          docker run --rm -v \
-            "$PWD/target:/target:Z" \
-            rustembedded/cross:arm-unknown-linux-gnueabihf \
-            arm-linux-gnueabihf-strip \
-            /target/arm-unknown-linux-gnueabihf/release/ein \
-            /target/arm-unknown-linux-gnueabihf/release/gix
+          "$CARGO" build --verbose --profile="$PROFILE" "$TARGET_FLAGS" --no-default-features --features="$FEATURE"
 
       - name: Determine archive basename
         run: echo "ARCHIVE=gitoxide-$FEATURE-$VERSION-$TARGET" >> "$GITHUB_ENV"
@@ -200,8 +187,8 @@ jobs:
       - name: Build archive (Windows)
         if: matrix.os == 'windows-latest'
         run: |
-          file -- "$TARGET_DIR"/release/{ein,gix}.exe
-          cp -- "$TARGET_DIR"/release/{ein,gix}.exe "$ARCHIVE/"
+          file -- "$TARGET_DIR/$PROFILE"/{ein,gix}.exe
+          cp -- "$TARGET_DIR/$PROFILE"/{ein,gix}.exe "$ARCHIVE/"
           7z a "$ARCHIVE.zip" "$ARCHIVE"
           /usr/bin/core_perl/shasum --algorithm=256 --binary "$ARCHIVE.zip" > "$ARCHIVE.zip.sha256"
           echo "ASSET=$ARCHIVE.zip" >> "$GITHUB_ENV"
@@ -210,8 +197,8 @@ jobs:
       - name: Build archive (Unix)
         if: matrix.os != 'windows-latest'
         run: |
-          file -- "$TARGET_DIR"/release/{ein,gix}
-          cp -- "$TARGET_DIR"/release/{ein,gix} "$ARCHIVE/"
+          file -- "$TARGET_DIR/$PROFILE"/{ein,gix}
+          cp -- "$TARGET_DIR/$PROFILE"/{ein,gix} "$ARCHIVE/"
           tar czf "$ARCHIVE.tar.gz" "$ARCHIVE"
           shasum --algorithm=256 --binary "$ARCHIVE.tar.gz" > "$ARCHIVE.tar.gz.sha256"
           echo "ASSET=$ARCHIVE.tar.gz" >> "$GITHUB_ENV"
@@ -346,7 +333,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Publish the release
+      - name: Publish the release  # FIXME: Reenable this.
+        if: false
         run: gh release --repo="$REPOSITORY" edit "$VERSION" --draft=false
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -139,7 +139,7 @@ jobs:
       TARGET: ${{ matrix.target }}
       TARGET_FLAGS: --target=${{ matrix.target }}
       TARGET_DIR: target/${{ matrix.target }}
-      PROFILE: release-opt
+      PROFILE: release-github
 
     steps:
       - name: Checkout repository
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -333,8 +333,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Publish the release  # FIXME: Reenable this.
-        if: false
+      - name: Publish the release
         run: gh release --repo="$REPOSITORY" edit "$VERSION" --draft=false
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -213,6 +213,8 @@ panic = "unwind"
 incremental = false
 build-override = { opt-level = 0 }
 
+# This profile is currently used in building releases for GitHub.
+# It may be removed at any time and should not otherwise be relied on.
 [profile.release-opt]
 inherits = "release"
 lto = "fat"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -206,17 +206,18 @@ sha1_smol = { opt-level = 3 }
 
 [profile.release]
 overflow-checks = false
-#lto = "fat"
-# this bloats files but assures destructors are called, important for tempfiles. One day I hope we
+lto = "thin"
+# This bloats files but assures destructors are called, important for tempfiles. One day I hope we
 # can wire up the 'abrt' signal handler so tempfiles will be removed in case of panics.
-panic = 'unwind'
-#codegen-units = 1
+panic = "unwind"
 incremental = false
 build-override = { opt-level = 0 }
 
-# It's not quite worth building dependencies with more optimizations yet. Let's keep it here for later.
-#[profile.dev.package."*"]
-#opt-level = 2
+[profile.release-opt]
+inherits = "release"
+lto = "fat"
+codegen-units = 1
+strip = "symbols"
 
 [workspace]
 members = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -211,15 +211,17 @@ lto = "thin"
 # can wire up the 'abrt' signal handler so tempfiles will be removed in case of panics.
 panic = "unwind"
 incremental = false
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -346,7 +333,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Publish the release
+      - name: Publish the release  # FIXME: Reenable this.
+        if: false
```
EliahKagan, line 337: Before merging, actual publishing should be reenabled:

```suggestion
      - name: Publish the release
```

I have no objection to this being merged at any time with that change, but as noted above I suspect you may want other changes, so for now this is still marked as a draft.

EliahKagan: Obsolete,  871eec7 includes this.

Cargo.toml
```diff
@@ -206,17 +206,20 @@ sha1_smol = { opt-level = 3 }
 
 [profile.release]
 overflow-checks = false
-#lto = "fat"
-# this bloats files but assures destructors are called, important for tempfiles. One day I hope we
+lto = "thin"
```
EliahKagan, line 209: It seems to me that `lto = "thin"` here in the `release` profile is reasonable, but as noted above it does slow down builds a bit. In case you don't want it, you can accept this suggestion:

```suggestion
```

This is for convenience if that is the case, but not to advocate this change.

It might make sense to wait until benchmarking is done before adding this, though I believe it usually improves runtime performance--unlike going from thin to fat LTO, which is not always the case--and decisively determining that it is not worthwhile may involve benchmarking on quite a wide variety of architectures.

(This change would not affect the `release-opt` profile, because while that inherits from this, it has its own value here, setting it to `fat`.)

Byron (approved):
Thanks a lot for proposing the change and for acquiring numbers!
I think going to LTO="thin" should be fine and maybe it's good for a speedboost at some point, even though maybe one day we might go back to the previous setting if it turns out not to be the case.
Out of ~~caution~~ fear I also removed the opt-level=0 statement to rather leave it at the default.

Finally, I renamed the profile to clearly state what it is for, `-github`, which I think should make it less likely to be used by others - `-opt` sounded general enough for usage by others who want an even more optimized build.

Please do double-check my changes in case they have a chance to break something, after all I didn't really test any of this.

With the current settings, on my M1 Pro a release build can be done in 58s, which seems fine. A `release-github` build takes 2m30s, for comparison.

As a side-note: The `inherits` directive is required even though I tried not to use it, so instead I repeated all values from the `release` profile to assure they are not actually inherited, but rather, controlled explicitly.

Cargo.toml
```diff
@@ -211,15 +211,17 @@ lto = "thin"
 # can wire up the 'abrt' signal handler so tempfiles will be removed in case of panics.
 panic = "unwind"
 incremental = false
-build-override = { opt-level = 0 }
```
EliahKagan, line 215: I'm not sure I understand the reason for removing this based on the information in the commit message, though I can guess a possible good reason. This was added in add2e3e (#2) and only reduces the optimization level for `build` dependencies, i.e., those that are used for building. So the resulting executables' level of optimization is unaffected by this. But a note in [the documentation](https://doc.rust-lang.org/cargo/reference/profiles.html#overrides) mentions:

> When a dependency is both a normal dependency and a build dependency, Cargo will try to only build it once when `--target` is not specified. When using `build-override`, the dependency may need to be built twice, once as a normal dependency and once with the overridden build settings. This may increase initial build times.

Is this the reason for removing it?

Cargo.toml
```diff
@@ -211,15 +211,17 @@ lto = "thin"
...
 # It may be removed at any time and should not otherwise be relied on.
-[profile.release-opt]
+[profile.release-github]
 inherits = "release"
+overflow-checks = false
+panic = "unwind"
 lto = "fat"
 codegen-units = 1
 strip = "symbols"
+build-override = { opt-level = 3 }
```
EliahKagan, line 224: Assuming `build-override` is no longer to be set in the `release` profile from which this one inherits, isn't this superfluous? In the absence of `build-override`, I think build dependencies have the same `opt-level` as the profile, and since that is `release` here, we already have `opt-level = 3`.

You mentioned:

> As a side-note: The `inherits` directive is required even though I tried not to use it, so instead I repeated all values from the `release` profile to assure they are not actually inherited, but rather, controlled explicitly.

But only the values changed in `Cargo.toml` were repeated, except this, which is explicit here but implicit in the `release` profile. (`opt-level` for the profiles themselves is implicit in both.)

EliahKagan (commented):
> Out of ~~caution~~ fear I also removed the opt-level=0 statement to rather leave it at the default.

I've requested clarification about this in my [first review comment](https://github.com/Byron/gitoxide/pull/1495/files#r1708741138) below. The presence of `build-override` with a lower `opt-level` can affect build times, usually decreasing them unless it causes the same crates to be built more than once and otherwise sometimes increasing them. It should not affect the runtime performance of the `gix` and `ein` binaries, regardless of whether and how it is set.

> Finally, I renamed the profile to clearly state what it is for

Good idea!

> Please do double-check my changes in case they have a chance to break something, after all I didn't really test any of this.

I haven't rerun the workflow yet, but on inspection it doesn't seem like anything would be broken.

> As a side-note: The `inherits` directive *[...}*

This relates to my [second review comment](https://github.com/Byron/gitoxide/pull/1495/files#r1708752857) below. I'm not sure the current amount of repetition is clearer, since it is only repeating what `release` overrides explicitly from the `cargo` defaults--quite a bit remains implicit in both profile definitions.

It doesn't seem particularly unclear though, except with respect to the intent of one change that the second comment concerns.

Cargo.toml
Byron: I completely misunderstood `build-override` and was thinking it's related to an opt-level override for anything that is built, not just build scripts.

Avoiding to possibly build dependencies twice It's probably a good reason to remove it, especially given that there are no build dependencies with a build-script that just uses the standard library.

Cargo.toml
Byron: With my recent gain of knowledge, I think this line should be removed. This was done in `main` now.

EliahKagan:
> > Please do double-check my changes in case they have a chance to break something, after all I didn't really test any of this.
>
> I haven't rerun the workflow yet, but on inspection it doesn't seem like anything would be broken.

I have actually [tested this](https://github.com/EliahKagan/gitoxide/actions/runs/10345803160) now, and [it seems fine](https://gist.github.com/EliahKagan/12398ba669c19cf1f94444969d0d55f0).

PR #1506, opened 2024-08-10
Don't show console on Windows

sheremetyev:
When used inside GUI application, gix causes console window to appear temporarily, due to execution of git.

Command needs special flags to prevent that. More details: https://stackoverflow.com/a/60958956


Byron (approved):
Thanks a lot, this is a much appreciated fix!

There is also the `gix-command` crate which creates a `std::process::Command` instance to create git-style invocations, and it looks like it also doesn't turn this 'feature' off - I changed this now and also will assure that `gix-command` is used everywhere.

PR #1510, opened 2024-08-11
gix-testtools: Upgrade versions of gix crates

joshtriplett:
This avoids duplicate dependencies in packages depending on both gix and
gix-testtools.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1675,22 +1675,6 @@ dependencies = [
 
 [[package]]
 name = "gix-discover"
-version = "0.32.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fc27c699b63da66b50d50c00668bc0b7e90c3a382ef302865e891559935f3dbf"
-dependencies = [
- "bstr",
- "dunce",
- "gix-fs 0.11.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "gix-hash 0.14.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "gix-path 0.10.9 (registry+https://github.com/rust-lang/crates.io-index)",
- "gix-ref 0.44.1",
- "gix-sec 0.10.7 (registry+https://github.com/rust-lang/crates.io-index)",
- "thiserror",
-]
-
-[[package]]
-name = "gix-discover"
 version = "0.33.0"
 dependencies = [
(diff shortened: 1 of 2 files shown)


joshtriplett:
(Would it be possible to get a gix-testtools release based on this?)

Byron:
Actually, I don't think I can do that as it will break the ability of `cargo smart-release` to perform a release of `gitoxide`. Publishes fail if the test-tools use dependencies in the workspace, and using a separate set of crate versions (that also have to be one major version in the past, ideally) was the only workaround I could find. *(`gix-testtools` seems to be special as it's used as dev-dependency in crates that are also depend on it, it's a cycle.)*

I'd also love to not have this special-case, but it will probably need more time than I can invest into `cargo smart-release` to find a proper fix for it - it's a complex tool that was rewritten once already, yet is totally under-tested. My hopes are that one day, `cargo smart-release` will only be used to set the crate versions update changelogs, while `cargo publish` can be used to publish everything in the right order.

Byron:
As this PR is blocked by technicalities and won't budge for that reason, I am closing it as the issue is likely to persist until `cargo smart-release` has been adapted to handle this kind of cyclic dependency correctly.

To re-state the problem:

* `gix-*` crates use `gix-testtools = { path = "…" }` as `[dev-dependency]`.
* if `gix-testtools` also uses `gix-* = { path = "…", version = "…" }` declarations then the operation.
    - The problem probably isn't `cargo publish`, but `cargo smart-release` which incorrectly edits the manifests when adjusting crate versions.

The workaround for the problem as it's employed now is to let `gix-testtools` refer to `gix-*` crates which have been published before, *and* are at least a major version in the past. For some reason, it also doesn't work when it refers to `crates.io` releases of `gix` crates that are compatible with the respective workspace crates.

Now that I am thinking about it, another solution would be to completely remove all `gix-*` dependencies from `gix-testtools`. But after evaluating this option, I had to conclude that it's not possible as it relies on the signal-aware `gix-tempfile` implementation to assure its tempfiles are always removed.

In theory, `gix-tempfile` and `gix-lock` are stable, but in practice they have to get republished due to shortcomings in `cargo smart-release` - it currently simply can't keep major versions stable, an issue that will need to be fixed before stabilising additional crates.

Thus it's probably better to wait until `cargo smart-release` gets some additional development time.

PR #1513, opened 2024-08-12
Use the stable Rust toolchain to build all targets

EliahKagan:
This uses stable rather than nightly Rust toolchains to build all binaries for GitHub releases in the `release.yml` workflow, as discussed in #1497. Prior to this, some of them already used stable toolchains, while others used nightly toolchains. No targets are changed here, only toolchains (and not the target triple part).

[Here is workflow run](https://github.com/EliahKagan/gitoxide/actions/runs/10345885471) with modified workflow, and [here's an examination of generated artifacts](https://gist.github.com/EliahKagan/d67785c5c1ef6ffc5f757f03707af2e0) from it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -86,30 +86,24 @@ jobs:
         # When changing these features, make the same change in build-macos-universal2-release.
         feature: [ small, lean, max, max-pure ]
         include:
+          - rust: stable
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thank you, that looks like a great improvement!

PR #1518, opened 2024-08-12
Drop `once_cell`

spenserblack:
The ecosystem seems to be moving away from using the `once_cell` crate towards using the types available in the standard library. In particular, `once_cell::sync::Lazy` is getting replaced by `std::sync::LazyLock`, which is mostly a drop-in replacement.

As `LazyLock` was stabilized in 1.80, this is a drastic increase in the MSRV, so this probably shouldn't be merged for a while. I'm not sure what the MSRV policy is, but I haven't really searched too deeply for it, yet. X versions back? X months back?

Leaving this in a draft status until:

- [ ] The new MSRV is no longer "too new" (see [helix](https://github.com/helix-editor/helix))
- [ ] `std::cell::OnceCell::get_or_try_init` is stabilized


Byron:
Thanks for the initiative, it's much appreciated!

It seems a bit early, especially since you have to keep maintaining this PR for however long it takes until the MSRV can be 1.80, but I suppose this is acceptable to you.

My recommendation is to track the MSRV of `helix`, which in turn is tracked by `gitoxide`.

spenserblack:
> My recommendation is to track the MSRV of `helix`, which in turn is tracked by `gitoxide`.

Does this mean that, as [helix](https://github.com/helix-editor/helix)'s MSRV is updated, `gitoxide` will respond by updating the MSRV?

Byron:
Yes, that's usually what I do.

Byron:
Would you consider reopneing the PR once it has a chance to actually be merged?

spenserblack:
Sure! Up to you how you want to manage this project, but what I had in mind is that this could sit around as a "to do" that I'd check in with occasionally.

Byron:
I think that works similarly (from your side) on a closed PR. If CI is needed, it's possible to enable it on your fork, and it will run automatically for all branches named `run-ci/**` (see [here](https://github.com/Byron/gitoxide/blob/main/.github/workflows/ci.yml#L11)).

Right now there are a lot of mostly stale PRs and this one I think will be around for quite a few months until 1.80 is the new baseline MSRV.

spenserblack:
No problem, I'll check in a while later. Unless someone else remakes this PR before I get back to it :laughing:

PR #1520, opened 2024-08-13
various improvements

Byron:
Various API changes to make diffing between trees and stats-retrieval less cumbersome.
Additionally, add more methods to make API more similar to `git2` where it's clearly for convenience.

Related to https://github.com/gitbutlerapp/gitbutler/pull/4670 .

### Tasks

* [x] more convenience methods
* [x] more convenience around tree diffing and stats


Byron:
@Jake-Shadle Would you be interested in adding `libz-ng` sans `cmake` to `flate2`? I have a feeling it's not straightforward though, but having it there under an equally unstable feature toggle would definitely help spread the goodness, I think.

Jake-Shadle:
Why would it not be straightforward?

Byron:
`flate2` uses the `libz-ng-sys` crate, which is a separate publish without the special feature. Probably there is a good reason for it. Then it might be required to work with all features still, so as always with cargo features, it's probably more finicky than one anticipates.

Your optimism is encouraging though.

PR #1521, opened 2024-08-13
Fix dir name tracking in the FileName mode

mvlabat:
The PR fixes directory name tracking in the `FileName` mode. Previously, `Change::location` would contain a name of a previous non-tree entry, instead of a corresponding directory name.

I'm not sure why existing tests where we compare newly added objects against an empty tree didn't catch this, so I added a new one. For that I had to edit fixtures a bit so that changes touch a tree as well, hence all the commit hash edits.


Byron (approved):
Thanks a lot, that's a great catch! And even better that there now is a test for that!

PR #1523, opened 2024-08-15
run `git config -l` in temp dir when looking up system config

martinvonz:
Commit a9a3545db2b2 made `gix::config::File::from_globals()` try to find the system config (typically `/etc/config`) by subprocessing to `git config -l --show-origin`. That command takes about 500 ms when run in certain directories in our VFS at Google. Since we don't need anything but the system config, let's run the command in the system's temporary directory instead.

I also added `--system` for good measure. I've confirmed that just adding `--system` is not enough, however - `git config -l --system --show-origin` is still slow in our VFS.

### DIFF
--- a/gix-path/src/env/git/mod.rs
+++ b/gix-path/src/env/git/mod.rs
@@ -1,3 +1,4 @@
+use std::env;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 
@@ -88,7 +89,8 @@ pub(super) static EXE_INFO: Lazy<Option<BString>> = Lazy::new(|| {
             const CREATE_NO_WINDOW: u32 = 0x08000000;
             cmd.creation_flags(CREATE_NO_WINDOW);
         }
-        cmd.args(["config", "-l", "--show-origin"])
+        cmd.args(["config", "-l", "--system", "--show-origin"])
+            .current_dir(env::temp_dir())
             .stdin(Stdio::null())
             .stderr(Stdio::null());
         cmd
--- a/gix-path/src/env/git/mod.rs
+++ b/gix-path/src/env/git/mod.rs
@@ -89,7 +89,7 @@ pub(super) static EXE_INFO: Lazy<Option<BString>> = Lazy::new(|| {
             const CREATE_NO_WINDOW: u32 = 0x08000000;
             cmd.creation_flags(CREATE_NO_WINDOW);
         }
-        cmd.args(["config", "-l", "--system", "--show-origin"])
+        cmd.args(["config", "-l", "--show-origin"])
             .current_dir(env::temp_dir())
             .stdin(Stdio::null())
             .stderr(Stdio::null());


Byron (dismissed):
Thanks a lot, a much appreciated change and: "Oh wow, this code runs at Google now!"

With that said, since `--system` doesn't make a performance difference and fails on MacOS (see below), I will have to remove that before merging.

----

```
❯ which git
/usr/bin/git

gitbutler ( missing-project) [$] via 
❯ git --version
git version 2.39.3 (Apple Git-146)

gitbutler ( missing-project) [$] via 
❯ git config -l --system --show-origin
fatal: unable to read config file '/etc/gitconfig': No such file or directory
```

EliahKagan:
The nature of the macOS failure is interesting:

```text
~> git config -l --show-origin --show-scope
unknown file:/Library/Developer/CommandLineTools/usr/share/git-core/gitconfig   credential.helper=osxkeychain
unknown file:/Library/Developer/CommandLineTools/usr/share/git-core/gitconfig   init.defaultbranch=main
global  file:/Users/ek/.gitconfig       user.name=Eliah Kagan
global  file:/Users/ek/.gitconfig       user.email=degeneracypressure@gmail.com
global  file:/Users/ek/.gitconfig       core.precomposeunicode=false
```

It has the usual value of `/etc/gitconfig` as the path of its system configuration file, which usually does not exist, and it has its installation configuration in a separate file (whose scope is shown as 'unknown`).

martinvonz:
> With that said, since `--system` doesn't make a performance difference and fails on MacOS (see below), I will have to remove that before merging.

It works on my Mac `git config -l --system` but that seems to be because we have a custom `git` at Google.

Actually, `--system` was not what I thought it was anyway, so I'm glad you're removing that! :) On my Linux machine, `--system` is `/usr/share/git-core/config`, while the first ("installation") path in `git config -l --show-origin`  is  `/etc/gitconfig`.

Byron:
Sorry for the early merge - apparently the "Require status checks before merging" option does not like I thought require all checks to pass (what I want), but is a no-op unless a check is chosen. Now I would have to select all checks and keep them in sync with what's actually there, which seems inconvenient at best.

Screenshot 2024-08-15 at 17 27 38

And I so wanted to have auto-merge 😭.

NobodyXu:
@Byron you can add a dummy job "test-pass":

https://github.com/cargo-bins/cargo-binstall/blob/c4abcb90a5fac18f5f8744cbf7bd11b00a37ca41/.github/workflows/ci.yml#L372

EliahKagan:
This pull request, in addition to improving performance, fixed the vulnerability **CVE-2024-45305** ([RUSTSEC-2024-0367](https://rustsec.org/advisories/RUSTSEC-2024-0367.html), https://github.com/Byron/gitoxide/security/advisories/GHSA-v26r-4c9c-h3j6).

(It may make sense to edit the PR description to mention this, though either way should be fine because this comment can itself be found by searching. Further related hardening, as well as a fix for a separate vulnerability, is included in #1567, but the fix for CVE-2024-45305 is here, and the changes there build on this.)

PR #1528, opened 2024-08-19
Improve path join test assertion messages and fix minor test bugs

EliahKagan:
This fixes inaccurate assertion messages in the path-joining tests, improves some messages that were okay but capable of being clarified, and renames a couple of variables to avoid making it seem like relative paths are absolute. Most of the changes are to the Windows tests. This relates to https://github.com/Byron/gitoxide/pull/1374#pullrequestreview-2244385346.

It also corrects a few tests where the same thing was asserted multiple times with a different message rather than asserting the subtly different claim that the message showed was intended, or where it tested paths with a different number of backslashes than the context and message show were meant.

Finally, though this is only a refactoring, it changes all string literals with backslash-containing paths to be raw string literals. I think this makes it easier and also faster to understand what paths are being tested, and generally improves readability a bit.

(Sometimes it makes sense to write all Windows paths as raw string literals, even those without backslashes, but I didn't do that, because in this case I think it would produce an effect that, one way or another, would be confusing or at least feel strange, when comparing corresponding tests in the Windows and non-Windows build configurations.)

Several assertion messages had contained expressions of surprise at the effects. When rewording made this clear, I removed such wording. Elsewhere I kept it. In one place, I even added it: I do not know why concatenating `\` and `\\localhost` on Windows produces `\localhost`, and it seems like the intent had even been to assert that it produces `\\localhost` (because it characterized the result as being UNC). But it does produce `\localhost` for some reason. Hopefully the message there can be improved further, but unless more is known about that now, I think my characerization as "strangely, single-backslashed host" may still be an improvement.

Some more details are available in commit messages and, in what may be an easier form to work with, in review comments I've posted below on particular changes.

### DIFF
--- a/gix-fs/tests/stack/mod.rs
+++ b/gix-fs/tests/stack/mod.rs
@@ -35,24 +35,24 @@ fn p(s: &str) -> &Path {
 #[test]
 #[cfg(windows)]
 fn path_join_handling() {
-    let absolute = p("/absolute");
+    let looks_absolute = p("/absolute");
     assert!(
-        absolute.is_relative(),
+        looks_absolute.is_relative(),
         "on Windows, absolute Linux paths are considered relative (and relative to the current drive)"
     );
-    let bs_absolute = p("\\absolute");
+    let bs_looks_absolute = p("\\absolute");
     assert!(
-        bs_absolute.is_relative(),
+        bs_looks_absolute.is_relative(),
         "on Windows, strange single-backslash paths are relative (and relative to the current drive)"
     );
     assert_eq!(
-        p("relative").join(absolute),
-        absolute,
+        p("relative").join(looks_absolute),
+        looks_absolute,
         "relative + absolute = absolute - however, they kind of act like they are absolute in conjunction with relative base paths"
     );
     assert_eq!(
-        p("relative").join(bs_absolute),
-        bs_absolute,
+        p("relative").join(bs_looks_absolute),
+        bs_looks_absolute,
         "relative + absolute = absolute - backslashes aren't special here, and it just acts like it's absolute"
     );
 
@@ -68,22 +68,22 @@ fn path_join_handling() {
     );
 
     assert_eq!(
-        p("\\\\?\\base").join(absolute),
+        p("\\\\?\\base").join(looks_absolute),
         p("\\\\?\\base\\absolute"),
         "absolute1 + unix-absolute2 = joined result with backslash"
     );
     assert_eq!(
-        p("\\\\.\\base").join(absolute),
+        p("\\\\.\\base").join(looks_absolute),
         p("\\\\.\\base\\absolute"),
         "absolute1 + absolute2 = joined result with backslash (device namespace)"
     );
     assert_eq!(
-        p("\\\\?\\base").join(bs_absolute),
+        p("\\\\?\\base").join(bs_looks_absolute),
         p("\\\\?\\base\\absolute"),
         "absolute1 + absolute2 = joined result"
     );
     assert_eq!(
-        p("\\\\.\\base").join(bs_absolute),
+        p("\\\\.\\base").join(bs_looks_absolute),
         p("\\\\.\\base\\absolute"),
         "absolute1 + absolute2 = joined result (device namespace)"
     );
--- a/gix-fs/tests/stack/mod.rs
+++ b/gix-fs/tests/stack/mod.rs
@@ -38,7 +38,7 @@ fn path_join_handling() {
     let looks_absolute = p("/absolute");
     assert!(
         looks_absolute.is_relative(),
-        "on Windows, absolute Linux paths are considered relative (and relative to the current drive)"
+        "on Windows, 'absolute' Linux paths are relative (and relative to the current drive)"
     );
     let bs_looks_absolute = p("\\absolute");
     assert!(
@@ -48,12 +48,12 @@ fn path_join_handling() {
     assert_eq!(
         p("relative").join(looks_absolute),
         looks_absolute,
-        "relative + absolute = absolute - however, they kind of act like they are absolute in conjunction with relative base paths"
+        "relative + unix-absolute = unix-absolute - the relative path without a drive is replaced"
     );
     assert_eq!(
         p("relative").join(bs_looks_absolute),
         bs_looks_absolute,
-        "relative + absolute = absolute - backslashes aren't special here, and it just acts like it's absolute"
+        "relative + unix-absolute = unix-absolute - the relative path without a drive is replaced - backslashes aren't special here"
     );
 
     assert_eq!(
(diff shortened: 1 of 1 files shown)


gix-fs/tests/stack/mod.rs
```diff
@@ -35,97 +35,97 @@ fn p(s: &str) -> &Path {
...
+        p("relative").join(bs_looks_absolute),
+        bs_looks_absolute,
+        "relative + unix-absolute = unix-absolute - the relative path without a drive is replaced - backslashes aren't special here"
     );
 
     assert_eq!(
         p("c:").join("relative"),
         p("c:relative"),
-        "drive + relative = strange joined result with missing backslash, but it's a valid path that works just like `c:\relative`"
+        "drive + relative = relative to the drive-specific current directory"
```
EliahKagan, line 62: This change relates to https://github.com/Byron/gitoxide/pull/1374#discussion_r1721045107.

gix-fs/tests/stack/mod.rs
```diff
@@ -35,97 +35,97 @@ fn p(s: &str) -> &Path {
...
+        p(r"\\.\base\absolute"),
         "absolute1 + absolute2 = joined result (device namespace)"
     );
 
     assert_eq!(p("/").join("C:"), p("C:"), "unix-absolute + win-drive = win-drive");
     assert_eq!(
         p("d:/").join("C:"),
         p("C:"),
-        "d-drive + c-drive = c-drive - interesting, as C: is supposed to be relative"
+        "d-drive + c-drive-relative = c-drive-relative - C: is relative but not on D:"
```
EliahKagan, line 95: This change relates to https://github.com/Byron/gitoxide/pull/1374#discussion_r1721055278.

gix-fs/tests/stack/mod.rs
```diff
@@ -35,97 +35,97 @@ fn p(s: &str) -> &Path {
...
         "unix-absolute + win-drive = win-drive, strangely enough it changed the trailing slash to backslash, so better not have trailing slashes"
     );
-    assert_eq!(p("/").join("C:\\"), p("C:\\"), "unix-absolute + win-drive = win-drive");
+    assert_eq!(p("/").join(r"C:\"), p(r"C:\"), "unix-absolute + win-drive = win-drive");
     assert_eq!(
-        p("\\\\.").join("C:"),
+        p(r"\\.").join("C:"),
         p("C:"),
-        "device-namespace-unc + win-drive-relative = win-drive-relative - c: was supposed to be relative, but it's not acting like it."
+        r"device-namespace-unc + win-drive-relative = win-drive-relative - C: as a relative path is not the C: device, so this is not \\.\C:"
```
EliahKagan, line 116: This change relates to https://github.com/Byron/gitoxide/pull/1374#discussion_r1721079186.

gix-fs/tests/stack/mod.rs
```diff
@@ -35,97 +35,97 @@ fn p(s: &str) -> &Path {
...
     );
     assert_eq!(p("relative").join("C:"), p("C:"), "relative + win-drive = win-drive");
 
     assert_eq!(
-        p("/").join("\\\\localhost"),
-        p("\\localhost"),
-        "unix-absolute + win-absolute-unc = win-absolute-unc"
+        p("/").join(r"\\localhost"),
+        p(r"\localhost"),
+        "unix-absolute + win-absolute-unc-host = strangely, single-backslashed host"
```
EliahKagan, line 123: This strange, and my puzzlement, relates to https://github.com/Byron/gitoxide/pull/1374#discussion_r1721079809.

EliahKagan (commented):
These identify specific comments in https://github.com/Byron/gitoxide/pull/1374#pullrequestreview-2244385346 to which these changes relate.

Byron (approved):
Thanks a lot for digging into this again and for the clarifications.

Indeed, these tests only exists to 'test' the `std::path::Path` implementation and get a feeling what happens in varying scenarios. Hopefully having written these down gives enough confidence that joining two relative paths won't accidentally create an absolute path, which could certainly be exploited.

EliahKagan:
> Thanks a lot for digging into this again and for the clarifications.

I just realized I  had forgotten to mention why I did not also add a clarifying comment for:

https://github.com/Byron/gitoxide/blob/c3f173c94968907ecc685c5383f2fe24c02dcfbf/gix-fs/tests/stack/mod.rs#L1

I had originally planned to comment that this was a workaround for https://github.com/rust-lang/rust-clippy/issues/12244. This would also add another source of information about how paths that start with `/` or `\` on Windows are not really absolute, except when they start with `\\`. Therefore I didn't further clarify that in any other ways, except through small adjustments to the custom assertion messages.

I didn't include this because the [`clippy::join_absolute_paths`](https://rust-lang.github.io/rust-clippy/master/index.html#/join_absolute_paths) suppression does not have an affect on that, though it might once have, in an earlier draft. So it would be wrong, or at least misleading, to say it is a workaround for that `clippy` issue.

On Windows, the leading-`/` and leading-`\` paths on Windows are placed in variables:

https://github.com/Byron/gitoxide/blob/c3f173c94968907ecc685c5383f2fe24c02dcfbf/gix-fs/tests/stack/mod.rs#L38

https://github.com/Byron/gitoxide/blob/c3f173c94968907ecc685c5383f2fe24c02dcfbf/gix-fs/tests/stack/mod.rs#L43

These variables were originally named `absolute` and `bs_absolute`, which was misleading since the paths are not absolute on Windows, and I wonder if the reason for putting them in variables had been that `clippy` diagnostic. But with the new variable names `looks_absolute` and `bs_looks_absolute`, I think it is clearer even than the use of literal strings for them.

I *think* that diagnostic is only for literals. Commenting out the suppression on Windows does not seem to give any new `cargo clippy` warnings. So at least with clippy 0.1.80, this seems to have an affect only on non-Windows systems, since in the non-Windows tests, some such literals are used (for comparison to the Windows tests).

Because the tests are supposed to speak for themselves, I am reluctant to add a big commented explanation about these issues to the file. But if I can think of something that would improve clarity without making the code more cumbersome to read or work with, then I'll open another PR.

> Indeed, these tests only exists to 'test' the `std::path::Path` implementation and get a feeling what happens in varying scenarios. Hopefully having written these down gives enough confidence that joining two relative paths won't accidentally create an absolute path, which could certainly be exploited.

Some relative paths would of course also be exploitable, such as if permitted paths could produce a relative path with a `..` component in it, or a more dotted component like `...` in it in a scenario where such a component performs upward traversal, or a relative path like `C:` or `C:foo` that specifies a drive, or a relative path like `/` or `/foo` that is from the root of a drive, or a path that contains a symlink component that is used in a way that traverses symlinks. I think we cover all that though.

Byron:
Thanks for sharing! Paths are…fascinating :).

> Because the tests are supposed to speak for themselves, I am reluctant to add a big commented explanation about these issues to the file. But if I can think of something that would improve clarity without making the code more cumbersome to read or work with, then I'll open another PR.

Since these tests are merely for documenting current behaviour, I think making it more obvious what they are testing (or what one should or could learn) is in their interest. Please feel free to open a PR with additional remarks, it's valuable to write it down while it's in active memory.

Issue #1524, opened 2024-08-15
copy detection differs from git weirdly


fowles:
### Current behavior 😯

```
cli/src/commands/{ => file}/chmod.rs
cli/{tests/test_cat_command.rs => src/commands/file/mod.rs}
cli/{tests/test_unsquash_command.rs => src/commands/file/print.rs}
cli/tests/{test_chmod_command.rs => test_file_chmod_command.rs}
cli/{src/commands/cat.rs => tests/test_file_print_command.rs}
```

### Expected behavior 🤔

```
cli/src/commands/{ => file}/chmod.rs
create mode 100644 cli/src/commands/file/mod.rs
cli/src/commands/{cat.rs => file/print.rs
cli/tests/{test_chmod_command.rs => test_file_chmod_command.rs}
cli/tests/{test_cat_command.rs => test_file_print_command.rs}
```

### Git behavior

Git does what I have listed under "expected behavior"

### Steps to reproduce 🕹

I am sorry that I don't have a super simple repro.  I only know how to cause this in a larger context.

1.  Install `jj` via `cargo install jj-cli`
2. `jj git clone --colocate https://github.com/martinvonz/jj`
3. `cd jj`
4. `cargo build`
5. `target/debug/jj diff -r 47bd6f4aa4a7eeef8b01ce168c6c771bdfffcbd3 --summary`
6.  `git diff 2de73f57fc9599602e001fc6331034749b2eacb0 47bd6f4aa4a7eeef8b01ce168c6c771bdfffcbd3 --summary`

The relevant code for how we use gix is https://github.com/martinvonz/jj/blob/95e8dd51ebfd61712d9fc5f6e19dd95d0026a004/lib/src/git_backend.rs#L1301

Byron:
Thanks for reporting! I'd think that it is possible to extract a test-case from the trees that show the issue.

Also, I am not surprised as Git implements a more complex algorithm to find the best possible matches. For that it keeps a list of candidates, which `gitoxide` does not currently do.

One cold hope that once `gitoxide` implements what Git does more closely, that it will get similar results.

For now I have no plans to improve this, but I should get there at some point. Contributions for this are very welcome, too.

fowles:
if you help with test reduction and point me to the code involved, I could take a crack at this, but I don't know the code base well, so something to give me a jump start would be appreciated

Byron:
[...]

PR #1529, opened 2024-08-19
tooling for better copy detection

Byron:
A PR to add tooling and a test created with it to faciliate changes to the copy-detection algorithm.

Related to #1524

### Tasks

* [x] royal-copy - degenerative copy from index with pathspecs
    - that away, the content of relevant files can be approximated and should still diff similarly
* [x] content-to-script - here-documents for all blobs of the listed trees
* [x] setup a fixture by manually creating indices of the respective tree, and commit them in order
* [x] create a tracked diff and assert on its renames (as they are now)

Documenting the current behaviour would allow changes to be expressed and then implemented.
Unfortuantely, for reasons of convenience, the test is on a higher level than the implementation has to be.


gix/tests/object/tree/diff.rs
```diff
@@ -378,6 +379,84 @@ mod track_rewrites {
         Ok(())
     }
 
+    #[test]
+    fn jj_realistic_needs_to_be_more_clever() -> crate::Result {
```
Byron, line 383: Due to the complexity of setting up the diffing in the tracker, I decided to add the test on `gix` level instead of on plumbing level, where everything is far more 'simulated'.

The benefit is that here one can define the final result exactly as one would want to see it, but with the downside that the implementation actually happens elsewhere, i.e. in the tracker, which has its own set of tests which should probably also be amended with boiled-down versions of this once the algorithm is adjusted.

One assume that that the algorithm changes will already change the outcome of the existing tests.

PR #1531, opened 2024-08-22
Fix typos in config support info

### LINKED ISSUE #1530: progress.rs misleadingly claims "we target 32bit systems only"
### Current behavior 😯

It's unclear why `core.packedGitLimit` is not applicable.

[`progress.rs`](https://github.com/Byron/gitoxide/blob/main/src/plumbing/progress.rs) gives the status of [**`core.packedGitLimit`**](https://git-scm.com/docs/git-config#Documentation/git-config.txt-corepackedGitLimit) as:

https://github.com/Byron/gitoxide/blob/242fedc973c56b6c1b6f150af99dda972a67f547/src/plumbing/progress.rs#L77-L80

This item was introduced in 0e9bd41 (#555).

I'm not clear on what "we target 32bit systems only" means here. It sounds like it is saying the code is not meant to be built and run on 64-bit targets, which is of course not the case--64-bit targets are more common, and this was already the case a couple of years ago.

### Expected behavior 🤔

Since gitoxide is most often used on 64-bit targets, `progress.rs` should not state "we target 32bit systems only."

Secondarily, the entry for `core.packedGitLimit` should more clearly explain why it isn't applicable. However, changing the string to just say "we don't use a windowing mechanism" would be sufficient to fix the main issue.

### Git behavior

Since the underlying context is documenting how and why gitoxide and Git differ with respect whether `core.packedGitLimit` is supported, the Git behavior, in documentation or otherwise, may not be directly relevant.

However, the Git documentation on `core.packedGitLimit` is linked above.

### Steps to reproduce 🕹

Examine `progress.rs` as linked and quoted above.

EliahKagan:
This fixes #1530 and makes some other minor improvements.

- `core.looseCompression` - use canonical capitalization for key
- `core.packedGitLimit` - work well on 64-bit, not 32-bit ([#1530](https://github.com/Byron/gitoxide/issues/1530))
- `core.sparseCheckoutCone` - fix misspelled `sparseCheckout` ref
- `submodule.recurse` - fix minor spelling typo

Further improvements are possible, such as to unify the style of the explanations so that they all follow the same capitalization and quoting conventions. I believe that should be done, but since I may inquire about other substantive areas soon anyway (as I did in #1530), I think that can wait until a subsequent PR.

This information is available by API, so maybe this should be a conventional commit? I can easily amend this so the commit message starts with `fix:` if you like. This seems like a gray area with respect to whether `fix:` ought to be used.

### DIFF
--- a/src/plumbing/progress.rs
+++ b/src/plumbing/progress.rs
@@ -76,7 +76,7 @@ static GIT_CONFIG: &[Record] = &[
     },
     Record {
         config: "core.packedGitLimit",
-        usage: NotApplicable("we target 32bit systems only and don't use a windowing mechanism")
+        usage: NotApplicable("we target 64-bit systems only and don't use a windowing mechanism")
     },
     Record {
         config: "core.alternateRefsCommand",
@@ -91,7 +91,7 @@ static GIT_CONFIG: &[Record] = &[
         usage: Planned("Allow to remove similar hardcoded value - passing it through will be some effort")
     },
     Record {
-        config: "core.loosecompression",
+        config: "core.looseCompression",
         usage: Planned("")
     },
     Record {
@@ -100,7 +100,7 @@ static GIT_CONFIG: &[Record] = &[
     },
     Record {
         config: "core.sparseCheckoutCone",
-        usage: Planned("this is a nice improvement over spareCheckout alone and should one day be available too")
+        usage: Planned("this is a nice improvement over sparseCheckout alone and should one day be available too")
     },
     Record {
         config: "core.gitProxy",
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot, appreciated!

> This information is available by API, so maybe this should be a conventional commit? I can easily amend this so the commit message starts with `fix:` if you like. This seems like a gray area with respect to whether `fix:` ought to be used.

That's an interesting question - I'd think of it as documentation change or fix, for which there could be the `doc: ` prefix, but I never tried it before and only hope it works.

I think that and more can be tried in a subsequent PR, let's not let this one wait any longer (as the effect of not having it in the changelog is probably very minor).

EliahKagan:
I've opened #1537 for those further changes.

PR #1535, opened 2024-08-22
fix similarity detection

fowles:
track all removed tokens and not the most recent batch only

### DIFF
--- a/gix-diff/src/rewrites/tracker.rs
+++ b/gix-diff/src/rewrites/tracker.rs
@@ -561,10 +561,10 @@ mod diff {
         type Out = usize;
 
         fn process_change(&mut self, before: Range<u32>, _after: Range<u32>) {
-            self.removed_bytes = self.input.before[before.start as usize..before.end as usize]
+            self.removed_bytes += self.input.before[before.start as usize..before.end as usize]
                 .iter()
                 .map(|token| self.input.interner[*token].len())
-                .sum();
+                .sum::<usize>();
         }
 
         fn finish(self) -> Self::Out {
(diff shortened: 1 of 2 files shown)


martinvonz:
@Byron: This fix seems pretty important to jj users. When do you expect it will be released? I didn't see an obvious pattern to Gitoxide releases (seems roughly around the middle of each month but not quite). Just trying to decide how to plan our releases for it.

gix-diff/src/rewrites/tracker.rs
```diff
@@ -561,10 +561,10 @@ mod diff {
         type Out = usize;
 
         fn process_change(&mut self, before: Range<u32>, _after: Range<u32>) {
-            self.removed_bytes = self.input.before[before.start as usize..before.end as usize]
+            self.removed_bytes += self.input.before[before.start as usize..before.end as usize]
```
Byron, line 564: It's amazing that it's just this!

Byron (approved):
Thanks, this looks great!

> @Byron: This fix seems pretty important to jj users. When do you expect it will be released? I didn't see an obvious pattern to Gitoxide releases (seems roughly around the middle of each month but not quite). Just trying to decide how to plan our releases for it.

I will try to just release `gix-diff` as patch release, which will probably allow you to pick that up and perform a release of `jj` as you see fit.
If that's not possible, I will try to create a new `gix` release.

gix-diff/src/rewrites/tracker.rs
fowles: I was genuinely surprised myself, but it makes all the tests pass and the old tests continue passing.  We will likely find more over time, but this will definitely improve things in the short term!

martinvonz:
> I will try to just release `gix-diff` as patch release, which will probably allow you to pick that up and perform a release of `jj` as you see fit.
> If that's not possible, I will try to create a new `gix` release.

Thanks!

Byron:
`gix` v0.66 is now available: https://github.com/Byron/gitoxide/releases/tag/gix-v0.66.0 - v0.65 was botched so I had to re-release. Quite a mess, once again, due to a breaking change not being signalled as one (or somehow the safety-bumping didn't work).

PR #1536, opened 2024-08-22
fixes

Byron:
The problem is that with them, we don't notice anymore if the crate changes, because a dependency changes. That also means that older versions of the dependency may stay even though some other crates might pick up a newer version.

Ultimately, this will lead to drift and subtle incompatibilities.

We declare this breaking to enforce a proper re-release.

**Note**: Had to disable fuzzing again as it fails due to some incompatibility with `serde` and the nightly compiler that it appears to be using. I really have had it all tonight 😁


EliahKagan:
Does this change have any bearing on the feasibility of #1510? (I'm asking because I notice that's still open, even though the comment discussion suggests it may not be something that can or should be done.)

EliahKagan:
It looks like the fuzzing failure may have been temporary, since I ran it in my fork, both on main and at [1757377](https://github.com/Byron/gitoxide/pull/1536/commits/17573779688e755a786546d5e42ab533088cd726) from this PR, a few times each, and it ran with no problems:

- https://github.com/EliahKagan/gitoxide/actions/runs/10515699495 (b3ff033)
- https://github.com/EliahKagan/gitoxide/actions/runs/10516173098 (1757377)

Maybe rerunning [the failing job here](https://github.com/Byron/gitoxide/actions/runs/10514997896/job/29134124511) would work as well, and the removal in [3d90ab0](https://github.com/Byron/gitoxide/pull/1536/commits/3d90ab05959f34803e26cffda6ac33993fb183fe) would not be needed.

Byron:
> Does this change have any bearing on the feasibility of #1510? (I'm asking because I notice that's still open, even though the comment discussion suggests it may not be something that can or should be done.)

I think that's unrelated.

Thanks for the hint about fuzzing - I will try again.

EliahKagan:
> Thanks for the hint about fuzzing - I will try again.
> 
> Edit: It still fails for me.

That's odd, and I wonder if it is somehow influenced by which event trigger is used. I'll see if I can figure anything out.

PR #1537, opened 2024-08-22
Unify style in config support info

EliahKagan:
This builds on b31d6b7 (#1531) by adjusting the capitalization and quoting style of string arguments in `progress.rs` that help document the status of what configuration options are and are not planned.

- Since I believe these strings are not usually rendered as Markdown, and most places that code formatting in Markdown would be used were already written with single quotes, this applies that quoting style consistently.
- This applies initial capitalization, which was sometimes but not always done. It does not apply a consistent style for when a string will end in a `.`, which seems already to differ deliberately for reasons of clarity. That can be adjusted later.
- A small amount of minor rewording for clarity is also included.

### DIFF
--- a/src/plumbing/progress.rs
+++ b/src/plumbing/progress.rs
@@ -64,27 +64,27 @@ struct Record {
 static GIT_CONFIG: &[Record] = &[
     Record {
         config: "core.symlinks",
-        usage: Planned("needed to handle checkouts faithfully")
+        usage: Planned("Needed to handle checkouts faithfully")
     },
     Record {
         config: "core.hideDotFiles",
-        usage: Planned("Seems useful, but needs demand from windows users")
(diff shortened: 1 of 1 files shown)


EliahKagan:
I've prefixed the commit message title with `doc:` as suggested in https://github.com/Byron/gitoxide/pull/1531#pullrequestreview-2254236622. However, your wording there suggests that the effect is not certain. If you prefer not to deal with that at the same time as the release-related complexities discussed in comments at https://github.com/Byron/gitoxide/commit/f21857844332f73a6c090f349aa6693684d443bb, then I can amend this to remove the `doc:` prefix. My guess is that it's fine and you don't want me to do that, though, so I'll leave it in place unless you request a change.

Byron (approved):
Thanks a lot for this improvement, it's much appreciated.

Let's try `doc:`, it should work.

PR #1540, opened 2024-08-23
Fix winnow version requirement in gix-object

### LINKED ISSUE #1538: gix-object v0.44.0 fail to compile
### Current behavior 😯

https://github.com/cargo-bins/simple-git/actions/runs/10517678165/job/29142235436?pr=35

```
error[E0599]: `(&[u8], impl Parser<_, <_ as Stream>::Slice, _>, &[u8], fn(&mut _) -> Result<..., ...> {rest::<..., ...>})` is not an iterator
  --> /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/gix-object-0.44.0/src/tag/decode.rs:59:26
   |
53 | /                     (
54 | |                         &PGP_SIGNATURE_BEGIN[1..],
55 | |                         take_until(0.., PGP_SIGNATURE_END),
56 | |                         PGP_SIGNATURE_END,
57 | |                         rest,
58 | |                     )
59 | |                         .take()
   | |                         -^^^^ `(&[u8], impl Parser<_, <_ as Stream>::Slice, _>, &[u8], fn(&mut _) -> Result<..., ...> {rest::<..., ...>})` is not an iterator
   | |_________________________|
   |
   |
   = note: the full type name has been written to '/Users/runner/work/simple-git/simple-git/target/aarch64-apple-darwin/debug/deps/gix_object-976e30f84e72c296.long-type-4190708012872330450.txt'
   = note: consider using `--verbose` to print the full type name to the console
   = note: the full name for the type has been written to '/Users/runner/work/simple-git/simple-git/target/aarch64-apple-darwin/debug/deps/gix_object-976e30f84e72c296.long-type-4190708012872330450.txt'
   = note: consider using `--verbose` to print the full type name to the console
   = note: the following trait bounds were not satisfied:
           `(&[u8], impl winnow::Parser<_, <_ as Stream>::Slice, _>, &'static [u8], for<'a> fn(&'a mut _) -> Result<<_ as Stream>::Slice, ErrMode<_>> {rest::<_, _>}): Iterator`
           which is required by `&mut (&[u8], impl winnow::Parser<_, <_ as Stream>::Slice, _>, &'static [u8], for<'a> fn(&'a mut _) -> Result<<_ as Stream>::Slice, ErrMode<_>> {rest::<_, _>}): Iterator`

error[E0599]: the method `take` exists for tuple `(impl Parser<_, <_ as Stream>::Slice, _>, &[u8], Map<..., ..., ..., ..., ..., ...>)`, but its trait bounds were not satisfied
  --> /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/gix-object-0.44.0/src/parse.rs:25:14
   |
20 | /         (
21 | |             take_till(1.., NL),
22 | |             NL,
23 | |             repeat(1.., terminated((SPACE, take_until(0.., NL)), NL)).map(|()| ()),
24 | |         )
25 | |             .take()
   | |             -^^^^ method cannot be called due to unsatisfied trait bounds
   | |_____________|
   |
   |
   = note: the full type name has been written to '/Users/runner/work/simple-git/simple-git/target/aarch64-apple-darwin/debug/deps/gix_object-976e30f84e72c296.long-type-16570815735586947285.txt'
   = note: consider using `--verbose` to print the full type name to the console
   = note: the following trait bounds were not satisfied:
           `(impl winnow::Parser<_, <_ as Stream>::Slice, _>, &'static [u8], winnow::combinator::Map<winnow::combinator::Repeat<impl winnow::Parser<_, (_, _), _>, _, (_, _), (), _>, {closure@/Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/gix-object-0.44.0/src/parse.rs:23:75: 23:79}, _, (), (), _>): Iterator`
           which is required by `&mut (impl winnow::Parser<_, <_ as Stream>::Slice, _>, &'static [u8], winnow::combinator::Map<winnow::combinator::Repeat<impl winnow::Parser<_, (_, _), _>, _, (_, _), (), _>, {closure@/Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/gix-object-0.44.0/src/parse.rs:23:75: 23:79}, _, (), (), _>): Iterator`
```

### Expected behavior 🤔

Should compile without error

### Git behavior

_No response_

### Steps to reproduce 🕹

_No response_

NobodyXu:
Fixed #1538

### DIFF
--- a/gix-object/Cargo.toml
+++ b/gix-object/Cargo.toml
@@ -50,7 +50,7 @@ bstr = { version = "1.3.0", default-features = false, features = [
     "std",
     "unicode",
 ] }
-winnow = { version = "0.6", features = ["simd"] }
+winnow = { version = "0.6.18", features = ["simd"] }
 smallvec = { version = "1.4.0", features = ["write"] }
 serde = { version = "1.0.114", optional = true, default-features = false, features = [
     "derive",


Byron (approved):
Thanks a lot!

I still think that `winnow` can't break between patches without being non-compliant to semver. 
In any case, I hope that this along with the improved `msrv `check will prevent such issues in future.

spenserblack:
Just ran into this issue.

I don't believe `winnow` released a breaking release as a patch. Cargo resolves the `0.6.0` version of `winnow` as `^0.6.0`, or `>= 0.6.0, < 0.7.0`. This is why `winnow` was `0.6.16` in your `Cargo.lock`, even though you specified it as `0.6.0`. To lock to *exactly* `0.6.0`, you would use `=0.6.0`.

This resulted in you using a higher version of `winnow` when working in your workspace, yet downstream users could have lower, potentially incompatible versions of `winnow`.

Bisected this to f5cd3baf57676bfee0c40fee577d2779958bbe72 by bisecting between gix 0.66 and 0.64, locking `winnow` to `=0.6.0` at each step. Looks like you need at least `winnow 0.6.15`. This makes sense, since `winnow` introduced the `.take()` method in https://github.com/winnow-rs/winnow/commit/28fdb2f08067ce1b02a90b4e1d0d39c14e258b09, which was then released in `0.6.15`. This issue is *not* that winnow broke usage. Just the opposite -- they renamed `.recognize` to `.take`, but they kept `.recognize` as a deprecated method to ensure backwards compatibility. The issue is that `gix-object` started calling a method that does not exist in `winnow < 0.6.15`, *without specifying that it requires `winnow >= 0.6.15`.* So downstream users continued to use `winnow <0.6.15`, causing the failure to compile.

In the future, to ensure the best compatibility, it's probably going to be best to not update direct dependencies directly in the `Cargo.lock`, but update the required version specs in your `Cargo.toml` files, instead. Especially if you're changing your usage of those dependencies. So, in this example, because you were dropping usage of the deprecated `.recognize` for `.take`, you should have updated your `Cargo.toml` to specify a version greater than or equal to the `winnow` version that introduced the deprecation.

Byron:
> Bisected this to [f5cd3ba](https://github.com/Byron/gitoxide/commit/f5cd3baf57676bfee0c40fee577d2779958bbe72) by bisecting between gix 0.66 and 0.64, locking `winnow` to `=0.6.0` at each step. Looks like you need at least `winnow 0.6.15`. This makes sense, since `winnow` introduced the `.take()` method in [winnow-rs/winnow@28fdb2f](https://github.com/winnow-rs/winnow/commit/28fdb2f08067ce1b02a90b4e1d0d39c14e258b09), which was then released in `0.6.15`. This issue is _not_ that winnow broke usage. Just the opposite -- they renamed `.recognize` to `.take`, but they kept `.recognize` as a deprecated method to ensure backwards compatibility. The issue is that `gix-object` started calling a method that does not exist in `winnow < 0.6.15`, _without specifying that it requires `winnow >= 0.6.15`._ So downstream users continued to use `winnow <0.6.15`, causing the failure to compile.

Thanks for the detective-work! I see now how this could happen, and that I should have adjusted the minimum required version of `winnow` in Cargo.toml accordingly.

> Hope this helps!

Yes, it does! Interestingly, I am very aware of the differences between Cargo.lock and Cargo.toml, yet in practice I have failed to do the right thing. Interesting.

PR #1541, opened 2024-08-23
Fix msrv: Run msrv checks with minimal versions

NobodyXu:
Since it is possible to for dependencies to bump MSRV in patch release, msrv checking should use the minimal versions supported by gix.

Users cares deeply about msrv can also pin them to the minimal versions to maintain their current msrv.

This would have also help discover bug in #1538

### DIFF
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -30,6 +30,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: extractions/setup-just@v2
       - run: |
-          rustup toolchain install ${{ env.rust_version }} --profile minimal --no-self-update
+          rustup toolchain install ${{ env.rust_version }} nightly --profile minimal --no-self-update
           rustup default ${{ env.rust_version }}
+          cargo +nightly update -Zminimal-versions
       - run: just ci-check-msrv
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1045,9 +1045,9 @@ dependencies = [
 
 [[package]]
 name = "flate2"
-version = "1.0.32"
+version = "1.0.33"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9c0596c1eac1f9e04ed902702e9878208b336edc9d6fddc8a48387349bab3666"
+checksum = "324a1be68054ef05ad64b861cc9eaf1d623d2d8cb25b4bf2cb9cdd902b4bf253"
 dependencies = [
  "crc32fast",
  "libz-ng-sys",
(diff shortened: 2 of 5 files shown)


Byron (approved):
Thanks a lot for the initiative! I can use any help I can get to get this dependency-monster under control.

Byron:
Even though it was initially reproducing the `gix-object` + `winnow` error, now it seems to fail in another crate entirely.

Is this something you can figure out? I tried, but it doesn't make sense after all.

NobodyXu:
I think the version of `yoke` might be too old?

Since the error is from `icu_locid_transform` I suppose try bumping it, if latest version doesn't work then I think it's report a bug in upstream.

NobodyXu:
```
error: failed to run custom build command for `libz-ng-sys v1.1.8`

Caused by:
  process didn't exit successfully: `D:\a\gitoxide\gitoxide\target\debug\build\libz-ng-sys-c79adf8887a64bd5\build-script-build_zng` (exit code: 101)
  --- stderr
  thread 'main' panicked at C:\Users\runneradmin\.cargo\registry\src\index.crates.io-6f17d22bba15001f\cmake-0.1.44\src\lib.rs:783:22:
  Visual studio version detected but this crate doesn't know how to generate cmake files for it, can the `cmake` crate be updated?
```

looks like libz-ng is too old?

Byron:
Thanks a lot for the encouragement!

I ended up running the `cargo +nightly update -Zminimal-versions` to reproduce it locally.

However, now it seems to show drift in the `flate2` crate, which apparently fails to build on Windows 2022, and I opened a PR against it. 

This shows that `minimal-versions` is also debugging the entire dependency tree, and I don't know if I like that ~~burden~~ responsibility 😅.

NobodyXu:
> This shows that minimal-versions is also debugging the entire dependency tree, and I don't know if I like that burden responsibility 😅.

Yeah it's somewhat transitive, I think we can reduce this burden by running `-Zminimal-versions` in more places, i.e. in flate2, url, etc

NobodyXu:
@Byron I opened rust-lang/flate2-rs#425 to add `-Zminimal-versions` to flate2

Byron:
Thanks  - once it fails I plan to put my 'fix' from the other PR on top of it, bump the version, merge and create a new release. Then this PR should be unblocked (or is blocked by something else :D).

Byron:
Needs https://github.com/rust-lang/flate2-rs/pull/425 to continue, which needs `minimal-version` checks for `libz-sys` along with a fix to raise the minimally required `vcpkg`.

PR #1543, opened 2024-08-23
Be as lenient as Git when parsing tags

### LINKED ISSUE #1542: gitoxide is still less lenient than git when parsing signatures, in particular in tags
### Current behavior 😯

In the wild, some tags exist that don't have a timestamp on the tagger line.
For instance this one:  https://kernel.googlesource.com/pub/scm/network/bridge/bridge-utils/+/refs/tags/ChangeLog

### Expected behavior 🤔

gitoxide should be as lenient as git when parsing tags so it can be used on the exotic git objects that one can find.
If the timestamp is garbage or missing, it should set it to zero instead of failing.

### Git behavior

git is very lenient when parsing signatures: on any sort of malformed timestamp, it simply sets the timestamp to 0.
See implementation [here](https://github.com/git/git/blob/master/tag.c#L116)

## To observe that this behaviour exists in the wild:
```
> git clone https://kernel.googlesource.com/pub/scm/network/bridge/bridge-utils
```
```
Cloning into 'bridge-utils'...
remote: Counting objects: 16, done
remote: Total 712 (delta 487), reused 712 (delta 487)
Receiving objects: 100% (712/712), 142.50 KiB | 2.30 MiB/s, done.
Resolving deltas: 100% (487/487), done.
warning: remote HEAD refers to nonexistent ref, unable to checkout
```
```
> cd bridge-utils
> git show-ref ChangeLog
```
```
e601dc1094107999de050b7104bf01ce865fe60f refs/tags/ChangeLog
```
```
> git cat-file -p e601dc1094107999de050b7104bf01ce865fe60f
```
```
object 4fcd840c4935e4c7a5ea3552710a0f26b9178c24
type commit
tag ChangeLog
tagger shemminger <shemminger>
```
## To repro the behaviour with gitoxide
I will share a couple of new unit tests that show that gitoxide fails to parse.

pierrechevalier83:
When Git parses tags, it ignores invalid or missing timestamps in the tagger line and simply zeroes out that field.
gitoxide currently requires that the tagger line be correct.

Since we can see tags in the wild that don't have a timestamp in the tagger line, make gitoxide able to parse such tags.

See issue #1542 for more details.

fixes #1542

### DIFF
--- a/gix-actor/src/signature/decode.rs
+++ b/gix-actor/src/signature/decode.rs
@@ -6,7 +6,7 @@ pub(crate) mod function {
     use winnow::error::{ErrMode, ErrorKind};
     use winnow::stream::Stream;
     use winnow::{
-        combinator::{alt, separated_pair, terminated},
+        combinator::{alt, opt, separated_pair, terminated},
         error::{AddContext, ParserError, StrContext},
         prelude::*,
         stream::AsChar,
@@ -21,8 +21,8 @@ pub(crate) mod function {
     ) -> PResult<SignatureRef<'a>, E> {
         separated_pair(
             identity,
-            b" ",
-            (
+            opt(b" "),
+            opt((
                 terminated(take_until(0.., SPACE), take(1usize))
                     .verify_map(|v| to_signed::<SecondsSinceUnixEpoch>(v).ok())
                     .context(StrContext::Expected("<timestamp>".into())),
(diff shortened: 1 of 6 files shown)


Byron (approved):
Thanks a lot for the fix!

Whenever parsers become more lenient I am wondering how `fsck` would be handling this - after all, there one wouldn't want leniency. 

Then I thought it would be easiest *and* cleanest to just re-serialize all objects, and show a diff if they differ. That wouldn't happen for correctly formed objects, but would happen in this case, for example.

Thus it should be good to have very lenient parsers without the need to add options for conditional leniency.

pierrechevalier83:
> Thanks a lot for the fix!
> 
> Whenever parsers become more lenient I am wondering how `fsck` would be handling this - after all, there one wouldn't want leniency.
> 
> Then I thought it would be easiest _and_ cleanest to just re-serialize all objects, and show a diff if they differ. That wouldn't happen for correctly formed objects, but would happen in this case, for example.
> 
> Thus it should be good to have very lenient parsers without the need to add options for conditional leniency.

As you're saying, with `git`, `fsck` is not lenient in this case:
taking the tag that I found in bridge-utils and that I showed in the issue, `fsck` fails:
```
> git cat-file -p e601dc1094107999de050b7104bf01ce865fe60f > tag
> git mktag < tag
error: tag input does not pass fsck: missingSpaceBeforeDate: invalid author/committer line - missing space before date
fatal: tag on stdin did not pass our strict fsck check
```

I agree with the suggestion you are making: if it doesn't round-trip cleanly, like here, something is up and `fsck` should make you aware of it.

Byron:
Unfortunately I can't make a patch release anymore, so the next major one is due 22nd of September.
Meantime, maybe you can use a `git = "<url>"` dependency?

PR #1546, opened 2024-08-24
Add missing semicolons

nyurik:
Automatically apply it with clippy:

```
 cargo clippy --fix --all-targets --workspace -- -A clippy::all -W clippy::semicolon-if-nothing-returned
```

This would close the biggest pedantic clippy complaint in #874


Byron (approved):
Thanks a lot for picking this up!

I wonder if this check could be added to the standard lints then? These days configuration for Clippy can very well be placed in the top-level Cargo.toml, but I never got around to actually do it.

Any help in that regard is also very much appreciated.

PR #1548, opened 2024-08-24
One fix of clippy::match_same_arms

nyurik:
Clippy suggests to collapse identical match arms ([info](https://rust-lang.github.io/rust-clippy/master/index.html#/match_same_arms)), but i am not too certain it will create more readable code everywhere. Though, while looking at it, this instance clearly is an accidental dup.


Byron (approved):
It definitely is a duplicate, great catch!

Thanks again!

PR #1549, opened 2024-08-24
 Support compile `gix-url` to wasm32-wasi

ChanTsune:
Hi! I tried to compile `gix-url` to wasm32-wasi but encountered an error during compilation with this error:
```
error[E0425]: cannot find function `home_dir_inner` in the crate root
  --> /.cargo/registry/src/index.crates.io-6f17d22bba15001f/home-0.5.9/src/env.rs:33:16
   |
33 |         crate::home_dir_inner()
   |                ^^^^^^^^^^^^^^ not found in the crate root
   |
note: found an item that was configured out
  --> /.cargo/registry/src/index.crates.io-6f17d22bba15001f/home-0.5.9/src/lib.rs:66:14
   |
66 | use windows::home_dir_inner;
   |              ^^^^^^^^^^^^^^
note: found an item that was configured out
  --> /.cargo/registry/src/index.crates.io-6f17d22bba15001f/home-0.5.9/src/lib.rs:69:4
   |
69 | fn home_dir_inner() -> Option<PathBuf> {
   |    ^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0425`.
error: could not compile `home` (lib) due to 1 previous error
```
The reason for this error is that `home` does not support wasm. However, `gix-path`, which `gix-url` depends on, has a function `gix_path::env::home_dir` that provides similar functionality.
So, I used `gix_path::env::home_dir` instead of `home::home_dir` and remove `home` from `gix-url` dependencies.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -209,7 +209,7 @@ jobs:
       - name: Install Rust
         run: rustup update stable && rustup default stable && rustup target add ${{ matrix.target }}
       - uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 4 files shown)


Byron (approved):

PR #1551, opened 2024-08-25
Add job test-pass for stable job name

NobodyXu:
Which can be used in github branch protection

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -246,3 +246,24 @@ jobs:
           set -x
           git status
           git diff --exit-code
+
+  # Dummy job to have a stable name for the "all tests pass" requirement
+  tests-pass:
+    name: Tests pass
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot!

I think I can use this *and* the main jobs of the other CI configurations to assure truly everything needs to pass.
And I should be sure to maintain it or there is a risk it auto-merges even though CI is still running.
But I guess it's all worth it to get auto-merge :).

NobodyXu:
Yes it's more convenient than specifying each job manually in github branch protection and a bit easy to update, you just need to update the test-pass to include the newly added jobs

From my experience it's easier

Byron:
Screenshot 2024-08-25 at 09 51 08

All setup now, let's see how it works. *("Allow Auto-Merge" is also enabled in the general settings)*.

PR #1552, opened 2024-08-25
Speedup clippy in CI

NobodyXu:
Use --no-deps


Byron (approved):
Also nice - thanks again!

Byron:
It's 2m 44s here vs 2m 49s in a run that doesn't have the settings. So it's probably not super effective here if at all, but probably still a good idea to keep around (even if it's just 1s faster).

PR #1555, opened 2024-08-25
Run job installation in workflow release

NobodyXu:
Fixed #1553

also rm rust-cache from job installation as it is not effective for cargo-install


Byron (commented):
Thanks a lot for your help!

Does it mean that `installation` can fail without failing any of the builds?
If so, it can be merged, but if not, it's probably better to put it into its own workflow file.

NobodyXu:
> Does it mean that installation can fail without failing any of the builds?

Yes as long as it's not in github branch protection.

Byron (approved):
Alright, let's try it then. It's not mentioned in GitHub branch protections.

PR #1556, opened 2024-08-25
Speedup ci

NobodyXu:
- Run doc-test in test-fast
 - Use cargo-nextest for job test
 - Do not remove `target`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,9 @@ jobs:
         run:
           sudo apt-get install tree
       - uses: extractions/setup-just@v2
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: nextest
       - name: test
         env:
           CI: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,7 +92,9 @@ jobs:
         with:
           tool: nextest
       - name: "Test (nextest)"
-        run: cargo nextest run --all --no-fail-fast
+        run: |
+          cargo nextest run --all --no-fail-fast
+          cargo test --doc
 
   test-32bit:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


NobodyXu:
@Byron it shows that this PR makes ci.yml 5m faster

Byron (approved):
This is a very impressive outcome, thanks so much!

I am also surprised that `cargo nextest` manages to do the same faster, but maybe it's just the removal of `clear-target` (which exists there for reasons lost to time) which so much.

In any case, I prefer the output of `cargo nextest` and wasn't aware it can be used as `cargo test` stand-in - it's very neat, indeed!

NobodyXu:
> I am also surprised that cargo nextest manages to do the same faster, but maybe it's just the removal of clear-target 

It could be caching, but nextest is indeed faster if you have lots of tests and more than 1 cores, I checked the log it does run tests for some crates faster by running them in parallel by spawning one process for each tests.

It might be more robust if your tests involve global variables or potential UBs, running in separate processes protects against that.

NobodyXu:
> wasn't aware it can be used as cargo test stand-in - it's very neat, indeed!

It's not a full drop-in, it can't run doc-test right now, you'd need `cargo test --doc`.

However I think you don't have the habit of using doc-test so often as most of the valuable tests is probably in `tests/` or under `#[cfg(test)] mod tests`

NobodyXu:
P.S. nextest also integrates with llvm-cov, miri, criterion and cargo-mutants

https://nexte.st/docs/integrations/

NobodyXu:
https://github.com/nextest-rs/reuse-build-partition-example/blob/main/.github/workflows/ci.yml

And it can:
 - compile the tests and run it on different machines
 - partition the tests, run tests in parallel on different machines/GHA runners

Byron:
Thanks again for the heads-up!
`cargo nextest` is truly fascinating, and I hope that one day it will also learn how to run `doc` tests (even though it now also improves to be more efficient when compiling and running these tests).

Issue #780, opened 2023-03-15
Blake3 support


yonas:
### Summary 💡

In addition to SHA3-256 (or just SHA-256), supporting Blake3 would be great.

### Motivation 🔦

Blake3 is extremely fast and secure.

![BLAKE3-benchmark](https://user-images.githubusercontent.com/110130/225387360-fa03d313-a660-49a2-9a58-65f830d3c5b3.svg)

Byron:
Once #281 lands it should be possible to more easily plug in many other hash algorithms at hearts content. To make these useful, the `git` project would have to adopt them first and SHA256 is already kind of stalled, so I don't see any of this happening anytime soon.

Issue #585, opened 2022-11-07
Use hardened `Sha1` implementations (collision detection)


Byron:
* [ ] [implement collision detection in `sha-1` crate](https://github.com/RustCrypto/hashes/issues/204)
* [ ] See if `sha1_smol` implements hardening (https://github.com/mitsuhiko/sha1-smol/issues/43)

At least add a feature toggle to support this crate:

* [ ] https://docs.rs/sha1collisiondetection

pascalkuthe:
Some additional details:

* git uses hardened sha1 by default so there might be a chance that gitoxide messes up an object graph by generating different hashes then git does (needs investigation)
* this case is extremely rare as the probability of differing in honest usage of less than $2^{-90}$
* [sha1collisiondetection](https://docs.rs/sha1collisiondetection) is a straight port from C to rust using `c2rust` without much modifications on top. So it is insanely unsafe (nearly 100% unsafe code). Its likely also slower then the current sha1 algorithm (implemented in assembly).
* a paper describing the algrithm can be found [here](https://marc-stevens.nl/research/papers/C13-S.pdf).
* GitHub an other hosting provides likely place a high priority on this (for example https://github.blog/2017-03-20-sha-1-collision-detection-on-github-com/)

I think implementing this in the `sha-1` crate would be the best option but also extremely challenging (its partly written in assembly). Perhaps adding this to the `sha1_soml` crate first would be easier (and adding some kind of safety feature). That library only contains two unsafe blocks so it might be preferable from a security point of view and also easier to contribute to (it also does not contain any assembly).

Byron:
I was wondering what should happen once a collision is detected, as it's critical to do the same thing that `git` does in that case.

Options are…

1. fail and abort the operation that caused the collision, effectively preventing the object to be added
2. generate a different, safer hash instead

It [looks like `git` is taking approach number 2](https://github.com/git/git/blob/3b08839926fcc7cc48cf4c759737c1a71af430c1/sha1dc/sha1.h#L60:L72), and creates a different hash instead. By the looks of it, they can control various flags related to collision detection at runtime, but initialize these to be as safe as possible (detect collision, do unavoidable attack condition checks, and generate a safe hash with additional rounds).

For our implementation to be valid, we'd have to be sure that our safe hashes are indeed the same as the ones that git produces. Unfortunately I wasn't able to find a test there.

[...]

Issue #281, opened 2021-12-21
Sha 256 support


Byron:
Even though the foundation is set, it needs another push to actually make it work with different kinds of hashes.

### Tasks

* [ ] remove hash-type specific methods from `git-hash` and replace them with parametric usage of `git_hash::Kind`
* [ ] all code assuming hashes of len 20 should receive this value as parameter instead. This is what git does for the old index and pack file formats.
* [ ] general tests for reading  refs and objects of different len
* [ ] tests for writing and reading objects of different len, maybe even write a conversion program

shea256:
+1 to this feature. Excited to see this implemented.

msrd0:
+1. Given that git no longer considers sha256 support experimental and forges such as forgejo added support, this has became more relevant for new git repositories.

D4V1D3-008:
+1. Without sha256 support Starship prompt cannot detect repositories using that algorithm.

jokeyrhyme:
@Byron 

> make sure once Sha256 is added as ObjectId variant, that it's behind a feature toggle to allow builds that opt-out of SHA256 support to not unnecessarily use more memory than needed. Maybe there are alternatives to this, too.

should the old SHA1 support also be put behind a feature toggle (enabled by default), so that builds can choose to _only_ support SHA256 ?

Byron:
> +1. Without sha256 support Starship prompt cannot detect repositories using that algorithm.

Starship would definitely be a great motivation to finally go through with this.

`gix-hash` has been created with that in mind:

https://github.com/GitoxideLabs/gitoxide/blob/ec69c88fc119f3aa1967a7e7f5fca30e3ce97595/gix-hash/src/object_id.rs#L13-L14

Right now it's an enum and one could add another variant to it, behind a feature toggle, as the size of these hash digests definitely affects the resource requirements of operations like `clone`. As the size of these hashes would increase, this would affect the memory consumption even when cloning SHA1 repositories (unfortunately).
Maybe there are better ways to represent this - I think Git simply allocates for it which would be one solution here as well.

In any case, if this were to move forward, I think it would be good to review the hash representation first, as it's very relevant for performance.

> [@Byron](https://github.com/Byron)
> 
> > make sure once Sha256 is added as ObjectId variant, that it's behind a feature toggle to allow builds that opt-out of SHA256 support to not unnecessarily use more memory than needed. Maybe there are alternatives to this, too.
> 
[...]

PR #1566, opened 2024-08-30
octopus-merge (part 1: tree-editing)

Byron:
Implement an octopus merge based on trees, and (mostly) equivalent to `merge-ORT` in Git.

Related to https://github.com/gitbutlerapp/gitbutler/pull/4793.

### Tasks

* [x] tree-editor plumbing
* [x] cursor support for editor
* [x] prevent writing grocely invalid trees (\0 in filename, should be validate path components?)
* [x] benchmark with/without cursor
* [x] tree-editor in `gix`
    - [x] proper validation with `gix_validate::path::component()`
    - [x] convenience by default by making it easier to pass relative paths?
    - [x] cursor support
    - [x] write in-memory

### Next PR

* git2-equivalent merge (first version essentially, hoping it's not as complex as a full merge-ORT)

### Research

Everything is about MergeORT.

* It puts conflict-markers in the result tree, with annotations to provide additional context

### Handle Special Cases

* A file was renamed differently
* deal with "merge.directoryRenames"

### References

* https://git-scm.com/docs/git-merge-tree
* [trivial three-way merge in Git](https://github.com/git/git/blob/master/builtin/merge-tree.c#L326), visual only, otherwise it uses Merge ORT


gix-object/src/tree/editor.rs
```diff
@@ -0,0 +1,267 @@
...
+/// all edits until everything is flushed to write all changed trees.
+///
+/// The editor is optimized to edit existing trees, but can deal with building entirely new trees as well
+/// with some penalties.
+///
+/// ### Note
+///
+/// For reasons of efficiency, internally a SHA1 based hashmap is used to avoid having to store full paths
+/// to each edited tree. The chance of collision is low, but could be engineered to overwrite or write into
+/// an unintended tree.
```
Byron, line 20: I think this note I wrote specifically for you @EliahKagan, and just wanted to say that I plan to add benchmark and see if an ordinary HashMap would also do the trick without being slower, effectively, to prevent clashes entirely.

EliahKagan: Is the concern that this is using object IDs in a way that goes beyond how Git relies on them and beyond how gitoxide is otherwise relying on them, such that it makes sense to avoid this usage even when object IDs are SHA-1 rather than Hardened SHA-1?

If not, then I think the more important thing to do is is #585, to make gitoxide readily use Hardened SHA-1, and preferably to do so by default.

From the comments there, I am having trouble telling what the status really is of Hardened SHA-1 support in Git and in gitoxide. I believe gitoxide currently does not ever use it, but I could be misunderstanding or #585 may not have been updated.

As for Git, https://github.com/Byron/gitoxide/issues/585#issuecomment-1978209802 suggests that it, with default build options, it does not use Hardened SHA-1. Unfortunately, it does not link to a specific commit in Git, so I don't know if the text it takes to indicate that is the same as the text I see now. But following [that link](https://github.com/git/git/blob/master/Makefile#L480-L534) now does not clearly support the claim that Hardened SHA-1 has to be enabled explicitly. It also does not clearly refute that claim. [This sentence](https://github.com/git/git/blob/4590f2e9412378c61eac95966709c78766d326ba/Makefile#L500-L501) seems to reference [this text](https://github.com/git/git/blob/4590f2e9412378c61eac95966709c78766d326ba/Makefile#L511-L512), which makes it seems like the default is non-hardened SHA-1, while in contrast [this sentence](https://github.com/git/git/blob/4590f2e9412378c61eac95966709c78766d326ba/Makefile#L520-L523) seems to say that the default is Hardened SHA-1.

So basically I don't know what's going on with that. It's something I've meant to look into for a while, and had planned to comment on #585 after looking into. But it seems relevant now so I figured I'd mention what I know about it, even though what I know is almost nothing.

Hardened SHA-1 of course does not necessarily protect against all collision attacks, but only against a class of attacks that [involve strange inputs](https://crypto.stackexchange.com/questions/44141/what-is-hardened-sha-1-how-does-it-work-and-how-much-protection-does-it-offer) that Hardened SHA-1 detects individually (whether or not there is a collision) and computes alternate hashes for. However, my understanding is that all successful published collision attacks for SHA-1 so far, and not just [SHAttered](https://shattered.io/), have been in this class. Certainly it is much better than non-hardened SHA-1.

> I plan to add benchmark and see if an ordinary HashMap would also do the trick without being slower

Regarding that choice, I can think of a few considerations that may be relevant:

- From the highlighted source code comment together with your review comment, I think the main choice here is between using object IDs and paths as keys. But the performance of storing paths may be affected by how the are represented, including possibly in ways that can be adjusted for better performance. In particular, I would expect that the small string optimization (or small vector optimization) would have some affect, one way or another.
- If the data structure itself can be changed, then my understanding is that a `BTreeMap`, even though its operations are asymptotically slower, sometimes outperforms a `HashMap`. What I am not sure of is how often this really happens in Rust where the `HashMap` is implemented in a way that is fast on modern hardware. (I am accustomed to using `HashMap` implementations that use separate chaining, so I am always thinking of whether hash maps are really the fastest data structure, which may be a habit I can relax in Rust. Then again, `BTreeMap` is also designed to be fast on modern hardware.)
- Per [#281](https://github.com/Byron/gitoxide/issues/281), eventually gitoxide will also support SHA-256. Based on the point in [#780 (comment)](https://github.com/Byron/gitoxide/issues/780#issuecomment-1470462470), it sounds like using SHA-256 may not be a major slowdown. But if `Editor` is an area where SHA-1 keys are used for performance, it is not obvious that the same performance advantages will occur with SHA-256.

Byron: > Is the concern that this is using object IDs in a way that goes beyond how Git relies on them and beyond how gitoxide is otherwise relying on them, such that it makes sense to avoid this usage even when object IDs are SHA-1 rather than Hardened SHA-1?

It's related to me using an inferior hash-map as it allows for unrecognized key-conflicts, after all, it stores `K where K = SHA1(path)`. An attacker could craft trees that have different paths `a` and `b` where `SHA1(a) = X` and `SHA1(b) = X`, both entries use the same hash which means one will be silently overwriting another.

I don't think that this happens in practice, and I don't think it's an actual attack vector.

My plan is to add benchmarks and see what the loss in performance is if a real hashmap is used where the whole key, a path, is stored. If it's no major reduction in performance, I would of course just use the proper version to make this a none-issue.

> From the comments there, I am having trouble telling what the status really is of Hardened SHA-1 support in Git and in gitoxide. I believe gitoxide currently does not ever use it, but I could be misunderstanding or #585 may not have been updated.

Despite being a different topic, Indeed, it was always ambiguous in Git and `gitoxide` doesn't have it as it came with a huge performance penalty.


> There also seems to be a widespread assumption that Git uses Hardened SHA-1 by default ([here's an example](https://radicle.xyz/faq#isnt-sha-1-broken-how-do-you-handle-hash-collisions)), so if that is not the case then that may be an area where Git's documentation could be improved.

It would certainly be good to clarify that. On the 19th there will be the GitMerge conference in Berlin and I will be sure to ask - plenty of Git maintainers will be there. If you have other questions for them, please let me know and I will do my best to get them answered.



> * In particular, I would expect that the small string optimization (or small vector optimization) would have some affect, one way or another.

That's a good point and I could experiment with it. One major problem is that there are no small-string implementations for byte-strings available, so whatever I come up with there would be some workaround to this.

> * f the data structure itself can be changed, then my understanding is that a `BTreeMap`, even though its operations are asymptotically slower, sometimes outperforms a `HashMap`.

A good point, I can try that too.

> * But if `Editor` is an area where SHA-1 keys are used for performance, it is not obvious that the same performance advantages (of object IDs as compared to paths) will occur with SHA-256.

I'd most definitely use SHA-256 if it was available to `gitoxide`, but that's not the case yet. Then again, any hash can collide and I'd be better to not optimize something without a benchmark. I will be sure to get that and try to not do what I do now.

EliahKagan: > On the 19th there will be the GitMerge conference in Berlin and I will be sure to ask - plenty of Git maintainers will be there. If you have other questions for them, please let me know and I will do my best to get them answered.

Thanks--I will think about this and let you know in the next few days what other questions I have!

Byron: About the original topic: it turns out that using a standard HashMap is 30% to 40% faster.

```
❯ cargo bench -p gix-object@0.44.0 --bench edit-tree
   Compiling gix-object v0.44.0 (/Users/byron/dev/github.com/Byron/gitoxide/gix-object)
   Compiling gix-pack v0.53.0 (/Users/byron/dev/github.com/Byron/gitoxide/gix-pack)
   Compiling gix-odb v0.63.0 (/Users/byron/dev/github.com/Byron/gitoxide/gix-odb)
    Finished `bench` profile [optimized] target(s) in 5.97s
     Running benches/edit_tree.rs (target/release/deps/edit_tree-6af6651a1c453a05)
editor/small tree (empty -> full -> empty)
                        time:   [2.5972 µs 2.6019 µs 2.6075 µs]
                        thrpt:  [3.8351 Melem/s 3.8434 Melem/s 3.8503 Melem/s]
                 change:
                        time:   [-32.618% -32.355% -32.038%] (p = 0.00 < 0.05)
                        thrpt:  [+47.142% +47.831% +48.409%]
                        Performance has improved.
Found 14 outliers among 100 measurements (14.00%)
  13 (13.00%) high mild
  1 (1.00%) high severe
editor/deeply nested tree (empty -> full -> empty)
                        time:   [8.2019 µs 8.2079 µs 8.2145 µs]
                        thrpt:  [5.5998 Melem/s 5.6043 Melem/s 5.6084 Melem/s]
                 change:
                        time:   [-33.517% -33.377% -33.246%] (p = 0.00 < 0.05)
                        thrpt:  [+49.804% +50.099% +50.415%]
                        Performance has improved.
Found 13 outliers among 100 measurements (13.00%)
  8 (8.00%) high mild
  5 (5.00%) high severe

cursor/small tree (empty -> full -> empty)
                        time:   [2.6911 µs 2.6935 µs 2.6961 µs]
                        thrpt:  [3.7090 Melem/s 3.7127 Melem/s 3.7160 Melem/s]
                 change:
                        time:   [-33.881% -33.546% -33.225%] (p = 0.00 < 0.05)
                        thrpt:  [+49.757% +50.480% +51.242%]
                        Performance has improved.
Found 14 outliers among 100 measurements (14.00%)
  4 (4.00%) high mild
  10 (10.00%) high severe
cursor/deeply nested tree (empty -> full -> empty)
                        time:   [1.3616 µs 1.3631 µs 1.3649 µs]
                        thrpt:  [33.703 Melem/s 33.747 Melem/s 33.783 Melem/s]
                 change:
                        time:   [-40.063% -39.675% -39.234%] (p = 0.00 < 0.05)
                        thrpt:  [+64.566% +65.769% +66.843%]
                        Performance has improved.
Found 20 outliers among 100 measurements (20.00%)
  18 (18.00%) high mild
  2 (2.00%) high severe
```

Who would have thought :D.

EliahKagan: Cool!

PR #1568, opened 2024-08-31
Slightly adjust internal `gix_path::env::git`-related organization

EliahKagan:
This is a small follow-up to #1567. As suggested in https://github.com/Byron/gitoxide/pull/1567#issuecomment-2322846485, it moves the `first_file_from_config_with_origin` test case into the nested `exe_info` module, and it changes `NULL_DEVICE` from `static` to `const`.

This also changes `EXE_NAME`, which provides `git` or `git.exe` and never holds a longer path even when one is ultimately used, from `static` to `const`. While `EXE_NAME` is non-private, it is crate-internal via `pub(super)`, so this change is still non-breaking and non-user-facing.

In contrast, the static items `ALTERNATIVE_LOCATIONS` and `EXE_INFO` cannot be made `const`. No change is made to them. It may make sense in the future for `EXE_INFO` to hold an `Option<PathBuf>` rather than an `Option<BString>`; depending in part on the future direction of changes related to #1523 and #1567 or to the question of which scopes are eligible to inform `GitInstallation`, that may be required in the future. But nothing toward that is included here.

### DIFF
--- a/gix-path/src/env/git/mod.rs
+++ b/gix-path/src/env/git/mod.rs
@@ -83,9 +83,9 @@ pub(super) static EXE_NAME: &str = "git";
 pub(super) static EXE_INFO: Lazy<Option<BString>> = Lazy::new(exe_info);
 
 #[cfg(windows)]
-static NULL_DEVICE: &str = "NUL";
+const NULL_DEVICE: &str = "NUL";
 #[cfg(not(windows))]
-static NULL_DEVICE: &str = "/dev/null";
+const NULL_DEVICE: &str = "/dev/null";
 
 fn exe_info() -> Option<BString> {
     let mut cmd = git_cmd(EXE_NAME.into());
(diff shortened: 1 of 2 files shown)


Byron (approved):
Thanks a lot for your help with this, and for catching that indentation issue. Cargo-fmt doesn't deal as well with macros as one would hope.

EliahKagan:
I agree with renaming `EXE_INFO`, since its name was not very descriptive, and since it holds information *from* but not actually *about* the executable, so that name was arguably even inaccurate. Furthermore, there is no compatibility reason to keep the name `EXE_INFO`, since it is `pub(super)`.

However, I think this name runs the risk making it seem like the opposite of what it really is:

https://github.com/Byron/gitoxide/blob/adbaa2ab6840fd247c14701c39dcda3d280d86c0/gix-path/src/env/git/mod.rs#L80-L83

Although the meaning may vary subtly across contexts, in my mind this is the configuration file whose configuration has the *lowest* priority, because the variables in it have the highest scope; setting those variables in any other scope takes precedence.

I am not sure what the best name is, but maybe a better name could be `GIT_HIGHEST_SCOPE_CONFIG_PATH`?

EliahKagan:
I have opened #1569 to propose the PRIORITY -> SCOPE naming adjustment I suggested above in https://github.com/Byron/gitoxide/pull/1568#issuecomment-2322870711.

PR #1569, opened 2024-08-31
Rename to `GIT_HIGHEST_SCOPE_CONFIG_PATH`

EliahKagan:
This renames the crate-internal static item with the cached path of the highest scope nonempty config, treated as associated with the `git` installation, from `GIT_HIGHEST_PRIORITY_CONFIG_PATH` to `GIT_HIGHEST_SCOPE_CONFIG_PATH`.

This reflects the propensity of "priority" to have the opposite of the intended meaning here, since configuration in this file stands to be superseded by configuration in any lower (i.e. any other) scope. See #1568 for context.

I think this change from "PRIORITY" to "SCOPE" in the name preserves the benefits of the change from `EXE_INFO` to `GIT_HIGHEST_PRIORITY_CONFIG_PATH` (dd2d666), while making clearer what is going on and avoiding misinterpretations.

### DIFF
--- a/gix-path/src/env/git/mod.rs
+++ b/gix-path/src/env/git/mod.rs
@@ -80,7 +80,7 @@ pub(super) const EXE_NAME: &str = "git";
 /// Invoke the git executable to obtain the origin configuration, which is cached and returned.
 ///
 /// The git executable is the one found in PATH or an alternative location.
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot - naming is hard 🤦‍♂️

EliahKagan:
> Thanks a lot - naming is hard 🤦‍♂️

Yes, and I'm not sure this one is the best possible name either. If I think of something better, I'll let you know!

Issue #1575, opened 2024-09-05
`jj_realistic_needs_to_be_more_clever` failure is not limited to CI


EliahKagan:
### Current behavior 😯

The test case `jj_realistic_needs_to_be_more_clever` was added in 7ef1e88 (#1529). It was known to fail in Windows on CI:

https://github.com/Byron/gitoxide/blob/d69c6175574f34d6df92b4488ed2c9a85df12c89/gix/tests/object/tree/diff.rs#L396-L401

As indicated in the comment there, it was found pass when run locally on some Windows system. However, it fails when I run in on my main Windows 10 (x86-64) development system. So the failure is not limited to CI.

```text
ek@Glub MINGW64 ~/source/repos/gitoxide (main)
$ RUST_BACKTRACE=1 cargo nextest run -p gix jj_realistic_needs_to_be_more_clever
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.66s
------------
 Nextest run ID 845dbb1b-4fc5-470c-a3b8-6daa5649fd54 with nextest profile: default
    Starting 1 test across 4 binaries (262 tests skipped)
        FAIL [   0.166s] gix::gix object::tree::diff::track_rewrites::jj_realistic_needs_to_be_more_clever

--- STDOUT:              gix::gix object::tree::diff::track_rewrites::jj_realistic_needs_to_be_more_clever ---

running 1 test
test object::tree::diff::track_rewrites::jj_realistic_needs_to_be_more_clever ... FAILED

failures:

failures:
    object::tree::diff::track_rewrites::jj_realistic_needs_to_be_more_clever

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 258 filtered out; finished in 0.14s


--- STDERR:              gix::gix object::tree::diff::track_rewrites::jj_realistic_needs_to_be_more_clever ---
thread 'object::tree::diff::track_rewrites::jj_realistic_needs_to_be_more_clever' panicked at gix\tests\object\tree\diff.rs:452:9:
assertion `left == right` failed
  left: {"cli/src/commands/file/print.rs": ("cli/src/commands/cat.rs", 90), "cli/tests/test_file_print_command.rs": ("cli/tests/test_cat_command.rs", 77), "cli/tests/test_file_chmod_command.rs": ("cli/tests/test_chmod_command.rs", 88), "cli/src/commands/file/chmod.rs": ("cli/src/commands/chmod.rs", 90)}
 right: {}
[...]

PR #1580, opened 2024-09-06
Append to preexisting `MSYS` env var even if ill-formed

### LINKED ISSUE #1574: An old value of `MSYS` with unpaired surrogates is silently replaced for fixtures
### Current behavior 😯

In `gix-testtools`, when test fixture scripts are run, the `configure_command` function customizes their environment. Since 0899c2e (#1444), this includes setting the `MSYS` variable to make it so that commands in shell scripts such as `ln -s` will really attempt to create symlinks rather than copies.

If `MSYS` was already set, the options held in its old value are kept, with the new option appended at the end so it will only suppress an option that conflicts with it, otherwise leaving the effects of preceding options in place.

But this does not happen if the old value of `MSYS` contained an unpaired surrogate code point. This is because `env::var` rather than `env::var_os` is used, so the `unwrap_or_default` method returns an empty string both when `MSYS` was absent and when it was present with a value that could not be represented as well-formed UTF-8.

https://github.com/Byron/gitoxide/blob/9bddeeb948397df3d1e391ec48597cc34d10acd0/tests/tools/src/lib.rs#L596-L597

https://github.com/Byron/gitoxide/blob/9bddeeb948397df3d1e391ec48597cc34d10acd0/tests/tools/src/lib.rs#L605

There are at least three scenarios where `MSYS` may contain an unpaired surrogate:

- By accident.
- The path to one's debugger that one specifies with `error_start:` contains one. This seems unlikely, except as a special case of...
- Intentionally to test the behavior of some component when `MSYS` contains an unpaired surrogate.

The impact of this bug is therefore small. However, fixing it would help clarify the intended semantics of how `gix-testtools` treats the `MSYS` environment variable when running test fixture scripts. (This may potentially be relevant to some changes made for #1578.)

### Expected behavior 🤔

Due to the nature of `gix-testtools` as a testing facility, both as part of `gitoxide`'s test suite and when the published crate is used, I think any of these behaviors are fine and should be considered to fix the bug:

1. Always panic when the previous value of `MSYS` contained an unpaired surrogate, whether or not this is documented.

2. Always accept and append to the previous value of `MSYS` when it contains an unpaired surrogate, such as by using `env::var_os` and calling `push` (rather than `env::var` and calling `push_str`), whether or not this is documented.

   If this is done, then `unwrap_or_default` could still be used, because it would only be falling back to the default empty value when the environment variable was absent.

   ```diff
   -    let mut msys_for_git_bash_on_windows = std::env::var("MSYS").unwrap_or_default();
   -    msys_for_git_bash_on_windows.push_str(" winsymlinks:nativestrict");
   +    let mut msys_for_git_bash_on_windows = std::env::var_os("MSYS").unwrap_or_default();
   +    msys_for_git_bash_on_windows.push(" winsymlinks:nativestrict");
   ```

3. Most other behaviors, including the current behavior, if it is documented.

While any of those would, in my opinion, satisfy users' expectations and eliminate the *bug* as such, I think the *best* behavior would be to do (1) or (2). Specifically:

- If appending a space followed by `nativesymlinks` to a preexisting value of `MSYS` that ends in a high surrogate is effective at causing symlinks to be created, then (2) is preferred.
- If that does not work, then (1) is preferred.

I plan to investigate this and open a PR that makes one of those changes. (This investigation has been slightly slowed by the problem described in https://github.com/rust-lang/libz-sys/issues/215, but I should be able to complete it.)

*Update:* I have opened #1580, which changes it to use `OsString` instead of `String` and to append with `OsString::push`. As noted there, I verified that a value of `MSYS` produced this way does still have the desired effect on symlink creation.

### Git behavior

Not directly applicable.

### Steps to reproduce 🕹

The nice way to reproduce this would be to make a fixture that displays the value, or to add a tracing statement, or to use a debugger.

Instead, I just added a `panic!` to print out whatever `MSYS` value will be used:

```diff
diff --git a/tests/tools/src/lib.rs b/tests/tools/src/lib.rs
index b3c0102ea..127e7cab1 100644
--- a/tests/tools/src/lib.rs
+++ b/tests/tools/src/lib.rs
@@ -595,6 +595,7 @@ fn configure_command<'a>(
 ) -> &'a mut std::process::Command {
     let mut msys_for_git_bash_on_windows = std::env::var("MSYS").unwrap_or_default();
     msys_for_git_bash_on_windows.push_str(" winsymlinks:nativestrict");
+    panic!("DEBUG: {msys_for_git_bash_on_windows:?}"); // FIXME: Remove after debugging.
     cmd.args(args)
         .stdout(std::process::Stdio::piped())
         .stderr(std::process::Stdio::piped())
```

Then I ran a test that, while conceptually unrelated to this issue, exercises that code, ensuring the `panic!` is hit. The following output was warnings induced by making that change, about unused variables and unreachable code, removed. [The full output can be seen in this gist.](https://gist.github.com/EliahKagan/823a06406ede7fdbc8ea71afb8405dcd)

First, a control, with well-formed Unicode:

```text
ek@Glub MINGW64 ~/source/repos/gitoxide/tests/tools (main)
$ MSYS='error_start:nonexistent' cargo nextest run configure_command_clears_external_config

    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.39s
------------
 Nextest run ID e36b5636-bba9-47d5-b9d5-0ae20e5c0f30 with nextest profile: default
    Starting 1 test across 3 binaries (6 tests skipped)
        FAIL [   0.015s] gix-testtools tests::configure_command_clears_external_config

--- STDOUT:              gix-testtools tests::configure_command_clears_external_config ---

running 1 test
test tests::configure_command_clears_external_config ... FAILED

failures:

failures:
    tests::configure_command_clears_external_config

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2 filtered out; finished in 0.01s


--- STDERR:              gix-testtools tests::configure_command_clears_external_config ---
thread 'tests::configure_command_clears_external_config' panicked at tests\tools\src\lib.rs:598:5:
DEBUG: "error_start:nonexistent winsymlinks:nativestrict"

   Canceling due to test failure
------------
     Summary [   0.017s] 1 test run: 0 passed, 1 failed, 6 skipped
        FAIL [   0.015s] gix-testtools tests::configure_command_clears_external_config
error: test run failed
```

A now, when the existing value contains an unpaired high surrogate:

```text
ek@Glub MINGW64 ~/source/repos/gitoxide/tests/tools (main)
$ MSYS=$'error_start:\uD800z' cargo nextest run configure_command_clears_external_config

    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.39s
------------
 Nextest run ID bc896e5b-fabd-45b5-8823-4889319cf522 with nextest profile: default
    Starting 1 test across 3 binaries (6 tests skipped)
        FAIL [   0.024s] gix-testtools tests::configure_command_clears_external_config

--- STDOUT:              gix-testtools tests::configure_command_clears_external_config ---

running 1 test
test tests::configure_command_clears_external_config ... FAILED

failures:

failures:
    tests::configure_command_clears_external_config

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2 filtered out; finished in 0.02s


--- STDERR:              gix-testtools tests::configure_command_clears_external_config ---
thread 'tests::configure_command_clears_external_config' panicked at tests\tools\src\lib.rs:598:5:
DEBUG: " winsymlinks:nativestrict"

   Canceling due to test failure
------------
     Summary [   0.026s] 1 test run: 0 passed, 1 failed, 6 skipped
        FAIL [   0.024s] gix-testtools tests::configure_command_clears_external_config
error: test run failed
```

Therefore, normally it is able to keep content from the old value:

```text
DEBUG: "error_start:nonexistent winsymlinks:nativestrict"
```

But when it cannot be converted to UTF-8, its silently discards it:

```text
DEBUG: " winsymlinks:nativestrict"
```

(The reason I added the `z` after the unpaired surrogate was to show a situation where it is not at the very end, which is the more common situation. The problem is not limited to when it is at the very end; rather, whether appending *ought* to be done, rather than panicking, should, I think, be determined by whether program built against the MSYS DLL are able to heed the appended `winsymlinks` option even when it the space that delimits it follows an unpaired high surrogate.)

EliahKagan:
Fixes #1574

When the `MSYS` variable is absent, we treat this the same as when it is present but empty. However, as described in #1574, an `MSYS` variable that is present but whose value contains an unpaired surrogate would also be replaced entirely, rather than appending to its old value. This changes that, to instead append, retaining whatever was there even if it was ill-formed Unicode.

An alternative change could be to panic when the old value is ill-formed Unicode. The change made here allows and appends to the old value, rather than panicking or keeping and documenting the previous behavior of discarding the old value, because the appended sequence ` winsymlinks:nativestrict` is effective at causing fixture scripts to attempt to create actual symlinks even if the preceding code point is an unpaired Unicode high surrogate.

---

To avoid confusion related to #1358, I tested without setting `GIX_TEST_IGNORE_ARCHIVES`, relying on how some fixtures must still run because archive generation is suppressed for them in `.gitignore` files. That failure would occur in connection with the `MSYS` variable even when run this way was observed in #1443 and #1444, but I have also verified that.

This PR adds and then removes tests, which were unsuitable for the reasons noted in #1578, and which I removed rather than attempting to fix due to the concerns in #1577 combined with there being little doubt, in this case, that the code using `OsString::push` works as it seems. If those abandoned tests are not wanted in the history, I have no objection to rebasing to remove those commits, which I would be pleased to do on request.

[**This gist**](https://gist.github.com/EliahKagan/fc6f81c4d0f1da5bd327772f3870f852) contains transcript of the manual verification done to decide on and check over this approach. The scripts run in a [normal environment](https://gist.github.com/EliahKagan/fc6f81c4d0f1da5bd327772f3870f852#file-1-no-msys-inherited-txt), as well as in an [environment in which the old value of `MSYS` has a trailing unpaired Unicode high surrogate](https://gist.github.com/EliahKagan/fc6f81c4d0f1da5bd327772f3870f852#file-2-ill-formed-msys-inherited-txt). Failures occur due to the [absence of `MSYS` when the code that sets it is removed](https://gist.github.com/EliahKagan/fc6f81c4d0f1da5bd327772f3870f852#file-3-no-msys-inherited-or-added-txt), which verifies both that this code does not break setting it, and that the appended space character and `winsymlinks` option takes effect properly even if it immediately follows an unpaired high surrogate that "tries" to combine with the next code point. [Testing with a `panic!` temporarily added](https://gist.github.com/EliahKagan/fc6f81c4d0f1da5bd327772f3870f852#file-4-with-panic-temporarily-added-to-check-value-txt), by the same technique as #1574, confirms that we really are concatenating.

All full test runs have an unrelated failure of `jj_realistic_needs_to_be_more_clever` locally, which is due to #1575 and can therefore be disregarded in the transcripts of local test runs here. Between tests, I removed generated fixture script output by using the technique discussed in #1435 of running `gix clean -xd -m '*generated*' -e`, which is shown in the gist transcripts. I did it that way rather than by running the stronger command `gix clean -xde` so the unpaired surrogate in the `MSYS` environment variable would not stall the build due to https://github.com/rust-lang/libz-sys/issues/215. The command used is strong enough, but to doubly ensure insufficient cleaning would not cause misinterpreted results, I ran the tests with the expected failure modification last of all the full runs, so that if left-over fixture output was going to cause tests to wrongly pass, then it would happen with the tests I was verifying should fail.

### DIFF
--- a/tests/tools/src/lib.rs
+++ b/tests/tools/src/lib.rs
@@ -593,8 +593,8 @@ fn configure_command<'a>(
     args: &[String],
     script_result_directory: &Path,
 ) -> &'a mut std::process::Command {
-    let mut msys_for_git_bash_on_windows = std::env::var("MSYS").unwrap_or_default();
-    msys_for_git_bash_on_windows.push_str(" winsymlinks:nativestrict");
+    let mut msys_for_git_bash_on_windows = std::env::var_os("MSYS").unwrap_or_default();
+    msys_for_git_bash_on_windows.push(" winsymlinks:nativestrict");
     cmd.args(args)
         .stdout(std::process::Stdio::piped())
         .stderr(std::process::Stdio::piped())
--- a/tests/tools/src/lib.rs
+++ b/tests/tools/src/lib.rs
@@ -872,7 +872,6 @@ impl<'a> Drop for Env<'a> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use std::ffi::OsStr;
 
     #[test]
     fn parse_version() {
@@ -932,45 +931,5 @@ mod tests {
         let status = output.status.code().expect("terminated normally");
         assert_eq!(lines, Vec::<&str>::new(), "should be no config variables from files");
         assert_eq!(status, 0, "reading the config should succeed");
-    }
-
-    #[test]
-    fn configure_command_msys() {
-        let temp = tempfile::TempDir::new().expect("can create temp dir");
-        let mut cmd = std::process::Command::new("not-actually-run");
-
-        let msys_option = configure_command(&mut cmd, &[], temp.path())
-            .get_envs()
-            .find(|(k, _)| *k == OsStr::new("MSYS"))
-            .expect("should customize MSYS variable")
-            .1
-            .expect("should set, not unset, MSYS variable")
-            .to_str()
-            .expect("valid UTF-8")
-            .trim_matches(' ');
-
-        assert_eq!(msys_option, "winsymlinks:nativestrict");
-    }
-
-    #[test]
-    fn configure_command_msys_extends() {
-        let old_msys = r"error_start:C:\gdb.exe";
-        let temp = tempfile::TempDir::new().expect("can create temp dir");
-        let mut cmd = std::process::Command::new("not-actually-run");
-        cmd.env("MSYS", old_msys);
-
-        let msys_options = configure_command(&mut cmd, &[], temp.path())
-            .get_envs()
-            .find(|(k, _)| *k == OsStr::new("MSYS"))
-            .expect("should customize MSYS variable")
-            .1
-            .expect("should set, not unset, MSYS variable")
-            .to_str()
-            .expect("valid UTF-8")
-            .split(' ') // Just spaces, not arbitrary whitespace.
-            .filter(|s| !s.is_empty())
-            .collect::<Vec<_>>();
-
-        assert_eq!(msys_options, vec![old_msys, "winsymlinks:nativestrict"]);
     }
 }
--- a/tests/tools/src/lib.rs
+++ b/tests/tools/src/lib.rs
@@ -872,6 +872,7 @@ impl<'a> Drop for Env<'a> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::ffi::OsStr;
 
     #[test]
     fn parse_version() {
@@ -931,5 +932,45 @@ mod tests {
         let status = output.status.code().expect("terminated normally");
         assert_eq!(lines, Vec::<&str>::new(), "should be no config variables from files");
         assert_eq!(status, 0, "reading the config should succeed");
+    }
+
+    #[test]
+    fn configure_command_msys() {
+        let temp = tempfile::TempDir::new().expect("can create temp dir");
+        let mut cmd = std::process::Command::new("not-actually-run");
+
+        let msys_option = configure_command(&mut cmd, &[], temp.path())
+            .get_envs()
+            .find(|(k, _)| *k == OsStr::new("MSYS"))
+            .expect("should customize MSYS variable")
+            .1
+            .expect("should set, not unset, MSYS variable")
+            .to_str()
+            .expect("valid UTF-8")
+            .trim_matches(' ');
+
+        assert_eq!(msys_option, "winsymlinks:nativestrict");
+    }
+
+    #[test]
+    fn configure_command_msys_extends() {
+        let old_msys = r"error_start:C:\gdb.exe";
+        let temp = tempfile::TempDir::new().expect("can create temp dir");
+        let mut cmd = std::process::Command::new("not-actually-run");
+        cmd.env("MSYS", old_msys);
+
+        let msys_options = configure_command(&mut cmd, &[], temp.path())
+            .get_envs()
+            .find(|(k, _)| *k == OsStr::new("MSYS"))
+            .expect("should customize MSYS variable")
+            .1
+            .expect("should set, not unset, MSYS variable")
+            .to_str()
+            .expect("valid UTF-8")
+            .split(' ') // Just spaces, not arbitrary whitespace.
+            .filter(|s| !s.is_empty())
+            .collect::<Vec<_>>();
+
+        assert_eq!(msys_options, vec![old_msys, "winsymlinks:nativestrict"]);
     }
 }


Byron (approved):
Thanks a lot!

I should definitely have used `push` and `var_os` from the beginning, and will try to remember that (at least by now) `OsString` has some ways of altering it. In my head, it's still so that `OsStr` is known to be able to do nothing, so it's hard to work with.

EliahKagan:
> that (at least by now) `OsString` has some ways of altering it

It occurs to me that I do not always check when features are introduced. I am hoping that if this feature came in later than the project MSRV, then one of the `clippy` checks on CI would catch it and fail. (If that is not the case, please let me know.)

PR #1581, opened 2024-09-06
Unset other env vars related to `GIT_DIR` for fixtures

EliahKagan:
This removes other environment variables that have an effect conceptually related to `GIT_DIR` even when `GIT_DIR` is not set. Most of them change where `git` will look for files that are ordinarily in a repository's `.git` directory. In contrast, `GIT_WORK_TREE` changes where the working tree is found.

Significant further details are in the commit message, in case they are of interest now or in the future. I did not add tests because the likelihood that there is a mistake here that new tests would find is low, and because tests that go beyond effectively repeating the changes to the code under test would effectively run into the concerns described in #1577 or #1578.


Byron (approved):
This is great, thanks a lot!

On another note, it's interesting to wonder if, or to what extend, `gitoxide` should respect such environment variables. If they are undocumented, the answer is not at all, but otherwise they would need to be given some consideration.

Facilities exist to control which environment variables `gitoxide` uses, and methods exist to respect environment variables specifically, and maybe some of that can be extended should there be any need.

EliahKagan:
I've noticed this PR has been open for a while since being approved. Was auto-merge intended to have picked it up?

Byron:
Strange, I completely missed this one (and your message) :/.

EliahKagan:
This could happen if you accidentally unsubscribed from notifications on this PR specifically, which I have found is easy to do on the Android app. But that wouldn't have an effect on auto-merge. All pushes here (including force pushes) substantially preceded the approving review, so I don't know what would cause that. But it may be that I am misunderstanding the circumstances required for auto-merge to take effect.

PR #1582, opened 2024-09-06
gix-path release

Byron:
In preparation for advisory.


Byron:
@EliahKagan I jumped the gun and published [the advisory](https://github.com/Byron/gitoxide/security/advisories/GHSA-m8rp-vv92-46c7) before adding comments, so I am adding the comment here instead.
The CVE is requested as well, and with the publish of `gix-path` done (it was a patch-release fortunately) I think everything worked out as planned.
Merging this PR is just a formality due to branch-protection now activated - auto-merge compensates well for that.

EliahKagan:
Thanks!

Seeing this, I have edited the "Affected versions" in the advisory metadata to change it from `<= 0.10.10` to `< 0.10.11`, which I think is clearer and also is similar to what we have been doing for the other advisories. (If you prefer this not be done, please certainly change it back and/or ask me to do so.)

EliahKagan:
I've opened https://github.com/rustsec/advisory-db/pull/2071 to create the corresponding RUSTSEC advisory.

EliahKagan:
The GitHub Advisory Database global advisory for CVE-2024-45405 has been released. I've updated the PR https://github.com/rustsec/advisory-db/pull/2071 (already updated with the CVE number) to include a reference to it. I've also opened https://github.com/github/advisory-database/pull/4768 to fix the rendering/linkifying of a commit hash that is intended to be a hyperlink, which rendered as intended automatically in the repo-local GHSA but not in the global GHSA. (This would not render automatically in a RUSTSEC advisory either, but I anticipated that when submitting it, so nothing further has to be done in that PR for this rendering issue.)

EliahKagan:
Unfortunately there were some problems revising the global CVE-2024-45405 advisory and the published version is now substantially broken.

Merging https://github.com/github/advisory-database/pull/4768 has also brought in a number of incorrect changes to parts of the advisory not edited in that PR. This appears to be due to different parsing or rendering rules applying before and after a PR is merged, and I have no idea how that happens, but it is something I've seen before in https://github.com/github/advisory-database/pull/3290. However, in that older PR on an older advisory, a member of the GitHub security curation team was visibly involved in reviewing the advisory, and was therefore available to the fix the problem. I don't know how to go about fixing it in this case. Presumably a manual PR with some syntax could fix it, but I have no idea what syntax changes would work because I have no way to preview it. A revert would not work: presumably the same rules would continue to apply, but even if not, a revert would also undo the date change in the metadata, thereby causing the previous version with the later date to still be considered authoritative.

This undesirable changes include adding backslashes and newlines to code blocks. This includes commands in the PoC section. As they currently appear in the global advisory, those commands will fail to trigger the vulnerability even when it is present and all other steps have been followed correctly. The risk is low, since the vulnerability it itself low risk as described in the "Impact" section, but this could lead to unjustified confidence that a patched version is in use.

For an example, though this is not the only breakage, these are the commands that "Ren" should run, as still correctly shown in the repo-local advisory:

```powershell
$d = "$HOME\303\251e\AppData\Local\Programs\Git\etc"
mkdir $d
git config --file $d\gitconfig core.sshCommand calc.exe
icacls $HOME\303 /grant 'Renée:(RX)' /T
```

This is what they have become in the global advisory since https://github.com/github/advisory-database/pull/4768 was merged:

```powershell
$d = \"$HOME\\303\\251e\\AppData\\Local\\Programs\\Git\\etc\"
mkdir $d
git config --file $d\\gitconfig core.sshCommand calc.exe
icacls $HOME\\303 /grant 'Renée:(RX)' /T
```

So the global advisory should be fixed in some way, but I don't know how. I can open an issue on the GitHub Advisory Database repository, but I think the turnaround time is fairly long for them, and I am also not confident that internal tooling is sufficient to allow this to be fixed on the first try.

EliahKagan:
https://github.com/github/advisory-database/issues/4777

EliahKagan:
The formatting and content problems that had arisen in the the global CVE-2024-45405 advisory have been completely fixed (https://github.com/github/advisory-database/issues/4777).

PR #1583, opened 2024-09-06
Don't read "installation" config from `GIT_CONFIG`

EliahKagan:
This makes it so `GIT_HIGHEST_SCOPE_CONFIG_PATH` never uses the value of the `GIT_CONFIG` environment variable. Previously it used it because `git config` uses it, treating it as though its value were an operand to `--file`. But no other `git` commands use it.

It was previously used in most situations when present, since it causes `git config -l` to list nothing else of any scope (even if very high scoped configuration is available). The main and possibly only significant situation where it was not used was the `EXEPATH` optimization on Windows (which in practice only occurs in Git Bash environments). I think this inconsistency--since it never has anything to do with whether a user would want `GIT_CONFIG` to have an effect--is one of the most significant reasons to omit it.

However, when reviewing this, I recommend considering if possible uses where a user would set `GIT_CONFIG` to view configuration from an arbitrary file with commands like `gix config`--even though it would not work from Git Bash--might be something people are doing. I have prefixed the commit message `fix:` but maybe I should have written `fix!:` for this reason? I am not sure. But as far as I know, this is not altering any behavior to become inconsistent with what was previously documented or otherwise stated.

There is more information in the commit messages, especially the one prefixed `fix:`, eb72d31a6. This also adds tests related to `GIT_CONFIG`, in the two commits that precede that, 340ff371d and f8e38d057.

I have also unset some other environment variables (that would only occasionally be set in the first place) besides `GIT_CONFIG`. This is in the first commit, 01a412feb, and it is for a different reason: it is to ensure that even if some version of `git` behaves weirdly, or if my understanding of its intended behavior is incorrect, then we still get the full effect of setting `GIT_DIR`. The commit message has more information about that. I mostly included that here rather than in a separate PR to avoid a merge conflict (or decreased code readability if the `remove_env` calls were to be merged in an unintuitive order).

### DIFF
--- a/gix-path/src/env/git/mod.rs
+++ b/gix-path/src/env/git/mod.rs
@@ -140,8 +140,10 @@ fn git_cmd(executable: PathBuf) -> Command {
     // scope. Although `GIT_CONFIG_NOSYSTEM` will suppress this as well, passing --system omits it.
     cmd.args(["config", "-lz", "--show-origin", "--name-only"])
         .current_dir(cwd)
-        .env_remove("GIT_COMMON_DIR") // We are setting `GIT_DIR`.
         .env_remove("GIT_DISCOVERY_ACROSS_FILESYSTEM")
+        .env_remove("GIT_OBJECT_DIRECTORY")
+        .env_remove("GIT_ALTERNATE_OBJECT_DIRECTORIES")
+        .env_remove("GIT_COMMON_DIR")
         .env("GIT_DIR", NULL_DEVICE) // Avoid getting local-scope config.
         .env("GIT_WORK_TREE", NULL_DEVICE) // Avoid confusion when debugging.
         .stdin(Stdio::null())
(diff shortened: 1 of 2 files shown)


Byron (approved):
Thanks so much, I really appreciate your work (and sometimes feel I don't make that explicit often enough 😅).

> There is more information in the commit messages, especially the one prefixed `fix:`, [eb72d31](https://github.com/Byron/gitoxide/commit/eb72d31a64b5041170e5903c328b246dab522067). This also adds tests related to `GIT_CONFIG`, in the two commits that precede that, [340ff37](https://github.com/Byron/gitoxide/commit/340ff371d181686afd80582f7d5c8f340aabd24c) and [f8e38d0](https://github.com/Byron/gitoxide/commit/f8e38d0575e7e898133aa471539ad79b7508c88e).

I would only indicate breaking changes if the API breaks - changes in behaviour could qualify, of course, but thus far I didn't encounter them.

PR #1584, opened 2024-09-06
Don't test `jj_realistic_needs_to_be_more_clever` on Windows

EliahKagan:
This conditionally marks the `jj_realistic_needs_to_be_more_clever` test as ignored, when the target is Windows. But the test is still built on Windows, and can be run with:

```powershell
cargo nextest run -p gix -- --ignored jj_realistic_needs_to_be_more_clever
```

This is a workaround for #1575.

When run on Windows, the test previously checked if it was running on CI and skipped its contents. This removes that logic, since the test no longer runs by default on CI or otherwise.

### DIFF
--- a/gix/tests/object/tree/diff.rs
+++ b/gix/tests/object/tree/diff.rs
@@ -392,13 +392,12 @@ mod track_rewrites {
     }
 
     #[test]
+    #[cfg_attr(
+        windows,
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot! Using a conditional `ignore` is so much better! Now that no runtime information is used anymore, this works.

Issue #1432, opened 2024-06-28
ssh clone does not correctly detect the location of ssh.exe


Mr-Ao-Dragon:
### Current behavior 😯

[![asciicast](https://asciinema.org/a/9mxSMvPg80KJu2biphHxWoNt3.svg)](https://asciinema.org/a/9mxSMvPg80KJu2biphHxWoNt3)

Byron:
Could you also show what Git does with `git clone …`? It might help to run it with `GIT_TRACE=1`.
Additionally, it would be good to see SSH related configuration of the output of `gix config | grep -i ssh`, if any is available (but probably there isn't).

Oh, and learning where `git.exe` and `ssh.exe` are located would also help.

Otherwise I could imagine that it finds `ssh.exe` somehow and all by itself, but in the wrong place:

https://github.com/Byron/gitoxide/blob/3e5c974dd62ac134711c6c2f5a5490187a6ea55e/gix-command/src/lib.rs#L221

The reason it has to do its own PATH lookup is to prevent Windows to launch executables that are in the current working directory, which is a blunder

My feeling is that it should be able to find `ssh.exe` from `git --exec-path`, but doesn't because it only searches `PATH`.

Also CC @EliahKagan

EliahKagan:
#### General Windows, Rust, and gitoxide path search subtleties

> The reason it has to do its own PATH lookup is to prevent Windows to launch executables that are in the current working directory, which is a blunder

It's true that the underlying [`CreateProcessW`](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw) function will search in the current directory before searching in `PATH` when running a program specified with no `\`. Fortunately, since https://github.com/rust-lang/rust/pull/87704, `std::process::Command` has performed its own path search to resolve full paths, so that what actually gets passed to `CreateProcessW` is safe.

For this reason, I am not sure if any of the manual `PATH` searching done in gitoxide is actually needed for security. However, I am reluctant to rush to change it, because:

[...]

PR #1585, opened 2024-09-07
octopus-merge (part 2: blob-merge)

Byron:
Implement an octopus merge based on trees, and (mostly) equivalent to `merge-ORT` in Git. The foundation, tree-editing, was implemented in #1566.

Related to https://github.com/gitbutlerapp/gitbutler/pull/4793.

### Tasks

This PR was re-focussed on blob-based conflict detection and the generation of blob merge-results with conflict markers.
* [x] sketch API and document all knowledge from Git docs through it
    - [x] ~~flatten `merge` into `platform` API, similar to how it's done in `diff`~~ Bring `merge::State` to being able to execute the merge.
* [x] `Pipeline` implementation and tests
    - [x] renormalization also works
    - [x] refactor to only lookup attributes once, on platform, as only a single path counts.
* [x] `Platform` implementation and tests
    - [x] generalize merge-command invocation - lots of context must be passed
        - [x] See if shell-quoting of the `%P` argument for merge-drivers is needed for it to get through to the driver program
        - [x] Allow to pass additional environment variables so `GIT_REFLOG_ACTION`, `GIT_EXEC_PATH` and `GIT_PREFIX` can be passed
    - [x] Can 'zealous' conflict resolution (or information) be part of that, whether it's used or not? Is it decided here?
        - Yes, we *could* provide access to our internal hunk information, but it seems like something that isn't desirable just yet.
  * [x] binary merge
      - [x] with large file and binary file support (i.e. it doesn't fail due to these conditions)
  * [x] text merge
     - [x] resolve conflicts with union, ours and theirs
     - [x] `merge` style
     - [x] `diff3` style
     - [x] `zdiff3` style
     - [x] properly deduce line endings
     - [x] integrate tests for `diff3-conflict-markers.sh` (some special cases)
         - ~~figure out `common` case, where single-line changes are merged with the common middle, probably unrelated to the actual diff as zdiff3 and diff3 work.~~ - skipped in interest of time
         - ~~Note that 'union' is also affected by this, which seems detremental~~ - skipped in interest of time
     - [x] tests for newline handling (in baseline)
     - [x] integrate tests for `merge-file` (less simplistic input)
     - [x] How are newlines written in the `<<<` lines that the text merge is adding? Probably configurable?
          - It's based on looking at the [hunks before and after](https://github.com/git/git/blob/master/xdiff/xmerge.c#L182-L194)
          - [x] a test or two specifically for newline handling - we are deviating quite a bit - would need to check Git code to do this right, skipping in the interest of time as this is probably minor.
     - [x] What about deletions when zealously handling hunks? Have a test for that!
  * [x] assure tests run into every condition - only one place is missing that the code doesn't run into, and for now I don't feel bold enough to make it an assertion.
* [x] validate that a difference in line-ending style will clash (it should, tokens-with-line-endings are used) (also for `gix-diff`.

### Questions

* [x] How to deal with MERGE_HEAD and additional files the merge might create?
    - For merging commits only
* [x] Does Git provide binary data to external merge programs (and diff drivers, for that matter?)
    - Yes, and it also does that for files that are too large (larger than 1023MB). It has no problem reading them from disk to memory though, which is inefficient, and we should do better.

### Next PR / Outscoped

* `textconv` with context, see [this gist](https://gist.github.com/EliahKagan/31517ea8d4bb0300a521eb63c6b2efd2) for details.
    - There seem to be different 'tiers' of tools, some don't get `GIT_DIR` set, others do.
    - It also seems that diff-programs get too much context right now, but that depends on how much is passed to them by the caller as `gix-command::Context`.
* How to model virtual-merge-bases? Can be none or many, user should have control over how this is done.
    - there is tests for that in `diff3-conflict-markers`
* Understand how conflicts are currently handled in GitButler.
* Actual tree-based merging
    - See also, all the [merge options](https://git-scm.com/docs/git-config#Documentation/git-config.txt-mergedirectoryRenames)
    - Submodule merges are also possible! Maybe outscope it though! `libgit2` also doesn't try it.
    -  `diff3-conflict-markers.sh`  - be sure to capture the 'empty tree' label , but also other special cases
* Make blob-merge based conflict handling possible in the tree merge from `gix` at least.

### Research

Everything is about MergeORT.

* It puts conflict-markers in the result tree, with annotations to provide additional context
* data stored by path, and is interned in the map to allow pointer-based comparisons
    - merge-info with everything one needs to know, also related to renames
    - or conflict information
    - it uses a memory-pool/arena to get memory for many paths all at once (and also release it like that)
* paths start out as conflicted, and then can later be changed to non-conflicting if a content-based merged succeeded.
    - If it remains conflicts, the meta-data is used to produce an 'as merged as possible' version with conflict markers that can be checked out to the working tree.
* hunks can partially overlap, but can also be resolved [line-by line to some extend](https://neg4n.dev/blog/understanding-zealous-diff3-style-git-conflict-markers).

### Crates

Mostly for performance optimization, also interesting for handling the index, and the tree::editor types.

#### Building Blocks for String Interning

* https://lib.rs/crates/bumpalo - re-use one allocation for multiple strings, but makes it hard to keep arena internal (see `yoke`.
    - https://lib.rs/crates/id-arena
    - https://lib.rs/crates/scoped-arena
    - **but it's not trivial to build interned string maps with that**
* https://docs.rs/slab/latest/slab - Similar to `bumpalo`, but returns an index to the item, can also remove things. Strings aren't really supported though (i.e. variable length slices)
* https://docs.rs/yoke/latest/yoke/ - make it possible to store something with its backing in the same struct.

#### String Interining

* https://lib.rs/crates/lasso - but only for `&str`
* ~~https://lib.rs/crates/interned~~ - also for `&[u8]` - but leaks the data to get pointer comparison
* https://lib.rs/crates/internment - with `ArenaInterner` also for `[u8]`
    - but seems to always use a mutex internally (double-check for AreanInterner)
    - but still allocates individual items on the heap
    - ✔️ - could be the basis for a version that uses a bump-allocator

### Handle Special Cases

* A file was renamed differently
* deal with "merge.directoryRenames"

#### Is `git2::merge_trees()` a trivial merge? Does it handle all the cases of MergeORT?

Maybe not, but it definitely handles rename tracking. See [these `git2` flags](https://github.com/libgit2/libgit2/blob/403a03b3beaea7d26b9515e27dd36553239647ca/include/git2/merge.h#L69-L102) for more information.

#### How does rename-tracking affect a tree-merge?

The first round is done without renames, then there is a second round to find renames, and perform the merge of renamed items.

#### How is an octopus merge implemented, particularly with Merge ORT?
- recursive mere-base merge to get virtual merge base
- three-way tree iterator merge
- rename tracking

### References

* https://git-scm.com/docs/git-merge-tree
* [trivial three-way merge in Git](https://github.com/git/git/blob/master/builtin/merge-tree.c#L326), visual only, otherwise it uses Merge ORT

<details>

### Removed Utility

Prints a patch with context 3, using bytes only, based on imara-diff.

```Rust
/// An adapted version of the unified diff writer to get a first idea.
        // TODO: remove this
        #[allow(dead_code)]
        mod unified_test {
            use bstr::ByteVec;
            use imara_diff::intern::{InternedInput, Interner, Token};
            use imara_diff::Sink;
            use std::ops::Range;

            /// A [`Sink`] that creates a textual diff
            /// in the format typically output by git or gnu-diff if the `-u` option is used
            pub struct UnifiedDiffBuilder<'a, W>
            where
                W: std::io::Write,
            {
                before: &'a [Token],
                after: &'a [Token],
                interner: &'a Interner<&'a [u8]>,

                pos: u32,
                before_hunk_start: u32,
                after_hunk_start: u32,
                before_hunk_len: u32,
                after_hunk_len: u32,

                buffer: Vec<u8>,
                dst: W,
            }

            impl<'a, W> UnifiedDiffBuilder<'a, W>
            where
                W: std::io::Write,
            {
                /// Create a new `UnifiedDiffBuilder` for the given `input`,
                /// that will writes it output to the provided implementation of [`Write`].
                pub fn with_writer(input: &'a InternedInput<&'a [u8]>, writer: W) -> Self {
                    Self {
                        before_hunk_start: 0,
                        after_hunk_start: 0,
                        before_hunk_len: 0,
                        after_hunk_len: 0,
                        buffer: Vec::with_capacity(8),
                        dst: writer,
                        interner: &input.interner,
                        before: &input.before,
                        after: &input.after,
                        pos: 0,
                    }
                }

                fn print_tokens(&mut self, tokens: &[Token], prefix: char) {
                    for &token in tokens {
                        self.buffer.push_char(prefix);
                        self.buffer.extend_from_slice(self.interner[token]);
                    }
                }

                fn flush(&mut self) {
                    if self.before_hunk_len == 0 && self.after_hunk_len == 0 {
                        return;
                    }

                    let end = (self.pos + 3).min(self.before.len() as u32);
                    self.update_pos(end, end);

                    writeln!(
                        &mut self.dst,
                        "@@ -{},{} +{},{} @@",
                        self.before_hunk_start + 1,
                        self.before_hunk_len,
                        self.after_hunk_start + 1,
                        self.after_hunk_len,
                    )
                    .unwrap();
                    self.dst.write_all(&self.buffer).unwrap();
                    self.buffer.clear();
                    self.before_hunk_len = 0;
                    self.after_hunk_len = 0
                }

                fn update_pos(&mut self, print_to: u32, move_to: u32) {
                    self.print_tokens(&self.before[self.pos as usize..print_to as usize], ' ');
                    let len = print_to - self.pos;
                    self.pos = move_to;
                    self.before_hunk_len += len;
                    self.after_hunk_len += len;
                }
            }

            impl<W> Sink for UnifiedDiffBuilder<'_, W>
            where
                W: std::io::Write,
            {
                type Out = W;

                fn process_change(&mut self, before: Range<u32>, after: Range<u32>) {
                    if before.start - self.pos > 6 {
                        self.flush();
                        self.pos = before.start - 3;
                        self.before_hunk_start = self.pos;
                        self.after_hunk_start = after.start - 3;
                    }
                    self.update_pos(before.start, before.end);
                    self.before_hunk_len += before.end - before.start;
                    self.after_hunk_len += after.end - after.start;
                    self.print_tokens(&self.before[before.start as usize..before.end as usize], '-');
                    self.print_tokens(&self.after[after.start as usize..after.end as usize], '+');
                }

                fn finish(mut self) -> Self::Out {
                    self.flush();
                    self.dst
                }
            }
        }
    }
```

</details>


Byron:
@EliahKagan Just as a note in case you'd like to submit a first patch to Git that will definitely be able to land, maybe to warm-up with the mailing-list workflow, then I have something for you.

The [attribute-documentation about three-way merges](https://git-scm.com/docs/gitattributes#_performing_a_three_way_merge) contains  a reference to `merge.default` which is indeed [used in code](https://github.com/git/git/blob/c5ee8f2d1c9d336e0a46139bd35236d4a0bb93ff/merge-ll.c#L283) as well. 

However, the [git-config documentation](https://git-scm.com/docs/git-config#Documentation/git-config.txt-mergedefaultToUpstream) doesn't mention it at all.

gix-diff/src/blob/pipeline.rs
```diff
@@ -533,6 +541,10 @@ impl Driver {
     pub fn prepare_binary_to_text_cmd(&self, path: &Path) -> Option<std::process::Command> {
         let command: &BStr = self.binary_to_text_command.as_ref()?.as_ref();
         let cmd = gix_command::prepare(gix_path::from_bstr(command).into_owned())
+            // This program intentionally doesn't get the whole Git context.
+            // In fact, when checked it only saw the `GIT_EXEC_PATH` and `COLORTERM` environment variables, nothing more.
+            // We by default let it inherit the entire environment. On Windows, Git seems to add the env though.
```
Byron, line 546: @EliahKagan I just pushed this intermediate commit to share this comment with you. I just noticed that `gix-command` may provide too much of the parent process environment to the spawned commands, and thus is less hygienic than Git seems to be.
Maybe attackers can abuse this, or even if it's not an attacker, malevolent players of developer software might use it to syphon off information from developer machines that Git doesn't provide to them.

My feeling is the `gix-command` should run `env_clear()` beforehand and then selectively re-add variables to match Git. That, however, probably has to be checked on a case-by-case basis.

Byron: The merge driver program, similar to the diff-driver (which isn't the `textconv` program) seem to see the inherited environment + the Git context. So it looks like the textconv is special, but technically still sees too much of the environment right now when invoked from `gix-diff`.

EliahKagan: What sanitization does `git` do here that we are missing? At least with `diff.*.command` and `diff.*.textconv`, I have not been able to observe any removed environment variables. [I did not find the environment to be sanitized on any platform.](https://gist.github.com/EliahKagan/31517ea8d4bb0300a521eb63c6b2efd2) As shown in that gist, I ran `printenv`:

- Directly.
- Through a custom Git command.
- Through a custom Git diff script.
- Through a custom Git textconv script.

All environment variables, whether Git-related or not, seem to be kept. This was never limited to `GIT_EXEC_PATH` and `COLORTERM`. Several have different values and some new variables are set, especially with Apple Git on macOS, but I did not observe any to be removed.

This was the case even when I set `GIT_DIR` and `GIT_WORK_TREE`, which are repository-specific and which are among the environment variables `git` clears when performing an operation on a different repository. However, I did not do that test on all platforms; in particular, I didn't get a chance to do it on macOS.

For simplicity, I did not test with explicitly specified merge tools, since from your description it sounds like diff and merge tools are expected to be treated similarly with respect to environment handling, while textconv tools might be treated differently.

There are [a number of other variables](https://github.com/git/git/blob/57974d46a4d5a079471c4f7eaa5933b1922299e4/environment.c#L128-L148), some [repository-specific](https://github.com/git/git/blob/57974d46a4d5a079471c4f7eaa5933b1922299e4/environment.c#L128-L148) and others not, that Git treats specially and I wouldn't be surprised if it unset them under some circumstances. But before testing with them, I figured I'd ask for clarification.

> or even if it's not an attacker, malevolent players of developer software might use it to syphon off information from developer machines that Git doesn't provide to them

I'm not sure what kind of threat you're describing here. Certainly if one expects some kind of sanitization to occur and it does not, then one might end up causing sensitive information to become available to a component one wishes to limit, or causing untrusted information *from* a component one wished to be limited to be used where it might be wrongly trusted. But that is completely general. If you are talking about malevolent development tools that have access to repositories or can run arbitrary commands, then I don't think this would affect their abilities.

I do recommend that any differences in sanitization from what `git` does, especially where `git` removes an environment variable and we do not, be considered potential vulnerabilities and investigated. But I haven't been able to observe that with diff and textconv tools.

> My feeling is the `gix-command` should run `env_clear()` beforehand and then selectively re-add variables to match Git. That, however, probably has to be checked on a case-by-case basis.

Except in circumstances where Git does that, this does not sound like it would necessarily be an improvement. Since the user may configure arbitrary diff and textconv tools, the user may configure a tool that takes configuration from arbitrary environment variables. If a tool uses environment variables that start with a widely recognized prefix, such as `GIT_`, and uses them in a way that is inconsistent with the way they are commonly used, then that tool is broken. Otherwise, though, sanitizing the environment unexpectedly seems like it could harm usability and correctness, and in cases where a less permissive configuration has been overridden using environment variables, may also harm security.

I suspect I may not fully understand what you are proposing, however.

Byron: Thanks so much for the detailed analysis - I didn't expect this, but am happy about such a large sample size.

It's astonishing that no environment was sanitised, even though the `textconv` tool in my run definitely didn't have any but two environment variables set.
Something very relevant I forgot to share is my using of a very recent Git executable directly built from `master` (c5ee8f2d1c9d336e0a46139bd35236d4a0bb93ff). It might be possible that something changed.
*Actually, scratch that, I just retried with the standard apple git installation, and it's the same result*.

This is the textconv filter program I am using:

```bash
#!/bin/bash

env
pwd
echo $@
```

> I'm not sure what kind of threat you're describing here.

The best description I can come up with is the loss of confidentiality, along the lines of developer tools making an extra buck by selling developer data. But it's true, if the program is running, some additional environment variables probably are the smallest thing one would worry about.

> Except in circumstances where Git does that, this does not sound like it would necessarily be an improvement. 

Thanks for sharing your thoughts. For now I was just passing this concern by you to see what you think, and it's good to see I can also be a little bit too paranoid. However, I'd definitely keep sharing even the smallest concerns, maybe even as practice, to maybe one day write secure code from the start 😅.

Then I will take no action, but instead pass the standard Git context to everything as it seems to have been passed as well in your examples.

Byron: 🤦‍♂️

I think I know now what happened here. This is the output I was looking at.

```
❯ git --no-pager diff HEAD^ HEAD
diff --git a/file b/file
index 1f7391f..67b7f61 100644
--- a/file
+++ b/file
@@ -49,4 +49,4 @@ WEZTERM_CONFIG_FILE=/Users/byron/.wezterm.lua
 GIT_EXEC_PATH=/Applications/Xcode.app/Contents/Developer/usr/libexec/git-core
 COLORTERM=truecolor
 /Users/byron/dev/github.com/Byron/small
-/var/folders/x7/jgy95vjs3v3ffszwn0wg0gz80000gn/T//git-blob-GMpYec/file
+/var/folders/x7/jgy95vjs3v3ffszwn0wg0gz80000gn/T//git-blob-490Xcv/file
```

After all, I wasn't outputting the information to a separate file, but *as* the file to be diffed.

Hence, what I saw was a patch with most information the same, 50 lines of that… my testing is entirely broken, and so are the results.

So it doesn't only inherit the existing env, it also adds the `git-core` tools to the `PATH`.

EliahKagan: > Thanks so much for the detailed analysis - I didn't expect this, but am happy about such a large sample size.

I usually test with only something like [`git-run`](https://gist.github.com/EliahKagan/31517ea8d4bb0300a521eb63c6b2efd2#file-script-1-git-run-sh), and usually only to observe what environment Git for Windows prepares. This was an opportunity to see, to some extent, how similar the environment for custom Git commands really is to the environment for external tools. So I was also motivated by that goal.

Accordingly, even though the cause of the environment seeming to be sanitized has been discovered, I have gone ahead and repeated the experiment with `GIT_DIR` and `GIT_WORK_TREE` set [on macOS](https://gist.github.com/EliahKagan/31517ea8d4bb0300a521eb63c6b2efd2#macos---with-git_dir-and-git_work_tree-set).

Of course, there are still specific additions and removals that this does not reveal, and some of them may be relevant to compatibility goals for gitoxide. For example, I have separately observed that, at least in recent versions of Git, the `GIT_CONFIG_COUNT` variable is removed from the environment for an external transport command such as `ssh`, even when it is present in the immediate parent of that transport command. I'm not sure why that is done, if it is important, or if `gix-transport` should behave similarly. I imagine there are a number of specific behaviors like this.

For `gix-transport`, it seems [already to remove a number of variables](https://github.com/Byron/gitoxide/blob/ab8880fbdf0f7af7b483f4e1f9adbb8e374183ee/gix-transport/src/client/blocking_io/file.rs#L17-L35). This has long included `GIT_CONFIG_COUNT`, which was added in https://github.com/Byron/gitoxide/commit/aa315b4a136bec57a1cce4b245c606904adc5c12. But I am unsure if it the list there is entirely current.

#### Whether we use a shell or not is significant to compatibility

Maybe a more significant compatibility concern, including for `gix-transport`, is that there seem to be some configurable external commands where whether or not a shell is used to run them is part of the semantics of how they are configured, but gitoxide does not currently honor those semantics.

For example, when `git` runs a SSH client from `GIT_SSH`, it never uses a shell, and when it runs a SSH client command from `GIT_SSH_COMMAND` (or `core.sshCommand`), it always uses a shell. [This is documented](https://git-scm.com/docs/git#Documentation/git.txt-codeGITSSHcode) and I think the `git` documentation intends that users be able to rely not just on `GIT_SSH_COMMAND` using a shell, but also on `GIT_SSH` either not using a shell or behaving impeccably as though it does not. I think that gitoxide not always adhering to this may actually be the cause of #1432.

Even if the effect of shell processing triggered by [special characters](https://github.com/Byron/gitoxide/blob/ab8880fbdf0f7af7b483f4e1f9adbb8e374183ee/gix-command/src/lib.rs#L101) in the original command name is correct, whether a command is run directly or through a shell can have important effects, at least on Windows where whether a command is run directly or through another program can have important effects even when that other program is not a shell.

This is because, on Windows, [programs with different "bit-ness" see some parts of the filesystem differently](https://learn.microsoft.com/en-us/windows/win32/winprog64/file-system-redirector). When `ssh` is `C:\Windows\System32\OpenSSH\ssh.exe` or `GIT_SSH` points there, and `gix` is a 64-bit build but the shell `gix-command` finds is a 32-bit program, then when it uses a shell, it will fail to find a usable `ssh.exe`, because there is no `C:\Windows\SysWOW64\OpenSSH\ssh.exe`.

A shell may be used more often on Windows when the caller permits it, because an environment variable like `GIT_SSH` may contain `\` as a path separator, which [is matched here](https://github.com/Byron/gitoxide/blob/ab8880fbdf0f7af7b483f4e1f9adbb8e374183ee/gix-command/src/lib.rs#L101). This may, in particular, lead to attempts to work around the problem by adding more backslashes, which may sometimes work, since suitably repeated non-leading backslashes cause backslashes to be treated literally by shells while being equivalent to single backslashes when a shell is not used, which is also consistent with the description in #1432.

#### The environment may be relevant to deniability for data exfiltration

> The best description I can come up with is the loss of confidentiality, along the lines of developer tools making an extra buck by selling developer data.

Unless isolated in a way that would presumably also sanitize the environment, I would expect such a tool could obtain that information if it has fairly unrestricted disk access compared to the user running it.

On the other hand, this may be a real concern in that including all environment variables and their values in telemetry or crash reporting data might not be identified as a malicious behavior.

I think the best protection against that, though, is to sanitize the environment before we are ourselves doing anything with it, rather than afterwards, and preferably to sanitize it before we even have access to it. That is, in the best case, when using a program to run tools whose environments one wishes to sanitize, one would sanitize the environment for that program. For example, one might run `gix` or `ein` (or `git`) with a sanitized environment if one wishes to use it to run tools whose environments should be limited.

In particular, there seems to be some overlap in environment variables one might want to sanitize and those one might need for a custom tools. For example, since a textconv tool is [typically lossy](https://git-scm.com/docs/gitattributes#_performing_text_diffs_of_binary_files), it may sometimes make sense for it to use a service that accesses real-time data, or that uses a machine learning model that might be infeasible or inconvenient to run locally. (For example, an image could be presented as an automatically generated caption describing it.) But these are also two kinds of use cases where one often has an API key in an environment variable that one would wish the tool to access and use.

So I don't know how much we can help with this, other than by doing all the sanitization `git` documents it does, or otherwise does, under similar circumstances, and by improving compatibility with sanitization that has already been performed by attempting to avoid malfunctioning even if our own environment is unusually sparse.

#### Miscellaneous

> However, I'd definitely keep sharing even the smallest concerns

Yes, definitely please do!

> After all, I wasn't outputting the information to a separate file, but *as* the file to be diffed.
>
> Hence, what I saw was a patch with most information the same, 50 lines of that… my testing is entirely broken, and so are the results.

Ah, that makes sense. (It was to avoid this sort of confusion that I had [`dumpenv`](https://gist.github.com/EliahKagan/31517ea8d4bb0300a521eb63c6b2efd2#file-script-2-dumpenv-sh) always dump the environment to a file of a fixed name in the current directory.)

> So it doesn't only inherit the existing env, it also adds the `git-core` tools to the `PATH`.

On macOS, at least on the system I tested on, some other environment variables were also added.

(The situation on Windows may be dissimilar and less relevant. On Windows, I tested in in a Git Bash environment so that I could run the exact same commands, and I believe there is some MSYS path normalization logic that affected what `git` itself was seeing, so some of the other changes may not have been done by `git`.)

#### `PATH`, but also `GIT_CONFIG_PARAMETERS`, are significant to compatibility

> It seems that some work will be needed to assure the environments are the same, for best compatibility.

I don't know what we should do, and I think it might even be best to deliberately refrain from doing so by default, maybe with the ability to change this behavior through `gitoxide.*` configuration variables, environment variables, or both.

For prepending the path of the `git-core` directory to `PATH` , I think the disadvantages of us doing that by default are:

1. There is no corresponding directory for a tool that performs Git operations with a `gix*` crate. It will not be "our" `git-core` directory, and the tools in it will not do anything to integrate back with gitoxide.
2. There are some guarantees relevant to safety that Git can provide and we cannot. Because we may be running on a system that does not have Git installed, or where we have not found Git, we cannot always prepend a `git-core` directory to `PATH`. But developers will likely have Git installed with `git` in their `PATH`, so this will make it very easy to write bugs that assume `PATH` starts with a `git-core` directory.
3. The consequences of guessing the wrong `git-core` directory may be worse than those of reading "installation" configuration from a different `git` installation than the one currently being used. In particular, guessing the wrong "installation" configuration is likely to be discovered sooner than guessing the wrong `git-core` directory, whose effects are minimal until a subprocess run via `gix-command` runs `git` or commands found only in `git-core`. Then again, this may be possible to avoid by unifying the way we determine both locations.
4. We do not currently support other environment changes made by `git`. We may be called *by* `git`. This is relevant both when we when have a descendant `git` process and when don't.

Regarding point 4, I am aware of two cases I believe to be significant:

- One is a a special case of the above point 3 above. If we prepend a `git-core` directory to `PATH` based on our search for `git`, we need to make sure we always find the `git` that called us. Ordinarily this would happen automatically, but I think it could go wrong in some cases with the `EXEPATH` optimization on Windows.

- Another case is when we are called in a custom `git` command, or otherwise as a tool used by `git`, and the `-c` option has been passed to `git` to set command scope configuration. `git` uses the `GIT_CONFIG_PARAMETERS` environment variable to pass such configuration to subprocesses--both those that are conceptually part of `git`, and custom scripts. With [that `git-run`](https://gist.github.com/EliahKagan/31517ea8d4bb0300a521eb63c6b2efd2#file-script-1-git-run-sh):

  ```text
  ek@Glub:~$ git -c foo.bar=baz run printenv | sort | grep ^GIT
  GIT_CONFIG_PARAMETERS='foo.bar'='baz'
  GIT_EXEC_PATH=/usr/lib/git-core
  ```

  But `gix-config` does not recognize `GIT_CONFIG_PARAMETERS`, as can be verified by running `git -c foo.bar=baz run gix config` in a repository and inspecting the output.

Then again, point 4 may be less of a disadvantage than an opportunity:

- If `PATH` starts with a `git-core` directory, and inspection of the directory reveals it has the expected contents, then that could probably be used to infer the same information we get on Windows from `EXEPATH` when `EXEPATH` is defined. Checking this before checking `EXEPATH` would reduce the risk of guessing wrong based on `EXEPATH`.

- Maybe gitoxide should recognize `GIT_CONFIG_PARAMETERS`. It seems to me that this would be a nice feature. But it should possibly be opt-in, since while the parsing rules for `GIT_CONFIG_PARAMETERS` will *probably* not change in an incompatible way, it is really an implementation detail.

  ([`GIT_CONFIG_{COUNT,KEY,VALUE}`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-GITCONFIGCOUNT) was introduced in part because the parsing rules for `GIT_CONFIG_PARAMETERS` are nonobvious and hard to handle manually, rather than out a desire to be able to make incompatible changes to `GIT_CONFIG_PARAMETERS`. But [the commit](https://github.com/git/git/commit/d8d77153eafdb0fc334e827976f09e4bdff26b58) that introduced `GIT_CONFIG_{COUNT,KEY,VALUE}` noted that `GIT_CONFIG_PARAMETERS` is "for internal use only," which seems still to be the case since it remains [undocumented](https://git-scm.com/docs/git-config#ENVIRONMENT) in [git-config(1)](https://git-scm.com/docs/git-config) and [seems not](https://git-scm.com/search/results?search=GIT_CONFIG_PARAMETERS) to be mentioned at all in any of the Git documentation.)

  That it is an implementation detail suggests we should not use it in any way, even to obtain configuration from it when present. But if we don't obtain configuration from it, then we ignore command-scope configuration passed via `-c` in a `git` invocation that calls us as a child process, and thus we do not have the *full* interoperability with `git` that prepending a detected `git` installation's `git-core` directory to `PATH` would be done to attempt to achieve.

Maybe gitoxide should have some default level of attempted interoperability with a detected installed `git`--similar to what we do now, looking for and using installation-level configuration but not going much beyond that--and also recognize one or more `GIX_*` environment variables that that allow the level of attempted interoperability to be decreased (not looking for or using anything from the installed `git`), increased (recognizing `GIT_CONFIG_PARAMETERS`, prepending `git-core` to `PATH`), or otherwise adjusted (specifying the path to a `git` installation to be used instead of the current heuristics). Or maybe these should each be individually configured knobs, with no broad "level" aggregating them.

On the other hand, I think there is an argument for reading configuration from `GIT_CONFIG_PARAMETERS` by default, even if none of the rest of these ideas are done. We already look for `git` configuration associated with the installation in ways that are not limited to guarantees provided by Git:

- This is most pronounced in Windows. On Windows, we look for Git in places where it it is installed by default but is not always installed. We assume that if Git is found there and is not in `PATH`, then its configuration should be used. We also avoid some `git` invocations on Windows, where they can be slow, by using information from `EXEPATH`, which I don't think is documented. `EXEPATH` is named generally enough that some other software might be using it incompatibly, though I think the practical risk from that is low.
- But it also applies to other systems. For example, as far as I know, the way `gix-path` obtains installation-level configuration from Apple Git relies on experimentally observed behavior that has not been fully documented. Furthermore, if some other downstream Git distribution added other custom scopes that were lower rather than higher, then it might be broken, even after [#1523](https://github.com/Byron/gitoxide/pull/1523) and [#1567](https://github.com/Byron/gitoxide/pull/1567).

So by gitoxide's current standards of how it obtains configuration, it seems like it should probably obtain configuration from `GIT_CONFIG_PARAMETERS` when present, as well. Furthermore, `GIT_CONFIG_PARAMETERS` is at least somewhat visible. As quoted in #147, changes to it have in the past had Git changelog entries.

Byron: > For `gix-transport`, it seems [already to remove a number of variables](https://github.com/Byron/gitoxide/blob/ab8880fbdf0f7af7b483f4e1f9adbb8e374183ee/gix-transport/src/client/blocking_io/file.rs#L17-L35). This has long included `GIT_CONFIG_COUNT`, which was added in [aa315b4](https://github.com/Byron/gitoxide/commit/aa315b4a136bec57a1cce4b245c606904adc5c12). But I am unsure if it the list there is entirely current.

Thanks again for researching this, and I glad `gix-transport` isn't entirely off when launching `ssh`. Being 90% there seems like a good start. Assuring that is good enough or even safe will be harder though.

> [..] That is, in the best case, when using a program to run tools whose environments one wishes to sanitize, one would sanitize the environment for that program. For example, one might run `gix` or `ein` (or `git`) with a sanitized environment if one wishes to use it to run tools whose environments should be limited.

That makes sense. However, it's also a stance that `gix` (the crate) itself may take, a configurable choice that is inherited by any program that is using it. In that sense, `git` can reasonably assume that users are expecting the entire environment to be inherited, and maybe `gix` (the CLI) should mirror that behavior. Probably there are also other cases when it's better to pass only the most necessary information.
Part of me thinks that I am too hung up on the confidentiality aspect of it, and that it would be best to just do what Git does and to pass any other handling of the environment to the user of `gix`, maybe even explicitly so by making this concern part of the documentation. After all, `gix` (the crate) can't really offer facilities for this as it's unclear what 'sanitized' has to look like.

> So I don't know how much we can help with this, other than by:
> 
> * Doing all the sanitization Git itself performs or documents as recommended, under similar circumstances.
> * Improving compatibility with sanitization that has already been performed by attempting to avoid malfunctioning even if our own environment is unusually sparse.

This convinces me to only document the behaviour, i.e. the attempt to produce similar environments as Git would, with everything else left to the user of the library.

> I don't know what we should do, and I think it might even be best to deliberately refrain from doing so by default, maybe with the ability to change this behavior through `gitoxide.*` configuration variables, environment variables, or both.

Without having read the remainder of the paragraph yet, my intuitive response is to be as compatible to Git as possible, maybe with a way to opt-out.
The reason for that is that tools that use `gix` expect to perform work on Git repositories, and should work similarly to Git by default. It's a selling point of `gitoxide` as a project - be able to 'be Git' but with your own tools.


> For prepending the path of the `git-core` directory to `PATH` , I think the disadvantages of us doing that by default are:
> 
> 1. There is no corresponding directory for a tool that performs Git operations with a `gix*` crate. It will not be "our" `git-core` directory, and the tools in it will not do anything to integrate back with gitoxide.
> 2. There are some guarantees relevant to safety that Git can provide and we cannot. Because we may be running on a system that does not have Git installed, or where we have not found Git, we cannot always prepend a `git-core` directory to `PATH`. But developers will likely have Git installed with `git` in their `PATH`, so this will make it very easy to write bugs that assume `PATH` starts with a `git-core` directory.
> 3. The consequences of guessing the wrong `git-core` directory may be worse than those of reading "installation" configuration from a different `git` installation than the one currently being used. In particular, guessing the wrong "installation" configuration is likely to be discovered sooner than guessing the wrong `git-core` directory, whose effects are minimal until a subprocess run via `gix-command` runs `git` or commands found only in `git-core`. Then again, this may be possible to avoid by unifying the way we determine both locations.
> 4. We do not currently support other environment changes made by `git`. We may be called _by_ `git`. This is relevant both when we when have a descendant `git` process and when don't.

I see that finding the right `git` executable and `git-core` location would be relevant to correctness, and that it's not trivial to do so in all the cases. Maybe you have ideas on how to collect the knowledge presented here in a tracking issue so we don't loose it? If you say you have it in your mind or in your own notes, that would already be good enough for me, I don't want to cause unnecessary overhead.

> Maybe gitoxide should have some default level of attempted interoperability with a detected installed `git`--similar to what we do now, looking for and using installation-level configuration but not going much beyond that--and also recognize one or more `GIX_*` environment variables that that allow the level of attempted interoperability to be decreased (not looking for or using anything from the installed `git`), increased (recognizing `GIT_CONFIG_PARAMETERS`, prepending `git-core` to `PATH`), or otherwise adjusted (specifying the path to a `git` installation to be used instead of the current heuristics). Or maybe these should each be individually configured knobs, with no broad "level" aggregating them.

I like the idea of planning to increase this kind of compatibility but also to provide configuration for it. The way to do this would be to create a `gitoxide.*` configuration key that has an environment variable associated with it. It is then picked up automatically, written into in-memory git-configuration, and available by the code that would have to prepare the command-context.

> On the other hand, I think there is an argument for reading configuration from `GIT_CONFIG_PARAMETERS` by default, even if none of the rest of these ideas are done. We already look for `git` configuration associated with the installation in ways that are not limited to guarantees provided by Git: [..]

My intuition here, still, is to aim for max compatibility and do as much as possible, even if it means to read `GIT_CONFIG_PARAMETERS` and rely on an implementation detail when trying to decode them.

> So by gitoxide's current standards of how it obtains configuration, it seems like it should probably obtain configuration from `GIT_CONFIG_PARAMETERS` when present, as well. Furthermore, `GIT_CONFIG_PARAMETERS` is at least somewhat visible. As quoted in #147, changes to it have in the past had Git changelog entries.

And that's a great find! The parameter parser [is here](https://github.com/gitext-rs/git-config-env/blob/main/src/param.rs) and it seems it would be straightforward to adapt it to not rely on `String`, and to use `BString` instead.

EliahKagan: > Maybe you have ideas on how to collect the knowledge presented here in a tracking issue so we don't loose it?

Yes, I will open at least two issues related to this: one issue specifically about `GIT_CONFIG_PARAMETERS`, because there are some further subtleties to iron out for that, and also because I think supporting that would be valuable even independently of these other compatibility considerations; and at least one issue for the other things.

I've also noticed #1432 has been closed. Since my hypothesis about the cause is not confirmed, and there may not be enough information there to know the cause, I am disinclined to reopen that (unless the reporter were to decide to reopen it). Instead, I can either include the shell bug in a combined issue or open one specifically for it, whichever seems best once I've researched it slightly more.

EliahKagan: A brief note, mostly to myself: When I open the issues mentioned above in https://github.com/GitoxideLabs/gitoxide/pull/1585#discussion_r1764341739, I'll also include something about the environment variable handling compatibility issue mentioned in https://github.com/GitoxideLabs/gitoxide/pull/1620#discussion_r1797634536.

PR #1590, opened 2024-09-11
Have CI check if we need to commit regenerated archives

EliahKagan:
This adds the ability of `gix-testtools` to generate archives of fixture output even when the environment is detected to be CI. This behavior is turned off by default, but enabled if `GIX_TEST_CREATE_ARCHIVES_EVEN_ON_CI` is set. A small amount of refactoring is included to avoid decreasing code readability while making this change.

Although there are a few plausible use cases for this, which along with other information are detailed in commit messages, the motivating use case is to let us use CI to find out if generated archives are out of date. This is something that has happened a number of times, and is not always immediately noticed, because it has had to be observed in the state of a repository after running tests locally.

On this project's CI, this newly recognized environment variable is:

- Left unset in the `test` job.
- Set in the `test-fast` jobs. The `test-fast` jobs will use existing generated archives if available, because `GIX_TEST_IGNORE_ARCHIVES` is not set for them, so this should only cause more work to be done on CI in situations where there something is wrong.

In addition to setting `GIX_TEST_CREATE_ARCHIVES_EVEN_ON_CI` for the `test-fast` jobs, this also adds a new step at the end to run `git diff --exit-code`. That actually detects if any new archives were create. This is because, in practice, as far as I know, they always end up being at least slightly different from whatever existed before for them, if anything.

There may be slightly more efficient ways to detect this, and maybe even slightly more reliable ways to detect it, but I think it makes sense to start with this way, because:

- As noted in commit messages, there are some other plausible applications of the new environment variable anyway, including if other projects that use `gix-testtools` ever want to use CI to generate their archives and save them as artifacts for reuse.
- This automates what is, in effect, the exact workflow that happens locally, for noticing that running the test suite produces a new or changed generated archive.

---

**As of now, CI will fail** on this PR, because there is an archive that needs to be regenerated. The new step can be expected [to fails with](https://github.com/EliahKagan/gitoxide/actions/runs/10803415921/job/29967218978):

```text
Run git diff --exit-code
  git diff --exit-code
  shell: /bin/bash -e {0}
  env:
    CARGO_TERM_COLOR: always
    CLICOLOR: 1
    CARGO_HOME: /Users/runner/.cargo
    CARGO_INCREMENTAL: 0
    CACHE_ON_FAILURE: false
diff --git a/gix-worktree-state/tests/fixtures/generated-archives/make_dangling_symlink_to_windows_reserved.tar b/gix-worktree-state/tests/fixtures/generated-archives/make_dangling_symlink_to_windows_reserved.tar
index 3d9c416..4d50761 100644
Binary files a/gix-worktree-state/tests/fixtures/generated-archives/make_dangling_symlink_to_windows_reserved.tar and b/gix-worktree-state/tests/fixtures/generated-archives/make_dangling_symlink_to_windows_reserved.tar differ
```

If #1589 is merged to main, and then this PR is rebased onto main (or main is merged into it afterwards), CI here should pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,9 +92,13 @@ jobs:
         with:
           tool: nextest
       - name: "Test (nextest)"
+        env:
+          GIX_TEST_CREATE_ARCHIVES_EVEN_ON_CI: 1
         run: cargo nextest run --all --no-fail-fast
       - name: Doctest
         run: cargo test --doc
+      - name: Check that tracked archives are up to date
+        run: git diff  # If this fails, the fix is usually to commit a regenerated archive.
 
   test-32bit:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,7 +98,7 @@ jobs:
       - name: Doctest
         run: cargo test --doc
       - name: Check that tracked archives are up to date
-        run: git diff  # If this fails, the fix is usually to commit a regenerated archive.
+        run: git diff --exit-code  # If this fails, the fix is usually to commit a regenerated archive.
 
   test-32bit:
     runs-on: ubuntu-latest
--- a/tests/tools/src/lib.rs
+++ b/tests/tools/src/lib.rs
@@ -569,7 +569,7 @@ fn scripted_fixture_read_only_with_args_inner(
                     output.stdout.as_bstr(),
                     output.stderr.as_bstr()
                 );
-                create_archive_if_not_on_ci(
+                create_archive_if_we_should(
                     &script_result_directory,
                     &archive_file_path,
                     script_identity_for_archive,
@@ -644,7 +644,7 @@ fn is_lfs_pointer_file(path: &Path) -> bool {
 
 /// The `script_identity` will be baked into the soon to be created `archive` as it identifies the script
 /// that created the contents of `source_dir`.
-fn create_archive_if_not_on_ci(source_dir: &Path, archive: &Path, script_identity: u32) -> std::io::Result<()> {
+fn create_archive_if_we_should(source_dir: &Path, archive: &Path, script_identity: u32) -> std::io::Result<()> {
     // On Windows, we fail to remove the meta_dir and can't do anything about it, which means tests will see more
     // in the directory than they should which makes them fail. It's probably a bad idea to generate archives on Windows
     // anyway. Either Unix is portable OR no archive is created anywhere. This also means that Windows users can't create
--- a/tests/tools/src/lib.rs
+++ b/tests/tools/src/lib.rs
@@ -642,12 +642,12 @@ fn is_lfs_pointer_file(path: &Path) -> bool {
         .map_or(false, |_| buf.starts_with(PREFIX))
 }
 
-/// The `script_identity` will be baked into the soon to be created `archive` as it identitifies the script
+/// The `script_identity` will be baked into the soon to be created `archive` as it identifies the script
 /// that created the contents of `source_dir`.
 fn create_archive_if_not_on_ci(source_dir: &Path, archive: &Path, script_identity: u32) -> std::io::Result<()> {
-    // on windows, we fail to remove the meta_dir and can't do anything about it, which means tests will see more
-    // in the directory than they should which makes them fail. It's probably a bad idea to generate archives on windows
-    // anyway. Either unix is portable OR no archive is created anywhere. This also means that windows users can't create
+    // On Windows, we fail to remove the meta_dir and can't do anything about it, which means tests will see more
+    // in the directory than they should which makes them fail. It's probably a bad idea to generate archives on Windows
+    // anyway. Either Unix is portable OR no archive is created anywhere. This also means that Windows users can't create
     // archives, but that's not a deal-breaker.
     if cfg!(windows) || is_ci::cached() || is_excluded(archive) {
         return Ok(());
--- a/tests/tools/src/lib.rs
+++ b/tests/tools/src/lib.rs
@@ -10,6 +10,7 @@
 
 use std::{
     collections::BTreeMap,
+    env,
     ffi::OsString,
     io::Read,
     path::{Path, PathBuf},
@@ -150,8 +151,8 @@ fn git_version_from_bytes(bytes: &[u8]) -> Result<(u8, u8, u8)> {
 
 /// Set the current working dir to `new_cwd` and return a type that returns to the previous working dir on drop.
 pub fn set_current_dir(new_cwd: impl AsRef<Path>) -> std::io::Result<AutoRevertToPreviousCWD> {
-    let cwd = std::env::current_dir()?;
-    std::env::set_current_dir(new_cwd)?;
+    let cwd = env::current_dir()?;
+    env::set_current_dir(new_cwd)?;
     Ok(AutoRevertToPreviousCWD(cwd))
 }
 
@@ -166,7 +167,7 @@ pub struct AutoRevertToPreviousCWD(PathBuf);
 
 impl Drop for AutoRevertToPreviousCWD {
     fn drop(&mut self) {
-        std::env::set_current_dir(&self.0).unwrap();
+        env::set_current_dir(&self.0).unwrap();
     }
 }
 
@@ -461,7 +462,7 @@ fn scripted_fixture_read_only_with_args_inner(
                 crc_digest.update(&std::fs::read(&script_path).unwrap_or_else(|err| {
                     panic!(
                         "file {script_path:?} in CWD {:?} could not be read: {err}",
-                        std::env::current_dir().expect("valid cwd"),
+                        env::current_dir().expect("valid cwd"),
                     )
                 }));
                 for arg in &args {
@@ -548,7 +549,7 @@ fn scripted_fixture_read_only_with_args_inner(
                         script_location.display()
                     );
                 }
-                let script_absolute_path = std::env::current_dir()?.join(script_path);
+                let script_absolute_path = env::current_dir()?.join(script_path);
                 let mut cmd = std::process::Command::new(&script_absolute_path);
                 let output = match configure_command(&mut cmd, &args, &script_result_directory).output() {
                     Ok(out) => out,
@@ -593,7 +594,7 @@ fn configure_command<'a>(
     args: &[String],
     script_result_directory: &Path,
 ) -> &'a mut std::process::Command {
-    let mut msys_for_git_bash_on_windows = std::env::var_os("MSYS").unwrap_or_default();
+    let mut msys_for_git_bash_on_windows = env::var_os("MSYS").unwrap_or_default();
     msys_for_git_bash_on_windows.push(" winsymlinks:nativestrict");
     cmd.args(args)
         .stdout(std::process::Stdio::piped())
@@ -632,6 +633,14 @@ fn write_failure_marker(failure_marker: &Path) {
     std::fs::write(failure_marker, []).ok();
 }
 
+fn should_skip_all_archive_creation() -> bool {
+    // On Windows, we fail to remove the meta_dir and can't do anything about it, which means tests will see more
+    // in the directory than they should which makes them fail. It's probably a bad idea to generate archives on Windows
+    // anyway. Either Unix is portable OR no archive is created anywhere. This also means that Windows users can't create
+    // archives, but that's not a deal-breaker.
+    cfg!(windows) || (is_ci::cached() && !env::var_os("GIX_TEST_CREATE_ARCHIVES_EVEN_ON_CI").is_some())
+}
+
 fn is_lfs_pointer_file(path: &Path) -> bool {
     const PREFIX: &[u8] = b"version https://gix-lfs";
     let mut buf = [0_u8; PREFIX.len()];
@@ -645,11 +654,7 @@ fn is_lfs_pointer_file(path: &Path) -> bool {
 /// The `script_identity` will be baked into the soon to be created `archive` as it identifies the script
 /// that created the contents of `source_dir`.
 fn create_archive_if_we_should(source_dir: &Path, archive: &Path, script_identity: u32) -> std::io::Result<()> {
-    // On Windows, we fail to remove the meta_dir and can't do anything about it, which means tests will see more
-    // in the directory than they should which makes them fail. It's probably a bad idea to generate archives on Windows
-    // anyway. Either Unix is portable OR no archive is created anywhere. This also means that Windows users can't create
-    // archives, but that's not a deal-breaker.
-    if cfg!(windows) || is_ci::cached() || is_excluded(archive) {
+    if should_skip_all_archive_creation() || is_excluded(archive) {
         return Ok(());
     }
     if is_lfs_pointer_file(archive) {
@@ -702,7 +707,7 @@ fn is_excluded(archive: &Path) -> bool {
     let mut lut = EXCLUDE_LUT.lock();
     lut.as_mut()
         .and_then(|cache| {
-            let archive = std::env::current_dir().ok()?.join(archive);
+            let archive = env::current_dir().ok()?.join(archive);
             let relative_path = archive.strip_prefix(cache.base()).ok()?;
             cache
                 .at_path(
@@ -746,7 +751,7 @@ fn extract_archive(
         let mut buf = Vec::new();
         #[cfg_attr(feature = "xz", allow(unused_mut))]
         let mut input_archive = std::fs::File::open(archive)?;
-        if std::env::var_os("GIX_TEST_IGNORE_ARCHIVES").is_some() {
+        if env::var_os("GIX_TEST_IGNORE_ARCHIVES").is_some() {
             return Err(std::io::Error::new(
                 std::io::ErrorKind::Other,
                 format!(
@@ -848,16 +853,16 @@ impl<'a> Env<'a> {
 
     /// Set `var` to `value`.
     pub fn set(mut self, var: &'a str, value: impl Into<String>) -> Self {
-        let prev = std::env::var_os(var);
-        std::env::set_var(var, value.into());
+        let prev = env::var_os(var);
+        env::set_var(var, value.into());
         self.altered_vars.push((var, prev));
         self
     }
 
     /// Unset `var`.
     pub fn unset(mut self, var: &'a str) -> Self {
-        let prev = std::env::var_os(var);
-        std::env::remove_var(var);
+        let prev = env::var_os(var);
+        env::remove_var(var);
         self.altered_vars.push((var, prev));
         self
     }
@@ -867,8 +872,8 @@ impl<'a> Drop for Env<'a> {
     fn drop(&mut self) {
         for (var, prev_value) in self.altered_vars.iter().rev() {
             match prev_value {
-                Some(value) => std::env::set_var(var, value),
-                None => std::env::remove_var(var),
+                Some(value) => env::set_var(var, value),
+                None => env::remove_var(var),
             }
         }
     }
--- a/tests/tools/src/lib.rs
+++ b/tests/tools/src/lib.rs
@@ -638,7 +638,7 @@ fn should_skip_all_archive_creation() -> bool {
     // in the directory than they should which makes them fail. It's probably a bad idea to generate archives on Windows
     // anyway. Either Unix is portable OR no archive is created anywhere. This also means that Windows users can't create
     // archives, but that's not a deal-breaker.
-    cfg!(windows) || (is_ci::cached() && !env::var_os("GIX_TEST_CREATE_ARCHIVES_EVEN_ON_CI").is_some())
+    cfg!(windows) || (is_ci::cached() && env::var_os("GIX_TEST_CREATE_ARCHIVES_EVEN_ON_CI").is_none())
 }
 
 fn is_lfs_pointer_file(path: &Path) -> bool {


Byron (approved):
Thanks a lot, I love that this part of the workflow is finally automated and a little safer.

As a matter of fact, due to me using nightly locally and `gix-macros` tests failing due to that I stopped running `just test` locally entirely. Thus I wouldn't even notice if some archives are regenerated anymore.

On the negative side, as Git is updated, there might be test-failures due to that, making CI tests less reliable. However, for now I'd think we want to know and typically update the archives accordingly.

EliahKagan:
> On the negative side, as Git is updated, there might be test-failures due to that, making CI tests less reliable.

I think that should only be expected to happen if both `GIX_TEST_IGNORE_ARCHIVES` and `GIX_TEST_CREATE_ARCHIVES_EVEN_ON_CI` are used together, which is why I did not do that on CI.

In that situation, the Git version and usually other details of the system would have to be the same to prevent changes to tracked files. For example, when I run `cargo nextest run --all` with `GIX_TEST_IGNORE_ARCHIVES=1` locally on any system besides Windows (where archive creation is suppressed in `gix-testtools`), I get a large number of changed archives.

But with only `GIX_TEST_CREATE_ARCHIVES_EVEN_ON_CI` and not `GIX_TEST_IGNORE_ARCHIVES`, I believe the only fixture scripts that should be run are those whose associated archives are:

- Excluded by being listed in a `.gitignore` file, in which case `git diff` won't show anything for them because they're ignored, *or*
- Missing or out of date relative to the most recent change of the fixture script, in which case that is the situation we are looking for, *or*
- Running because a problem in `gix-testtools` prevents it from identifying an archive or reuses an identity across multiple archives, but that also seems like something that ought to cause a failure. The latter was the cause of [#1440](https://github.com/Byron/gitoxide/issues/1440) if I understand [#1441](https://github.com/Byron/gitoxide/pull/1441) correctly.

However, if there are other situations where just the version of Git changing will cause a failure here, then I think a comment or something should be added to warn about it, possibly in the CI workflow where I already have a comment suggesting to regenerate the archive.

Byron:
> In that situation, the Git version and usually other details of the system would have to be the same to prevent changes to tracked files. For example, when I run `cargo nextest run --all` with `GIX_TEST_IGNORE_ARCHIVES=1` locally on any system besides Windows (where archive creation is suppressed in `gix-testtools`), I get a large number of changed archives.

Ah, I see, thanks for elaborating. It's interesting how I am already not truly aware of how this all works anymore, and just take it for granted assuming it works correctly.

Thanks again for etching out the shortcomings!

EliahKagan:
One interesting aspect of this is that even on the same system, if I run `GIX_TEST_IGNORE_ARCHIVES=1 cargo nextest run --all`, stage, and then run it again, I get further changes. (Presumably, if I did not stage in between, I would also get further changes. Staging just lets me see the changes.)

I am not clear on why that is, and I would like to understand it better. I'd also like to make it so changes to generated archives are easier to examine, which might just entail documenting somewhere what configuration to use, but might also (or instead) involve changing `.gitattributes`, I am not sure.

After both runs (in case somehow it would affect them), I made this staged but uncommitted change to `.gitattributes` for simplicity, though I'm not sure it's really required:

```diff
diff --git a/.gitattributes b/.gitattributes
index 50d83ecdf..6245dce7c 100644
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,5 @@
 **/generated-archives/*.tar* filter=lfs-disabled diff=lfs merge=lfs -text
+*.tar diff=tar

 # assure line feeds don't interfere with our working copy hash
 **/tests/fixtures/**/*.sh text crlf=input      eol=lf
```

And also added this local configuration (I could have added an `lfs` configuration instead, which is why I am not sure the above change is required):

```gitconfig
[diff "tar"]
        textconv = tar -tvf
        binary = true
```

So one question I have is what changes, if any, should be committed to `.gitattributes` to make it easier to investigate this.

But the other question is why I get the diffs I do. I investigated this on an Ubuntu 24.04 LTS system [with git](https://launchpad.net/ubuntu/noble/+source/git) 2.43.0-1ubuntu7.1. Full results are [in this gist](https://gist.github.com/EliahKagan/7564db6b9e52b099c17aea1d0a6c92ca).

[The first diff](https://gist.github.com/EliahKagan/7564db6b9e52b099c17aea1d0a6c92ca#file-run1-diff), which is the output of `git diff --staged` except that I have removed the part showing changes to `.gitattributes` applied after the runs, represents the difference between the archives are they were committed, and the archives after I run the whole test suite and force them to be regenerated. It makes intuitive sense that there would be many changes, and a cursory examination of the changes is unsurprising: many seem like the sort of thing one would expect to be different across platforms or versions, such as what `.sample` hooks exist with what contents, though some of the other changes I am less clear on.

[The second diff](https://gist.github.com/EliahKagan/7564db6b9e52b099c17aea1d0a6c92ca#file-run2-diff), which is the output of `git diff`, represents the difference between the archives after being recreated, versus after being recreated again in an immediately subsequent full test run. These changes are shorter, so I've included them below, and I also do not understand the reason for them:

```diff
diff --git a/gix-index/tests/fixtures/generated-archives/v2_split_index.tar b/gix-index/tests/fixtures/generated-archives/v2_split_index.tar
index 409da3baf..00983e9a7 100644
--- a/gix-index/tests/fixtures/generated-archives/v2_split_index.tar
+++ b/gix-index/tests/fixtures/generated-archives/v2_split_index.tar
@@ -5,6 +5,7 @@ drwxr-xr-x 0/0               0 2006-07-24 01:21 .git
 -rw-r--r-- 0/0              21 2006-07-24 01:21 .git/HEAD
 drwxr-xr-x 0/0               0 2006-07-24 01:21 .git/info
 -rw-r--r-- 0/0             240 2006-07-24 01:21 .git/info/exclude
+-rw-r--r-- 0/0              96 2006-07-24 01:21 .git/sharedindex.a7c54197d242f14f5c246cda5be747aee80393b8
 drwxr-xr-x 0/0               0 2006-07-24 01:21 .git/hooks
 -rwxr-xr-x 0/0             189 2006-07-24 01:21 .git/hooks/post-update.sample
 -rwxr-xr-x 0/0             424 2006-07-24 01:21 .git/hooks/pre-applypatch.sample
@@ -20,7 +21,6 @@ drwxr-xr-x 0/0               0 2006-07-24 01:21 .git/hooks
 -rwxr-xr-x 0/0            1492 2006-07-24 01:21 .git/hooks/prepare-commit-msg.sample
 -rwxr-xr-x 0/0            1643 2006-07-24 01:21 .git/hooks/pre-commit.sample
 -rwxr-xr-x 0/0            1374 2006-07-24 01:21 .git/hooks/pre-push.sample
--rw-r--r-- 0/0              96 2006-07-24 01:21 .git/sharedindex.772b543d373251f25f394435e8ff6a0ad429b5c1
 drwxr-xr-x 0/0               0 2006-07-24 01:21 .git/logs
 -rw-r--r-- 0/0             156 2006-07-24 01:21 .git/logs/HEAD
 drwxr-xr-x 0/0               0 2006-07-24 01:21 .git/logs/refs
diff --git a/gix-index/tests/fixtures/generated-archives/v2_split_index_recursive.tar b/gix-index/tests/fixtures/generated-archives/v2_split_index_recursive.tar
index d7b836330..80745dfa2 100644
--- a/gix-index/tests/fixtures/generated-archives/v2_split_index_recursive.tar
+++ b/gix-index/tests/fixtures/generated-archives/v2_split_index_recursive.tar
@@ -1,8 +1,8 @@
 drwxr-xr-x 0/0               0 2006-07-24 01:21 ./
 drwxr-xr-x 0/0               0 2006-07-24 01:21 .git
 -rw-r--r-- 0/0              73 2006-07-24 01:21 .git/description
+-rw-r--r-- 0/0             205 2006-07-24 01:21 .git/sharedindex.f3e525d52de36e08a116a91a59b84ae2512073d3
 -rw-r--r-- 0/0             113 2006-07-24 01:21 .git/config
--rw-r--r-- 0/0             205 2006-07-24 01:21 .git/sharedindex.b03111262ebac0866ea3bae2ae5987cd2abdcaef
 -rw-r--r-- 0/0              21 2006-07-24 01:21 .git/HEAD
 drwxr-xr-x 0/0               0 2006-07-24 01:21 .git/info
 -rw-r--r-- 0/0             240 2006-07-24 01:21 .git/info/exclude
diff --git a/gix-index/tests/fixtures/generated-archives/v2_split_vs_regular_index.tar b/gix-index/tests/fixtures/generated-archives/v2_split_vs_regular_index.tar
index 78c144b13..25f67056d 100644
--- a/gix-index/tests/fixtures/generated-archives/v2_split_vs_regular_index.tar
+++ b/gix-index/tests/fixtures/generated-archives/v2_split_vs_regular_index.tar
@@ -72,7 +72,6 @@ drwxr-xr-x 0/0               0 2006-07-24 01:21 split
 -rw-r--r-- 0/0               2 2006-07-24 01:21 split/z
 drwxr-xr-x 0/0               0 2006-07-24 01:21 split/.git
 -rw-r--r-- 0/0              73 2006-07-24 01:21 split/.git/description
--rw-r--r-- 0/0             416 2006-07-24 01:21 split/.git/sharedindex.1183b4544429fd6795bd9c3e1babd628a19bf369
 -rw-r--r-- 0/0             129 2006-07-24 01:21 split/.git/config
 -rw-r--r-- 0/0              21 2006-07-24 01:21 split/.git/HEAD
 drwxr-xr-x 0/0               0 2006-07-24 01:21 split/.git/info
@@ -92,6 +91,7 @@ drwxr-xr-x 0/0               0 2006-07-24 01:21 split/.git/hooks
 -rwxr-xr-x 0/0            1492 2006-07-24 01:21 split/.git/hooks/prepare-commit-msg.sample
 -rwxr-xr-x 0/0            1643 2006-07-24 01:21 split/.git/hooks/pre-commit.sample
 -rwxr-xr-x 0/0            1374 2006-07-24 01:21 split/.git/hooks/pre-push.sample
+-rw-r--r-- 0/0             416 2006-07-24 01:21 split/.git/sharedindex.b65ac769815289980f011f9ae3406be2c8fa8407
 drwxr-xr-x 0/0               0 2006-07-24 01:21 split/.git/logs
 -rw-r--r-- 0/0             305 2006-07-24 01:21 split/.git/logs/HEAD
 drwxr-xr-x 0/0               0 2006-07-24 01:21 split/.git/logs/refs
diff --git a/gix-negotiate/tests/fixtures/generated-archives/make_repos.tar b/gix-negotiate/tests/fixtures/generated-archives/make_repos.tar
index f4b7cc55f..021b86233 100644
--- a/gix-negotiate/tests/fixtures/generated-archives/make_repos.tar
+++ b/gix-negotiate/tests/fixtures/generated-archives/make_repos.tar
@@ -758,9 +758,9 @@ drwxr-xr-x 0/0               0 2006-07-24 01:21 multi_round/client/.git/objects/
 -rw-r--r-- 0/0           10712 2006-07-24 01:21 multi_round/client/.git/objects/info/commit-graph
 -rw-r--r-- 0/0              54 2006-07-24 01:21 multi_round/client/.git/objects/info/packs
 drwxr-xr-x 0/0               0 2006-07-24 01:21 multi_round/client/.git/objects/pack
--rw-r--r-- 0/0            1972 2006-07-24 01:21 multi_round/client/.git/objects/pack/pack-fe61307136aecbdd8c8dfe398d26986291921283.rev
--rw-r--r-- 0/0           59451 2006-07-24 01:21 multi_round/client/.git/objects/pack/pack-fe61307136aecbdd8c8dfe398d26986291921283.pack
--rw-r--r-- 0/0           14512 2006-07-24 01:21 multi_round/client/.git/objects/pack/pack-fe61307136aecbdd8c8dfe398d26986291921283.idx
+-rw-r--r-- 0/0            1972 2006-07-24 01:21 multi_round/client/.git/objects/pack/pack-0d92e2ea6533a6622cdab3bd74bccdce3d85f34e.rev
+-rw-r--r-- 0/0           14512 2006-07-24 01:21 multi_round/client/.git/objects/pack/pack-0d92e2ea6533a6622cdab3bd74bccdce3d85f34e.idx
+-rw-r--r-- 0/0           59340 2006-07-24 01:21 multi_round/client/.git/objects/pack/pack-0d92e2ea6533a6622cdab3bd74bccdce3d85f34e.pack
 -rw-r--r-- 0/0               7 2006-07-24 01:21 multi_round/client/b8.c12.t
 -rw-r--r-- 0/0               6 2006-07-24 01:21 multi_round/client/b8.c3.t
 -rw-r--r-- 0/0               7 2006-07-24 01:21 multi_round/client/b8.c17.t
diff --git a/gix-odb/tests/fixtures/generated-archives/make_repo_multi_index.tar b/gix-odb/tests/fixtures/generated-archives/make_repo_multi_index.tar
index 92d7ef83a..332922625 100644
--- a/gix-odb/tests/fixtures/generated-archives/make_repo_multi_index.tar
+++ b/gix-odb/tests/fixtures/generated-archives/make_repo_multi_index.tar
@@ -691,42 +691,42 @@ drwxr-xr-x 0/0               0 2006-07-24 01:21 .git/objects/cf
 -rw-r--r-- 0/0              24 2006-07-24 01:21 .git/objects/cf/476b3404986ec61542145808a5c404d2a42789
 drwxr-xr-x 0/0               0 2006-07-24 01:21 .git/objects/pack
 -rw-r--r-- 0/0            4669 2006-07-24 01:21 .git/objects/pack/pack-2bbdc0e28b43868ee99e4be43a1792a40b4e9129.pack
+-rw-r--r-- 0/0            3368 2006-07-24 01:21 .git/objects/pack/pack-1ad50de26d8845bd6bb80a060f7d7c3f00505fb7.idx
 -rw-r--r-- 0/0           26060 2006-07-24 01:21 .git/objects/pack/multi-pack-index
 -rw-r--r-- 0/0             264 2006-07-24 01:21 .git/objects/pack/pack-9b7062fbd0582882cab3454d7b3de09f1578bfa8.rev
 -rw-r--r-- 0/0            1828 2006-07-24 01:21 .git/objects/pack/pack-8c07ea08d9a9a4be5bca3e7fc246893243950f6d.idx
 -rw-r--r-- 0/0             136 2006-07-24 01:21 .git/objects/pack/pack-2aa9aa8c891b01a5dfd811c9ad9e1381dd469e39.rev
 -rw-r--r-- 0/0            3172 2006-07-24 01:21 .git/objects/pack/pack-572114382e4d12da25a0587fbc715315ece056bb.idx
--rw-r--r-- 0/0            3900 2006-07-24 01:21 .git/objects/pack/pack-e3a08db09c6b3290bca8c291d47e2810d6075392.idx
 -rw-r--r-- 0/0             236 2006-07-24 01:21 .git/objects/pack/pack-f8777e0a8830020ac1c2e49d399356bbb0f4d48b.rev
+-rw-r--r-- 0/0            3536 2006-07-24 01:21 .git/objects/pack/pack-1153fd15cfceb0c7f356c0d63111b0654c495a8b.idx
 -rw-r--r-- 0/0            2556 2006-07-24 01:21 .git/objects/pack/pack-9b7062fbd0582882cab3454d7b3de09f1578bfa8.idx
 -rw-r--r-- 0/0            2221 2006-07-24 01:21 .git/objects/pack/pack-5bc9b6bf1bf64398bfb3e435518e93d3e145382e.pack
 -rw-r--r-- 0/0            3854 2006-07-24 01:21 .git/objects/pack/pack-9b7062fbd0582882cab3454d7b3de09f1578bfa8.pack
 -rw-r--r-- 0/0             428 2006-07-24 01:21 .git/objects/pack/pack-9c7e898ea061ab4af575def7f9d0e6836d727871.rev
 -rw-r--r-- 0/0            1238 2006-07-24 01:21 .git/objects/pack/pack-2aa9aa8c891b01a5dfd811c9ad9e1381dd469e39.pack
 -rw-r--r-- 0/0            4931 2006-07-24 01:21 .git/objects/pack/pack-dd8f4f04f6c665f86f8a954204f89a3807fc890f.pack
--rw-r--r-- 0/0            7087 2006-07-24 01:21 .git/objects/pack/pack-818fb3b664fab9435175f66012273ef903ea893c.pack
--rw-r--r-- 0/0             456 2006-07-24 01:21 .git/objects/pack/pack-e3a08db09c6b3290bca8c291d47e2810d6075392.rev
--rw-r--r-- 0/0             380 2006-07-24 01:21 .git/objects/pack/pack-818fb3b664fab9435175f66012273ef903ea893c.rev
+-rw-r--r-- 0/0            7457 2006-07-24 01:21 .git/objects/pack/pack-1153fd15cfceb0c7f356c0d63111b0654c495a8b.pack
+-rw-r--r-- 0/0            7198 2006-07-24 01:21 .git/objects/pack/pack-1ad50de26d8845bd6bb80a060f7d7c3f00505fb7.pack
 -rw-r--r-- 0/0             304 2006-07-24 01:21 .git/objects/pack/pack-2bbdc0e28b43868ee99e4be43a1792a40b4e9129.rev
+-rw-r--r-- 0/0             380 2006-07-24 01:21 .git/objects/pack/pack-1ad50de26d8845bd6bb80a060f7d7c3f00505fb7.rev
 -rw-r--r-- 0/0             352 2006-07-24 01:21 .git/objects/pack/pack-572114382e4d12da25a0587fbc715315ece056bb.rev
 -rw-r--r-- 0/0             276 2006-07-24 01:21 .git/objects/pack/pack-8b66663f0e277f430c0574bca2ad2d27af318b0b.rev
 -rw-r--r-- 0/0            2024 2006-07-24 01:21 .git/objects/pack/pack-5bc9b6bf1bf64398bfb3e435518e93d3e145382e.idx
 -rw-r--r-- 0/0            1660 2006-07-24 01:21 .git/objects/pack/pack-2aa9aa8c891b01a5dfd811c9ad9e1381dd469e39.idx
--rw-r--r-- 0/0           10084 2006-07-24 01:21 .git/objects/pack/pack-e3a08db09c6b3290bca8c291d47e2810d6075392.pack
--rw-r--r-- 0/0            7550 2006-07-24 01:21 .git/objects/pack/pack-1e0a20554ca41743e224be67a9f273e253f5b461.pack
diff --git a/gix-index/tests/fixtures/generated-archives/v2_split_index.tar b/gix-index/tests/fixtures/generated-archives/v2_split_index.tar
index 409da3baf..00983e9a7 100644
--- a/gix-index/tests/fixtures/generated-archives/v2_split_index.tar
+++ b/gix-index/tests/fixtures/generated-archives/v2_split_index.tar
@@ -5,6 +5,7 @@ drwxr-xr-x 0/0               0 2006-07-24 01:21 .git
 -rw-r--r-- 0/0              21 2006-07-24 01:21 .git/HEAD
 drwxr-xr-x 0/0               0 2006-07-24 01:21 .git/info
 -rw-r--r-- 0/0             240 2006-07-24 01:21 .git/info/exclude
+-rw-r--r-- 0/0              96 2006-07-24 01:21 .git/sharedindex.a7c54197d242f14f5c246cda5be747aee80393b8
 drwxr-xr-x 0/0               0 2006-07-24 01:21 .git/hooks
 -rwxr-xr-x 0/0             189 2006-07-24 01:21 .git/hooks/post-update.sample
 -rwxr-xr-x 0/0             424 2006-07-24 01:21 .git/hooks/pre-applypatch.sample
@@ -20,7 +21,6 @@ drwxr-xr-x 0/0               0 2006-07-24 01:21 .git/hooks
 -rwxr-xr-x 0/0            1492 2006-07-24 01:21 .git/hooks/prepare-commit-msg.sample
 -rwxr-xr-x 0/0            1643 2006-07-24 01:21 .git/hooks/pre-commit.sample
 -rwxr-xr-x 0/0            1374 2006-07-24 01:21 .git/hooks/pre-push.sample
--rw-r--r-- 0/0              96 2006-07-24 01:21 .git/sharedindex.772b543d373251f25f394435e8ff6a0ad429b5c1
 drwxr-xr-x 0/0               0 2006-07-24 01:21 .git/logs
 -rw-r--r-- 0/0             156 2006-07-24 01:21 .git/logs/HEAD
 drwxr-xr-x 0/0               0 2006-07-24 01:21 .git/logs/refs
diff --git a/gix-index/tests/fixtures/generated-archives/v2_split_index_recursive.tar b/gix-index/tests/fixtures/generated-archives/v2_split_index_recursive.tar
index d7b836330..80745dfa2 100644
--- a/gix-index/tests/fixtures/generated-archives/v2_split_index_recursive.tar
+++ b/gix-index/tests/fixtures/generated-archives/v2_split_index_recursive.tar
@@ -1,8 +1,8 @@
 drwxr-xr-x 0/0               0 2006-07-24 01:21 ./
 drwxr-xr-x 0/0               0 2006-07-24 01:21 .git
 -rw-r--r-- 0/0              73 2006-07-24 01:21 .git/description
+-rw-r--r-- 0/0             205 2006-07-24 01:21 .git/sharedindex.f3e525d52de36e08a116a91a59b84ae2512073d3
 -rw-r--r-- 0/0             113 2006-07-24 01:21 .git/config
--rw-r--r-- 0/0             205 2006-07-24 01:21 .git/sharedindex.b03111262ebac0866ea3bae2ae5987cd2abdcaef
 -rw-r--r-- 0/0              21 2006-07-24 01:21 .git/HEAD
 drwxr-xr-x 0/0               0 2006-07-24 01:21 .git/info
 -rw-r--r-- 0/0             240 2006-07-24 01:21 .git/info/exclude
diff --git a/gix-index/tests/fixtures/generated-archives/v2_split_vs_regular_index.tar b/gix-index/tests/fixtures/generated-archives/v2_split_vs_regular_index.tar
index 78c144b13..25f67056d 100644
--- a/gix-index/tests/fixtures/generated-archives/v2_split_vs_regular_index.tar
+++ b/gix-index/tests/fixtures/generated-archives/v2_split_vs_regular_index.tar
@@ -72,7 +72,6 @@ drwxr-xr-x 0/0               0 2006-07-24 01:21 split
 -rw-r--r-- 0/0               2 2006-07-24 01:21 split/z
 drwxr-xr-x 0/0               0 2006-07-24 01:21 split/.git
 -rw-r--r-- 0/0              73 2006-07-24 01:21 split/.git/description
--rw-r--r-- 0/0             416 2006-07-24 01:21 split/.git/sharedindex.1183b4544429fd6795bd9c3e1babd628a19bf369
 -rw-r--r-- 0/0             129 2006-07-24 01:21 split/.git/config
 -rw-r--r-- 0/0              21 2006-07-24 01:21 split/.git/HEAD
 drwxr-xr-x 0/0               0 2006-07-24 01:21 split/.git/info
@@ -92,6 +91,7 @@ drwxr-xr-x 0/0               0 2006-07-24 01:21 split/.git/hooks
 -rwxr-xr-x 0/0            1492 2006-07-24 01:21 split/.git/hooks/prepare-commit-msg.sample
 -rwxr-xr-x 0/0            1643 2006-07-24 01:21 split/.git/hooks/pre-commit.sample
 -rwxr-xr-x 0/0            1374 2006-07-24 01:21 split/.git/hooks/pre-push.sample
+-rw-r--r-- 0/0             416 2006-07-24 01:21 split/.git/sharedindex.b65ac769815289980f011f9ae3406be2c8fa8407
 drwxr-xr-x 0/0               0 2006-07-24 01:21 split/.git/logs
 -rw-r--r-- 0/0             305 2006-07-24 01:21 split/.git/logs/HEAD
 drwxr-xr-x 0/0               0 2006-07-24 01:21 split/.git/logs/refs
diff --git a/gix-negotiate/tests/fixtures/generated-archives/make_repos.tar b/gix-negotiate/tests/fixtures/generated-archives/make_repos.tar
index f4b7cc55f..021b86233 100644
--- a/gix-negotiate/tests/fixtures/generated-archives/make_repos.tar
+++ b/gix-negotiate/tests/fixtures/generated-archives/make_repos.tar
@@ -758,9 +758,9 @@ drwxr-xr-x 0/0               0 2006-07-24 01:21 multi_round/client/.git/objects/
 -rw-r--r-- 0/0           10712 2006-07-24 01:21 multi_round/client/.git/objects/info/commit-graph
 -rw-r--r-- 0/0              54 2006-07-24 01:21 multi_round/client/.git/objects/info/packs
 drwxr-xr-x 0/0               0 2006-07-24 01:21 multi_round/client/.git/objects/pack
--rw-r--r-- 0/0            1972 2006-07-24 01:21 multi_round/client/.git/objects/pack/pack-fe61307136aecbdd8c8dfe398d26986291921283.rev
--rw-r--r-- 0/0           59451 2006-07-24 01:21 multi_round/client/.git/objects/pack/pack-fe61307136aecbdd8c8dfe398d26986291921283.pack
--rw-r--r-- 0/0           14512 2006-07-24 01:21 multi_round/client/.git/objects/pack/pack-fe61307136aecbdd8c8dfe398d26986291921283.idx
+-rw-r--r-- 0/0            1972 2006-07-24 01:21 multi_round/client/.git/objects/pack/pack-0d92e2ea6533a6622cdab3bd74bccdce3d85f34e.rev
+-rw-r--r-- 0/0           14512 2006-07-24 01:21 multi_round/client/.git/objects/pack/pack-0d92e2ea6533a6622cdab3bd74bccdce3d85f34e.idx
+-rw-r--r-- 0/0           59340 2006-07-24 01:21 multi_round/client/.git/objects/pack/pack-0d92e2ea6533a6622cdab3bd74bccdce3d85f34e.pack
 -rw-r--r-- 0/0               7 2006-07-24 01:21 multi_round/client/b8.c12.t
 -rw-r--r-- 0/0               6 2006-07-24 01:21 multi_round/client/b8.c3.t
 -rw-r--r-- 0/0               7 2006-07-24 01:21 multi_round/client/b8.c17.t
diff --git a/gix-odb/tests/fixtures/generated-archives/make_repo_multi_index.tar b/gix-odb/tests/fixtures/generated-archives/make_repo_multi_index.tar
index 92d7ef83a..332922625 100644
--- a/gix-odb/tests/fixtures/generated-archives/make_repo_multi_index.tar
+++ b/gix-odb/tests/fixtures/generated-archives/make_repo_multi_index.tar
@@ -691,42 +691,42 @@ drwxr-xr-x 0/0               0 2006-07-24 01:21 .git/objects/cf
 -rw-r--r-- 0/0              24 2006-07-24 01:21 .git/objects/cf/476b3404986ec61542145808a5c404d2a42789
 drwxr-xr-x 0/0               0 2006-07-24 01:21 .git/objects/pack
 -rw-r--r-- 0/0            4669 2006-07-24 01:21 .git/objects/pack/pack-2bbdc0e28b43868ee99e4be43a1792a40b4e9129.pack
+-rw-r--r-- 0/0            3368 2006-07-24 01:21 .git/objects/pack/pack-1ad50de26d8845bd6bb80a060f7d7c3f00505fb7.idx
 -rw-r--r-- 0/0           26060 2006-07-24 01:21 .git/objects/pack/multi-pack-index
 -rw-r--r-- 0/0             264 2006-07-24 01:21 .git/objects/pack/pack-9b7062fbd0582882cab3454d7b3de09f1578bfa8.rev
 -rw-r--r-- 0/0            1828 2006-07-24 01:21 .git/objects/pack/pack-8c07ea08d9a9a4be5bca3e7fc246893243950f6d.idx
 -rw-r--r-- 0/0             136 2006-07-24 01:21 .git/objects/pack/pack-2aa9aa8c891b01a5dfd811c9ad9e1381dd469e39.rev
 -rw-r--r-- 0/0            3172 2006-07-24 01:21 .git/objects/pack/pack-572114382e4d12da25a0587fbc715315ece056bb.idx
--rw-r--r-- 0/0            3900 2006-07-24 01:21 .git/objects/pack/pack-e3a08db09c6b3290bca8c291d47e2810d6075392.idx
 -rw-r--r-- 0/0             236 2006-07-24 01:21 .git/objects/pack/pack-f8777e0a8830020ac1c2e49d399356bbb0f4d48b.rev
+-rw-r--r-- 0/0            3536 2006-07-24 01:21 .git/objects/pack/pack-1153fd15cfceb0c7f356c0d63111b0654c495a8b.idx
 -rw-r--r-- 0/0            2556 2006-07-24 01:21 .git/objects/pack/pack-9b7062fbd0582882cab3454d7b3de09f1578bfa8.idx
 -rw-r--r-- 0/0            2221 2006-07-24 01:21 .git/objects/pack/pack-5bc9b6bf1bf64398bfb3e435518e93d3e145382e.pack
 -rw-r--r-- 0/0            3854 2006-07-24 01:21 .git/objects/pack/pack-9b7062fbd0582882cab3454d7b3de09f1578bfa8.pack
 -rw-r--r-- 0/0             428 2006-07-24 01:21 .git/objects/pack/pack-9c7e898ea061ab4af575def7f9d0e6836d727871.rev
 -rw-r--r-- 0/0            1238 2006-07-24 01:21 .git/objects/pack/pack-2aa9aa8c891b01a5dfd811c9ad9e1381dd469e39.pack
 -rw-r--r-- 0/0            4931 2006-07-24 01:21 .git/objects/pack/pack-dd8f4f04f6c665f86f8a954204f89a3807fc890f.pack
--rw-r--r-- 0/0            7087 2006-07-24 01:21 .git/objects/pack/pack-818fb3b664fab9435175f66012273ef903ea893c.pack
--rw-r--r-- 0/0             456 2006-07-24 01:21 .git/objects/pack/pack-e3a08db09c6b3290bca8c291d47e2810d6075392.rev
--rw-r--r-- 0/0             380 2006-07-24 01:21 .git/objects/pack/pack-818fb3b664fab9435175f66012273ef903ea893c.rev
+-rw-r--r-- 0/0            7457 2006-07-24 01:21 .git/objects/pack/pack-1153fd15cfceb0c7f356c0d63111b0654c495a8b.pack
+-rw-r--r-- 0/0            7198 2006-07-24 01:21 .git/objects/pack/pack-1ad50de26d8845bd6bb80a060f7d7c3f00505fb7.pack
 -rw-r--r-- 0/0             304 2006-07-24 01:21 .git/objects/pack/pack-2bbdc0e28b43868ee99e4be43a1792a40b4e9129.rev
+-rw-r--r-- 0/0             380 2006-07-24 01:21 .git/objects/pack/pack-1ad50de26d8845bd6bb80a060f7d7c3f00505fb7.rev
 -rw-r--r-- 0/0             352 2006-07-24 01:21 .git/objects/pack/pack-572114382e4d12da25a0587fbc715315ece056bb.rev
 -rw-r--r-- 0/0             276 2006-07-24 01:21 .git/objects/pack/pack-8b66663f0e277f430c0574bca2ad2d27af318b0b.rev
 -rw-r--r-- 0/0            2024 2006-07-24 01:21 .git/objects/pack/pack-5bc9b6bf1bf64398bfb3e435518e93d3e145382e.idx
 -rw-r--r-- 0/0            1660 2006-07-24 01:21 .git/objects/pack/pack-2aa9aa8c891b01a5dfd811c9ad9e1381dd469e39.idx
--rw-r--r-- 0/0           10084 2006-07-24 01:21 .git/objects/pack/pack-e3a08db09c6b3290bca8c291d47e2810d6075392.pack
--rw-r--r-- 0/0            7550 2006-07-24 01:21 .git/objects/pack/pack-1e0a20554ca41743e224be67a9f273e253f5b461.pack
--rw-r--r-- 0/0            3536 2006-07-24 01:21 .git/objects/pack/pack-1e0a20554ca41743e224be67a9f273e253f5b461.idx
 -rw-r--r-- 0/0            2640 2006-07-24 01:21 .git/objects/pack/pack-8b66663f0e277f430c0574bca2ad2d27af318b0b.idx
 -rw-r--r-- 0/0            1608 2006-07-24 01:21 .git/objects/pack/pack-8c07ea08d9a9a4be5bca3e7fc246893243950f6d.pack
 -rw-r--r-- 0/0            2360 2006-07-24 01:21 .git/objects/pack/pack-f8777e0a8830020ac1c2e49d399356bbb0f4d48b.idx
+-rw-r--r-- 0/0             404 2006-07-24 01:21 .git/objects/pack/pack-1153fd15cfceb0c7f356c0d63111b0654c495a8b.rev
 -rw-r--r-- 0/0            3004 2006-07-24 01:21 .git/objects/pack/pack-dd8f4f04f6c665f86f8a954204f89a3807fc890f.idx
--rw-r--r-- 0/0            3368 2006-07-24 01:21 .git/objects/pack/pack-818fb3b664fab9435175f66012273ef903ea893c.idx
 -rw-r--r-- 0/0            2795 2006-07-24 01:21 .git/objects/pack/pack-f8777e0a8830020ac1c2e49d399356bbb0f4d48b.pack
 -rw-r--r-- 0/0            1492 2006-07-24 01:21 .git/objects/pack/pack-b736b4d7459841f7c6d1181c18c9685a2b1e75c2.idx
 -rw-r--r-- 0/0            3704 2006-07-24 01:21 .git/objects/pack/pack-9c7e898ea061ab4af575def7f9d0e6836d727871.idx
 -rw-r--r-- 0/0            2836 2006-07-24 01:21 .git/objects/pack/pack-2bbdc0e28b43868ee99e4be43a1792a40b4e9129.idx
 -rw-r--r-- 0/0             112 2006-07-24 01:21 .git/objects/pack/pack-b736b4d7459841f7c6d1181c18c9685a2b1e75c2.rev
+-rw-r--r-- 0/0             456 2006-07-24 01:21 .git/objects/pack/pack-265d5a5a1b49208356d72008af5870937f764bef.rev
 -rw-r--r-- 0/0             160 2006-07-24 01:21 .git/objects/pack/pack-8c07ea08d9a9a4be5bca3e7fc246893243950f6d.rev
 -rw-r--r-- 0/0            2906 2006-07-24 01:21 .git/objects/pack/pack-8b66663f0e277f430c0574bca2ad2d27af318b0b.pack
+-rw-r--r-- 0/0            3900 2006-07-24 01:21 .git/objects/pack/pack-265d5a5a1b49208356d72008af5870937f764bef.idx
 -rw-r--r-- 0/0             212 2006-07-24 01:21 .git/objects/pack/pack-202904d12f68b77afa1da3d9708144023d6a999c.rev
 -rw-r--r-- 0/0            2192 2006-07-24 01:21 .git/objects/pack/pack-202904d12f68b77afa1da3d9708144023d6a999c.idx
 -rw-r--r-- 0/0            2518 2006-07-24 01:21 .git/objects/pack/pack-202904d12f68b77afa1da3d9708144023d6a999c.pack
@@ -734,8 +734,8 @@ drwxr-xr-x 0/0               0 2006-07-24 01:21 .git/objects/pack
 -rw-r--r-- 0/0             188 2006-07-24 01:21 .git/objects/pack/pack-5bc9b6bf1bf64398bfb3e435518e93d3e145382e.rev
 -rw-r--r-- 0/0             328 2006-07-24 01:21 .git/objects/pack/pack-dd8f4f04f6c665f86f8a954204f89a3807fc890f.rev
 -rw-r--r-- 0/0            5214 2006-07-24 01:21 .git/objects/pack/pack-572114382e4d12da25a0587fbc715315ece056bb.pack
--rw-r--r-- 0/0             404 2006-07-24 01:21 .git/objects/pack/pack-1e0a20554ca41743e224be67a9f273e253f5b461.rev
 -rw-r--r-- 0/0             800 2006-07-24 01:21 .git/objects/pack/pack-b736b4d7459841f7c6d1181c18c9685a2b1e75c2.pack
+-rw-r--r-- 0/0            9951 2006-07-24 01:21 .git/objects/pack/pack-265d5a5a1b49208356d72008af5870937f764bef.pack
 drwxr-xr-x 0/0               0 2006-07-24 01:21 .git/objects/c8
 -rw-r--r-- 0/0              38 2006-07-24 01:21 .git/objects/c8/cc550204e043706fa1618ef7fb9c450f04d22b
 -rw-r--r-- 0/0             146 2006-07-24 01:21 .git/objects/c8/553151f7776f597a3a3fe80d070aa3c8c5d689
```

In some cases, hashes shown in pack filenames differ. In others, filenames remain the same yet are for some reason listed in a different order. Is some of this related to the effect of running tests concurrently? (With `cargo nextest`, they are either mostly or entirely run in different processes rather than in different threads of the same process, so I expect that multiple fixture scripts actually can run at the same time.)

Byron:
Thanks for sharing - I think I also noticed that the archives could change quite a bit and appreciate looking into 'stabilizing' archive generation.

> So one question I have is what changes, if any, should be committed to .gitattributes to make it easier to investigate this.

Please feel free to add these proposed changes to `.gitattributes`, maybe even with some garnish of comments to share which git-configuration would be needed to make it effective.

> In some cases, hashes shown in pack filenames differ. In others, filenames remain the same yet are for some reason listed in a different order. Is some of this related to the effect of running tests concurrently?

I didn't notice that they just moved while filenames stayed the same, but also find it strange that the hashes differ. That would mean the objects differ, and those should always be frozen to a specific date. I also don't know of any file format that would contain the time (or any other variable), so formats should be stable if the content is stable.

`v2_split_index.tar` seems like a good candidate to investigate as only one file seemingly changed. Another reason for changes might be an unsorted traversal over the contents of directories when creating the `tar` archive, maybe that's something that can be changed as well to get stable archives.

> (With `cargo nextest`, they are either mostly or entirely run in different processes rather than in different threads of the same process, so I expect that multiple fixture scripts actually can run at the same time.)

Multiple *different* fixture scripts can run at the same time, as each holds a file-based lock to assure exclusive access to the target directory. This is to assure that only the first thread or process will create the result of a read-only script, while all that follow wait and just use the result. Writeable scripts don't perform locking as their target directory is assume to be exclusive already.

EliahKagan:
> `v2_split_index.tar` seems like a good candidate to investigate as only one file seemingly changed

Here's some preliminary information about that, using [`gin`](https://github.com/sbp/gin) on the extracted `sharedindex.*` files:

```text
ek@walpurti:~$ gin sharedindex.772b543d373251f25f394435e8ff6a0ad429b5c1
[header]
  signature = DIRC
  version = 2
  entries = 1

[entry]
  entry = 1
  ctime = 1726081166.7460086
  mtime = 1726081166.7460086
  dev = 2049
  ino = 1356183
  mode = 100644
  uid = 1000
  gid = 1000
  size = 0
  sha1 = e69de29bb2d1d6434b8b29ae775ad8c2e48c5391
  flags = 1
  assume-valid = False
  extended = False
  stage = (False, False)
  name = a

[checksum]
  checksum = True
  sha1 = 772b543d373251f25f394435e8ff6a0ad429b5c1
```

```text
ek@walpurti:~$ gin sharedindex.a7c54197d242f14f5c246cda5be747aee80393b8
[header]
  signature = DIRC
  version = 2
  entries = 1

[entry]
  entry = 1
  ctime = 1726082332.4457066
  mtime = 1726082332.4457066
  dev = 2049
  ino = 1393893
  mode = 100644
  uid = 1000
  gid = 1000
  size = 0
  sha1 = e69de29bb2d1d6434b8b29ae775ad8c2e48c5391
  flags = 1
  assume-valid = False
  extended = False
  stage = (False, False)
  name = a

[checksum]
  checksum = True
  sha1 = a7c54197d242f14f5c246cda5be747aee80393b8
```

```diff
ek@walpurti:~$ git diff --no-index <(gin sharedindex.772b543d373251f25f394435e8ff6a0ad429b5c1) <(gin sharedindex.a7c54197d242f14f5c246cda5be747aee80393b8)
diff --git a/dev/fd/63 b/dev/fd/62
--- a/dev/fd/63
+++ b/dev/fd/62
@@ -5,10 +5,10 @@

 [entry]
   entry = 1
-  ctime = 1726081166.7460086
-  mtime = 1726081166.7460086
+  ctime = 1726082332.4457066
+  mtime = 1726082332.4457066
   dev = 2049
-  ino = 1356183
+  ino = 1393893
   mode = 100644
   uid = 1000
   gid = 1000
@@ -22,4 +22,4 @@

 [checksum]
   checksum = True
-  sha1 = 772b543d373251f25f394435e8ff6a0ad429b5c1
+  sha1 = a7c54197d242f14f5c246cda5be747aee80393b8
```

So it looks like changes seem to be, or at least include, metadata on the actual filesystem that the V2 index files maintain for the purpose of speeding up change detection.

Because these are filesystem metadata (some of which--at least the inode numbers--Git does not even represent in repositories), it makes sense to me that it efforts to ensure configuration and metadata *pertaining to Git* are the same every time a fixture script is run would be insufficient to ensure these filesystem metadata are the same.

Byron:
Thanks for looking into this! With the index changes, I could have known 😅. Indeed there is nothing that can easily be done. For a moment I thought one could rewrite the index to unify these value, but I don't think it's worth the effort.

What's left for me to understand is how packs can be different. Timestamps and author names are controlled, so all objects should always be the same. Maybe it's not the content of the packs that change, but only their checksum which is made over compressed data as well. And the compressor may always be a little different for many reasons, particularly across systems.

Assuming the above is correct, I think this explains the differences, and would clearly put them into the 'something to accept' category. Maybe that's OK as well?

EliahKagan:
> Thanks for looking into this! With the index changes, I could have known 😅. Indeed there is nothing that can easily be done. For a moment I thought one could rewrite the index to unify these value, but I don't think it's worth the effort.

> Assuming the above is correct, I think this explains the differences, and would clearly put them into the 'something to accept' category. Maybe that's OK as well?

I agree that modifying the index files once they are created is not worthwhile. If this were done, then when investigating future strange things that turn out to be unrelated, one would occasionally wonder if the modifications were themselves somehow wrong, even in the absence of any actual bugs in how the modifications were made. In contrast, keeping the files are they were really created by `git` makes it so there will be fewer things to investigate. Keeping them "honest" also has the benefit that they can be used more readily (and with less explanation) when reporting bugs outside gitoxide, such as when reporting bugs in Git.

Keeping things as they are seems okay to me, but there are three other options:

1. I don't recommend this, but we can probably cause the device and inode numbers to *actually* be the same: the inode numbers could at least in principle be made the same by using a carefully constructed squashfs filesystem, and the inode number for that filesystem may be possible to make the same even when regenerating archives on different systems by using containerization and making sure it is always mapped the same way. I am not actually 100% confident that this will reliably work, but the bigger issue is that it is incurs substantial effort, and substantial limitation on the environment in which archives are regenerated to be committed.
2. Ensure archives *always* in practice differ, but in a way that does not make diffs hard to read, so that non-intentional differences can still be spotted when desired. So timestamps, at least for files that any tests potentially use (i.e. files currently in the archives), should be kept the same. But a new file could be added to each archive with a current timestamp and/or with a random value as its content. By making them always differ, the technique introduced in this PR would always be effective at detecting when a currently committed generated archives will not be used.
3. Keep archive generation the same, but refine or replace the technique introduced in this PR so it more directly checks--or includes a more direct check--what archives are actually generated. The best way to do that might be to take a wholly different approach from what I did here, and add functionality to `gix-testtools` to quickly check if any archives, other than those that are `.gitigore`d, would need to be generated. Another approach could be to do this outside `gix-testtools` by checking for new, changed, or removed fixture scripts in the absence of what appears to be a corresponding change to generated archives in the repository.

> What's left for me to understand is how packs can be different. Timestamps and author names are controlled, so all objects should always be the same. Maybe it's not the content of the packs that change, but only their checksum which is made over compressed data as well. And the compressor may always be a little different for many reasons, particularly across systems.

`make_repos.tar` was the archive in which [one pack and its associated files, but nothing else, was changed](https://gist.github.com/EliahKagan/7564db6b9e52b099c17aea1d0a6c92ca#file-run2-diff-L55-L71), so I looked into that a bit, so far only examining the `.pack` files. [Details are in this gist.](https://gist.github.com/EliahKagan/c9ee13f421e25549035a5fe73f7c89e0)

I don't know what's going on here. Cursory examination of the actual repositories suggests that the objects are the same, though I have not really verified that this is the case. But I think the mystery can be solved by examining the changes in the output of `git verify-pack -v` on the `.pack` files, possibly even without checking the `.idx` and `.rev` files. I did that in the gist, and also included a `diff` of the output, though the diff would be more useful if I had made it more granular.

Starting in the middle of the output at the first changed line, here are the first 10 lines before regenerating the archives (i.e., actually after regenerating them the first time but not the second, per the procedure described above):

```text
84f9c8767d96afe6e080271e613e51be5bd829ac tree   566 434 36051
9e88bc3f090bb37f8890abdc29f4375e0d806760 tree   12 23 36485 1 0a10084d765d90854e2bc548df400f23812680cf
71f30135b6e3ada2575c01919d3ebccfdaea79da tree   12 23 36508 1 a6db8c57d2358a97a00db2102926402ac0504751
f41b7b3c34a99d01ea840e13f9e82e1b7a962161 tree   12 23 36531 1 dd2143845727fb6adc72f7b053b9093ca9c05160
748c524eb71e57e3b4d105ba29db24e1fefd335c tree   670 515 36554
e09a018c615016f37b31421e52e0929f613f25bb tree   635 487 37069
ce8a2638c9cc62255b55fde7b2a1533fffaa4cc4 tree   600 463 37556
4b410c35e9ea1c4b03c28886aae751726acc420d tree   565 436 38019
ee139939d2141fa9b5abf467bbddab5c53882244 tree   530 409 38455
4a735977214a2df6acac52ca5ae1bc9dd12db700 tree   12 23 38864 1 0a10084d765d90854e2bc548df400f23812680cf
```

And here are the first 10 lines after regenerating the archives (i.e., after re-regenerating them):

```text
84f9c8767d96afe6e080271e613e51be5bd829ac tree   12 23 36051 1 c3adda84e145206e063dadb5702cc84e42068d71
9e88bc3f090bb37f8890abdc29f4375e0d806760 tree   12 23 36074 1 0a10084d765d90854e2bc548df400f23812680cf
71f30135b6e3ada2575c01919d3ebccfdaea79da tree   12 23 36097 1 a6db8c57d2358a97a00db2102926402ac0504751
f41b7b3c34a99d01ea840e13f9e82e1b7a962161 tree   12 23 36120 1 dd2143845727fb6adc72f7b053b9093ca9c05160
748c524eb71e57e3b4d105ba29db24e1fefd335c tree   670 515 36143
e09a018c615016f37b31421e52e0929f613f25bb tree   635 487 36658
ce8a2638c9cc62255b55fde7b2a1533fffaa4cc4 tree   600 463 37145
4b410c35e9ea1c4b03c28886aae751726acc420d tree   12 23 37608 1 879c3581b71bdbae7cbf69718960ab1d478c769e
ee139939d2141fa9b5abf467bbddab5c53882244 tree   12 23 37631 1 c3adda84e145206e063dadb5702cc84e42068d71
4a735977214a2df6acac52ca5ae1bc9dd12db700 tree   12 23 37654 1 0a10084d765d90854e2bc548df400f23812680cf
```

Simplifying even further by looking at just the *first* changed line:

```diff
-84f9c8767d96afe6e080271e613e51be5bd829ac tree   566 434 36051
+84f9c8767d96afe6e080271e613e51be5bd829ac tree   12 23 36051 1 c3adda84e145206e063dadb5702cc84e42068d71
```

The output of `git ls-tree 84f9c8767d96afe6e080271e613e51be5bd829ac` is the same in both repositories.

Details of packs are a weak area for me in Git, and I suspect that the situation may well be obvious. I can look into this further, and I'm willing to do so, but I figured I'd post that now in case the cause can already be inferred.

Is the difference due to different ways of computing deltas that, when resolved, give the same objects? That is, do the packs have the same objects, differently deltafied?

Byron:
Thanks for sharing additional options!

For now I also feel to just keep it the way it is, but gain a better understanding of how scripts can be made (more) idempotent.

> Is the difference due to different ways of computing deltas that, when resolved, give the same objects? That is, do the packs have the same objects, differently deltafied?

Yes, that would be my guess as well. Besides that, I am surprised that this operation isn't deterministic, and something we will have to live with.

There is a way to disable deltification for the entire blob-content of a repository using `* delta` in `.gitattributes`, but there is still trees and commits to deal with. Maybe there are other ways to discourage deltification in `git gc` invocations.

EliahKagan:
Since it is possible to have a regression where gitoxide does not properly resolve deltas, I'm not sure if we should try to avoid deltafication. I wonder, though, if running `git gc --aggressive` at the end of fixture scripts would be sufficient to make it so deltafication is done the same almost every time. From [git-gc(1)](https://git-scm.com/docs/git-gc#_aggressive):

> When the `--aggressive` option is supplied, [git-repack[1]](https://git-scm.com/docs/git-repack) will be invoked with the `-f` flag, which in turn will pass `--no-reuse-delta` to [git-pack-objects[1]](https://git-scm.com/docs/git-pack-objects). This will throw away any existing deltas and re-compute them, at the expense of spending much more time on the repacking.

But even aside from how this would slow down the fixture scripts, I am not at all sure this would be worthwhile.

Byron:
That's a good point, some fixtures in `gix-pack` are likely to rely on delta-compression.

> But even aside from how this would slow down the fixture scripts, I am not at all sure this would be worthwhile.

If it would make the output of scripts more stable in terms of the files it produces, I'd like it, particularly if these are used read-only. After all, these scripts are rarely big or have much actual content, so compression will be really fast no matter what.

If you think it's not worth it, I'd go with that - after all, delta-compression is just a part of the problem which by itself isn't reasonably solvable due to the reliance on the filesystem.

PR #1413, opened 2024-06-22
Fix missing license files in the released gix-testtools crate

musicinmybrain:
The contents of `tests/tools/` are released as https://crates.io/crates/gix-testtools, so license-file symlinks are needed to ensure the license texts are distributed.

PR #1591, opened 2024-09-11
In gix-testtools, use `ignore` and `index` via `gix_worktree`

EliahKagan:
This removes the `gix-ignore` and `gix-index` direct dependencies of `gix-testtools`, by having `gix_testtools` use them through `gix-worktree`, accessing `gix_worktree::ignore` for `gix_ignore` and `gix_worktree::index` for `gix_index`.

The rationale is that various other `gix-*` dependencies were used this way already, and also that this specific change appears to have been planned, based on the TODO comment in ddaacda (#1413).

Just in case this were to have unanticipated ramifications, I've rerun the full test suite, after `gix clean -xde`, both with and without `GIX_TESTS_IGNORE_ARCHIVES=1`, locally on Ubuntu 22.04 LTS and Windows 10.

([On Windows 10](https://gist.github.com/EliahKagan/1aa24e5ba916ede6c9a1788d08c10aff) without using pre-generated archives, there are failures, but the failures are the same as in #1358 and https://github.com/Byron/gitoxide/issues/1358#issuecomment-2342635661 in [this gist](https://gist.github.com/EliahKagan/1ff070df2c144f26ffacc7c73ff14143), indicating that the changes in this PR have not broken Windows either.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2577,8 +2577,6 @@ dependencies = [
  "fs_extra",
  "gix-discover 0.32.0",
  "gix-fs 0.11.3 (registry+https://github.com/rust-lang/crates.io-index)",
- "gix-ignore 0.11.4 (registry+https://github.com/rust-lang/crates.io-index)",
- "gix-index 0.33.1",
  "gix-lock 14.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "gix-tempfile 14.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "gix-worktree 0.34.1",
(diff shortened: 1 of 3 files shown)


EliahKagan:
The linked gists relate to https://github.com/Byron/gitoxide/issues/1358#issuecomment-2342635661.

(This comment is a workaround for https://github.com/orgs/community/discussions/138229, since cross-linking seems valuable here. I will hide this comment almost immediately.)

Byron (approved):
Thanks a lot for all this cleanup, it's much appreciated!

PR #1594, opened 2024-09-14
Improve comments in a couple private functions

EliahKagan:
This tiny PR has some comment edits that I would ideally have included in previous PRs but didn't notice the need for at the time, and that don't seem like they are particularly suited to any other PRs I would likely make in the next couple days. So I'm opening this for them.


Byron (approved):
Thanks a lot!
I am also looking forward to the other PRs.

Issue #463, opened 2022-07-19
WASM support


Byron:
The ability to compile to `--target WASM`. Here we track what's needed to make this happen as well as the progress thus far.

### Tasks

Analysis TBD.

dzmitry-lahoda:
you may consider support of `no_std`(it is the proper way to do wasm rather than just compile to target). afaik embedded and blockchain targets are `no_std`  and these (wasm stems from being no_std

dzmitry-lahoda:
on top of allowing in-browser ide to work better, this feature will allow more secure software release and dependency verifications. (there is version of `git` in blockchain)

dzmitry-lahoda:
after no_std is done for core libs, can support `wasi` (os/fs interface of some no_std hosts). that is separate task.

uberroot4:
Hi @Byron, as there was not much effort/interest in this issue the last two years I want to ask if there are any plans currently?
Looking at the list of crates to be fixed, are they still up-to-date? As there was much change since opening the issue, I assume it is not exhaustive anymore?

I executed `cargo tree` in the `gix` package and tried to come up with a bottom-up list to identify an order of the crates to be fixed:
- [ ] gix-hash
- [ ] gix-validate
- [ ] gix-trace
- [ ] gix-tempfile
- [ ] gix-utils
- [ ] gix-actor
- [ ] gix-date
- [ ] gix-chunk
- [ ] gix-config-value
- [ ] gix-quote
- [ ] gix-features
- [ ] gix-path
- [ ] gix-sec
- [ ] gix-fs
- [ ] gix-traverse
- [ ] gix-object
- [ ] gix-attributes
- [ ] gix-index
- [ ] gix-commitgraph
- [ ] gix-ref
- [ ] gix-discover
- [ ] gix-revwalk
- [ ] gix-credentials
- [ ] gix-archive
- [ ] gix-diff
- [ ] gix-hashtable
- [ ] gix-ignore
- [ ] gix-config
- [ ] gix-filter
- [ ] gix-odb
- [ ] gix-pack
- [ ] gix-negotiate
- [ ] gix-worktree
- [ ] gix-pathspec
- [ ] gix-mailmap
- [ ] gix-command
- [ ] gix-revision
- [ ] gix-status
- [ ] gix-url
- [ ] gix-refspec
- [ ] gix-dir
- [ ] gix-merge
- [ ] gix-submodule
- [ ] gix-worktree-state
- [ ] gix-worktree-stream
[...]

PR #1600, opened 2024-09-21
Implemented `is_path_owned_by_current_user` for wasi.

Earthmark:
Not that it's doing very much when wasi doesn't have a concept of current user.

For more context:

* Wasi filesystems: https://github.com/WebAssembly/wasi-filesystem/blob/main/wit/types.wit
  * File stat doesn't include a user ID.
* Wasi CLI: https://github.com/WebAssembly/wasi-cli/tree/main/wit
  * I have yet to find any reference to a user.

Emscripten does have a concept of user, so this isn't a valid solution for that target.

This is part of trying to get this to build with wasi as a build target, AKA: https://github.com/Byron/gitoxide/issues/463

This gets gix-sec building for wasi.


Earthmark:
Technically this helps with https://github.com/Byron/gitoxide/issues/463

Earthmark:
It turns out the `wasm32-unknown-emscripten` target already builds successfully for gix-sec.

I don't know if `wasm32-unknown-unknown` is a valid target, I don't believe that target has a concept of file IO, let alone users.

Byron (approved):
Thanks a lot!

I think this should work and I have added a (hopefully working) test that will validate that.

Issue #1608, opened 2024-09-24
"Tests pass" blocks PRs that don't need tests


EliahKagan:
### Current behavior 😯

Tests run on pull requests when changes are made to files in:

https://github.com/Byron/gitoxide/blob/612896d7ec15c153d3d48012c75aaf85d10a5abe/.github/workflows/ci.yml#L28-L37

Otherwise they are intentionally skipped as unnecessary.

That applies to the `ci.yml` workflow as a whole. The "Tests pass" check, which is defined in that workflow, as well as the actual test jobs it depends on, do not run on pull requests that don't make any of those modifications. An example of such a pull request is #1607.

### Expected behavior 🤔

If there is a way to do it that works well, only relevant checks should be required for pull requests.

A possible alternative is to keep things as they are and handle merging of such pull requests manually. Most pull requests do run tests, so auto-merging is still very useful even if this is not fixed. But I suspect there may be a reasonable way to achieve the effect of only requiring checks when they are relevant.

Another consideration, though I'm not sure if it matters, is that PRs are sometimes opened with minimal initial contents, sometimes even empty. Such a PR intentionally does not run unnecessary tests when opened. There might be a disadvantage to such a PR being auto-mergeable, even if it has a merging review from a user whose reviews would otherwise permit that.

### Git behavior

As far as I can quickly tell, neither [git/git](https://github.com/git/git) nor [git-for-windows/git](https://github.com/git-for-windows/git) use the GitHub auto-merging feature. ([git/git](https://github.com/git/git) is also a mirror, and its PRs work via [GitGitGadget](https://gitgitgadget.github.io/), so it wouldn't necessarily be comparable in this area.)

### Steps to reproduce 🕹

There are two ways to verify this:

- Open a new PR that does not modify any files whose paths match the patterns listed above. For a compelling example, the PR should make *some* change, so as to resemble a PR that might be in a mergeable state. *Or*
- Just examine #1607.

PR #1607, opened 2024-09-24
Tell github-linguist to count fixture shell scripts

EliahKagan:
By default, github-linguist [treats source code files inside `tests/fixtures` directories as vendored](https://github.com/github-linguist/linguist/blob/9b50e9e46bc0fb2451f5deee612a36c386aae6cf/lib/linguist/vendor.yml#L342-L344), thereby declining to consider them when computing the percentage each language contributes to a repository:

```yml
# Test fixtures
- (^|/)[Tt]ests?/fixtures/
- (^|/)[Ss]pecs?/fixtures/
```

But in this repository, the shell scripts inside `tests/fixtures` are code that is part of the test suite. The effect of treating them as vendored is that only the comparably few shell scripts that are not test fixture scripts contribute to the language computation (and possibly other uses of github-linguist by various tooling? I am not sure).

Thus, the "Languages" bar on GitHub for this repository (or in forks) currently shows 97.5% Rust, 1.3% HTML, and 1.2% Other. But "Shell" should really be shown as a the second most prevalent language here, with a percentage significantly exceeding that of HTML.

[Running `github-linguist -b` locally](https://gist.github.com/EliahKagan/23d667cfdd2cffa38eff2579a86e4c11) verifies the situation.

Marking the affected files as *not* vendored, with the usual pattern we have been using to adjust their attributes in other ways, in the top-level `.gitattributes` file, fixes that. This PR makes that change.

This changes the output of `github-linguist` from:

```text
97.63%  7154452    Rust
1.32%   96933      HTML
0.94%   69205      Shell
0.10%   7100       Makefile
0.00%   108        Dockerfile
```

To:

```text
94.98%  7154452    Rust
3.64%   274106     Shell
1.29%   96933      HTML
0.09%   7100       Makefile
0.00%   108        Dockerfile
```

I have verified that `github-linguist` identifies all the expected files by running:

```sh
find . -name '*.sh' -printf '%P\n' | sort >a
github-linguist -b | grep -E '\.sh$' | sort >b
diff a b
```

(Where the reason for `-printf '%P\n'` was to omit the leading `./` that would otherwise stymie `diff`. I don't know if this is portable; I used GNU find.)

On the `count-sh` branch--the feature branch for this PR--that gives no output.

Regarding non-obvious specifics of the edit to `.gitattributes`:

- While I don't think the pattern `**/tests/fixtures/**/*.sh` confers a benefit over `tests/fixtures/**/*.sh`, I've used it because it was already in use.
[...]

PR #1609, opened 2024-09-24
Only fuzz test PRs that run the test suite

EliahKagan:
This adds a `paths` restriction on the CIFuzz job to avoid the situation where a PR that doesn't modify anything that causes the usual tests to run on it would still run fuzz tests.

An example of a such a PR is #1607, where, as things are set up now, the CIFuzz job ran, taking 19 minutes, even though the only other CI jobs that needed to run there were the MSRV jobs. (I'm not entirely sure they need to run either, but it seems okay since they are rather light weight.)

Note that while #1608 also relates to a potential non-ideal CI-related behavior observed in #1607, that is unrelated to this, and this PR does *not* fix issue #1608.

This should still have CIFuzz run even when the only change a PR makes is to adjust how CI fuzzing itself works, because one of the patterns is matched by any change to any CI workflow file. In particular, this PR itself should be fuzzed under these rules.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,7 +1,18 @@
 name: CIFuzz
 on:
   pull_request:
-    branches: [ main ]
+    branches:
+      - main
+    paths:
+      - '.github/**'
+      - 'ci/**'
+      - 'etc/**'
+      - 'src/**'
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot! This will save some time.

PR #1610, opened 2024-09-26
feat: add option to traverse commits from oldest to newest

nrdxp:
This change introduces an enum to control commit traversal order for Simple. Users can now choose between newest-first or oldest-first traversal. The default behavior remains newest-first, but it can be toggled by passing a CommitTimeOrder to a Sorting::ByCommitTime* variant.

This feature is particularly useful for searching early repository history, which should be orders of magnitude faster with this variant. In my benchmarking I was able to identify the original commit in a large mono-repo source tree using the new OldestFirst variant in 12ms down from 2.6s with NewestFirst.

The implementation logic remains largely agnostic to this change, with only minor adjustments in key areas as necessary.

The reversed order is achieved by inverting the PriorityQueue key with std::cmp::Reverse when an oldest-first traversal is requested. This is what allows the bulk of the original logic to remain unchanged.

### DIFF
--- a/gix-traverse/src/commit/simple.rs
+++ b/gix-traverse/src/commit/simple.rs
@@ -6,12 +6,13 @@ use std::cmp::Reverse;
 use std::collections::VecDeque;
 
 #[derive(Default, Debug, Copy, Clone)]
-/// The order with which to prioritize the search
+/// The order with which to prioritize the search.
 pub enum CommitTimeOrder {
     #[default]
-    /// sort commits by newest first
+    /// Sort commits by newest first.
     NewestFirst,
-    /// sort commits by oldest first
+    /// Sort commits by oldest first.
+    #[doc(alias = "Sort::REVERSE", alias = "git2")]
     OldestFirst,
 }
 
@@ -31,7 +32,7 @@ pub enum CommitTimeOrder {
 pub enum Sorting {
     /// Commits are sorted as they are mentioned in the commit graph.
     ///
-    /// In the *sample history* the order would be `8, 6, 7, 5, 4, 3, 2, 1`
+    /// In the *sample history* the order would be `8, 6, 7, 5, 4, 3, 2, 1`.
     ///
     /// ### Note
     ///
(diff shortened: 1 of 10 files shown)


gix-traverse/src/commit/simple.rs
```diff
@@ -117,19 +145,23 @@ mod init {
...
-                        match cutoff_time {
-                            Some(cutoff_time) if time >= cutoff_time => {
-                                state.queue.insert(time, commit_id);
+                        let ordered_time = order_time(time, order);
+                        match (cutoff_time, order) {
+                            (Some(cutoff_time), CommitTimeOrder::NewestFirst) if time >= cutoff_time => {
+                                state.queue.insert(ordered_time, commit_id);
+                            }
+                            (Some(cutoff_time), CommitTimeOrder::OldestFirst) if time <= cutoff_time => {
+                                state.queue.insert(ordered_time, commit_id);
```
Byron, line 160: Did you invert the logic because it made sense or because it *seemed* right?

The reason I am curious is:

* there are no tests for this
* the way I think this could work with similar semantics is if the traversal would actually start from the back, and move forward in time, something that's not possible with a commitgraph.

My intuition here is to keep the cutoff logic the same, as the traversal direction is always from front to back, or newest to youngest.

Byron (commented):
Thanks a lot for setting this up - it's great to see that this sorting mode along can make such a difference!

I will keep working on this PR until it's ready for merging, by adding some tests and probably fix one logic-issue I am seeing. But tests should say what's right or wrong here.

Byron (approved):
I think that's the way to go, after removing the inverted cutoff logic essentially and adding tests.

Despite merging this, I am particularly interested to hear if this is the wrong way to go about it, and potential fixes are very welcome in follow-up PRs.

Thanks again for kicking off this feature!

gix-traverse/src/commit/simple.rs
nrdxp: I was considering that as well, I actually wasn't 100% confident either way, so I'm happy to keep it simple unless something breaks. I have a real-world usecase for this, so I will let you know. But I can also devise some tests for this in a follow up PR as well

gix-traverse/src/commit/simple.rs
```diff
@@ -61,11 +75,14 @@ pub enum Error {
     ObjectDecode(#[from] gix_object::decode::Error),
 }
 
+use Result as Either;
+type QueueKey<T> = Either<T, Reverse<T>>;
```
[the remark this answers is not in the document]
nrdxp: That was the idea yes, also just wanted to make it clear that this is not an error condition.

PR #1618, opened 2024-10-09
octopus-merge (part 5: tree-merge-ORT three-way)

Byron:
Three-way merging of trees.

Follow-up of #1612.

### Tasks

* [x] baseline tests for tree-merging
* [x] successful first tree-merge without conflicts
* [x] make all tests pass *(catch-all task that will be refined on the fly)*
     - [x] fully parse all information provided by Git in baseline
     - [x] update the conflict-lookup table to reflect changes to the tree as well - avoid double conflicts or missed conflicts if something clashes with a newly added rename, for instance.
     - [x] conflict-tree by index to allow getting 'next' of ours for lookahead
     - [x] `tree::Editor::get()` to find unique path names
     - ~~multi-tree traversal (without wildcard support!)~~ - for now, let's just do it 'the simple' way and perform two diffs 
    - See also, all the [merge options](https://git-scm.com/docs/git-config#Documentation/git-config.txt-mergedirectoryRenames)
* [x] gracefully handle submodules (by not attempting resolution)
* [x] assure content-merges only happen across the same type/mode (and what happens with symlinks)
* [x] assure we handle mode-merges similarly
* [x] support for `merge` attribute
* [x] there is tests for merging merge-bases in `diff3-conflict-markers`
    - `diff3-conflict-markers.sh`  - be sure to capture the 'empty tree' label , but also other special cases
    - [x] binary merges are pre-empted as using 'ours' for the merge somehow…
* [x] ~~have motivating tests to put `possibly_rewritten_location()` into all the right places~~ - not feasible for now, let's get this merged and tested in the real world
* [x] treat symlink merges like content merges, and respect the text-merge resolution setting
    - [x] when resolving merge-bases, symlink conflicts are resolved with the ancestor.
    - [x] otherwise there is a choice which symlink edit to chose, similar to binary resolution
* [x] Allow to turn off renaming to use the version from the merge-base (and only that)
* [x] replace any kind of assertion with conflicts so it can't crash, and won't screw up, AKA fail gracefully
* [x] ~~test `copy` versions of rewrites~~ - it's not in merge-ORT so let's skip that
* [x] ~~see if `remove_existing_leaf()` calls are necessary~~ - they are needed
* [x] fix reversed tests for blobs
* [x] fuzzing for blob merges
* [x] remove TODOs - everything not specifically handled is a conflict
* [x] `Repository::merge_trees` 
* [x] A trivial `gix merge tree` implementation, based on commits. Maybe create something to easily merge multiple commits while at it (in `gix`).
* [x] speed up writes

### Next PR / Outscoped

* `Repository::merge_commits()`
* A trivial `gix merge commit`
* A nicely wrapped `gix` API - for now all plumbing options and return values are exposed.
* Submodule merges are also possible! Maybe outscope it though! `libgit2` also doesn't try it.
* `textconv` with context, see [this gist](https://gist.github.com/EliahKagan/31517ea8d4bb0300a521eb63c6b2efd2) for details.
    - There seem to be different 'tiers' of tools, some don't get `GIT_DIR` set, others do.
    - It also seems that diff-programs get too much context right now, but that depends on how much is passed to them by the caller as `gix-command::Context`.
* How to model virtual-merge-bases? Can be none or many, user should have control over how this is done.
* Actual tree-based merging
* ~~Make blob-merge based conflict handling possible in the tree merge from `gix` at least.~~ - not needed for now
* consider finding a way to more clearly say what happened, maybe with some additional explanation so it's more of a drill-down way of working, instead of a super-enum with a lot of very similar cases.

### Research

* *optimization idea for when there are numbers*: a 'brute force' implementation that uses threads would benefit from the 

#### Everything is about MergeORT.

ability to re-use object caches of a repo that has seen the base-tree already, but overall, who knows*
* merge-options passed with `-X ours` for instance don't affect tree-related auto-resolutions, just the ones related to content. This could be implemented when there is demand though.
* Note that `merge-ORT` cannot *properly* handle renames within renamed directories, ending up with the source of the subdir-rename still present.
 ```
❯ git ls-tree -r $(git merge-tree main feat)
100644 blob e69de29bb2d1d6434b8b29ae775ad8c2e48c5391    a
100644 blob e69de29bb2d1d6434b8b29ae775ad8c2e48c5391    git-sec-renamed/2
100644 blob e69de29bb2d1d6434b8b29ae775ad8c2e48c5391    git-sec-renamed/7
100644 blob e69de29bb2d1d6434b8b29ae775ad8c2e48c5391    git-sec-renamed/subdir/6
100644 blob e69de29bb2d1d6434b8b29ae775ad8c2e48c5391    git-sec/subdir-renamed/6
100644 blob e69de29bb2d1d6434b8b29ae775ad8c2e48c5391    git-sequencer
100644 blob e69de29bb2d1d6434b8b29ae775ad8c2e48c5391    gix/5
100644 blob e69de29bb2d1d6434b8b29ae775ad8c2e48c5391    h
```
* It puts conflict-markers in the blobs of the result tree, with annotations to provide additional context
* data stored by path, and is interned in the map to allow pointer-based comparisons
    - merge-info with everything one needs to know, also related to renames
    - or conflict information
    - it uses a memory-pool/arena to get memory for many paths all at once (and also release it like that)
* paths start out as conflicted, and then can later be changed to non-conflicting if a content-based merged succeeded.
    - If it remains conflicts, the meta-data is used to produce an 'as merged as possible' version with conflict markers that can be checked out to the working tree.
* hunks can partially overlap, but can also be resolved [line-by line to some extend](https://neg4n.dev/blog/understanding-zealous-diff3-style-git-conflict-markers).


EliahKagan:
As detailed in #1623, which provides a fix, the failure observed here in the CI `test` job is actually not due to any of the changes in this PR, and also occurs if CI is re-run on the tip of main. It is instead due to the upgraded runner image not having the headers needed for building with `-llzma`, which is needed for the `xz` feature of `gix-testtools`. This is also entirely unrelated to #1622, which does it yet affect CI.

Assuming all checks pass on #1623, merging it and then rebasing this onto main should fix the the `test` failure here. The other failure here is in the `lint` job and unrelated.

PR #1620, opened 2024-10-11
fix discovery

Byron:
### Tasks

* [x] try in GitButler

* Discovered in https://github.com/gitbutlerapp/gitbutler/pull/5100


gix-discover/src/lib.rs
```diff
@@ -39,6 +39,8 @@ pub mod is_git {
         Metadata { source: std::io::Error, path: PathBuf },
         #[error("The repository's config file doesn't exist or didn't have a 'bare' configuration or contained core.worktree without value")]
         Inconclusive,
+        #[error("Could not obtain current directory when conforming repository path")]
```
EliahKagan, line 42: Is this, and the related `conformed_git_dir` variable, about something conforming to a standard or expectation? Or are they typos, where "confirming" and `confirmed_git_dir` should be written instead?

If "conform" is the intended word, then I suggest rephrasing this message for clarity and probably also renaming the `conformed_git_dir` variable to `conforming_git_dir`. If "confirm" is intended--i.e. if the intended meaning is that we are verifying that something is a git dir--then I recommend changing this to "confirming" and renaming the variable to `confirmed_git_dir`.

I'm not sure which meaning is intended, so I haven't proposed a specific edit at this time. Also, I couldn't propose it in a review comment easily, because at least as applied to the variable name, this is a preexisting ambiguity, and some occurrences of the variable name are in places this PR does not modify. If it is known that one of these two changes is desired, though, then I'd be pleased to open a PR for them ~~after~~ now that this PR is merged.

gix-discover/src/repository.rs
```diff
@@ -112,8 +114,11 @@ pub enum Kind {
     /// Note that this is merely a guess at this point as we didn't read the configuration yet.
     ///
     /// Also note that due to optimizing for performance and *just* making an educated *guess in some situations*,
-    /// we may consider a non-bare repository bare if it it doesn't have an index yet due to be freshly initialized.
-    /// The caller is has to handle this, typically by reading the configuration.
+    /// we may consider a non-bare repository bare if it doesn't have an index yet due to be freshly initialized.
+    /// The caller has to handle this, typically by reading the configuration.
+    ///
+    /// It could also be a directory which is non-bare by configuration, but is *not* named `.git`.
+    /// Unusual, but it's possible that a worktree is configured via `core.worktree`.
```
EliahKagan, line 121: Is there also the possibility that the worktree path is established by the presence of a [`GIT_WORK_TREE`](https://git-scm.com/docs/git#Documentation/git.txt-codeGITWORKTREEcode) environment variable? (Or by the presence of a [`GIT_DIR`](https://git-scm.com/docs/git#Documentation/git.txt-codeGITDIRcode) environment variable together with no `GIT_WORK_TREE` variable? In this case, I think the worktree is implicitly taken to be the process CWD, at least in Git.)

gix-discover/src/lib.rs
Byron: Thanks for the hint - the affected variable name is definitely leaking here and I improved the message to be much more specific [in this PR](https://github.com/Gitoxide-Labs/gitoxide/pull/1618).

gix-discover/src/repository.rs
Byron: In `gitoxide`, plumbing crates aren't allowed to read environment variables directly, so this is not handled here.
There is [a function in `gix`](https://github.com/Gitoxide-Labs/gitoxide/blob/f8952e4cbfaf9ab7ddc12a028a1cdb821ac9a3b1/gix/src/open/repository.rs#L114) though that reads these variables and uses them detect repositories from there. 
It's most definitely not fully conforming with what Git does.

EliahKagan: Thanks--I've commented https://github.com/GitoxideLabs/gitoxide/pull/1585#discussion_r1797734926 with a note about including or otherwise revisiting this when I open the other environment-related compatibility issues per the plan in https://github.com/GitoxideLabs/gitoxide/pull/1585#discussion_r1764341739.

Byron: Maybe as an additional note, despite claiming it's 'most definitely' not conforming, what I tried to say is that I didn't really try to make it conforming in the first place. To me it felt like environment variables are only needed in certain contexts, for instance, when creating a Git hook with `gitoxide`, and the implementation should rather be so that `gitoxide` can be used in certain situations (of which I probably only know one), and otherwise never be able to lead to something surprising. This is a general problem with environment variables, their implicitness, and `gitoxide` already makes them explicit by [mapping every one of them to a git configuration variable](https://github.com/GitoxideLabs/gitoxide/blob/d2ae9d5f11be9f2561f6799d88804d0d8eae33ef/gix/src/config/cache/init.rs#L311), which is used in its place.

Issue #1622, opened 2024-10-13
regex_matches baseline check fails due to rev-parse changes in git 2.47.0


EliahKagan:
*Update:* See https://github.com/GitoxideLabs/gitoxide/issues/1622#issuecomment-2529580735 on the cause of this bug.

### Current behavior 😯

When `git` is git 2.47.0 and tests are run with `GIX_TEST_IGNORE_ARCHIVES=1` to force fixture scripts--and in particular the [`make_rev_spec_parse_repos.sh`](https://github.com/GitoxideLabs/gitoxide/blob/main/gix/tests/fixtures/make_rev_spec_parse_repos.sh) script--to run, the `find_youngest_matching_commit::regex_matches` rev-spec test fails:

```text
     Summary [  72.831s] 2560 tests run: 2559 passed (1 leaky), 1 failed, 3 skipped
        FAIL [   0.024s] gix::gix revision::spec::from_bytes::regex::find_youngest_matching_commit::regex_matches
error: test run failed
```

More specifically:

```text
        FAIL [   0.024s] gix::gix revision::spec::from_bytes::regex::find_youngest_matching_commit::regex_matches

--- STDOUT:              gix::gix revision::spec::from_bytes::regex::find_youngest_matching_commit::regex_matches ---

running 1 test
test revision::spec::from_bytes::regex::find_youngest_matching_commit::regex_matches ... FAILED

failures:

failures:
    revision::spec::from_bytes::regex::find_youngest_matching_commit::regex_matches

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 297 filtered out; finished in 0.01s


--- STDERR:              gix::gix revision::spec::from_bytes::regex::find_youngest_matching_commit::regex_matches ---
thread 'revision::spec::from_bytes::regex::find_youngest_matching_commit::regex_matches' panicked at gix/tests/gix/revision/spec/from_bytes/util.rs:181:13:
[...]

PR #1623, opened 2024-10-13
Install `liblzma-dev` in CI `test` job

EliahKagan:
Per https://github.com/actions/runner-images/issues/10636, the `ubuntu-latest` GitHub Actions hosted runner is being migrated from `ubuntu-22.04` to `ubuntu-24.04`. The `ci.yml` workflow, including the full `test` job, uses `ubuntu-latest`, and recently has switched over to using the Ubuntu 24.04 runner.

It makes sense to use this newer version, but that runner apparently does not preinstall the headers required to build with `-llzma`. That causes `just ci-test` to fail at:

```sh
cargo nextest run -p gix-testtools --features xz
```

The error message is:

```text
/usr/bin/ld: cannot find -llzma: No such file or directory
```

This PR installs the `liblzma-dev` package that provides the required headers, so we can use Ubuntu 24.04 LTS while continuing to test the `xz` feature of `gix-testtools`.

The failure this addresses was first observed in #1618 at 64e0f78: https://github.com/GitoxideLabs/gitoxide/actions/runs/11304216949/job/31442302018?pr=1618

But it is not caused by any of the changes there, and it now occurs whenever the `test` job is run, including if it is re-run at the tip of the main branch (or any other branch), such as in: https://github.com/EliahKagan/gitoxide/actions/runs/11317510844/job/31471040361

In both cases, expanding *Set up job* > *Operating system* shows Ubuntu 24.04.1 LTS.

(The failure this addresses is also not related to #1622, which documents an unrelated failure that is possible to observe locally and that, if no change is made related to it, can be expected to occur on CI in the `test` job starting sometime soon, but that is not yet seen on CI.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - name: Setup dependencies
         run:
-          sudo apt-get install tree
+          sudo apt-get install -y --no-install-recommends liblzma-dev tree
       - uses: extractions/setup-just@v2
       - uses: taiki-e/install-action@v2
         with:


Byron:
Thanks a million, much appreciated!

Issue #1406, opened 2024-06-17
Sovereign Tech Fund and NLNet Application


Byron:
This issue tracks the status for the application to https://www.sovereigntechfund.de (STF)  as well as to https://nlnet.nl .

Note that the STF is mutually exclusive, but the idea is to have options.

### Tasks

* [ ] Apply to STF
* [ ] Apply to NlNet
* [ ] figure out how to receive funds
* [ ] figure out how to spend funds

EliahKagan:
Thanks! I've joined the [organization](https://github.com/GitoxideLabs).

(By the way, it looks like the link above, in the 2024-10-11 section, is broken: Even though it looks like the organization was initially named `Gitoxide-Labs` and then renamed to `GitoxideLabs`, [the URL with `Gitoxide-Labs`](https://github.com/organizations/Gitoxide-Labs) does not redirect to [the URL with `GitoxideLabs`](https://github.com/GitoxideLabs).)

Byron:
Thank you for joining, and thanks for all the useful hints :). The link above is now fixed, and I set my group membership to public as well.

Byron:
Moved to [this Tasks Discussion](https://github.com/GitoxideLabs/gitoxide/discussions/2804). This issue is being closed because task tracking now lives in Discussions.

_AI disclosure: This comment was written and posted by OpenAI Codex on behalf of Byron._

PR #1624, opened 2024-10-14
Update gitoxide repository URLs

EliahKagan:
This changes `Byron/gitoxide` URLs to `GitoxideLabs/gitoxide` in documentation and a few other places, now that the repository has been moved into an organization as detailed in #1406. The old URLs still work, redirecting to the new ones, so that change is not needed to fix broken links, but I think it is still a good idea. This makes the change in Markdown docs except those such as changelogs where it should not be done, in documentation comments, in manifest files, and in the documentary top-comment of the `pre-push` sample hook. See 64ff0a7 for details.

This also makes a few other changes to URLs that *do* fix broken links, which had been broken before and the above change led me to discover the breakage. For example, some links had previously had their paths updated to account for a rename without their OIDs being updated. See 65f69f7 for details. There are some more broken links I found during this process where the solution is [less readily](https://gist.github.com/EliahKagan/2460ac1f7459215d40aee02af55a503f) apparent; for those, I plan to open a separate PR and/or an issue requesting what should be done. It seemed to me that that those further hoped-for changes are non-blocking for this PR, since they pertain to links that were already broken and are made no worse here.

I think that when `Byron/gitoxide` appears in *test data*, it should usually be left unchanged, or at least not changed unless something nearby is already being changed. But in some places I think it makes sense to change it. See 507579e for details, including on why I did not change it in more places. However, I did update the URLs in the `Makefile`, as detailed in 1ea08bf.


Byron:
Thanks again for this very mindful update, something I conveniently skipped over entirely. In my mind I already moved on and probably would have relied on the redirect way too long 😅.

Also I am looking forward to the follow-up that deals with the previously broken links - I'd just trust your judgement with these as well.

EliahKagan:
> Also I am looking forward to the follow-up

I've opened #1627 and #1628 to discuss and propose further improvements.

Please note that #1628 does not fix issue #1627 because it does not have anything for #1627's first two tasks.

Issue #1627, opened 2024-10-14
Outdated broken links from documentation into code


EliahKagan:
### Current behavior 😯

There were three broken links found while working on #1624 that I didn't include fixes for in 65f69f7 because there was no clear best way to fix them. For two of them, which will probably require updating planning information. I am not at all sure what should be done, though I've tried to suggest broad possibilities. Those are the ones that motivate this issue. For the third, I'm pretty sure what the correct fix is and I'll go ahead and open a PR for it, but I've included it here as well so it can be easily reviewed.

- [ ] crate-status.md:232 ([#1631](https://github.com/GitoxideLabs/gitoxide/pull/1631))
- [ ] general-tasks.md:19 ([#1631](https://github.com/GitoxideLabs/gitoxide/pull/1631))
- [x] gix-diff/src/tree/function.rs:27 ([#1628](https://github.com/GitoxideLabs/gitoxide/pull/1628))

Please note that the broken links described in this issue were effectively unaffected by the change of `Byron/gitoxide` to `GitoxideLabs/gitoxide` in 64ff0a7: although that part of their URLs changed, they were broken in exactly the same way both before and after that change to their URLs.

#### In `crate-status.md`

[That link](https://github.com/GitoxideLabs/gitoxide/blob/8f9a55bb31af32b266d7c53426bc925361a627b2/gix-pack/src/cache/delta/from_offsets.rs#L101-L105) is a 404, because it specifies a `gix-pack` path at a commit before `git-pack` was renamed to `gix-pack`:

https://github.com/GitoxideLabs/gitoxide/blob/795962b107d86f58b1f7c75006da256d19cc80ad/crate-status.md?plain=1#L232

Changing the path back to `git-path` gives:

https://github.com/GitoxideLabs/gitoxide/blob/8f9a55bb31af32b266d7c53426bc925361a627b2/git-pack/src/cache/delta/from_offsets.rs#L101-L105

With more context, this is:

https://github.com/GitoxideLabs/gitoxide/blob/8f9a55bb31af32b266d7c53426bc925361a627b2/git-pack/src/cache/delta/from_offsets.rs#L100-L110

But the code comment, which `crate-status.md` links to, went away in https://github.com/GitoxideLabs/gitoxide/commit/84ade1d23060f10bf6c8529f8f693d06660b4f4e. The commit message there is:

> fix: Allow resolution of in-pack ref-deltas (#287)
>
[...]

PR #1628, opened 2024-10-14
Update and improve another few URLs

EliahKagan:
This builds on 65f69f7 (#1624, see also https://github.com/GitoxideLabs/gitoxide/pull/1624#issuecomment-2410324422), completing the third task in #1627 (`gix-diff/src/tree/function.rs:27`), fixing a broken link and updating another. See 991c939 for details.

This does not do the first or second task of #1627 because I have not (yet?) figured out how. I think those improvements could go in another PR, once the best solution is figured out.

I've also improved on 1ea08bf (#1624) by giving the nonexistent repo used for a 401 response the more self-documenting name `nonexistent`. See baa6f23 for the rationale and a few details.


Byron (approved):
Thanks a lot for finding the new spot of these references :)!

PR #1629, opened 2024-10-15
Ignore non-regular files

krobelus:
(very much unfinished, I will probably finish it later this week).

My "mkfifo fifo && cargo publish" hangs because cargo tries to tar
the fifo as well.  This is surprising because neither "git status"
nor "git check-ignore -v fifo" list the untracked fifo.

Fifo type files are completely ignored by Git since commit 8695c8bfe1
(Add "show-files" command to show the list of managed (or non-managed)
files., 2005-04-11) which states the reason:

> * [...] We currently ignore anything but
> * directories and regular files. That's because git doesn't
> * handle them at all yet. Maybe that will change some day.

It didn't change yet[*]. Try to match Git to minimize surprise.
Combined with some extra error handling, this enables some cleanup.

[*] But the logic from read_directory() has moved into treat_path().

### DIFF
--- a/gix-dir/src/walk/readdir.rs
+++ b/gix-dir/src/walk/readdir.rs
@@ -3,7 +3,7 @@ use std::borrow::Cow;
 use std::path::{Path, PathBuf};
 use std::sync::atomic::Ordering;
 
-use crate::entry::{PathspecMatch, Status};
+use crate::entry::{Kind, PathspecMatch, Status};
 use crate::walk::function::{can_recurse, emit_entry};
 use crate::walk::EmissionMode::CollapseDirectory;
 use crate::walk::{
@@ -59,12 +59,24 @@ pub(super) fn recursive(
         );
         current.push(file_name);
 
+        let disk_kind = {
+            let m = current.metadata().map_err(|err| Error::DirEntryFileType {
+                path: std::mem::take(current),
+                source: err,
+            })?;
+            match Kind::try_from(m.file_type()) {
+                Ok(kind) => kind,
+                Err(_err) => {
+                    continue;
+                }
+            }
+        };
+
         let mut info = classify::path(
             current,
             current_bstr,
             if prev_len == 0 { 0 } else { prev_len + 1 },
-            None,
-            || entry.file_type().ok().map(Into::into),
+            disk_kind,
             opts,
             ctx,
         )?;
@@ -94,7 +106,7 @@ pub(super) fn recursive(
             }
         } else {
             if opts.for_deletion == Some(ForDeletionMode::IgnoredDirectoriesCanHideNestedRepositories)
-                && info.disk_kind == Some(entry::Kind::Directory)
+                && info.disk_kind == entry::Kind::Directory
                 && matches!(info.status, Status::Ignored(_))
             {
                 info.disk_kind = classify::maybe_upgrade_to_repository(
@@ -227,7 +239,7 @@ impl Mark {
                 property: {
                     assert_ne!(
                         dir_info.disk_kind,
-                        Some(entry::Kind::Repository),
+                        entry::Kind::Repository,
                         "BUG: it shouldn't be possible to classify an empty dir as repository"
                     );
                     if !state.may_collapse(dir_path) {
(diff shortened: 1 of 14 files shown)


gix-dir/src/walk/readdir.rs
```diff
@@ -59,12 +59,24 @@ pub(super) fn recursive(
         );
         current.push(file_name);
 
+        let disk_kind = {
+            let m = current.metadata().map_err(|err| Error::DirEntryFileType {
```
Byron, line 63: I really tried hard to avoid checking each file on disk, but maybe that's ultimately futile as one can't even assume the disk isn't changed when finding a file in the index. Maybe that's also the source of subtle (and untested) bugs.

Also I hope that in practice, there will be no noticeable performance penalty.

gix-status/src/index_as_worktree/function.rs
```diff
@@ -270,16 +270,12 @@ impl<'index> State<'_, 'index> {
...
-                path,
-                Some(entry.mode.is_submodule()),
-                &mut |relative_path, case, is_dir, out| {
-                    self.attr_stack
-                        .set_case(case)
-                        .at_entry(relative_path, Some(is_dir_to_mode(is_dir)), objects)
-                        .map_or(false, |platform| platform.matching_attributes(out))
-                },
-            )
+            .pattern_matching_relative_path(path, true, &mut |relative_path, case, is_dir, out| {
```
Byron, line 273: I don't get why this defaults to `true` now, and would hope there is a test to catch it. If not, definitely a reason to add a couple, probably including a baseline to validate Git works similarly.

Byron (commented):

I am surprised these seemingly sweeping changes are needed when trying to ignored 'special' files, but would always be open to improvements. And changing `Option<bool>` to `bool` certainly counts as such.

However, I need this PR to change in structure so that…

* …there is one commit per crate when dealing with plumbing crates.
    - if there is a breaking change, that must be indicated as conventional commit message. That's critical to the release process.
* …there is one commit over all changes necessary to adapt to breaking changes (i.e. make it compile after breaking plumbing changes).

I already left a couple of comments, maybe it's too early though and in that case, please ignore.

It's really important to reorganise the commits before continuing though.

Thank you

gix-dir/src/walk/readdir.rs
krobelus: I'm definitely happy to restore the ability to skip stat().
It's just that all existing callers in the repo call stat() regardless (not sure, will re-check).
Obviously that doesn't apply to public-facing APIs so I might need to revisit

gix-dir/src/walk/readdir.rs
```diff
@@ -59,12 +59,24 @@ pub(super) fn recursive(
...
+                path: std::mem::take(current),
+                source: err,
+            })?;
+            match Kind::try_from(m.file_type()) {
+                Ok(kind) => kind,
+                Err(_err) => {
+                    continue;
```
krobelus, line 70: > I am surprised these seemingly sweeping changes are needed when trying to ignored 'special' files

Yeah they're not, the only change that's needed is this "continue", which should probably be its own commit.

But I believe other places can be simplified and made more consistent. I may be wrong, I just need to find time to finish it

Byron: One will definitely need tests for that as well, and in the crates you are working in, they are basically done in a test-driven fashion exclusively as the code is way to complex to understand exactly what's going to happen.
But with the re-organization of commits I think all this will come quite naturally.

krobelus:
Sorry I got busy with other stuff, I'm not sure when I'll get around to this.
Probably we can make a minimal fix and handle the rest later.

Byron:
No worries- do what you can, and I will pick it up asap as well.

Byron:
Thanks for getting this started!

Now I have to close this PR in favor of #1727 which implements it in a more minimal fashion and with tests-first. That way we can be far more certain that nothing else breaks.

A new release with the fix should be out on the 22nd of December, and I will assure Cargo also gets the update then.

PR #1633, opened 2024-10-16
Reword security contact description for clarity

### LINKED ISSUE #1632: GitoxideLabs-related updates other than URLs in this repo
This repository was moved from `Byron/gitoxide` to `GitoxideLabs/gitoxide` when the GitoxideLabs organization was created, as described in #1406. Not all references have to be updated immediately and some never need to be updated, because `Byron/gitoxide` URLs redirect to the new URL, which I think will remain in place for a very long time (ideally, for at least as long as `GitoxideLabs` URLs themselves work). Nonetheless there is a benefit to referring people to the canonical URL.

#1624 makes these changes where applicable *inside* files in this repository, which also indirectly takes care of updating them in a number of other places; for example, information on crates.io will have the new repository URLs for subsequent crate releases. But there are a few more things that may benefit from being updated. I propose:

- [ ] Update summary metadata in [Codeberg](https://codeberg.org/Byron/gitoxide) and [gitee](https://gitee.com/byronbates/gitoxide) mirrors
- [ ] Maybe resolve newly possible future ambiguity in `SECURITY.md`
- [ ] Figure out under what conditions references in advisories should be updated
- [x] Don't update things where's there's clearly no benefit (e.g. links in issues)

#### Summary metadata in mirrors

The [Codeberg](https://codeberg.org/Byron/gitoxide) mirror has the summary/about text:

> A mirror of https://github.com/Byron/gitoxide

The [gitee](https://gitee.com/byronbates/gitoxide) mirror has the summary/about text:

> An idiomatic, lean, fast & safe pure Rust implementation of Git

I suggest updating them both to this, or something much like it:

> An idiomatic, lean, fast & safe pure Rust implementation of Git - mirror of https://github.com/GitoxideLabs/gitoxide

#### Ambiguity in `SECURITY.md`

This is unrelated to hyperlinks, arising differently from moving the repository into the newly created GitoxdeLabs org. The second paragraph has been unambiguous, even when read by someone with very minimal familiarity with the project:

> If this doesn't seem like the right approach or there are questions, please feel free to reach out to the @icloud.com email used in my commits.

In my opinion, this is a very useful thing to have. I furthermore recognize the benefit of having one fewer heavily indexed place with that email address, as well as the benefit of its indirectness to foregrounding the primary approach of [making a draft advisory](https://github.com/GitoxideLabs/gitoxide/security/advisories/new) through the GitHub interface.

However, now that the repository is under an org rather than a specific user (you), people with very low familiarity with the project--which could happen if someone unfamiliar with this project discovers a vulnerability in gitoxide while using an application that uses it--may not immediately know who "my" is.

This is a very low risk, especially right now--right now, the risk is effectively zero, because there are no other actors with @icloud.com email addresses associated with any commits in the history of gitoxide's main branch.

#### References in advisories

There are three kinds of relevant advisories: repository-level GHSA advisories, global GHSA advisories (in the GitHub Advisory Database), and RUSTSEC advisories.

It seems to me that, when already editing an advisory for another reason, it would be a good idea to update the links, and otherwise they need not be updated. Exceptions to this could be made on an ad-hoc basis motivated by need or convenience.

I don't think we need an actual *policy* about this, but the above suggestion is not necessarily the only reasonable approach, and I will tend to lean toward following it unless I know I shouldn't.

EliahKagan:
Closes #1632

This doesn't change the contact, just rewords it to avoid possible future ambiguity, now that the repository is in an org.

This is one of several possible wording changes that I think could be done in accordance with https://github.com/GitoxideLabs/gitoxide/issues/1632#issuecomment-2415818357. Another possibility could be to keep all the previous wording and just replace "my". I'd be pleased to make adjustments to this PR and also I of course do not mind if you make them.

It's just occurred to me that I don't know if changes apply only to `SECURITY.md` should have a `doc:` prefix in the commit message title. I think they don't need one but I'm not sure.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -5,6 +5,6 @@
 Please feel free to [draft a GitHub advisory](https://github.com/GitoxideLabs/gitoxide/security/advisories/new), and I will work with you to disclose and or resolve the issue
 responsibly.
 
-If this doesn't seem like the right approach or there are questions, please feel free to reach out to the @icloud.com email used in my commits.
+If this doesn't seem like the right approach or there are questions, please feel free to reach out to the email used in Sebastian Thiel's commits.
 
 Thank you.


Byron:
Thanks a lot, this works perfectly!

PR #1635, opened 2024-10-19
fix CI

Byron:
CI images now us Git 2.47, which causes one assertion to fail.

CC @EliahKagan - it's happening :).

### DIFF
--- a/gix/tests/gix/revision/spec/from_bytes/regex.rs
+++ b/gix/tests/gix/revision/spec/from_bytes/regex.rs
@@ -95,7 +95,7 @@ mod find_youngest_matching_commit {
     fn regex_matches() {
         let repo = repo("complex_graph").unwrap();
 
-        // On the CI linux workflow, archived baseline aren't used - instead fixtures will be re-evaluated.
+        // On the full linux CI `test` workflow, archived baseline aren't used - instead fixtures will be re-evaluated.
         // As of Git 2.47, its behaviour changed which makes the following assertion fail.
         // We decided to just ignore it until it's clear that this isn't a bug - obviously the traversal order changed.
         let is_in_test_ci_workflow = is_ci::cached() && std::env::var_os("GIX_TEST_IGNORE_ARCHIVES").is_some();
--- a/gix/tests/gix/revision/spec/from_bytes/regex.rs
+++ b/gix/tests/gix/revision/spec/from_bytes/regex.rs
@@ -95,20 +95,38 @@ mod find_youngest_matching_commit {
     fn regex_matches() {
         let repo = repo("complex_graph").unwrap();
 
-        assert_eq!(
-            parse_spec(":/mes.age", &repo).unwrap(),
-            Spec::from_id(hex_to_id("ef80b4b77b167f326351c93284dc0eb00dd54ff4").attach(&repo))
-        );
+        // On the CI linux workflow, archived baseline aren't used - instead fixtures will be re-evaluated.
+        // As of Git 2.47, its behaviour changed which makes the following assertion fail.
+        // We decided to just ignore it until it's clear that this isn't a bug - obviously the traversal order changed.
+        let is_in_test_ci_workflow = is_ci::cached() && std::env::var_os("GIX_TEST_IGNORE_ARCHIVES").is_some();
+        if is_in_test_ci_workflow {
+            assert_eq!(
+                parse_spec_no_baseline(":/mes.age", &repo).unwrap(),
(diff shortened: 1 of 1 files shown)


EliahKagan (commented):
Yes, the problem this works around is #1622. But I think there are actually two assertions that fail. I was about to open a PR with a18cc22ab5d08a2fcb01e82697edc9ac247a8892, but this PR came first. The commit message and diff there may shed further light on exactly what is affected.

I think the approach here of limiting the skipping to CI is better than what I have in a18cc22ab5d08a2fcb01e82697edc9ac247a8892. But I think the approach in a18cc22ab5d08a2fcb01e82697edc9ac247a8892 of changing `parse_spec` to `parse_spec_no_baseline` is better than the approach of skipping the full assertions. As detailed in #1622, it is only `parse_spec` that fails, and in it, what fails is only the baseline check. In #1622 I had suggested an overcomplicated approach to doing it without the baseline check; actually, replacing `parse_spec` with `parse_spec_no_baseline` achieves this.

Byron:
Thanks a lot for chiming in!
I actually prefer your solution and now we have a mix of both which is more complex, but at least tests both cases still where possible.

gix/tests/gix/revision/spec/from_bytes/regex.rs
```diff
@@ -95,20 +95,38 @@ mod find_youngest_matching_commit {
...
         let repo = repo("complex_graph").unwrap();
 
-        assert_eq!(
-            parse_spec(":/mes.age", &repo).unwrap(),
-            Spec::from_id(hex_to_id("ef80b4b77b167f326351c93284dc0eb00dd54ff4").attach(&repo))
-        );
+        // On the CI linux workflow, archived baseline aren't used - instead fixtures will be re-evaluated.
+        // As of Git 2.47, its behaviour changed which makes the following assertion fail.
+        // We decided to just ignore it until it's clear that this isn't a bug - obviously the traversal order changed.
+        let is_in_test_ci_workflow = is_ci::cached() && cfg!(target_os = "linux");
```
EliahKagan, line 101: If this were to also require that `GIX_TEST_IGNORE_ARCHIVES` be set (or, if that is judged too complex, then to replace the `linux` check with such a check), then `parse_spec` would continue to be used in the CI `test_fast` jobs that don't need it because their baselines were computed with an earlier versions of Git and agree with the current gitoxide behavior, while `parse_spec_no_baseline` would be used in the CI `test` job that needs it.

EliahKagan (commented):
This looks great, and definitely includes the good parts of a18cc22ab5d08a2fcb01e82697edc9ac247a8892.

As for the complexity, in terms of reading ease it should be possible to avoid the *duplication* by conditionally assigning either `parse_spec` or `parse_spec_no_baseline` to a variable. But I think I slightly prefer the way you have done it, since I expect that assertion messages will be clearer this way where it's more explicit, in the assertions, what is being done.

I have one thought about a way to make the change even narrower while still working, commented below.

gix/tests/gix/revision/spec/from_bytes/regex.rs
Byron: Thank you, that's a much better selector actually.
This PR is a good example for quick & sloppy, but thanks to you it's not going to be merged that  way :).

gix/tests/gix/revision/spec/from_bytes/regex.rs
```diff
@@ -95,20 +95,38 @@ mod find_youngest_matching_commit {
     fn regex_matches() {
         let repo = repo("complex_graph").unwrap();
 
-        assert_eq!(
-            parse_spec(":/mes.age", &repo).unwrap(),
-            Spec::from_id(hex_to_id("ef80b4b77b167f326351c93284dc0eb00dd54ff4").attach(&repo))
-        );
+        // On the CI linux workflow, archived baseline aren't used - instead fixtures will be re-evaluated.
```
EliahKagan, line 98: I just noticed that this wording is slightly misleading because the `test-fast` workflow uses archived baselines on all three tested platforms including Linux--this had at one time not been the case, but as far as I know it has been the case ever since the committed archives were moved out of LFS--while really the most relevant distinction is between the `test` and `test-fast` workflows, where the `test` workflow re-evaluates fixtures.

```suggestion
        // On the full linux CI `test` workflow, archived baseline aren't used - instead fixtures will be re-evaluated.
```

I'm not sure what the best wording is, but maybe that slightly improves clarity.

EliahKagan (approved):
Looks great!

PR #1638, opened 2024-10-21
Allow construct CustomFormat outside the crate

tisonkun:
@Byron One question is that shall we use `Cow<'static, str>` instead to allow define a non-static format?

It breaks the `Copy` trait somehow so I don't make this change. And a static format is sufficient for my use cases.

### DIFF
--- a/gix-date/src/time/mod.rs
+++ b/gix-date/src/time/mod.rs
@@ -33,6 +33,13 @@ pub enum Format {
 #[derive(Clone, Copy, Debug)]
 pub struct CustomFormat(pub(crate) &'static str);
 
+impl CustomFormat {
+    /// Create a new custom format.
+    pub const fn new(format: &'static str) -> Self {
+        Self(format)
+    }
+}
+
 impl From<CustomFormat> for Format {
     fn from(custom_format: CustomFormat) -> Format {
         Format::Custom(custom_format)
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot!

Note that I adjusted the commit-message so this becomes a feature to be added to the changelog.

> One question is that shall we use `Cow<'static, str>` instead to allow define a non-static format?

It's probably something for another time once there is demand. That would require dynamic formats to be used, which probably is quite rare.

tisonkun:
@Byron Thank you!

I checked the logs and didn't find a consistent pattern on messages: Some follow the conventional commits flavor and others follow a capital message.

From your comment it seems conventional commits should be used consistently, if so I'll apply it for the potential following contributions :D

Byron:
> From your comment it seems conventional commits should be used consistently, if so I'll apply it for the potential following contributions :D

Actually that's not the case. They are only used if something should go into the changelog, and to signal breaking changes as well. That's why it's not always used, just sometimes.

tisonkun:
> just sometimes.

OK.

Besides, I'm using this crate with `gix = { ... }`. Is there an estimated date when I can expect to get a new version with this change? Or, how the gitoxide's crates typically get released?

Byron:
Usually on the 22nd of each month there is a new release.

tisonkun:
Looks like I'm lucky 😆

PR #1639, opened 2024-10-21
Respect env variables when running gix

cruessler:
While working on `gix blame` I found that, for testing purposes, I wanted to run `env GIT_WORK_TREE=… gix blame` which did not work as expected. This PR changes that. I don’t know whether it was intentional not to take env variables into account. If it was, feel free to just close this PR!


Byron:
Thanks a lot, it's appreciated!

It doesn't seem to be working though and I think I know why. In theory, it should already be working anyway, but… let me try to make this work, I will push on top (without force).

cruessler:
Fascinating! This looked like such an innocuous change! :-)

PR #1645, opened 2024-10-24
feat: handle ReflogLookup::Date

dvtkrlbs:
Handles the missing Date based reflog lookup. The target date is compared against the signature time field. This is my first contribution and was kinda a blind one so open to some guidance (especially on testing this addition)


dvtkrlbs:
also just realized my min_by logic is wrong since it only deal with values smaller than the target date I think

dvtkrlbs:
Hmm I see there is an implementation for to_time in gix_date however this is not public should I just implement Sub for Time where the Output is i64?

gix-date/src/lib.rs
```diff
@@ -29,6 +32,24 @@ pub struct Time {
...
+impl Sub for Time {
+    type Output = i64;
+
+    fn sub(self, rhs: Self) -> Self::Output {
+        self.total() - rhs.total()
+    }
+}
+
+impl Time {
+    fn total(&self) -> i64 {
```
Byron, line 44: This wants to be `local_time()` really, as it's the amount of seconds one would see with the offset applied.

gix-date/src/lib.rs
```diff
@@ -29,6 +32,24 @@ pub struct Time {
...
+        if self.sign == Sign::Minus {
```
Byron, line 45: The sign is just for serialization, what counts here is if `seconds` itself is negative.

gix/src/revision/spec/parse/delegate/revision.rs
```diff
@@ -106,12 +106,51 @@ impl delegate::Revision for Delegate<'_> {
...
+                        .min_by_key(|l| (date - l.signature.time).abs())
+                    {
+                        Some(closest_line) => {
+                            self.objs[self.idx]
+                                .get_or_insert_with(HashSet::default)
+                                .insert(closest_line.new_oid);
+                            Some(())
+                        }
+                        None => {
+                            // do we need an another error variant?
```
Byron, line 141: Yes, good call!

gix-date/src/lib.rs
```diff
@@ -29,6 +32,24 @@ pub struct Time {
     pub sign: time::Sign,
 }
 
+impl Sub for Time {
```
Byron, line 35: Impls want to be in the `time` module.

gix/tests/gix/revision/spec/from_bytes/reflog.rs
```diff
@@ -78,10 +75,13 @@ fn by_index() {
...
+fn by_date() {
     let repo = repo("complex_graph").unwrap();
-    assert!(matches!(
-        parse_spec_no_baseline("main@{1979-02-26 18:30:00}", &repo).unwrap_err(),
-        Error::Planned { .. }
-    ));
+
+    let spec = parse_spec_no_baseline("main@{1979-02-26 18:30:00}", &repo).unwrap();
+
+    assert_eq!(
```
Byron, line 83: Now there can easily be more tests which are then compared to what Git does - here this can be done by hand, but it's easy to also write a baseline tests for this where Git itself provides the expectations.

I would only want to test the that the branch-handling is correct, and the time-computation itself.

Byron (commented):
Thanks a lot, this improvement - this probably means that the `Planned` variant can be removed as well, nothing else uses it.

Further, please have one commit per crate, with `gix-date` getting its own `feat: …` commit.

Thanks again.

gix/tests/gix/revision/spec/from_bytes/reflog.rs
dvtkrlbs: similar to the other test where check the first_refence name is "refs/heads/main" right? For git based testing I would like some pointers on how to do that.

gix-date/src/lib.rs
dvtkrlbs: this implies also needs to be in the time module right?

Byron: Yes.

gix/src/revision/spec/parse/delegate/revision.rs
```diff
@@ -106,12 +106,51 @@ impl delegate::Revision for Delegate<'_> {
...
+                        Ok(None) => {
+                            self.err.push(Error::UnbornHeadsHaveNoRefLog);
+                            return None;
+                        }
+                        Err(err) => {
+                            self.err.push(err.into());
+                            return None;
+                        }
+                    },
+                };
```
dvtkrlbs, line 126: question since this is duplicated with the other part should I create a closure. This duplication is probably fine but it kinda bothers me.

Byron: Refactor at will, but please do so in a separate commit.

gix/tests/gix/revision/spec/from_bytes/reflog.rs
dvtkrlbs: i looked at the code a bit but I understand the how the complex_graph baseline works (I am on windows so I can't just run the script locally) to see how many entries it has and do exact time test etc (especially how the dates for test entries are created). We also need a baseline with no entries and I have no idea on how to do that.

Byron: Let's keep it with hand-written tests - these are enough to get this right I am sure. Baselines are usually for very complex behaviours that should be mimicked, but I don't think that's the case here.

dvtkrlbs: so do I just create an history using the gitoxide api. I was trying to to mimick the other test https://github.com/dvtkrlbs/gitoxide/blob/6b2643530e04bccfc741777a4b84d645d1537574/gix/tests/gix/revision/spec/from_bytes/reflog.rs#L41 but I don't don't understand the repo creation part.

Byron: No worries - if you think this could work, and you have verified it manually with `cargo run --bin gix -- rev parse`  a little, then I think this PR can be ready for review and I take it from there.

dvtkrlbs: I will do more manual testing and mark it as ready for review once I am confident about the changes. Thank you

Byron:
Is this ready for review?

dvtkrlbs:
hey sorry was not able to test it properly since I am not actually knowledgable about the internals of git that much. I would appreaciate if someone would test it since they would do it better than me.

Byron:
Thanks for letting me know - I will take care of it, hopefully so this PR is merged before next year.

PR #1657, opened 2024-11-07
Run CI on Windows with `GIX_TEST_IGNORE_ARCHIVES=1`

EliahKagan:
This adds a CI test job, `test-fixtures-windows`, that runs tests on Windows with `GIX_TEST_IGNORE_ARCHIVES=1`, allows the job to succeed even when the test running step fails, but then, if it failed (as currently it always does), runs another step that checks that no tests reported *errors* and that the number of tests reporting *failures* does not exceed 14.

Although there are currently 15 tests that regularly fail when run this way, as listed in #1358, the performance test does currently seem ever to fail *on CI*. Setting it to 14 carries a risk of failure even when nothing new is broken, but should shed light on whether it does ever fail on CI. (If not, then it might be that my test machine is inadequate, rather than an actual performance bug.) This also makes it less likely to miss a new failure, per https://github.com/GitoxideLabs/gitoxide/issues/1358#issuecomment-2457946710 and https://github.com/GitoxideLabs/gitoxide/issues/1358#issuecomment-2459710189.

The approach taken is to add a `with-xml` nextest profile and use it for the `cargo nextest` step of the new job, then have the subsequent step use PowerShell to parse the generated XML as discussed in https://github.com/GitoxideLabs/gitoxide/issues/1358#issuecomment-2459072454 and https://github.com/GitoxideLabs/gitoxide/issues/1358#issuecomment-2459710189.

This feature branch includes the intermediate approaches rather than squashing them away. I think the details about their effects and disadvantages may be of value in the future. There are also failures that only occur with the other approach I attempted first, presumably relating either to the differing environments of PowerShell and Git Bash or to pecularities arising from `|&` redirection. I'll look into this in more detail sometime fairly soon and I'm not requesting that you examine it, but full details are in the commit messages if you wish to.

The main commit message of interest is 145ae49 (before the rebase, a5d3698), which describes the specific advantages of the XML approach over the ad-hoc parsing approach.

This may not be immediately ready, because:

- [ ] The new job is deliberately not a required job. But we should make sure it passes in *this* PR before the PR is merged (even though it has already been run in my fork), since the purpose of this PR is to introduce the job and demonstrate that it passes under existing conditions. (In particular, I just rebased this branch.)
- [ ] Relating to [#1358 (comment)](https://github.com/GitoxideLabs/gitoxide/issues/1358#issuecomment-2459710189), it may make sense to make the new job depend on an existing job, but as discussed there I don't know what the best way to do this is. Right now, no such dependency is expressed.
- [ ] This job seems to take the same amount of time as the `test-fast` job for `windows-latest`, when tested on my fork. Here, it is slower, but I think that is related to caching and would go away once merged. The new job omits some of what is done in the `test-fast` job--for example, it does not run doctests--and that seems to roughly cancel out the added time to run the fixture scripts. In view of this, maybe it should be made a required check? On the other hand, that could slow some things down if we want to allow PRs to be quickly merged even if they introduce new failures on Windows when `GIX_TEST_IGNORE_ARCHIVES=1`.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,2 @@
+[profile.with-xml.junit]
+path = "junit.xml"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,6 +80,37 @@ jobs:
       - name: Check that tracked archives are up to date
         run: git diff --exit-code  # If this fails, the fix is usually to commit a regenerated archive.
 
+  test-fixtures-windows:
+    runs-on: windows-latest
+    env:
+      # FIXME: Change to 15 after verifying that the last step can fail.
+      # See https://github.com/GitoxideLabs/gitoxide/issues/1358.
+      EXPECTED_FAILURE_COUNT: 13
+    defaults:
+      run:
+        shell: bash  # Includes `-o pipefail`.
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: Swatinem/rust-cache@v2
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: nextest
+      - name: "Test (nextest)"
+        env:
+          GIX_TEST_IGNORE_ARCHIVES: 1
+        run: |
+          cargo nextest --color=always run --workspace --no-fail-fast |& tee nextest.log
+        continue-on-error: true
+      - name: Allow up to ${{ env.EXPECTED_FAILURE_COUNT }} failures
+        if: failure()
+        run: |
+          set -x
+          sed -Ei 's/\x1B\[[[:digit:];]*m//g' nextest.log  # Remove ANSI color codes.
+          pattern='\n-{10,}\r?\n[ \t]+Summary\b[^\n]+[ \t]\K\d+(?= failed\b)'
+          count="$(grep -zoP "$pattern" nextest.log)"
+          ((count <= EXPECTED_FAILURE_COUNT))
+
   test-32bit:
     runs-on: ubuntu-latest
     strategy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,7 +103,7 @@ jobs:
       - name: Check how many tests failed
         if: steps.nextest.outcome == 'failure'
         env:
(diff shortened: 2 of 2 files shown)


Byron:
Thanks so much for setting this up!

> [ ] This job seems to take the same amount of time as the `test-fast` job for `windows-latest`, when tested on my fork. Here, it is slower, but I think that is related to caching and would go away once merged. The new job omits some of what is done in the `test-fast` job--for example, it does not run doctests--and that seems to roughly cancel out the added time to run the fixture scripts. In view of this, maybe it should be made a required check? On the other hand, that could slow some things down if we want to allow PRs to be quickly merged even if they introduce new failures on Windows when `GIX_TEST_IGNORE_ARCHIVES=1`.

The timing is indeed very promising as it could fit into the envelope we already have once caching kicks in. Then again, caching, or the lack thereof, should be the same for each job in a new PR, so I'd still think that the new job is 3-4 minutes slower than the longest-running job thus far. [Here is the summary](https://github.com/GitoxideLabs/gitoxide/actions/runs/11719387357/usage?pr=1657) that I base this on.

Maybe I am wrong about this, so I think it's OK to merge this PR as is and possibly, in a follow up, make it a required job like any other.

I'd also think that it sends an email if the job fails, even though the workflow didn't fail, so it won't go unnoticed once I inevitably break something.

EliahKagan:
> I'd also think that it sends an email if the job fails, even though the workflow didn't fail

I think that workflow is considered to fail when this job fails, as with the other jobs in the workflow. The workflow status showed as failing in this situation when I tested it on my fork. (Some of this can be seen before the rebase; I did the rebase just after opening this PR instead of just before, in case it would be valuable to ever look at that.)

The only way this job is distinguished from most others in the workflow is that it is not listed as a dependency of a required check that attests that all conceptually required checks have completed. But if I understand correctly this is separate from the status of the workflow run: if all the required jobs complete successfully, then the PR is eligible to be merged when auto-merge is enabled, regardless of the status of the workflow run some of whose jobs were required checks.

Byron:
Thanks for the clarification! It's true that it is failing the workflow if the dependent check fails, and this will mean an email will be sent. It will still auto-merge earlier as it's not part of the check, which is exactly what it should do.
I think I got it now.

PR #1662, opened 2024-11-08
upgrade thiserror to v2.0.0

paolobarbolini:
This upgrades every crate to thiserror v2.0.0 [^1], trying to reduce the number of crates that cause duplicate dependency versions. thiserror v2 also supports `core::error::Error`, in case you may need it in some of the gix crates :-)

[^1]: https://github.com/dtolnay/thiserror/releases/tag/2.0.0


Byron:
That's awesome, thanks so much!

Byron (approved):

PR #1663, opened 2024-11-09
Check precise Windows `GIX_TEST_IGNORE_ARCHIVES` expectations on CI

EliahKagan:
This modifies the CI `test-fixtures-windows` job introduced in #1657, by making it so that, instead of checking that no more than the expected number of test cases fail, it checks that the *exact* test cases expected to fail are the ones that do fail. (The other check done before as part of this job, that no tests report *errors*, is still done.) The approach taken is:

1. **To collect *actual* failures**, parse the XML output saved by `cargo nextest` to get the package name and test name of each failing test and save them, one per line, to a file.

   Getting this information is pretty straightforward in PowerShell. When iterating through `testcase` nodes, which are filtered by the presence of a `failure` subnode, `$_.name` gives the name of the test case, including `::` qualifications within the containing package, and `$_.classname` gives the name of the containing package. Thus `"$($_.classname) $($_.name)"` identifies a test in the same way it is identified in human-readable `cargo nextest` output.

   The unintuitive attribute name `classname` is because `cargo nextest` XML is in the JUnit format, which originated for Java tests where test cases are methods in classes. An alternative approach is to use `$_.ParentNode.name` instead of `$_.classname`. I've verified that this works too, but I don't know of any strong reason to prefer one over another, so I went with the more compact `$_.classname`.

2. **To collect *expected* failures**, use `gh` (the GitHub CLI) to download the text of [#1358](https://github.com/GitoxideLabs/gitoxide/pull/1358), and parse the first ```` ```text ```` code block with regex, extracting the list of failing tests and saving them, one per line, to a file.

   One of the tests currently listed there, `gix-ref-tests::refs packed::iter::performance`, is a performance test that seems not to fail on CI (anymore?), though I still find it to fail locally. If it were to start failing on CI, we would want to know about it. So, before these changes, it was not included in the count. I've carried that over to the precise matching done here: tests that appear to be performance tests due to having `performance` in their names (not as part of any longer `\w+` word) are not regarded as expected to fail.

   It is not obvious that the general approach of downloading the information from an issue is the best one. Although I like this approach and I think it's not too cumbersome, this is admittedly the less elegant step. Other approaches, including hard-coding the expected failures in the workflow or another file, could be used if this approach is not wanted.

3. **To compare them**, use `git diff --no-index` with various other options, printing the entire list of failures as context if they differ. The job fails if there are any differences.

   Because this neither a required check for PR auto-merge nor a dependency of one, I think failing not only if tests unexpectedly fail but also if any unexpectedly pass is unlikely to cause any significant problems, and that knowing whenever the set of failing tests changes in any way is worthwhile.

I intentionally excluded performance tests late in the process, so that there would be a commit whose CI results could be inspected that would verify that the check is capable of failing when the diff is nonempty (and that the output would display in a useful way, with effective colorization, etc., in this situation). That can be observed in [this run output](https://github.com/EliahKagan/gitoxide/actions/runs/11749877635/job/32736979905) and compared to passing runs, such as the runs in this PR.


.github/workflows/ci.yml
```diff
@@ -89,21 +89,49 @@ jobs:
...
+          Write-Output $expected_failures
+          Set-Content -Path 'expected-failures.txt' -Value $expected_failures
+      - name: Compare expected and actual failures
         run: |
-          [xml]$junit = Get-Content -Path 'target/nextest/with-xml/junit.xml'
-          if ($junit.testsuites.errors -ne 0) { exit 1 }
-          if ($junit.testsuites.failures -gt $env:EXPECTED_FAILURE_COUNT) { exit 1 }
+          # Fail on any differences, even unexpectedly passing tests, so they can be investigated.
+          # (If the job is made blocking for PRs, it may make sense to make this less stringent.)
+          git --no-pager diff --no-index --exit-code --unified=1000000 --color=always -- `
```
Byron, line 133: Using git for a general diff between two files is very interesting, I didn't even know that's possible.

.github/workflows/ci.yml
```diff
@@ -89,21 +89,49 @@ jobs:
...
+
+          Write-Output $actual_failures
+          Set-Content -Path 'actual-failures.txt' -Value $actual_failures
+      - name: Collect expected failures
         env:
-          # See https://github.com/GitoxideLabs/gitoxide/issues/1358.
-          EXPECTED_FAILURE_COUNT: 14
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          $issue = 1358  # https://github.com/GitoxideLabs/gitoxide/issues/1358
```
Byron, line 116: This approach, as interesting as it may seem, breaks isolation of the test a bit too much for my taste and I'd prefer hardcoding it. The issue could then refer to the file to display the lines here, or link to the separate file that is read by the workflow.

Byron (changes requested):

I am very impressed by what can be done with powershell, it clearly is `.NET` in a shell :).

And indeed, like predicted, reading the text block from the issue breaks isolation too much and is the reason I'd love to see these extracted into a file, for example if that's OK with you.

Thanks again.

.github/workflows/ci.yml
EliahKagan: Sounds good. I've pushed [99238a7](https://github.com/GitoxideLabs/gitoxide/pull/1663/commits/99238a72894595e95d8c03df51fda02285447589), which adds a file with the list and has the job use that, removing altogether the retrieval step.

A subtlety here is that the list of expected failures is no longer being sorted. I can do that if desired, but it seemed like it would be preserving unnecessary complexity. Even if for some strange reason `cargo nextest` when run in manual testing were to output results not in this order, this is a file being committed to the repository, so the entries can be ordered lexicographically when committing changes to them.

(The list of actual failures is still sorted, so there is no doubt that the CI test run itself is producing output that is ready for comparison.)

I'm not sure what the file should be called, but this gigantic name seemed reasonable to me, since it makes everything clear. What I'm less sure of is it's location, but the existing `etc` directory seemed like it might be the best place.

.github/workflows/ci.yml
EliahKagan: `diff` in PowerShell is an alias for the `Compare-Object` cmdlet. Of course, the `diff` program on Windows can still be run as `diff.exe`, though I am uncertain if that will find the one installed as part of the Git Bash environment, or another one, when run from PowerShell on GHA runners.

More broadly, over time I've increasingly come to like `git diff`, since it seems to work more similarly across systems than `diff`, which is not a single program but any of a number of implementations. `git diff` is also feature rich and has the benefit of being configurable with Git configuration variables.

This can be made more compact, if desired, by omitting `--exit-code`, which is implied by `--no-index`, omitting `--no-pager` since the inability to page interactively will be detected, by expressing `--color=always` as just `--color`, and possibly even by omitting `--no-index`, since a diff between a tracked and an untracked file works fine without it (though I would want to check if that is still guaranteed to imply `--exit-code`, I am not sure). However, all of these seem to be making clearer how the command is supposed to work.

Byron (approved):
Wonderful, thanks so much!

PR #1668, opened 2024-11-11
Improve CI `permissions`, auto-merge maintainability, and clarity

EliahKagan:
This makes several changes to GitHub Actions workflows, which would create merge conflicts if submitted separately, and which even aside from that seemed, to varying degrees, like they would make more sense to do together than separately. But the commits are divided in such a way that breaking this up into a sequence of a few separate PRs would be easy to do, if preferred. There is also an associated recommendation for a change in the repository's settings.

The changes in this PR are as follows:

#### Use read-only `github.token` everywhere feasible

This PR sets `permissions:` explicitly in each workflow and, where different jobs don't all need the same permissions, in individual jobs as well. The permissions are only what each job needs and, in particular, `write` permissions are given only to the jobs that really require it (the jobs that create or modify releases in the release workflow).

This change is not needed to safeguard against threats from malicious pull requests, and has no effect on such risks. Its primary significance to security is that it slightly decreases the risk from supply-chain attacks. Many GitHub Actions jobs run code from numerous transitive dependencies and other sources. (I say "slightly" because it does not affect the risk when running commands locally, when creating releases, or when using releases.) Secondarily, this may also decrease risks from unintentional vulnerabilities or other bugs in dependencies or in workflows themselves.

The reason this is unrelated to threats from malicious PRs is that, when PRs that originate from forks trigger the `pull_request` event, jobs in workflows triggered by that event already automatically get a read-only `github.token`. (This repository currently does not define any workflows that trigger on the `pull_request_target` or `workflow_run` events, which would have additional security implications.)

Specific details of this change are given in f41a58c.

I recommend also setting the *default* permissions of GitHub tokens in GitHub Actions jobs to read-only, by going to [Settings → Actions → General](https://github.com/GitoxideLabs/gitoxide/settings/actions) and, under *Workflow permissions*, changing it from "Read and write permissions" to "Read repository contents and package permissions". That way, if a job is ever introduced with no explicit `permissions` in its definition or the containing workflow, the token will be read-only rather than read-write. That can be done at any time. Doing so without making any other changes will break the release workflow until the workflow is modified to specify the necessary write permissions (a change that this PR includes).

#### Style workflows more consistently and slightly more spaciously

This applies the prevailing style everywhere, except that it puts blank lines between the top-level nodes of a job. (This does not change the spacing between steps, but rather between, for example, `steps` itself and other things at the same level.)

It also makes some other improvements, such as replacing the last `actions/checkout@master` with `actions/checkout@v4`, quoting `env` values when the absence of quotes makes them boolean or numeric in YAML (since they are still turned into strings to be placed in the environment, and the string representations are not always obvious), and removing an unnecessary `env` key of `CI`, which is guaranteed to be set to `true` by the runner.

Some of these improvements are directly motivated by the goal of preventing the addition of explicit `permissions` from making workflows and job definitions harder to read. Those are in 7eeeee2. In view of those, it seemed like a good idea to include the others, which otherwise would conflict with this PR or might conflict with other subsequent PRs, and which are conceptually along the same lines, but because they are not needed to allow `permissions` to be added without confusion, they are done in a separate commit, 44ff412.

#### Bump cargo-diet version and express the logic more cleanly

87670a6 bumps cargo-diet to the latest stable version and makes some improvements to the steps, including splitting it into separate installation and running steps and making the running step `continue-on-error`, instead of writing `|| true` after the command that (through a `just` rule) runs `cargo diet`.

#### Provide richer log output for `tests-pass`

20794ff adapts a comment in `tests-pass` to be a step name, so everything is clear in logs even if the reader is unfamiliar with the workflow.

#### Detect new jobs that may be intended to block auto-merge

This only affects the `ci.yml` workflow, where the `tests-pass` job stands in for all the jobs that are meant to block PR auto-merge, by being set up as a required job and depending on the others (#1551). It intentionally does not depend on all jobs in the workflow: of course it does not depend on itself, and it also does not depend on `test-fixtures-windows` (#1657, #1663). Another change in this PR, described below, also causes it no longer to depend on `wasm`.

That existing situation unfortunately makes it easy to add a job that it doesn't depend on, by simply forgetting to add it or forgetting to make a decision on whether the new job should block PR auto-merge.

The solution is to maintain two lists of jobs that are checked to ensure are disjoint and form, as their union, all jobs defined in `ci.yml`. A job to check this is added in 4e672ef, 7b7a819, and eba50ae.

#### Make the `wasm` jobs fail the workflow but not block auto-merge

Prior to the changes in this PR, the WebAssembly jobs would report a successful conclusion even if they failed, by using the `continue-on-error` key at the job level. This was needed to keep them from blocking auto-merge, but only because `wasm` was listed as a dependency of `tests-pass`. This changes it so that it will report a failing conclusion (and thus fail the workflow) but not prevent PR auto-merge from working.

Since these jobs are expected to pass--if they fail then it indicates a regression in WASM compatibility, and they rarely fail--this seems like a more suitable approach. See 253b716e for full details.

More broadly, although I do *not* propose adopting any rule about it, I think we should be reluctant to use `continue-on-error` at the job level in `ci.yml`, ever since branch protection rules and required checks for auto-merge have been set up and (since #1551) made partially configurable in the workflow file:

- The granularity of having jobs that unambiguously fail but do not block automatic merging seems like it should usually capture the situation where the failure represents something important but that nonetheless would not be out of the ordinary to dismiss.
- Other than to represent something that is temporarily broken but intended to be fixed immediately, it would be strange to have a job triggered by `push` or `pull_request` whose failure doesn't represent something important, since such a job should usually either not exist or not run automatically.
- This reasoning only applies to `continue-on-error` at the job level and not, for example, in a specific step of a job that has at least one step not marked `continue-on-error`.

But I do not know of any way for one job to depend on specific matrix jobs while not depending on others. This is inapplicable to the `wasm` job definition because, although it uses a matrix strategy, the current intent is that neither job generated from it block PRs. But it does apply to the `cargo deny` job definition, whose `advisories` job is marked `continue-on-error`, while its `bans licenses sources` job is not.

This is a simple job definition, which could be broken up into two non-matrix job definitions with minimal code duplication. It seems to me that this should be done, so that a new advisory causes the job to have a failing conclusion and fail the workflow even as it does not block PRs. I had originally planned to include that in this PR, but I decided not to do so, since the question of whether to do shoud probably be considered separately. There are some interesting subtleties, such as in the effect on when commands like `@dependabot merge` will go ahead with a merge (relevant because Dependabot security updates are created for all ecosystems, even though `github-actions` is the only ecosystem that gets Dependabot version updates).

That `advisories` job is currently failing--both here and if it is [re-run on `main`](https://github.com/EliahKagan/gitoxide/actions/runs/11763341412/job/32780079585)--due to [RUSTSEC-2024-0384](https://rustsec.org/advisories/RUSTSEC-2024-0384.html) (https://github.com/sebcrozet/instant/issues/52), announcing that `instant` is unmaintained and `web-time` should be used instead. I had initially hoped to fix that here, but it is not feasible to do so: we use `instant` only as a transitive dependency through other crates that are not part of this repository and whose dependencies should be updated to use `web-time`. That situation is unaffected by any changes in this PR.

#### Update other GitHub-specific YAML configuration

e45ea25 removes the old Dependabot configuration file for the `cargo` ecosystem, and adds a comment to the current Dependabot configuration file to explain why that ecosystem is not included. The explanation only links to #144, but accounts for the full current situation (see #143, #144, https://github.com/GitoxideLabs/gitoxide/pull/1254#issuecomment-1898089062, and #1357).

4564a64 removes the empty pull request template file, as discussed in #1667.

(These changes are done here rather than in a separate PR because I was already editing `dependabot.yml` for style, and also because, as detailed above, I had originally planned to include other changes that might affect Dependbot indirectly, but that are deferred so they can be reviewed separately.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,14 +180,17 @@ jobs:
         run: just doc
       - name: Run cargo fmt
         run: cargo fmt --all -- --check
-      - name: Run cargo diet
+      - name: Install cargo diet
+        env:
+          CARGO_DIET_TAG: v1.2.7
         run: |
-          curl -LSfs https://raw.githubusercontent.com/the-lean-crate/cargo-diet/master/ci/install.sh |
-            sh -s -- --git the-lean-crate/cargo-diet --target x86_64-unknown-linux-musl --tag v1.2.4
-
-          # Let's not fail CI for this, it will fail locally often enough, and a crate a little bigger
-          # than allows is no problem either if it comes to that.
-          just check-size || true
+          curl -LSfs "https://raw.githubusercontent.com/the-lean-crate/cargo-diet/refs/tags/$CARGO_DIET_TAG/ci/install.sh" |
+            sh -s -- --git the-lean-crate/cargo-diet --target x86_64-unknown-linux-musl --tag "$CARGO_DIET_TAG"
+      - name: Run cargo diet
+        run: just check-size
+        # Let's not fail CI for this, it will fail locally often enough, and a crate a little bigger
+        # than allows is no problem either if it comes to that.
+        continue-on-error: true
 
   cargo-deny:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -297,10 +297,9 @@ jobs:
 
     env:
       # List all jobs that are intended NOT to block PR auto-merge here.
-      # FIXME: After ensuring this can fail, add `test-pass` below to fix it.
       EXPECTED_NONBLOCKING_JOBS: |-
         test-fixtures-windows
-      #  tests-pass
+        tests-pass
 
     defaults:
       run:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -315,12 +315,12 @@ jobs:
         with:
           sparse-checkout: ${{ env.WORKFLOW_PATH }}
       - name: Get all jobs
-        run: yq '.jobs | keys.[]' -- "$WORKFLOW_PATH" | sort > all-jobs.txt
+        run: yq '.jobs | keys.[]' -- "$WORKFLOW_PATH" | sort | tee all-jobs.txt
       - name: Get blocking jobs
-        run: yq '.jobs.tests-pass.needs.[]' -- "$WORKFLOW_PATH" | sort > blocking-jobs.txt
+        run: yq '.jobs.tests-pass.needs.[]' -- "$WORKFLOW_PATH" | sort | tee blocking-jobs.txt
       - name: Get jobs we intend do not block
-        run: sort <<<"$EXPECTED_NONBLOCKING_JOBS" > expected-nonblocking-jobs.txt
-      - name: Check that each job blocks PRs xor intentionally doesn't
+        run: sort <<<"$EXPECTED_NONBLOCKING_JOBS" | tee expected-nonblocking-jobs.txt
+      - name: Each job must block PRs or be declared not to
         run: |
           sort -m blocking-jobs.txt expected-nonblocking-jobs.txt |
             diff --color=always -U1000 -- - all-jobs.txt
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -215,8 +215,6 @@ jobs:
 
     runs-on: ubuntu-latest
 
-    continue-on-error: true
-
     strategy:
       matrix:
         target: [ wasm32-unknown-unknown, wasm32-wasi ]
@@ -302,6 +300,7 @@ jobs:
       # List all jobs that are intended NOT to block PR auto-merge here.
       EXPECTED_NONBLOCKING_JOBS: |-
         test-fixtures-windows
+        wasm
         tests-pass
 
     defaults:
@@ -338,7 +337,6 @@ jobs:
       - test-32bit
       - lint
       - cargo-deny
-      - wasm
       - check-packetline
       - check-blocking
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,9 @@ on:
       - main
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 env:
   CARGO_TERM_COLOR: always
   CLICOLOR: 1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -289,6 +289,42 @@ jobs:
           git status
           git diff --exit-code
 
+  # Check that only jobs intended not to block PR auto-merge are omitted as
+  # dependencies of the `tests-pass` job below, so that whenever a job is
+  # added, a decision is made about whether it must pass for PRs to merge.
+  check-blocking:
+    runs-on: ubuntu-latest
+
+    env:
+      # List all jobs that are intended NOT to block PR auto-merge here.
+      # FIXME: After ensuring this can fail, add `test-pass` below to fix it.
+      EXPECTED_NONBLOCKING_JOBS: |-
+        test-fixtures-windows
+      #  tests-pass
+
+    defaults:
+      run:
+        shell: bash  # Without specifying this, we don't get `-o pipefail`.
+
+    steps:
+      - name: Find this workflow
+        run: |
+          relative_workflow_with_ref="${GITHUB_WORKFLOW_REF#"$GITHUB_REPOSITORY/"}"
+          echo "WORKFLOW_PATH=${relative_workflow_with_ref%@*}" >> "$GITHUB_ENV"
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: ${{ env.WORKFLOW_PATH }}
+      - name: Get all jobs
+        run: yq '.jobs | keys.[]' -- "$WORKFLOW_PATH" | sort > all-jobs.txt
+      - name: Get blocking jobs
+        run: yq '.jobs.tests-pass.needs.[]' -- "$WORKFLOW_PATH" | sort > blocking-jobs.txt
+      - name: Get jobs we intend do not block
+        run: sort <<<"$EXPECTED_NONBLOCKING_JOBS" > expected-nonblocking-jobs.txt
+      - name: Check that each job blocks PRs xor intentionally doesn't
+        run: |
+          sort -m blocking-jobs.txt expected-nonblocking-jobs.txt |
+            diff --color=always -U1000 -- - all-jobs.txt
+
   # Dummy job to have a stable name for the "all tests pass" requirement
   tests-pass:
     name: Tests pass
@@ -302,6 +338,7 @@ jobs:
       - cargo-deny
       - wasm
       - check-packetline
+      - check-blocking
 
     if: always() # always run even if dependencies fail
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -308,7 +308,8 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      # fail if ANY dependency has failed or cancelled
-      - if: "contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')"
+      - name: Fail if ANY dependency has failed or cancelled
+        if: "contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')"
         run: exit 1
-      - run: exit 0
+      - name: OK
+        run: exit 0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ permissions:
 
 env:
   CARGO_TERM_COLOR: always
-  CLICOLOR: 1
+  CLICOLOR: '1'
 
 jobs:
   pure-rust-build:
@@ -36,14 +36,14 @@ jobs:
         run: |
           set -x
           for pattern in cmake g++ libssl-dev make pkgconf pkg-config; do
-              if dpkg-query --status -- "$pattern"; then
-                  exit 1
-              fi
+            if dpkg-query --status -- "$pattern"; then
+              exit 1
+            fi
           done
           for cmd in cmake g++ make pkgconf pkg-config; do
-              if command -v -- "$cmd"; then
-                  exit 1
-              fi
+            if command -v -- "$cmd"; then
+              exit 1
+            fi
           done
       - name: Install Rust via Rustup
         run: curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --profile minimal
@@ -65,8 +65,7 @@ jobs:
           tool: nextest
       - name: test
         env:
-          CI: true
-          GIX_TEST_IGNORE_ARCHIVES: 1
+          GIX_TEST_IGNORE_ARCHIVES: '1'
         run: just ci-test
 
   test-fast:
@@ -95,7 +94,7 @@ jobs:
           tool: nextest
       - name: "Test (nextest)"
         env:
-          GIX_TEST_CREATE_ARCHIVES_EVEN_ON_CI: 1
+          GIX_TEST_CREATE_ARCHIVES_EVEN_ON_CI: '1'
         run: cargo nextest run --workspace --no-fail-fast
       - name: Doctest
         run: cargo test --workspace --doc --no-fail-fast
@@ -127,8 +126,8 @@ jobs:
           [xml]$junit_xml = Get-Content -Path 'target/nextest/with-xml/junit.xml'
 
           $actual_failures = $junit_xml.SelectNodes("//testcase[failure]") |
-              ForEach-Object { "$($_.classname) $($_.name)" } |
-              Sort-Object
+            ForEach-Object { "$($_.classname) $($_.name)" } |
+            Sort-Object
 
           Write-Output $actual_failures
           Set-Content -Path 'actual-failures.txt' -Value $actual_failures
@@ -137,7 +136,7 @@ jobs:
           # Fail on any differences, even unexpectedly passing tests, so they can be investigated.
           # (If the job is made blocking for PRs, it may make sense to make this less stringent.)
           git --no-pager diff --no-index --exit-code --unified=1000000 --color=always -- `
-              etc/test-fixtures-windows-expected-failures-see-issue-1358.txt actual-failures.txt
+            etc/test-fixtures-windows-expected-failures-see-issue-1358.txt actual-failures.txt
 
   test-32bit:
     runs-on: ubuntu-latest
@@ -183,8 +182,8 @@ jobs:
         run: cargo fmt --all -- --check
       - name: Run cargo diet
         run: |
-          curl -LSfs https://raw.githubusercontent.com/the-lean-crate/cargo-diet/master/ci/install.sh | \
-           sh -s -- --git the-lean-crate/cargo-diet --target x86_64-unknown-linux-musl --tag v1.2.4
+          curl -LSfs https://raw.githubusercontent.com/the-lean-crate/cargo-diet/master/ci/install.sh |
+            sh -s -- --git the-lean-crate/cargo-diet --target x86_64-unknown-linux-musl --tag v1.2.4
 
           # Let's not fail CI for this, it will fail locally often enough, and a crate a little bigger
           # than allows is no problem either if it comes to that.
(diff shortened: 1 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -249,22 +290,63 @@ jobs:
...
+          echo "WORKFLOW_PATH=${relative_workflow_with_ref%@*}" >> "$GITHUB_ENV"
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: ${{ env.WORKFLOW_PATH }}
+      - name: Get all jobs
+        run: yq '.jobs | keys.[]' -- "$WORKFLOW_PATH" | sort | tee all-jobs.txt
+      - name: Get blocking jobs
+        run: yq '.jobs.tests-pass.needs.[]' -- "$WORKFLOW_PATH" | sort | tee blocking-jobs.txt
+      - name: Get jobs we intend do not block
+        run: sort <<<"$EXPECTED_NONBLOCKING_JOBS" | tee expected-nonblocking-jobs.txt
```
Byron, line 323: Triple `<<<`? It's amazing that I `bash` can still surprise me :D.

EliahKagan:
I didn't expect that updating dependencies in `Cargo.lock` (5173e9a) would be able to fix the `cargo deny advisories` error about depending on the unmaintained `instant` crate yet, but it looks like it has. Thanks!

I had totally missed that this could already be fixed easily, because I didn't realize that `fastrand` had completely stopped depending on `instant` since 2.0.0 due to https://github.com/smol-rs/fastrand/commit/eb78d89d4a25157778b744a54b8cc207c7480170 (https://github.com/smol-rs/fastrand/pull/60), and we had been depending on an earlier version than that. Sorry about my inaccurate claim that the dependency on `instant` was not possible to remove in this PR!

Byron (approved):
Thanks so much for this massive improvement! I can't wait to see it merged and adjust the auto-merge configuration to listen to `tests-pass` instead of a long list of individual jobs.

Screenshot 2024-11-12 at 08 07 09

Thanks a lot for the hint the additional reduction of risk. I assume that this repository setting is only doable if permissions are set correctly on a PR level, so I am glad this is possible now.

> This is a simple job definition, which could be broken up into two non-matrix job definitions with minimal code duplication. It seems to me that this should be done, so that a new advisory causes the job to have a failing conclusion and fail the workflow even as it does not block PRs. I had originally planned to include that in this PR, but I decided not to do so, since the question of whether to do shoud probably be considered separately.

I agree, let's merge this and have the WASM-dematrixification (if that's a word ;)) in a separate PR. There has been some interest in continuing the development of WASM-support lately so improving CI in that regard will certainly be helpful to them as well.

> I didn't expect that updating dependencies in `Cargo.lock` ([5173e9a](https://github.com/GitoxideLabs/gitoxide/commit/5173e9a2d8464beb7b18de56ac31c04491108381)) would be able to fix the `cargo deny advisories` error about depending on the unmaintained `instant` crate yet, but it looks like it has. Thanks!

No worries at all - at this time I didn't even read up to your assessment of the situation yet and just responded with my standard reflex in these situations 😁. Luckily, it worked.

EliahKagan:
> Thanks so much for this massive improvement! I can't wait to see it merged and adjust the auto-merge configuration to listen to `tests-pass` instead of a long list of individual jobs.

You give me too much credit, because the improvements here are not needed for that, though I believe they do make it easier to keep it working properly. `tests-pass` was introduced in #1551 and has been appropriate to mark as the solely required check in `ci.yml` since there. I am also pretty sure that you have already made it so, since none of the other checks show as required in PRs since around that time.

#1551 did not make it possible for `tests-pass` to substitute for listing checks from other workflows as required, but nothing in this PR extends it in that way, either. As far as I know, there is no change to this repository's required checks (or any other branch protection related settings) that should be done following the merge of this PR.

EliahKagan:
CI failed here on 5173e9a with a [very strange error](https://github.com/GitoxideLabs/gitoxide/actions/runs/11792655044/job/32847342666?pr=1668). It can be examined by expanding the annotations show at the top of that page. The error reports that the GHA runner machine was out of disk space. Although it's possible for a new dependency to have a bug that causes massively increased disk space usage, I think that's probably not the cause, because the error does not happen for that same commit [on my fork](https://github.com/EliahKagan/gitoxide/actions/runs/11792654501/attempts/1). My guess is that it might go away if the workflow is re-run, which I recommend trying.

Byron:
> You give me too much credit, because the improvements here are not needed for that, though I believe they do make it easier to keep it working properly. `tests-pass` was introduced in #1551 and has been appropriate to mark as the solely required check in `ci.yml` since then. I am also pretty sure that you have already made it so, since none of the other checks show as required in PRs since around that time.
> 
> _Edit:_ Yes, this was reported as done in [#1551 (comment)](https://github.com/GitoxideLabs/gitoxide/pull/1551#issuecomment-2308720001).
> 
> #1551 did not make it possible for `tests-pass` to substitute for listing checks from other workflows as required, but nothing in this PR extends it in that way, either. As far as I know, there is no change to this repository's required checks (or any other branch protection related settings) that should be done following the merge of this PR.

Oh, I see, thanks for the clarification. I was on an entirely different page there apparently.

##### No space left on device

For posterity, here is the way the runner itself crashes:

> System.IO.IOException: No space left on device : '/home/runner/runners/2.320.0/_diag/Worker_20241112-075012-utc.log'
   at System.IO.RandomAccess.WriteAtOffset(SafeFileHandle handle, ReadOnlySpan`1 buffer, Int64 fileOffset)
   at System.IO.Strategies.BufferedFileStreamStrategy.FlushWrite()
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.Diagnostics.TextWriterTraceListener.Flush()
   at GitHub.Runner.Common.HostTraceListener.WriteHeader(String source, TraceEventType eventType, Int32 id)
   at GitHub.Runner.Common.HostTraceListener.TraceEvent(TraceEventCache eventCache, String source, TraceEventType eventType, Int32 id, String message)
   at System.Diagnostics.TraceSource.TraceEvent(TraceEventType eventType, Int32 id, String message)
   at GitHub.Runner.Worker.Worker.RunAsync(String pipeIn, String pipeOut)
   at GitHub.Runner.Worker.Program.MainAsync(IHostContext context, String[] args)
System.IO.IOException: No space left on device : '/home/runner/runners/2.320.0/_diag/Worker_20241112-075012-utc.log'
   at System.IO.RandomAccess.WriteAtOffset(SafeFileHandle handle, ReadOnlySpan`1 buffer, Int64 fileOffset)
   at System.IO.Strategies.BufferedFileStreamStrategy.FlushWrite()
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.Diagnostics.TextWriterTraceListener.Flush()
   at GitHub.Runner.Common.HostTraceListener.WriteHeader(String source, TraceEventType eventType, Int32 id)
   at GitHub.Runner.Common.HostTraceListener.TraceEvent(TraceEventCache eventCache, String source, TraceEventType eventType, Int32 id, String message)
   at System.Diagnostics.TraceSource.TraceEvent(TraceEventType eventType, Int32 id, String message)
   at GitHub.Runner.Common.Tracing.Error(Exception exception)
   at GitHub.Runner.Worker.Program.MainAsync(IHostContext context, String[] args)
Unhandled exception. System.IO.IOException: No space left on device : '/home/runner/runners/2.320.0/_diag/Worker_20241112-075012-utc.log'
   at System.IO.RandomAccess.WriteAtOffset(SafeFileHandle handle, ReadOnlySpan`1 buffer, Int64 fileOffset)
   at System.IO.Strategies.BufferedFileStreamStrategy.FlushWrite()
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.Diagnostics.TextWriterTraceListener.Flush()
   at System.Diagnostics.TraceSource.Flush()
   at GitHub.Runner.Common.TraceManager.Dispose(Boolean disposing)
   at GitHub.Runner.Common.TraceManager.Dispose()
   at GitHub.Runner.Common.HostContext.Dispose(Boolean disposing)
   at GitHub.Runner.Common.HostContext.Dispose()
   at GitHub.Runner.Worker.Program.Main(String[] args)

So it looks like writing a trace to a log fails during normal operation, and that even before anything happens.

Screenshot 2024-11-12 at 11 21 13

The cache is pretty big, but no reason to fail this soon.

On the current try, it gets to running the tests at least and I will keep watching it - maybe it works now and all this was no more than a bad (and transient) dream.

Byron:
It really does fail when trying to compile:

```
   Compiling gix-protocol v0.46.0 (/home/runner/work/gitoxide/gitoxide/gix-protocol)
   Compiling gix-status v0.14.0 (/home/runner/work/gitoxide/gitoxide/gix-status)
   Compiling gix-merge v0.0.0 (/home/runner/work/gitoxide/gitoxide/gix-merge)
   Compiling gix-worktree-state v0.14.0 (/home/runner/work/gitoxide/gitoxide/gix-worktree-state)
   Compiling gix-negotiate v0.16.0 (/home/runner/work/gitoxide/gitoxide/gix-negotiate)
   Compiling gix-mailmap v0.25.0 (/home/runner/work/gitoxide/gitoxide/gix-mailmap)
   Compiling tracing-forest v0.1.6
   Compiling gix v0.67.0 (/home/runner/work/gitoxide/gitoxide/gix)
   Compiling gix-fsck v0.7.0 (/home/runner/work/gitoxide/gitoxide/gix-fsck)
   Compiling env_logger v0.10.2
   Compiling gitoxide-core v0.42.0 (/home/runner/work/gitoxide/gitoxide/gitoxide-core)
rustc-LLVM ERROR: IO failure on output stream: No space left on device
rustc-LLVM ERROR: IO failure on output stream: No space left on device
error: could not compile `gitoxide-core` (lib)
warning: build failed, waiting for other jobs to finish...
error: could not compile `gix` (lib)
error: Recipe `journey-tests-async` failed on line 220 with exit code 101
Error: Process completed with exit code 101.
```

And then the log gets truncated so when looking at it again, it seems like not much happened. Let me try something….

EliahKagan:
:tada:

Byron:
> The #1671 test PR here [passes initially](https://github.com/GitoxideLabs/gitoxide/actions/runs/11794934959/job/32853600572?pr=1671), and [then fails](https://github.com/GitoxideLabs/gitoxide/actions/runs/11795027359/job/32853886359?pr=1671) after cherry-picking [5173e9a](https://github.com/GitoxideLabs/gitoxide/commit/5173e9a2d8464beb7b18de56ac31c04491108381) from here. So the increased space usage is triggered by upgrading the crates, somehow. Maybe this is due to dependencies being heavier, but my guess is that it is instead due to there being more stuff that has to be generated rather than obtained from the GitHub Actions cache. Alternatively, maybe the problem is that it retrieves data from the cache and saves them, and the size of those data, some of which can't be used, combined with the files that are generated, is too big.

Yes, I think it's that!

Now the cache is basically useless, yet it consumes a lot of space. Alternatively, I could have cleared it, probably lovering the footprint significantly.

Now I deleted all caches, 35GB of 10GB available, way above the limit, and think that it might now work with the extra clearing which can cut down the build time a little, presumably.

Also, I think caches will become more inefficient over time as also Cargo doesn't delete old caches, so keeps various versions of the same crate around, each time it changes. At least…so I think, judging by the continuously growing crashes on my disk.

Byron:
> Why is it useless? Is this because, since [132696d](https://github.com/GitoxideLabs/gitoxide/commit/132696dce95ce8d79e279978d82f9f038f41a9a4), the data that make it worthwhile are now deleted in the `ci-test` rule when it calls `clear-target` before calling `journey-tests`? If so, this would apply only to the `test` job, and not to any other jobs, right?

Thanks for picking this up. I was blunt with words, and think "inefficient" would have been better. It's probably never useless, it is just reduced in value to the point where it might cost more to maintain the cache than the time it saves.


> If the files where this happens are in the `target` directory, then I think at least in principle the [exclusion of files whose modification time is older than 2 weeks](https://github.com/Swatinem/rust-cache?tab=readme-ov-file#cache-details) should prevent it from growing in this way. That assumes, of course, that the original mtimes metadata are set on files retrieved from the GHA cache. I have not checked to see if that is the case.

While looking up the docs I did realize that I knew nothing at all about how it works - it's much more adapted to the task than I thought. Here is an excerpt:

> In particular, the workspace crates themselves are not cached since doing so is https://github.com/Swatinem/rust-cache/issues/37#issuecomment-944697938. For this reason, this action automatically sets CARGO_INCREMENTAL=0 to disable incremental compilation, so that the Rust compiler doesn't waste time creating the additional artifacts required for incremental builds.

If that environment variable actually carries over, then the worst offender of huge `target` directories would be no more. However, it will still hash its output, even if regenerated from scratch, so multiple versions of the same crate will still be created.
If the 2-week rule holds, then at least there is a limit on how many of those duplicates it would pick up. This also means that dependencies that don't change a lot and remain stable - the case for the majority of the dependency graph - they might fall out of the cache even though they are still very useful.

> Whether that would make anything faster or make the total size of all caches smaller is another question; I am not sure. In addition, there is the problem that the `rust-cache` action seems like it might not be designed to easily facilitate sharing across separate jobs ([Swatinem/rust-cache#205](https://github.com/Swatinem/rust-cache/issues/205)).

Making Windows jobs faster would of course be great, but I wouldn't push it if `rust-cache` isn't made for that.
Regarding the total size of caches, here it was 35GB of 10GB, so I think they are generous in enforcing such limits and I personally wouldn't care about cache-size reduction related to the overall amount stored by GitHub. I would care, however, about the size of the individual cache, but don't think there is any influence. A world where the cache only contains files that are actually used is probably never going to happen.

EliahKagan:
I certainly have no objection to characterizing caching as useless when its goal is to decrease overall load or running time but instead it increases it. I would also call that useless. (The software used to do the caching can of course still be useful in many, even most, other cases, where it does provide savings.) So it didn't seem to me that you were too blunt at all. Instead, I wanted to understand the source of the inefficiency, as well as figure out if there are any current uses of the `rust-cache` action in this project's workflows that should be removed.

As it currently stands (with 132696d), the benefit of caching build artifacts in the CI `test` job seems fairly low, because we:

1. Retrieve cached artifacts, which are those that support whatever build steps were run between the last `cargo clean` and writing the cache.
2. Run `check` `doc` `unit-tests`.
3. Run `clear-target`, which runs `cargo clean`.
4. Run `ci-journey-tests`.
5. Cache artifacts from builds done as part of `ci-journey-tests`, since others were cleaned.

This is to say that we cache artifacts for `ci-journey-tests`, but `ci-journey-tests` never sees those artifacts. Instead, `check`, `doc`, and `unit-tests` get those artifacts, and they may be able to use some of them, since we are talking about dependencies, but it is probably not nearly as good as if they got the artifacts produced from *those* steps. Then both the artifacts from `ci-journey-tests` that were retrieved from the cache, and other artifacts from `check` `doc` `unit-tests`, are removed, and `ci-journey-tests` has to start off with a fresh build.

If the foregoing analysis is correct, then we should try out at least one of these changes in the hope that it will improve running time:

- Not use the GitHub Actions cache for the CI `test` job, *or*
- Split the `test` job into two jobs, at the breaking point where we are running `ci-journey-tests`, by splitting out the running of journey tests to a new job. Then it would no longer be necessary to delete artifacts.

The second approach is elegant in that it splits the jobs where they are effectively almost being split by the running of `cargo clean`. At first glance it seems like a perfect solution. But I am actually skeptical: it might turn out that, with what seems to be an increased size of build artifacts for dependencies ever since `Cargo.lock` was updated, this would--precisely because it would manage to cache everything we want to cache--cause the cache to grow too big.

But I think it's worth trying, so I'll open a PR for it.

In addition, this change might be justified for reasons *unrelated* to caching, in that it would allow for more parallelism across runners.

> If that environment variable actually carries over

Fortunately, [it does](https://github.com/GitoxideLabs/gitoxide/actions/runs/11811519914/job/32905230811#step:8:8).

> However, it will still hash its output, even if regenerated from scratch, so multiple versions of the same crate will still be created.

Is this due to the effect of builds being nondeterministic, or something else?

> Making Windows jobs faster would of course be great, but I wouldn't push it if `rust-cache` isn't made for that.

In that case, I will hold off, for now, on attempting to cause different jobs to share caches.

> Regarding the total size of caches, here it was 35GB of 10GB, so I think they are generous in enforcing such limits and I personally wouldn't care about cache-size reduction related to the overall amount stored by GitHub.

They are generous in the sense that new cache entries are created and old entries are not always immediately dropped. But my understanding of [the documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy) is that, while it's fine for the cache to grow bigger than 10 GB from time to time, this could significantly worsen performance when it happens frequently:

> If you exceed the limit, GitHub will save the new cache but will begin evicting caches until the total size is less than the repository limit. The cache eviction process may cause cache thrashing, where caches are created and deleted at a high frequency. To reduce this, you can review the caches for a repository and take corrective steps, such as removing caching from specific workflows. For more information, see "[Managing caches](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#managing-caches)."

Byron:
> This is to say that we cache artifacts for `ci-journey-tests`, but `ci-journey-tests` never sees those artifacts. Instead, `check`, `doc`, and `unit-tests` get those artifacts, and they may be able to use some of them, since we are talking about dependencies, but it is probably not nearly as good as if they got the artifacts produced from _those_ steps. Then both the artifacts from `ci-journey-tests` that were retrieved from the cache, and other artifacts from `check` `doc` `unit-tests`, are removed, and `ci-journey-tests` has to start off with a fresh build.

I think on top of that is the issue of having so many feature toggles, which make reuse of workspace dependencies (i.e. `gix*` crates) unlikely between unit tests and journey tests. Ultimately, I'd think that the only effective portion of the cache is the one for external dependencies.

> But I think it's worth trying, so I'll open a PR for it. The running time of the CI `test` job has increased to about 20 minutes, and maybe this can speed things up. Even if not, it could help with that for reasons unrelated to caching, in that it would allow for more parallelism across runners.

That's exciting, let's try that!

> > However, it will still hash its output, even if regenerated from scratch, so multiple versions of the same crate will still be created.
> 
> Is this due to the effect of builds being nondeterministic, or something else?

It's deterministic, at least that's a goal, but what I am suggesting is that each code change will cause a new file to be created. The build output changes, and so does the hashed portion of the filename of the build artefact.

But I'd take my 'ideas' with a grain of salt as it's really just that, an idea on how this works based on using it. There is certainly more to it, and I have seen cache-trashing happening when switching between compiler versions in the same project, or when changing environment variables, which is something I'd expect not to be present if that hashing-story was entirely true.

> hey are generous in the sense that new cache entries are created, and in that old entries are sometimes dropped gradually rather than immediately. But my understanding of [the documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy) is that, while it's fine for the cache to grow bigger than 10 GB from time to time, this could significantly worsen performance when it happens frequently:

Even that didn't happen here. Or if it happened, it happened too slowly. After all it seemed easy to generated caches larger than 10GB. After deleting the cache a couple of days ago entirely, it's now approaching 10GB already and I didn't work on `gitoxide` at all. I will keep watching.

EliahKagan:
> That's exciting, let's try that!

I hope to open a PR soon. In the mean time, I've opened the question #1673 to inquire about something that may help keep things clear in the changes in that forthcoming PR, but that I believe need not block that PR.

Byron:
Thanks for the hint - it's disabled now.

PR #1671, opened 2024-11-12
Investigate "No space left on device" CI error on PRs

EliahKagan:
The problem seems to happen only on the upstream repository, or maybe only on PRs from forks. This initially makes no changes, to verify that the problem does not simply happen regardless of content. After that, it cherry-picks 5173e9a from #1668 (to f491909) to see if the problem can be observed with just that.

See #1668, and in particular https://github.com/GitoxideLabs/gitoxide/pull/1668#issuecomment-2469868499, as well as EliahKagan#3, and EliahKagan#4, for details.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,12 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.22.0"
+version = "0.24.2"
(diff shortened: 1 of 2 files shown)


EliahKagan:
Per https://github.com/GitoxideLabs/gitoxide/pull/1668#issuecomment-2470189216, I'll probably close this shortly.

PR #1674, opened 2024-11-14
Split `test` CI job into `test` and `test-journey`

EliahKagan:
As detailed in 711c2f5, this splits out the journey tests into their own job, making changes to both `justfile` and the CI test workflow to do so. This is the change recently discussed in #1668 (https://github.com/GitoxideLabs/gitoxide/pull/1668#issuecomment-2474788049, https://github.com/GitoxideLabs/gitoxide/pull/1668#issuecomment-2475485053). I've tried to avoid making code comments in the `justfile` more confusing when doing this, but I may end up opening a subsequent PR to fix them up, depending on the answer to #1673.

This also fixes up the remaining `with_program tree` in the journey tests, changing it to `with_program find` (f1a171b). All uses of `tree` in the journey tests were replaced with `find` in 3c1bfd5 (#861), but this particular `with_program` call was not updated to reflect that. This allows the test to run when `tree` is absent, and nothing else is using it, so this also no longer installs it in any CI jobs. The reason for including this change in this PR is that it was necessary to figure out which commands had to be run to prepare each of `test` and `test-journey`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
       - name: Setup dependencies
-        run: sudo apt-get install -y --no-install-recommends liblzma-dev tree
+        run: sudo apt-get install -y --no-install-recommends liblzma-dev
       - uses: extractions/setup-just@v2
       - uses: taiki-e/install-action@v2
         with:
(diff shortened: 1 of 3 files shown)


Byron (approved):
Thanks a lot, this is fantastic!

By the looks of it, the separation brings the CI runtime down to about 15 minutes, which is now dominated by the Windows jobs.

Once the journey-test cache kicks in, this particular job should also be faster, and I'd expect it to be faster than the 7 minutes I saw now.

In any case, now I'd expect no disk-space related issues to occur anymore, at least not for a while.

Edit: Merging, as follow-ups could be made in a separate PR.

PR #1676, opened 2024-11-14
Let journey tests fail without `find` and enforce LF line endings

EliahKagan:
f1a171b (#1674) included the conservative step of correcting the last remaining `with_program tree` to `with_program find` and no longer installing `tree` on CI. However, we can go further: any environment without `find` is unsuitable for running journey tests (and, more broadly, broken for general purpose usage). `find` is required by POSIX, is in practice present on all but the most minimal Unix-like systems (and installable on those), and is present in Git Bash.

`with_program` skips tests when the named tool is not found, so using it does not prevent bugs from being concealed. More importantly, all the assertions covered under `with_progam find` were `expect_run_sh $SUCCESSFULLY` assertions, so if `find` were absent, they would fail, revealing the problem. I think that it is better for journey tests that use `find` to fail, when run in strange environments that don't have `find`.

This removes all such calls in bc6e4be. As detailed in that commit message, even with scoping considerations in mind, the subshells introduced with `with_find` calls are no longer needed either, so this removes those and dedents. (This, the changes here are significantly less than a +/- line count would suggest.) These were also the only remaining calls to `with_program`, so 05e176c removes that function definition.

While beginning to work on this, I noticed that shell scripts other than those related to fixtures don't have LF line endings enforced in `.gitattributes` and would end up being checked out with CRLF line endings on Windows. This is undesirable even if `bash` on Windows typically tolerates it, as detailed in 1a3f4e3, which therefore simplifies that part of `.gitattributes` to force LF for all shell scripts.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,8 +1,8 @@
 **/generated-archives/*.tar* filter=lfs-disabled diff=lfs merge=lfs -text
 
 # assure line feeds don't interfere with our working copy hash
-**/tests/fixtures/**/*.sh text crlf=input	eol=lf
-/justfile text crlf=input	eol=lf
+*.sh text crlf=input eol=lf
+justfile text crlf=input eol=lf
 
 # have GitHub include fixture-making scripts when it counts code
 **/tests/fixtures/**/*.sh linguist-vendored=false
--- a/tests/utilities.sh
+++ b/tests/utilities.sh
@@ -7,21 +7,6 @@ RED="$(tput setaf 1 2>/dev/null || echo -n '')"
 OFFSET=( )
 STEP="  "
 
-function with_program () {
-  local program="${1:?}"
-  hash "$program" &>/dev/null || {
-    function expect_run () {
-      echo 1>&2 "${WHITE} - skipped (missing program)"
-    }
-    function expect_run_sh () {
-      echo 1>&2 "${WHITE} - skipped (missing program)"
-    }
-    function expect_run_sh_no_pipefail () {
-      echo 1>&2 "${WHITE} - skipped (missing program)"
-    }
-  }
-}
-
 function on_ci () {
   [ -n "${CI-}" ] || {
     function expect_run () {
--- a/tests/journey/ein.sh
+++ b/tests/journey/ein.sh
@@ -28,144 +28,142 @@ title "Porcelain ${kind}"
     )
   )
   snapshot="$snapshot/porcelain"
-  (with_program find
-    (when "using the 'tool' subcommand"
-      title "ein tool"
-      (with "a repo with a tiny commit history"
-        (small-repo-in-sandbox
-          title "ein tool estimate-hours"
-          (when "running 'estimate-hours'"
-            snapshot="$snapshot/estimate-hours"
-            (with "no arguments"
-              it "succeeds and prints only a summary" && {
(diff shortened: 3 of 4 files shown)


.gitattributes
```diff
@@ -1,8 +1,8 @@
 **/generated-archives/*.tar* filter=lfs-disabled diff=lfs merge=lfs -text
 
 # assure line feeds don't interfere with our working copy hash
-**/tests/fixtures/**/*.sh text crlf=input	eol=lf
```
Byron, line 4: A great catch! I also stumbled over it some time ago but was too afraid to try it 😅.

tests/utilities.sh
```diff
@@ -7,21 +7,6 @@ RED="$(tput setaf 1 2>/dev/null || echo -n '')"
 OFFSET=( )
 STEP="  "
 
-function with_program () {
```
Byron, line 10: > Could this be re-added please?
Even though the name of the file is a bit generic, it should probably rather be "my-test-framework.sh", which I keep copying from project to project. Typically I assume that it's best on my recent projects so can be copied from there.

The above is what I wanted to write, but then again, I thought there isn't too much reason to copy it around anymore, and that I'd probably rather want to try something new, like in-Rust testing of CLIs for which options exist now.

Byron (approved):
You are on fire! Thanks again, it's much appreciated.

tests/utilities.sh
EliahKagan: `git revert 05e176c687cddf9b165bbfc4cfffc03373053bfa` will bring the `with_program` function definition back if it is wanted again, without undoing any of this PR's other changes. I would be pleased to do that immediately and open a PR with it, if you'd like to keep it in.

---

For general use, it seems to me that `with_program`, if kept, should have a strict mode where it fails immediately, rather than defining functions that skip when the program is not found; and that this strict mode should be the default when running on CI.

I had considered doing this, but I noticed that `find` should always be present, and that all assertions with it that were governed by `with_program find` would fail anyway if `find` is absent, since they happened to assert the success of `find` commands. So I removed it, which may be better since it also fails locally if `find` is absent.

If 05e176c is reverted, then that might be a future direction to go in, but no such change would need to be included in a PR that just reverts 05e176c.

.gitattributes
EliahKagan: It might be enough to further simplify this by removing `crlf=input` and just having the pattern and `eol=lf`. I'm unaccustomed to using `crlf=input` and it looks like, today, it might actually be equivalent to `eol=lf` and preserved for [backward compatibility](https://git-scm.com/docs/gitattributes#_backwards_compatibility_with_crlf_attribute).

Byron: Yes, that's probably possible and `gitoxide` should certainly try to be a better example :D.

tests/utilities.sh
Byron: Thanks for elaborating! Now it's even clearer that the removal was the right step and `with_program` is an anti-feature at least if running tests on CI which is a controlled environment - there one would want to assure the needed programs are present.

And like said before, I'd certainly prefer to try to stay in Rust even for testing CLI programs in future.

.gitattributes
EliahKagan: Among other changes, I've removed `crlf=input` from both `.gitattributes` files that had it, in #169. (They both also had `eol=lf`.)

Issue #1681, opened 2024-11-16
`max-pure` feature is not pure (pulls in `libsqlite3-sys`)


tanriol:
### Current behavior 😯

```
$ cargo build --target aarch64-unknown-linux-musl --release --no-default-features --features=max-pure
[...]
error: failed to run custom build command for `libsqlite3-sys v0.30.1`
```

### Expected behavior 🤔

`max-pure` feature should be actually pure (ideally, CI should ensure it works without a C compiler available at all)

Byron:
Thanks for bringing this to my attention.

It's true that `max-pure` is over-promising at the moment and what really is tested on CI is that a build is possible without additional tooling, like `cmake`. Here is the listing of prerequisites that are currently required to make any Rust build:

https://github.com/GitoxideLabs/gitoxide/blob/2fc93f76798ce35f038d52a85b1841fff5280837/.github/workflows/ci.yml#L33-L41

Note that `curl` is used to install Rust via `rustup.rs`.

As of now, it's not possible to make any Rust build without a C compiler as it invokes the linker through it. Thanks to that, C-dependencies can always sneak in and build as long as they don't invoke more than `make` to compile.

Even though this particular issue can be solved by turning off `gitoxide-core-tools-query` for the `max-pure` build, it seems that what this issue really is about is to be able to build without a functional `gcc`.

Maybe the CI run could be tuned to replace `gcc` with a wrapper that only allows linking - then an attempt to build an actual C file would fail naturally, and one would maybe notice even more such invocations.
Maybe there is also a way to force Rust to use `ld` directly?

NobodyXu:
There's nightly flag for using rustup distributed lld, honestly I wish something like zig-cc is provided via rustup.

Byron:
That should be possible to integrate on CI and would solve the problem. Thanks for suggesting!

Byron:
Thanks a lot for your help!

I have set this issue to `blocked` once again as `ring` currently builds some C code as part of its own compilation. It probably does so for very good reason, but while it's doing so a truly pure Rust build can't be accomplished unless one turns off HTTPS transport.

I could also imagine an intermediate step which involves setting up a `pure` target, one that tries to achieve a pure Rust build without having to provide `max` features.

tanriol:
[...]

PR #1682, opened 2024-11-17
Have `pure-rust-build` job find what compiles C code

EliahKagan:
This causes the `pure-rust-build` CI job to provide more insight about what and how some of gitoxide's dependencies, even with `max-pure`, compile C code. This is related to, and hoped to help with, #1681, but it is not a fix for it, because this does not change any of the project's dependencies. It only augments that CI job. The approach is twofold:

1. Look for crates in the dependency tree that very strongly suggest that this is happening. Unless none are found, show the full tree and highlight the lines that show such dependencies.
2. Instrument `cc1`, which `gcc` calls to do the actual work of compiling C and otherwise is rarely called, interleaving a log of its calls with high-level `cargo` output (while still omitting all other forms of verbose output), and also displaying the `cc1` log afterwards.

The new steps detect when and, to a large extent, how and why building `max-pure` builds C code. They would be able to fail the job except that, for now, I have used step-level `continue-on-error` to prevent this.

The commit messages have further details, and some rationale and coverage of subtleties. They also reflect an initially different approach that I replaced with (2). Before moving aside `cc1` and putting in place a script that mocks it, reports the call, and delegates to it, I had deleted `cc1` to observe the first error from the build (e1555bd).

That approach is currently not nearly as good, because it requires two builds (to maintain full coverage of what was previously tested) and because it only reveals the *first* failure, whereas instrumenting `cc1` reveals all calls to it (f0c7d42).

However, I've left it in the history, in part because it's a technique that might revisited in the future when a fix or greater progress on #1681 are in place. I've also left in some small false starts related to logging, since I think that make it easier to discern why a few things are being done the way they are.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,7 +70,7 @@ jobs:
           #!/bin/sh -e
           printf '%s\n' "$0 $*" |
             flock /run/lock/wrapper1.fbd136bd-9b1b-448d-84a9-e18be53ae63c.lock \
-            tee -a /var/log/wrapper1.log /dev/tty >/dev/null
+            tee -a -- /var/log/wrapper1.log ~/ci-stdout >/dev/null
           exec "$0.orig" "$@"
           EOF
 
@@ -85,6 +85,7 @@ jobs:
 
           chmod +x /usr/local/bin/wrap1 /usr/local/bin/wrapper1
           mkdir /run/lock/wrapper1.fbd136bd-9b1b-448d-84a9-e18be53ae63c.lock
+          ln -s -- "/proc/$$/fd/1" ~/ci-stdout
 
           find /usr/lib/gcc \( -name cc1 -o -name cc1plus \) \
             -print -exec /usr/local/bin/wrap1 {} \;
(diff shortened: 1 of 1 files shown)


NobodyXu:
https://github.com/rust-lang/cc-rs/pull/1292

P.S. cc-rs will have the ability to disable compilation soon.

it will be released on next Fri/Sat

Byron (approved):
This is wonderful, and I am amazed how quickly you could cook this up - after all there is considerable finesse involved here and I don't claim to understand it in detail just yet.

But nonetheless, I have seen the output and it's perfectly fits its purpose, showing that there is indeed more C being compiled than originally thought, with some of it being impossible to remove unless `ring` stops relying on it.

Thanks so much!

NobodyXu:
> with some of it being impossible to remove unless ring stops relying on it.

ring uses of c is a bit different, it actually uses c code generated from some perl script.

The perl script is verified to generate correct code, so I think they probably won't remove it.

Also ring depends [aws-lc-rs](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/) by default which also need c compiler

Byron:
Thanks for elaborating! 

It seems that `max-pure` was a pipe-dream to begin with, and what's possible is a `pure` build that drops everything that needs a C compiler. That one wouldn't be able to deal with HTTPS then and thus confined to local operation in practice.

NobodyXu:
Well if you use http only, use the raw git protocol or use the ssh protocol with external binary, then that can work without c compiler.

It'd work for some very specialised workflow, though yeah, for most workload and the general use case it is not possible.

EliahKagan:
> and I don't claim to understand it in detail just yet.

I've opened #1872 to make this clearer.

PR #1687, opened 2024-11-18
Fix 32-bit test failures and test workspace on a 32-bit target on CI

EliahKagan:
Currently, on 32-bit systems, [19 to 22 tests fail](https://gist.github.com/EliahKagan/dd3712ae37a64d434235742026659523) because they assert that data structures have exactly the size they are known to have on 64-bit systems, which the `test-32bit` job does not catch because it only runs tests in `gix-hashtable`.

At one time, `test-32bit` had included an i686 job, but this was removed because of problems related to glibc versioning in the image used by `cross`. This brings that back, splits it out to own job--naming the new job `test-32bit` and renaming the `cross`-using armv7 job `test-32bit-cross`--and solves that problem for it by using a 32-bit Debian container for the job with the `container` key rather than using `cross`.

This reveals a few bugs, which this PR includes fixes or partial workarounds for.

#### Data structure size assertions

As noted above, 22 tests of data structure size have been failing on 32-bit GNU/Linux systems. (Only 19 of the 22 have been failing on 32-bit Windows.) Of these:

1. 18 are mainly to ensure data structures do not grow larger than they have been, but there is some value in immediately finding out if they shrink as well, as described in 77c3c59. So, as described in fc13fc3, the assertions continue to do exact equality comparisons on 64-bit targets, but do `<=` comparisons on 32-bit targets.
2. 4 guard against unanticipated changes to internal data representations, such that it seems like it may be equally useful to find out if they change in size on any target. So for those, exact equality comparisons continue to be done on all targets, but separate expected values are used for 64-bit and 32-bit targets, as detailed in daf9990.

Various other assertions of data structure size already do `<=` comparisons. I think most of those should be changed to the approach in (1)--and perhaps some even to (2)--but that change is not necessary to get tests to run on CI or anywhere, so I haven't included it in this PR.

Even as used in this PR, it is not immediately obvious that the approach in (1) makes sense, because it is in principle possible for different 64-bit targets to have differently sized data structures. This does happen on 32-bit targets--sometimes even for the same architecture--which is why three fewer of these tests fail on 32-bit Windows systems than 32-bit GNU/Linux. However, the sizes have all been the same on four separate 64-bit architectures tested:

- x86-64 (amd64) - most test runs, including on CI (except macOS)
- AArch64 (arm64) - macOS test runs on CI, and [Windows 11 arm64](https://gist.github.com/EliahKagan/7a03f0b6e986774ac228d88fc78ca6c3)
- riscv64 - tested while working on [http://rust-lang/libz-sys#218](https://github.com/rust-lang/libz-sys/pull/218), I don't recall any of these tests failing
- s390x - tested recently as part of research for this, with the idea that it might reveal a problem due to being bigendian - although [some other tests fail](https://gist.github.com/EliahKagan/a107ea7502f79dfd7eea0f94b68689bf), all the size tests pass

So at least for now I think it's reasonable to continue viewing our data structures as generally not varying in size across 64-bit targets.

#### Coupling to gitoxide's own repo

The gitoxide test suite currently depends, I believe mostly unintentionally, on the `gitoxide` repository itself existing and having various properties. Various fixture scripts and tests fail if there is no outer repository; or if the outer repository is blocked by configured ceiling directories; or if the tests are run from a linked worktree, rather than its main worktree; or if `git` operations cannot be performed on the outer repository due to `safe.directory` errors.

As far as I know, the only coupling to the `gitoxide` repository that is firmly intended in the design of the test suite is the use of `.gitignore` information to determine which fixtures are allowed to generate archives. A few uses of the outer repository seem intentional, but not essential to the design, and they can be fixed using a fixture with nested repositories, decoupling completely from `gitoxide`'s own repository. More often, I think this coupling is entirely unintended.

I haven't documented this in full detail yet, and maybe it should have an issue, but see [this gist](https://gist.github.com/EliahKagan/bae50cd40a5558698cf599869e4251b1) and [this repo](https://github.com/EliahKagan/gitoxide-test-runs-outside-main-worktrees). (The latter is a full repo rather than a gist to facilitate visual diff via the web editor, and because it benefits from a non-flat directory structure.)

Fortunately, for this 32-bit Debian container on CI, only one case of this had to be addressed, largely because `GIX_TEST_IGNORE_ARCHIVES` is not being used. `git fetch-pack` commands that were run to produce baseline output--in a fixture script that run even without `GIX_TEST_IGNORE_ARCHIVES`--were actually operations on the `gitoxide` repository itself.

Because of the way ownership works with `actions/checkout` in jobs that use a `container` key, unless customized or `chown`ed, the owner and process UIDs differ. This could be worked around with `chown` (47f3819), but after verifying that works I instead fixed the bug in the fixture script by introducing a temporary nesting repository for `fetch-pack` to operate on (d07c32d).

#### Tests of "GitInstallation" when it is empty

Independent of the above, six tests failed in the container due to weaknesses of tests I introduced in #1567: when there is no identifiable "GitInstallation" scope, some of the tests of "GitInstallation" scope detection fail. This was an intentional decision at the time, but less sound than I had thought.

My thinking in #1567 had been that it will be natural to fix this at the same time as we exclude the `global` scope from ever being detected as "GitIntallation" (or if for some reason that is not to be done then the reason it shouldn't be done would inform how the tests should be improved). However, I had not anticipated that there were already prominent cases where these tests regularly fail even though there is no bug in the code under test. One such case is running the tests in a mostly minimal container that has `git`.

See 178cf25 for details. That commit works around the problem by defining a variable in the `system` scope in the container. It also describes the current situation in detail and describes the significance of likely future developments that would improve or worsen the situation.

This bug is not related to the container being 32-bit, just to it not containing `system` or `global` configuration (and not being macOS with Apple Git, thus not having a high `unknown` scope) as is typical of containers.

#### Duplicated tests

Some data structure size tests were duplicated within their package, as described in #1685. Code duplication would be increased by the changes here, since the tests use more code (see the note at the bottom of fc13fc3). The deduplication resolving #1685 is therefore also included in this PR (dc0a73a).


Byron (approved):
Thanks a lot for improving 32 bit support!

> Various other assertions of data structure size already do `<=` comparisons. I think most of those should be changed to the approach in (1)--and perhaps some even to (2)--but that change is not necessary to get tests to run on CI or anywhere, so I haven't included it in this PR.

Thank you - not only does that save time, but it's likely these started out with equality comparisons, but ended up a bit more flexible due to frequent failures on various platforms.

> [about reliance on the parent `gitoxide` repository] More often, I think this coupling is entirely unintended.

I think so, too! 

> `git fetch-pack` commands that were run to produce baseline output--in a fixture script that run even without `GIX_TEST_IGNORE_ARCHIVES`--were actually operations on the `gitoxide` repository itself.

That's probably also a great example for how easily one could accidentally create a huge `tar` file, after fetching the parent repository without noticing it happened. Even though in this case it appears `git fetch-pack` was called merely to invoke `--diag-url`, which is probably mostly unrelated.

> Because of the way ownership works with `actions/checkout` in jobs that use a `container` key, unless customized or `chown`ed, the owner and process UIDs differ. This could be worked around with `chown` ([47f3819](https://github.com/GitoxideLabs/gitoxide/commit/47f3819d6d0f7a29f735ae4bd5844ba32a1030f3)), but after verifying that works I instead fixed the bug in the fixture script by introducing a temporary nesting repository for `fetch-pack` to operate on ([d07c32d](https://github.com/GitoxideLabs/gitoxide/commit/d07c32d2795e964b6296a4d6dba9b8b197438bb6)).

Thank you.

Something I found generally interesting was that 32bit data structures were only about 20% smaller, not at all close to 50% like one would intuitively expect.

EliahKagan:
> Thank you - not only does that save time, but it's likely these started out with equality comparisons, but ended up a bit more flexible due to frequent failures on various platforms.

Because of the way sizes can shrink and later grow again, as described in 77c3c59, I think it's probably worthwhile to bring back equality comparisons for them on 64-bit systems, while still using `<=` on 32-bit systems, as [`size_ok`](https://github.com/GitoxideLabs/gitoxide/blob/438ee4718b67d0eba8f4c6ebdfac64bd39f68ec7/tests/tools/src/lib.rs#L902-L907) helps do. (Whether `size_ok` is the best way to express this, moving forward, I'm not sure. Maybe an assertion macro should be introduced for this?) However, that only applies if the size is expected to be the same on different 64-bit targets.

> > `git fetch-pack` commands that were run to produce baseline output--in a fixture script that run even without `GIX_TEST_IGNORE_ARCHIVES`--were actually operations on the `gitoxide` repository itself.
> 
> That's probably also a great example for how easily one could accidentally create a huge `tar` file, after fetching the parent repository without noticing it happened. Even though in this case it appears `git fetch-pack` was called merely to invoke `--diag-url`, which is probably mostly unrelated.

Although I can imagine that happening, the more likely kind of problem, with the kinds of coupling I've observed here, would be fetching something *into* the gitoxide repository. (That's what the code changed here had, notionally, done; but, as you mention, nothing was actually ever fetched.) As far as I know, the outer gitoxide repository is never added as a remote of a test repository.

> Something I found generally interesting was that 32bit data structures were only about 20% smaller, not at all close to 50% like one would intuitively expect.

Although it may be unintuitive, it seems that, at least in the data structures in gitoxide that have tests with size assertions, the "lower level" a data structure is, the less it is likely to shrink going from 64-bit to 32-bit targets:

- Data structures that contain lots of handles to data elsewhere, such as `Vec` objects, will shrink more, since such objects consists of pointers and sizes that are meant to participate in arithmetic with the pointers.
- In contrast, data structures that contain a lot of scalar values only decrease in size if some of the data types vary in size across architectures, and for numbers, fixed-size data types, e.g. `u32`, `u64`, seem more common.

But there may also be correlation between a data structure holding values of fixed size and having its size considered important enough to be asserted in a test.

Byron:
> Although it may be unintuitive, it seems that, at least in the data structures in gitoxide that have tests with size assertions, the "lower level" a data structure is, the less it is likely to shrink going from 64-bit to 32-bit targets:

Thanks for elaborating! Now that I read it it makes perfect sense and I am feeling very naive for having that intuition, without even realising (or caring to think about) the most obvious possible reason for this 🤦‍♂️.

EliahKagan:
I should mention--so as to avoid confusion for anyone looking at this later--that in most cases I have not checked the reason for specific data structure size differences. Sometimes the difference is small or zero because the data structure consists mostly or entirely of scalar types whose sizes are not architecture dependent. And it may be that this account for the entire effect.

But another possible contributing factor is packing: some data structures may have less efficient layouts on 32-bit architectures, with holes in some places, to satisfy alignment requirements, to satisfy an ABI convention, or because the Rust compiler decides to lay them out that way. I found this to be why some data structures are different sizes on different 32-bit targets, though I didn't spend time examining all their layouts.

PR #1688, opened 2024-11-18
Check for unrecognized `*-sys` dependencies

EliahKagan:
This is a regression test for the improvement in 3506afb (#1684) that fixed the unintended `libsqlite3-sys` dependency of `max-pure` reported in #1681. This also tries to guard against the introduction of other such crates as `max-pure` dependencies.

Specifically, this builds on #1682 by adding another step to `pure-rust-build` -- this one short, and allowed to fail the job (i.e. not `continue-on-error`) -- that verifies there are no dependencies named like `*-sys`, other than `linux-raw-sys`, which is known about.

I've verified in my fork that the new step fails when applied prior to 3506afb (#1684), and passes afterwards.

*Edit:* I had meant to do things in such a way as to verify that here, too, but I forgot about how checks are actually run as if on a merge commit that would integrate a PR rather than at the tip of the PR (https://github.com/actions/checkout/issues/504), even though that behavior is something I had recently reviewed for something else. I've edited out the misleading details that were inaccurate with respect to upstream checks.

I've also taken this opportunity to improve a shell variable I had somewhat misnamed, in another `pure-rust-build` step.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,9 @@ jobs:
           pattern='.*\b(-sys|cc|cmake|pkg-config|vcpkg)\b.*'
           ! GREP_COLORS='ms=30;48;5;214' grep --color=always -Ex -C 1000000 -e "$pattern" tree.txt
         continue-on-error: true
+      - name: Check for unrecognized *-sys dependencies
+        run: |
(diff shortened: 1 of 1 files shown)


Byron (approved):

Indeed, it should be much harder to sneak sqlite (or similar packages) back in now than it was before, where this could easily have happened without being noticed.

PR #1690, opened 2024-11-18
Fix marking `gix-packetline-blocking` copy auto-generated

EliahKagan:
In c957ab8 (#1340), I intended to mark the contents of `gix-packetline-blocking/src` as auto-generated for github-linguist, but I did not do so correctly, because giving the directory as a path does not specify the `linguist-generated` attribute as being unset for the contents.

This fixes that. The bug, and fix, is revealed by comparing the output of `github-linguist --breakdown` before and after this:

```diff
--- a   2024-11-18 15:45:57.376285107 -0500
+++ b   2024-11-18 15:46:44.808328857 -0500
@@ -1,6 +1,6 @@
-94.14%  Rust
-4.58%   Shell
-1.19%   HTML
+94.08%  Rust
+4.63%   Shell
+1.20%   HTML
 0.09%   Makefile

 Makefile:
@@ -682,23 +682,6 @@
 gix-pack/tests/pack/multi_index/mod.rs
 gix-pack/tests/pack/multi_index/verify.rs
 gix-pack/tests/pack/multi_index/write.rs
-gix-packetline-blocking/src/decode.rs
-gix-packetline-blocking/src/encode/async_io.rs
-gix-packetline-blocking/src/encode/blocking_io.rs
-gix-packetline-blocking/src/encode/mod.rs
-gix-packetline-blocking/src/lib.rs
-gix-packetline-blocking/src/line/async_io.rs
-gix-packetline-blocking/src/line/blocking_io.rs
-gix-packetline-blocking/src/line/mod.rs
-gix-packetline-blocking/src/read/async_io.rs
-gix-packetline-blocking/src/read/blocking_io.rs
-gix-packetline-blocking/src/read/mod.rs
-gix-packetline-blocking/src/read/sidebands/async_io.rs
-gix-packetline-blocking/src/read/sidebands/blocking_io.rs
-gix-packetline-blocking/src/read/sidebands/mod.rs
-gix-packetline-blocking/src/write/async_io.rs
-gix-packetline-blocking/src/write/blocking_io.rs
-gix-packetline-blocking/src/write/mod.rs
 gix-packetline/src/decode.rs
 gix-packetline/src/encode/async_io.rs
 gix-packetline/src/encode/blocking_io.rs
```

(The test was done on Ubuntu 18.04 LTS ESM with `github-linguist` provided by the `ruby-github-linguist` package, version 5.3.3-1.)

In addition to fixing that, this also simplifies `.gitattributes` files throughout the repository. This simplification includes removing redundant `clrf=input` (discussed in #1676) and a spurious unrecognized `-eof` (see #1689), changing lone `**` to `*` where equivalent, and some other changes.

This also makes the style of `.gitattributes` files more consistent, using `foo`/`-foo` rather than `foo=true`/`foo=false` everywhere even where both work, and using whitespace more consistently.

---

Regarding the use of `-linguist-vendored` instead of `linguist-vendored=false`, maybe that will get fixture shell scripts recognized as non-vendored in the language breakdown at [the gitee mirror](https://gitee.com/byronbates/gitoxide). The language breakdown at [the Codeberg mirror](https://codeberg.org/Byron/gitoxide) already recognizes them. Both seem to use `github-linguist`.

However, the motivation for the change is more stylistic consistency, and I doubt this will affect the language breakdown on gitee. I have tested with 7.9.0 (as well as even earlier versions, as above), and both forms seem to be recognized, though it seems there is a separate [7.9.0_gitee](https://gitee.com/oscstudio/linguist/commits/v7.9.0_gitee) version, which I haven't tested with. The language is just shown as Rust if I fork gitoxide on gitee.

I mainly mention the mirrors in connection with Codeberg: that doesn't need fixing, but if this is merged and if, once the Codeberg mirror is updated, the language breakdown is correctly shown on both GitHub and Codeberg, then that will confirm the changes here, at least in the fix to `linguist-generated`, are working.


Byron (approved):
This is great, thank you!
Special thanks for the extra diligence around figuring out if other forges will handle the change correctly.

I will update the mirrors right after merging, too.

EliahKagan:
Thanks! It looks like the language breakdown is working everywhere, including gitee (though only upstream; my fork there still just shows "Rust").

Issue #1696, opened 2024-11-22
Relative time ambiguity across DST adjustments


EliahKagan:
### Current behavior 😯

When [`gix-date::date time::parse::relative::various`](https://github.com/GitoxideLabs/gitoxide/blob/197d31aff5a602c3107c32661340e89781ad0b33/gix-date/tests/time/parse.rs#L179-L201) is run in an environment that uses local time in a time zone that has daylight savings clock adjustments, it fails in the two weeks after daylight saving time changes. All other `gix-date` tests pass.

For example, my time zone is [Eastern Time](https://en.wikipedia.org/wiki/Eastern_Time_Zone), which had a DST clock adjustment from EDT (UTC-4) to EST (UTC-5) on 3 November 2024. Since then, and until two weeks later, the test failed when I ran it locally. Here's a run from 9 November 2024:
 
```text
        FAIL [   0.310s] gix-date::date time::parse::relative::various

--- STDOUT:              gix-date::date time::parse::relative::various ---

running 1 test
test time::parse::relative::various ... FAILED

failures:

failures:
    time::parse::relative::various

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 27 filtered out; finished in 0.31s


--- STDERR:              gix-date::date time::parse::relative::various ---
thread 'time::parse::relative::various' panicked at gix-date/tests/time/parse.rs:195:9:
assertion `left == right` failed: relative times differ
  left: 2024-10-26T20:22:51Z
 right: 2024-10-26T19:22:51Z
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

------------
     Summary [   0.325s] 29 tests run: 28 passed, 1 failed, 0 skipped
        FAIL [   0.310s] gix-date::date time::parse::relative::various
error: test run failed
```

Under those conditions, the time is off by one hour, due to ambiguity of phrases like `2 weeks ago`. If there was a recent daylight savings adjustment to local time, then was `2 weeks` ago the current time minus the [usual](https://www.wolframalpha.com/input?i=2+weeks+in+hours) 336 hours? Or was `2 weeks ago` the time, fourteen calendar days prior, at which local clocks showed the time that local clocks show now? Often these two interpretations of relative times coincide, but they do not always.

[...]

PR #1697, opened 2024-11-23
Reveal local-time DST ambiguity where it occurs

EliahKagan:
This more effectively observes #1696 by expanding `gix-date::date time::parse::relative::various` to have multiple sub-cases. That issue describes a procedure to modify the test code to bring back the failure. This supersedes that procedure. This is to say that, with the changes in this PR, a system where the bug can be observed at any time of the year should always observe it.

Two very important things this does *not* do:

- This neither fixes nor in any way mitigates [#1696](https://github.com/GitoxideLabs/gitoxide/issues/1696).

- No failure occurs, and thus [#1696](https://github.com/GitoxideLabs/gitoxide/issues/1696) remains unobserved, *when*

  1. The system is set (or overridden via `TZ`) to UTC. Thus, this still does not cause a failure on CI. *Or*
  2. The system is set to local time, but in a time zone that does not make daylight savings related clock adjustments. I believe most countries actually do not observe daylight savings clock adjustments, though many do.

To make the results easy to understand and to show full results in the known failure cases, this does not place the existing code inside an outer loop, but instead populates arrays corresponding to the cases, so that assertions can be made comparing the arrays. `pretty_assertions::assert_eq!` is used, in the same manner as elsewhere in the test suite, to allow such output to be easily readable.

In addition to testing more cases, for each "*X* weeks ago" case, a corresponding "*Y* minutes ago" case is also included, where $Y = 10080 X$. Failures occur only in "*X* weeks ago" cases (that include exactly one daylight saving clock adjustment).

Even accounting for the weeks/minutes doubling up of sub-cases, and for variations in daylight savings rules creating the need to check more intervals than one might intuitively expect to need to check, I think this this has more sub-cases than needed. ac17b62 includes a description of why I did it this way and how it might be improved.

I have verified on GNU/Linux, macOS, and Windows that these tests fail in the expected way when run in North American Eastern Time (`America/New York`), but pass when `TZ=UTC` is set. That failure does not occur on CI is further demonstration that when UTC is configured there is not failure (which we should expect, since UTC has no DST-related adjustments). Because CI does not observe the failures, relevant failure output is included in each commit message.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1690,6 +1690,7 @@ dependencies = [
  "itoa",
  "jiff",
  "once_cell",
+ "pretty_assertions",
  "serde",
  "thiserror 2.0.3",
 ]
--- a/gix-date/Cargo.toml
+++ b/gix-date/Cargo.toml
@@ -28,9 +28,10 @@ thiserror = "2.0.0"
 document-features = { version = "0.2.0", optional = true }
 
 [dev-dependencies]
+gix-hash = { path = "../gix-hash" }
 gix-testtools = { path = "../tests/tools" }
 once_cell = "1.12.0"
-gix-hash = { path = "../gix-hash" }
+pretty_assertions = "1.4.1"
 
 [package.metadata.docs.rs]
 all-features = true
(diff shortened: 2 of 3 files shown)


Byron (approved):
Thanks a lot for improving the observability of the test!

I will attempt to share more in the related ticket.

PR #1700, opened 2024-11-24
reduce memory consumption during clone

Byron:
This PR attemps to reuse memory consumption when dealing with pack traversal by trying to avoid over-allocations
due to `with_capacity()`.


EliahKagan:
Am I correct to think that this helps with https://github.com/GitoxideLabs/gitoxide/discussions/1699#discussioncomment-11361517?

Byron:
It's an attempt to improve memory consumption, but it's also ineffective.

Unfortunately I am unable to profile memory consumption on MacOS as instruments fails to attach itself to the binary, despite having given it the respective entitlements. It seems ad-hoc signatures aren't supported, maybe.
I could probably use a linux VM to do the same.

PR #1716, opened 2024-12-09
Enhance builder-likeness of `commit::topo::Builder` in `gix-traverse`

neithernut:
Previously, the only way to create a `Builder` would be via `Builder::from_iters`. That function takes as arguments both the tips and ends, and used to be the only possibility for specifying either of them during the building process.

I would argue that the whole point of a builder is the ability to start with a minimal, setup and configure the resulting entity bit by bit. This would imo include the optional ends of a `Topo`, and I would also argue that there are valid use-cases in which a user would also like to add multiple distinct sets of tips without the need for `Iterator` chaining or, in the worst case, collecting them in another data structure first.

These changes introduce a set of functions enabling alternative usage patterns.


Byron (approved):
Thanks a lot for contributing - these changes make perfect sense!

PR #1720, opened 2024-12-10
Allow `topo::Builder::with_tips` and `with_ends` for non-default pred

neithernut:
We recently (in #1716) introduced the associated fns `with_tips` and `with_ends` for adding tips and ends after initial creation of a `Builder`. We erroneously added them to the one impl constraining the predicate to a specific function pointer type for the default predicate. This meant that tips and ends could only be added until the predicate is set to a `FnMut` that can not be coerced to that same type.

This change lifts this limitation by moving `with_tips` and `with_ends` to the more generic impl for `Builder`.

--

Sorry for the noise. I should have paid more attention where I put those functions. Btw: the rationale for this change also holds for `with_predicate` imo. Typically users will only set a predicate once and not change it afterwards, but it doesn't exactly hurt to have it in the impl generic over `Predicate`.

### DIFF
--- a/gix-traverse/src/commit/topo/init.rs
+++ b/gix-traverse/src/commit/topo/init.rs
@@ -45,23 +45,6 @@ where
         }
     }
 
-    /// Add commits to start reading from.
-    ///
-    /// The behavior is similar to specifying additional `ends` in `git rev-list --topo-order ^ends tips`.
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks for the fix, and… I also missed this in my armchair review. 

One would have thought that using the functions oneself in the implementation would catch all compiler related issues and that this is all trivial, but think again :D.

Issue #307, opened 2022-01-22
Server-side of fetch/pull


Byron:
What would be needed to allow a server to send a pack?

### Server fetch/pull (server to client)

* **git-odb**

The below is a very early draft - it would be better to study existing implementations first to get a better overview on what (not) to do.
This one starts with the fun part to allow writing tests early and experiment with different diff algorithms and potentially their performance.
  
* [x] generate a pack from objects received by an iterator producing (see [issue][pack-mvp])
  * [x] base objects only
  * [x] re-use existing delta objects
  * [x] A mechanism to declare some bases to be 'out of pack' for thin pack support
* [x] **Iterator** to feed pack generation efficiently
* [x] pack creation

[pack-mvp]: https://github.com/Byron/gitoxide/issues/67

* **git-transport**

Certainly needs more research, but roughly…
  
* [ ] Server side `accept()`
  * [ ] http(s)
  * [ ] ssh
  * [ ] ~~daemon~~ probaby only used in testing, and we might implement it if it's useful for us as well
  
* **git-protocol**
  
  * [ ] Server side chatter to negotiate a pack for
    * [ ] protocol V2
    * [ ] protocol V1 (_probably not worth it, let's see_)
  
* **gix-serve**

Probably more like a toy at first merely for testing operation against various git clients.
  
  * [ ] A server able to answer via
    * [ ] http(s)
    * [ ] file protocol (or remote invocation via SSH)
  
[sim-crate]: https://crates.io/crates/similar

Byron:
If I understand the question correctly, the answer is yes. The client of `git-transport` could be handled with what's provided by `accept()`, while the actual interaction patterns would be abstracted in `git-protocol`.

yurigeinish:
This is very desirable to be able to quickly implement your own self-hosted git server (existing ones in other langs require too much memory and therefore are a poor fit for cheap vms).

willstott101:
I've been experimenting with writing a server (in a private repo so far). My interest here is in using parts of gix for the protocol, but leaving the storage up to pluggable backends. I'm very curious about git-on-db, and git-on-object-storage, and git-on-kv-store, and mixtures of the three.

I think I have `ls-refs` working and I'm starting to investigate `fetch`. The protocol is very command-based, and in HTTP AsyncRead & AsyncWrite can't really exist at the same time. Regardless we'd want to leave it up to server implementations to authenticate during the connection, find the relevant repo, authorize the parsed command, then hand back to gitoxide to respond.

[...]

PR #1726, opened 2024-12-17
radicle tuning

Byron:
Make adjustments needed to make the final integration into [`heartwood`](https://app.radicle.xyz/nodes/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5) smoother.

### Tasks

* [x] make all necessary changes
* [x] bypass flaky CI

### Notes

* `gix-negotiate` is pulled in even though it's not used at all. Some types make this crate necessary though and I don't think it's worth trying to factor it out.

### DIFF
--- a/tests/journey/gix.sh
+++ b/tests/journey/gix.sh
@@ -109,7 +109,7 @@ title "gix (with repository)"
 
         # for some reason, on CI the daemon always shuts down before we can connect,
         # or isn't actually ready despite having accepted the first connection already.
-        (with "git:// protocol"
+        (not_on_ci with "git:// protocol"
           launch-git-daemon
           (with "version 1"
             it "generates the correct output" && {
@@ -278,7 +278,7 @@ title "gix commit-graph"
           )
         )
         fi
-        (with "git:// protocol"
+        (not_on_ci with "git:// protocol"
           launch-git-daemon
           (with "version 1"
             (with "NO output directory"
--- a/gitoxide-core/src/pack/receive.rs
+++ b/gitoxide-core/src/pack/receive.rs
@@ -94,7 +94,6 @@ where
     .await?;
     let mut negotiate = Negotiate { refmap: &refmap };
     gix::protocol::fetch(
-        &refmap,
         &mut negotiate,
         |read_pack, progress, should_interrupt| {
             receive_pack_blocking(
@@ -123,7 +122,6 @@ where
             shallow_file: "no shallow file required as we reject it to keep it simple".into(),
             shallow: &Default::default(),
             tags: Default::default(),
-            expected_object_hash: Default::default(),
             reject_shallow_remote: true,
         },
     )
--- a/gix-protocol/src/fetch/error.rs
+++ b/gix-protocol/src/fetch/error.rs
@@ -4,11 +4,6 @@
 pub enum Error {
     #[error("Could not decode server reply")]
     FetchResponse(#[from] crate::fetch::response::Error),
-    #[error("Cannot fetch from a remote that uses {remote} while local repository uses {local} for object hashes")]
-    IncompatibleObjectHash {
-        local: gix_hash::Kind,
-        remote: gix_hash::Kind,
-    },
     #[error(transparent)]
     Negotiate(#[from] crate::fetch::negotiate::Error),
     #[error(transparent)]
@@ -26,11 +21,6 @@ pub enum Error {
     LockShallowFile(#[from] gix_lock::acquire::Error),
     #[error("Receiving objects from shallow remotes is prohibited due to the value of `clone.rejectShallow`")]
     RejectShallowRemote,
-    #[error("None of the refspec(s) {} matched any of the {num_remote_refs} refs on the remote", refspecs.iter().map(|r| r.to_ref().instruction().to_bstring().to_string()).collect::<Vec<_>>().join(", "))]
-    NoMapping {
-        refspecs: Vec<gix_refspec::RefSpec>,
-        num_remote_refs: usize,
-    },
     #[error("Failed to consume the pack sent by the remove")]
     ConsumePack(Box<dyn std::error::Error + Send + Sync + 'static>),
     #[error("Failed to read remaining bytes in stream")]
--- a/gix-protocol/src/fetch/function.rs
+++ b/gix-protocol/src/fetch/function.rs
@@ -1,13 +1,12 @@
 use crate::fetch::{
-    negotiate, Context, Error, Negotiate, NegotiateOutcome, Options, Outcome, ProgressId, RefMap, Shallow, Tags,
+    negotiate, Context, Error, Negotiate, NegotiateOutcome, Options, Outcome, ProgressId, Shallow, Tags,
 };
 use crate::{fetch::Arguments, transport::packetline::read::ProgressAction};
 use gix_features::progress::DynNestedProgress;
 use std::path::Path;
 use std::sync::atomic::{AtomicBool, Ordering};
 
-/// Perform one fetch operation, relying on a `transport`, right after a [`ref_map`](RefMap::new()) was created so
-/// it's clear what the remote has.
+/// Perform one fetch operation, relying on a `transport`.
 /// `negotiate` is used to run the negotiation of objects that should be contained in the pack, *if* one is to be received.
 /// `progress` and `should_interrupt` is passed to all potentially long-running parts of the operation.
 ///
@@ -31,7 +30,6 @@ use std::sync::atomic::{AtomicBool, Ordering};
 /// to inform about all the changes that were made.
 #[maybe_async::maybe_async]
 pub async fn fetch<P, T, E>(
-    ref_map: &RefMap,
     negotiate: &mut impl Negotiate,
     consume_pack: impl FnOnce(&mut dyn std::io::BufRead, &mut dyn DynNestedProgress, &AtomicBool) -> Result<bool, E>,
     mut progress: P,
@@ -46,7 +44,6 @@ pub async fn fetch<P, T, E>(
         shallow_file,
         shallow,
         tags,
-        expected_object_hash,
         reject_shallow_remote,
     }: Options<'_>,
 ) -> Result<Option<Outcome>, Error>
@@ -57,16 +54,6 @@ where
     E: Into<Box<dyn std::error::Error + Send + Sync + 'static>>,
 {
     let _span = gix_trace::coarse!("gix_protocol::fetch()");
-
-    if ref_map.mappings.is_empty() && !ref_map.remote_refs.is_empty() {
-        let mut specs = ref_map.refspecs.clone();
-        specs.extend(ref_map.extra_refspecs.clone());
-        return Err(Error::NoMapping {
-            refspecs: specs,
-            num_remote_refs: ref_map.remote_refs.len(),
-        });
-    }
-
     let v1_shallow_updates = handshake.v1_shallow_updates.take();
     let protocol_version = handshake.server_protocol_version;
 
@@ -92,13 +79,6 @@ where
         arguments.use_include_tag();
     }
     let (shallow_commits, mut shallow_lock) = add_shallow_args(&mut arguments, shallow, &shallow_file)?;
-
-    if ref_map.object_hash != expected_object_hash {
-        return Err(Error::IncompatibleObjectHash {
-            local: expected_object_hash,
-            remote: ref_map.object_hash,
-        });
-    }
 
     let negotiate_span = gix_trace::detail!(
         "negotiate",
--- a/gix-protocol/src/fetch/types.rs
+++ b/gix-protocol/src/fetch/types.rs
@@ -12,8 +12,6 @@ pub struct Options<'a> {
     pub shallow: &'a Shallow,
     /// Describe how to handle tags when fetching.
     pub tags: Tags,
-    /// The hash the remote repository is expected to use, as it's what the local repository is initialized as.
-    pub expected_object_hash: gix_hash::Kind,
     /// If `true`, if we fetch from a remote that only offers shallow clones, the operation will fail with an error
     /// instead of writing the shallow boundary to the shallow file.
     pub reject_shallow_remote: bool,
--- a/gix/src/remote/connection/fetch/error.rs
+++ b/gix/src/remote/connection/fetch/error.rs
@@ -40,6 +40,11 @@ pub enum Error {
         feature: &'static str,
         description: &'static str,
     },
+    #[error("None of the refspec(s) {} matched any of the {num_remote_refs} refs on the remote", refspecs.iter().map(|r| r.to_ref().instruction().to_bstring().to_string()).collect::<Vec<_>>().join(", "))]
+    NoMapping {
+        refspecs: Vec<gix_refspec::RefSpec>,
+        num_remote_refs: usize,
+    },
     #[error("Could not write 'shallow' file to incorporate remote updates after fetching")]
     WriteShallowFile(#[from] crate::shallow::write::Error),
     #[error("'shallow' file could not be locked in preparation for writing changes")]
--- a/gix/src/remote/connection/fetch/receive_pack.rs
+++ b/gix/src/remote/connection/fetch/receive_pack.rs
@@ -73,14 +73,32 @@ where
         P: gix_features::progress::NestedProgress,
         P::SubProgress: 'static,
     {
+        let ref_map = &self.ref_map;
+        if ref_map.mappings.is_empty() && !ref_map.remote_refs.is_empty() {
+            let mut specs = ref_map.refspecs.clone();
+            specs.extend(ref_map.extra_refspecs.clone());
+            return Err(Error::NoMapping {
+                refspecs: specs,
+                num_remote_refs: ref_map.remote_refs.len(),
+            });
+        }
+
         let mut con = self.con.take().expect("receive() can only be called once");
         let mut handshake = con.handshake.take().expect("receive() can only be called once");
         let repo = con.remote.repo;
+
+        let expected_object_hash = repo.object_hash();
+        if ref_map.object_hash != expected_object_hash {
+            return Err(Error::IncompatibleObjectHash {
+                local: expected_object_hash,
+                remote: ref_map.object_hash,
+            });
+        }
+
         let fetch_options = gix_protocol::fetch::Options {
             shallow_file: repo.shallow_file(),
             shallow: &self.shallow,
             tags: con.remote.fetch_tags,
-            expected_object_hash: repo.object_hash(),
             reject_shallow_remote: repo
                 .config
                 .resolved
@@ -95,7 +113,7 @@ where
             user_agent: repo.config.user_agent_tuple(),
             trace_packetlines: con.trace,
         };
-        let ref_map = &self.ref_map;
+
         let negotiator = repo
             .config
             .resolved
@@ -137,7 +155,6 @@ where
         let mut write_pack_bundle = None;
 
         let res = gix_protocol::fetch(
-            ref_map,
             &mut negotiate,
             |reader, progress, should_interrupt| -> Result<bool, gix_pack::bundle::write::Error> {
                 let mut may_read_to_end = false;


tests/journey/gix.sh
```diff
@@ -109,7 +109,7 @@ title "gix (with repository)"
 
         # for some reason, on CI the daemon always shuts down before we can connect,
         # or isn't actually ready despite having accepted the first connection already.
-        (with "git:// protocol"
+        (not_on_ci with "git:// protocol"
```
EliahKagan, line 112: Could the journey test CI flakiness observed here be related to the journey test CI problems discussed in #1634 and worked around in #1725? Since, here, it is connecting, and just not seeing the same refs as are expected, my first thought is that this is separate. But if somehow a previous run of `git-daemon` could still be listening, then maybe it would be possible to connect to it and not are the expected refs.

In my experiments on the earlier breakage, I was unable to observe any process with an image name like `git-daemon`, and in your experiments commented in #1634, while it seemed like the port was not yet available to be bound by a new process, it was not actually still open for connections (which agrees with what I observed). So I suspect that the problem here, with connecting but seeing different refs, may be unrelated. But I figured I'd bring it up just in case.

Byron: Thanks for chiming in!

I think the successful connection is `nc` that connected to a server that is on the way down.

And then the `gix` even managed to connect, but it was doing so while the server shut down, so it unwound while the connection was used.

```

Caused by:
    Broken pipe (os error 32)
```

This probably means that `gix` should offer `gix serve` soon to be able to be able to deal with sockets properly :D.

EliahKagan: > I think the successful connection is `nc` that connected to a server that is on the way down.

In that case, I think running `git daemon` in such a way that we can detect if it is immediately failing may make it possible to avoid this problem entirely. There are a few ways to do that:

1. Keep having the shell create the asynchronous process, but in a different way that incorporates reporting what happened.
2. Keep having the shell create the asynchronous process with `&` in the same way as that is done now, but check the status immediately.
3. Run `git daemon` with `--detach`. I think this might be sufficient to ensure it synchronously fails if it cannot bind to the port, I'm not sure. This would then require different logic to stop the daemon.
4. Run `git daemon` with `--inetd`. Something to look at if way 3 seems like it's on the right track but ends up being unsuitable.

I'll look into this further to see if it can be done in a simple and reliable way.

In addition, we may be able to have to choose a different port based on what is available. But I am unsure if that can easily be done in a portable (no pun intended) way. It would have the benefit of allowing the journey tests to be run in parallel, such as with each other on different builds, and of allowing them to be run at the same time as other tests or processes that use `git-daemon` with a different port. The latter seems potentially valuable.

> And then the `gix` even managed to connect, but it was doing so while the server shut down, so it unwound while the connection was used.

Ah, I see. What's interesting is that, [in my testing](https://github.com/EliahKagan/gitoxide/commit/4dd0942bd251f18642d59c0576f1eb45efd2aa54), `git-daemon` was never running (according to `pgrep`) when a new `git -c ... daemon ...` was about to be run, as well as not accepting any connections. However, it may be that the race condition that happened here could in principle have happened before but (at least in my testing) never did.

> This probably means that `gix` should offer `gix serve` soon to be able to be able to deal with sockets properly :D.

That sounds like a great idea. I had the thought that it should not be specific to the `git://` protocol, since there would be a benefit in testing to being able to use it to test multiple protocols. But if I understand the description in #307, that seems already to be the plan (or maybe the plan, at that time, even excluded `git://`--should #307 be updated?).

Byron: > 3\. Run `git daemon` with `--detach`. I think this might be sufficient to ensure it synchronously fails if it cannot bind to the port, I'm not sure. This would then require different logic to stop the daemon.

It can write a `--pid-file` which should make it trivial to take it down.

I tried this:

```patch
diff --git a/tests/helpers.sh b/tests/helpers.sh
index 2f435a553..d6233ccb6 100644
--- a/tests/helpers.sh
+++ b/tests/helpers.sh
@@ -60,10 +60,13 @@ function small-repo-in-sandbox() {
 }
 
 function launch-git-daemon() {
-    git -c uploadpack.allowrefinwant daemon --verbose --base-path=. --export-all --user-path &>/dev/null &
-    daemon_pid=$!
+    if ! git -c uploadpack.allowrefinwant daemon --verbose --base-path=. --export-all --pid-file=git-daemon.pid --user-path --detach; then
+      echo "failed to start daemon"
+      return 1
+    fi
+    local daemon_pid=$(cat git-daemon.pid)
     while ! nc -z localhost 9418; do
       sleep 0.1
     done
-    trap 'kill $daemon_pid' EXIT
+    trap "kill $daemon_pid" EXIT
 }
```

However, it never failed to startup, and still ran into the same issue repeatedly. The first time it usually worked, but the second launch now sees "Connection reset by peer" repeatedly.

```
                 [it] generates the correct output in JSON format
Connection to localhost port 9418 [tcp/git] succeeded!
        [with] version 1
           [with] NO output directory
              [with] no wanted refs
                 [it] generates the correct output
              [with] wanted refs
                 [it] generates the correct output1c1,5
< Error: None of the refspec(s) =refs/heads/main matched any of the 5 refs on the remote
\ No newline at end of file
---
> Error: Could not decode server reply
>
> Caused by:
>     0: Failed to read from line reader
>     1: Connection reset by peer (os error 54)
\ No newline at end of file
 - FAIL
$ /Users/byron/dev/github.com/GitoxideLabs/gitoxide/target/debug/gix --no-verbose free pack receive -p 1 git://localhost/ -r =refs/heads/main
Output snapshot did not match snapshot at '/Users/byron/dev/github.com/GitoxideLabs/gitoxide/tests/snapshots/plumbing/no-repo/pack/receive/file-v-any-no-output-wanted-ref-p1'
Error: Could not decode server reply

Caused by:
    0: Failed to read from line reader
    1: Connection reset by peer (os error 54)
```

> and of allowing them to be run at the same time as other tests or processes that use `git-daemon` with a different port. The latter seems potentially valuable.

That would be useful, but would also make it necessary to configure the port used for the `git` protocol. It's no big deal, just something I thought I'd mention.

> at that time, even excluded `git://`--should #307 be updated?).

Right, I forgot about that issue 😅. But at least I have updated it - it's a server and it can reasonably handle multiple transports by abstracting over them, similar on how the client already does it. It's still far enough away to have to use `git-daemon` for the foreseeable future though.

EliahKagan: > However, it never failed to startup, and still ran into the same issue repeatedly. The first time it usually worked, but the second launch now sees "Connection reset by peer" repeatedly.

Hmm.

Could the `nc -z` connection(s) actually be triggering a bug in `git-daemon` where, if it receives a connection with no data, it malfunctions later, such as by confusing later connections with the trivial `nc` connection and sending them RST packets?

Byron: The [handler code](https://github.com/git/git/blob/cc01bad4a9f566cf4453c7edd6b433851b0835e2/daemon.c#L873-L915) of the daemon seems quite straightforward - all it does is to hook up the `git upload-pack` command with the incoming socket. But who knows, maybe there is a subtle bug there somehow… but I clearly don't get any of this as the daemon seems to go down without any kind of message.
It doesn't look like the daemon actively resets connections either.
I don't see what it could be and keep thinking that it would be nice to get the actual exit code of the daemon after it detached. With `--detach` it supposedly uses syslog, but messages still appeared in the terminal.
Oh, and I just found [a commit](https://github.com/git/git/commit/02d57da4a5c238eff7ea115c69d2e5c977bf1adb) which seems to indicate the daemon is fine with non-client connections.

Hmm, maybe this is a possible explanation… `nc` causes it to spawn a git subprocess, but later it somehow detects that it's nothing and it goes in for the kill, accidentally killing good subprocesses as well. It should be easy to test that by building a custom version with the killing removed.

I will try that in a bit.

EliahKagan: > PS: Even if that's related, the real question later is: why now? What changed in `gix` that could cause this to be a problem now? Very strange, still.

Has the hypothesis that the change is due solely to the timing of the test runs been ruled out? Prior to the changes here, the workaround of turning off caching (#1725) had seemed to work, which I had suspected (39a1cc7, unmerged) might due merely to extra delay incurred by building dependencies in between the justfile sub-recipes that ran the journey tests in different ways.

Byron: I never understood why that worked in the first place (after all I had to deactivate the tests again on CI), as it's not really about the time between `*.sh` runs. The time between two launches of `git-daemon` is what might matter, but that's not affected by Rust-built binaries as these have already been built. Maybe they now launch much faster, but even that I find hard to believe.
It seems like it's only a problem of `git-daemon`, but the facts seem to indicate it's something that happened in that PR. And it still makes no sense 😅.

EliahKagan: > I never understood why that worked in the first place (after all I had to deactivate the tests again on CI)

Yes. Although I ran it many times, however it affected things was probably nondeterministic and definitely brittle.

Since the only reason for disabling caching was to avoid this problem, and the relevant tests are turned back off on CI now, I'll open a PR shortly to reenable caching in the `test-journey` job.

> as it's not really about the time between `*.sh` runs

It seems to me that the time between `*.sh` runs may be very relevant here.

> The time between two launches of `git-daemon` is what might matter, but that's not affected by Rust-built binaries as these have already been built.

I do not mean immediately preceding test cases in the same run of the journey test script, but rather earlier runs of the journey tests in builds with different features.

On CI, within a single run of the same `test-journey` job, nontrivial `cargo build` commands that compile artifacts do run between separate invocations of `journey.sh`. The helper function that launches `git-daemon` is called at least once in each of the `journey.sh` runs that tests at least one thing related to `git://`.

So the timing between two launches of `git-daemon` is sometimes affected by how long those build commands take. This will happen when one of the launches is the last one before a nontrivial build command and the other launch is the first one after that same nontrivial build command. (I don't know if it happens at other times, but I don't think so.)

Details follow.

#### The per-feature recipes

On CI, several different feature sets are tested by running the journey tests, in the same CI job, and artifacts are compiled in between the runs. There are four feature sets for which `journey.sh` is run separately and, ahead of each one, there are explicit `cargo build` commands:

https://github.com/GitoxideLabs/gitoxide/blob/69ee6a32dd221a1aae7b8c3817f90feacf577598/justfile#L200-L222

All four of those build + `journey.sh` recipes are run in sequence (though in a different order) from the [`ci-journey-tests`](https://github.com/GitoxideLabs/gitoxide/blob/69ee6a32dd221a1aae7b8c3817f90feacf577598/justfile#L18-L19) recipe, which the `test-journey` CI job [runs](https://github.com/GitoxideLabs/gitoxide/blob/69ee6a32dd221a1aae7b8c3817f90feacf577598/.github/workflows/ci.yml#L128). The build commands are non-trivial; they build artifacts, and sometimes download source code. Without caching, I believe they take longer.

#### Verification in logs

That `git://` protocol journey tests are run, then later artifacts are compiled, then later more `git://` protocol journey tests run, can be verified in the logs of CI runs that have journey tests enabled, [such as this run](https://github.com/GitoxideLabs/gitoxide/actions/runs/12368039610/job/34517339672) in the last commit on main that journey tests.

Specifically, in [this log](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6), produced by viewing and saving the full log from that CI run, `git://` protocol tests are logged at lines [1236](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L1236), [1295](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L1295), [1783](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L1783), [1815](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L1815), [2128](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L2128), and [2187](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L2187), while builds that report compiling at least one artifact are logged at lines [717](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L717), [1055](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L1055), [1431](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L1431), [1651](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L1651), and [1919](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L1919). (Searching for `git://` and `Compiling` may be easier, that checking those, though.)

#### A more useful view of the logs

The progression of the most relevant events is:

1. [A `max-pure` build](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L717-L1118).
2. [A `./tests/journey.sh` run](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L1119-L1430) for `max-pure`, in which `git-daemon` is used in a [`gix remote refs`](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L1226-L1246) test and again a [`gix free pack receive`](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L1272-L1339) test.
3. [A `small` build](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L1431-L1498).
4. [A `./tests/journey.sh` run](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L1501-L1650) for `small`, in which `git-daemon` is *not* used.
5. [A `lean-async` build](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L1651-L1742).
6. [A `./tests/journey.sh` run](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L1745-L1918) for `lean-async`, in which `git-daemon` is used in a [`gix remote refs`](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L1780-L1788) test and a [`gix free pack receive`](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L1812-L1835) test.
7. [An all default features plus `http-client-curl-rustls` build](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L1919-L2006).
8. [A `./tests/journey.sh` run](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L2009-L2322) for all default features plus `http-client-curl-rustls`, in which `git-daemon` is used in a [`gix remote refs`](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L2118-L2138) test and a [`gix-free-pack-receive`](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L2164-L2231) test.

Within the individual steps listed above that run `./tests/journey.sh`, nothing is built. But in between them, builds take place, and I suspect that the variation in how long those take is a factor that contributes to whatever strange behavior of `git-daemon` we are seeing here.

#### This progression also applies in failing runs

The log examined above is from a run that succeeded, so a bit more is needed to make the case that build times are plausibly relevant to failures, in the way that failures have actually occurred.

Consider [the failing run](https://github.com/GitoxideLabs/gitoxide/actions/runs/12359365805/job/34491924319?pr=1634#step:6:1049) from #1634, linked from https://github.com/GitoxideLabs/gitoxide/pull/1634#discussion_r1887358877. This runs journey tests for different feature sets on in the same order (no changes to `justfile` have been made since somewhat before then), with the test of all default features plus `http-client-curl-rustls` coming last. GitHub Actions logs are eventually deleted, so just in case, [here's a gist](https://gist.github.com/EliahKagan/c7384bc5616ebdb146d7793d3a03ea63) of that run's log. (It should not be confused with the above gist.)

The same progression is happening there: journey tests on builds with some feature sets run, including journey tests that run and use `git-daemon`, then other builds run, then the journey test of default features plus `http-client-curl-rustls` runs `git-daemon` and fails to communicate with it.

#### Interleaved builds cover some cache-eligible artifacts

I counted build steps are non-trivial above when they compiled at least one artifact. I did this to make clear that artifacts really do get built between runs of `git-daemon` in the journey tests, due to the way the journey tests are run in four sub-recipes defined in the `justfile`, each of which starts with `cargo build` commands.

But `rust-cache` [avoids](https://github.com/Swatinem/rust-cache?tab=readme-ov-file#cache-details) caching workspace crates (https://github.com/Swatinem/rust-cache/issues/37#issuecomment-944697938, https://github.com/GitoxideLabs/gitoxide/pull/1668#issuecomment-2472511656). So the kind of caching `rust-cache` provides is unlikely to have a significant effect on build times unless the artifacts that have to be rebuilt include artifacts for crates that are obtained from the registry rather than being part of this workspace.

It turns out they do. Probably most significantly, because it precedes the `./tests/journey.sh` command in which failures occur in runs that have failures, the `cargo build --features http-client-curl-rustls` command builds [29 crates](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L1927-L1955) eligible for caching, as well as a few more interleaved with `gix-*` crates that are ineligible for caching. Of the crates eligible for caching that it builds, it even downloads [6 of them](https://gist.github.com/EliahKagan/2d0d93f7d0b896ba2a85da7a88844fa6#file-test-journey-log-L1921-L1926).

(I think there may be ways other than timing that this affects networking on the test machine. They would also be very strange and unintuitive. But the problem usually or always happens on GNU/Linux, at least on CI, and does not usually happen on macOS, at least in CI tests of macOS, as mentioned in https://github.com/GitoxideLabs/gitoxide/pull/1634#discussion_r1887990759. So the cause, whatever it is, is something unintuitive.)

#### Synthesis (not really a conclusion)

I do very much agree that some bug in `git-daemon`, or some problem related to it that it not specific to gitoxide, is probably at the root of this.

But the order in which `cargo build` and `./tests/journey.sh` commands run, as well as their observed effects, shows that some delays between `git-daemon` runs are affected by how much work `cargo build` steps do. Furthermore, a `cargo build` step between `git daemon` using tests that always succeed, and those that have been observed to fail, has to download some cache-eligible crates and to build artifacts for them and a number of other cache-eligible crates, when `rust-cache` is omitted.

Both due to those builds and more generally, timing could, as far as I know, still be a significant trigger.

I should be able to test how plausible this is by splitting the tests into one job per feature-set sub-recipe, or by changing the order in which the `ci-journey-tests` recipe lists those recipes.

#### Addendum 1: the `127.0.0.1` vs. `::1` connection (pun intended)

Although I didn't know until this PR that the workaround in #1725 would prove insufficient, I had hoped to find a better way, and while looking for one I [tried forcing](https://github.com/EliahKagan/gitoxide/commit/387a106be1c5565f3f6779c9707494c3404624fb) `git-daemon` to listen on the IPv6 `::1` rather than the IPv4 `127.0.0.1`. The most relevant of the [changed code](https://github.com/EliahKagan/gitoxide/blob/387a106be1c5565f3f6779c9707494c3404624fb/tests/helpers.sh#L62-L84) was this (where `$port` just expands to `9418`):

https://github.com/GitoxideLabs/gitoxide/blob/387a106be1c5565f3f6779c9707494c3404624fb/tests/helpers.sh#L77-L82

In https://github.com/GitoxideLabs/gitoxide/pull/1634#discussion_r1887990759, I had made this partially inaccurate claim:

> The reason I think--or, I should say, thought--`gix` is treating `localhost` to mean `::1`, even when `nc` has treated it to mean `127.0.0.1`, is that it looks like the problem went away when I added `--listen=::1` to the `git daemon` command and modified the `nc -z` loop that follows it to connect to `::1`. But I am not sure that was a correct observation, because when I do this now, it is taking forever, even on tests that had worked before with `::1`: somehow, even where `nc` had reported connecting to `::1`, it cannot do so when passed `::1`.

I now believe that what was really happening is that whenever the problem connecting to `git-daemon` is going to occur, the changing of forcing `::1` causes `nc` to block forever. Runs that succeeded before appear to have succeeded, still to have succeeded, while the journey tests after `cargo build --features http-client-curl-rustls` have it block forever.

This is shown [in the logs](https://github.com/EliahKagan/gitoxide/actions/runs/12393815980/job/34595873344#step:6:981) from the commit with the `::1` experiment (not to be confused with any other logs cited above). The "The operation was canceled" message documents automatic cancellation of the job after it blocked for 360 minutes.

One way to describe this is that forcing IPv6 causes the `nc` command, which is intended to ensure `gix` is able to connect to `git-daemon`, to fail when a `gix` command would fail. That is, if `gix` would not be able to use `git-daemon`, at least as tested `nc` under the `::1` modifications seems to be doing its intended job of blocking--just, the condition it is waiting for never happens. Whether this `::1` experiment really illuminated anything meaningful about the cause of the problem, I am not sure.

#### Addendum 2: `rust-cache` also affects code generation

This seems even weirder and less plausible, but maybe code in gitoxide crates or their dependencies really is the issue and the problem varies depending on whether intermediate artifacts are created.

The `rust_cache` action [sets](https://github.com/Swatinem/rust-cache?tab=readme-ov-file#cache-details) `CARGO_INCREMENTAL=0`. If unsafe code somewhere has undefined behavior and optimizations are turned on, then I think that could cause a bug due to UB to manifest. This seems unlikely for debug builds--such as those exercised in the journey tests. But some crates [are configured to use](https://github.com/GitoxideLabs/gitoxide/blob/69ee6a32dd221a1aae7b8c3817f90feacf577598/Cargo.toml#L197-L208) `opt-level = 3` even in debug builds.

(This is also something I should be able to test, by creating a situation where the tests run but the failures do not occur and then setting `CARGO_INCREMENTAL=0` to see if it has any effect.)

Byron: Thanks a lot for the detailed analysis - stunning work!

> Within the individual steps listed above that run `./tests/journey.sh`, nothing is built. But in between them, builds take place, and I suspect that the variation in how long those take is a factor that contributes to whatever strange behavior of `git-daemon` we are seeing here.

As the removal of `rust_cache` has seemingly solved the problem, it very much felt like a timing issue, possibly also related to network use when downloading crates. 


> The `rust_cache` action [sets](https://github.com/Swatinem/rust-cache?tab=readme-ov-file#cache-details) `CARGO_INCREMENTAL=0`. If unsafe code somewhere has undefined behavior and optimizations are turned on, then I think that could cause a bug due to UB to manifest. This seems unlikely for debug builds--such as those exercised in the journey tests. But some crates [are configured to use](https://github.com/GitoxideLabs/gitoxide/blob/69ee6a32dd221a1aae7b8c3817f90feacf577598/Cargo.toml#L197-L208) `opt-level = 3` even in debug builds.

Even though I definitely like this explanation more, that artefacts differ in meaningful ways when a cache is involved.
It still doesn't make much sense to me though, especially since most failures seem to stem from a trivial connection error, something that I'd trust to be 'correct'.

> But the problem usually or always happens on GNU/Linux, at least on CI, and does not usually happen on macOS, at least in CI tests of macOS, [..]

These days it's easy for me to reproduce the issue, it's basically not working for me anymore at all, most of the time.

Something that changed is that it now errors due to connection reset, not to connection refused.

```
Error: Could not decode server reply

Caused by:
    0: Failed to read from line reader
    1: Connection reset by peer (os error 54)
```

It's also always the second time the `git-daemon` starts up when it fails.

Just for fun I added this patch:

```
diff --git a/tests/journey/gix.sh b/tests/journey/gix.sh
index a7acfc8f7..b79a7a7c0 100644
--- a/tests/journey/gix.sh
+++ b/tests/journey/gix.sh
@@ -280,6 +280,7 @@ title "gix commit-graph"
         fi
         (not_on_ci with "git:// protocol"
           launch-git-daemon
+          sleep 10
           (with "version 1"
             (with "NO output directory"
               (with "no wanted refs"
```

It still failed, so timing shouldn't be an issue. I could also see, while it was waiting, that `git-daemon` was up and running.

Inversely, putting the `sleep 10` before `launch-git-daemon` allowed to validate that the previous `git-daemon` wasn't running anymore.

Both invocations should be completely independent.

To see if it's related to `remote refs` or `free pack receive`, I replaced the `free pack receive` portion with a copy of `remote refs`. *And that works*.

A clear indication that the `git-daemon` is crashing or terminating the connection forcefully.

When replacing the `remote refs` part with a copy of the seemingly failing `free pack receive` part, it seems to work but hangs at this line:

```
Connection to localhost port 9418 [tcp/git] succeeded!
        [with] version 1
           [with] NO output directory
              [with] no wanted refs
                 [it] generates the correct output
              [with] wanted refs
                 [it] generates the correct output
           [with] output directory
              [it] generates the correct output
        [with] version 2
           [with] NO output directory
              [with] NO wanted refs
                 [it] generates the correct output
              [with] wanted refs
                 [it] generates the correct output
                 [when] ref does not exist
                    [it] fails with a detailed error message including what the server said
```

After cranking up debug-logging in the daemon I get this:

```
                    [it] fails with a detailed error message including what the server said07:29:41.100591 run-command.c:655       trace: run_command: REMOTE_ADDR='[::1]' REMOTE_PORT=57324 /Applications/Xcode.app/Contents/Developer/usr/libexec/git-core/git-daemon --serve --verbose --base-path=. --export-all --user-path
07:29:41.102904 exec-cmd.c:139          trace: resolved executable path from Darwin stack: /Applications/Xcode.app/Contents/Developer/usr/libexec/git-core/git-daemon
07:29:41.103190 exec-cmd.c:238          trace: resolved executable dir: /Applications/Xcode.app/Contents/Developer/usr/libexec/git-core
[16726] Connection from [::1]:57324
07:29:41.103327 pkt-line.c:80           packet:          git< git-upload-pack /\0host=localhost\0\0version=2\0agent=git/oxide-0.68.0\0
[16726] Extended attribute "host": localhost
[16726] Extended attribute "protocol": version=2:agent=git/oxide-0.68.0
[16726] Request upload-pack for '/'
07:29:41.103487 run-command.c:655       trace: run_command: GIT_PROTOCOL=version=2:agent=git/oxide-0.68.0 git upload-pack --strict --timeout=0 .
[16726] 07:29:41.106433 exec-cmd.c:139          trace: resolved executable path from Darwin stack: /Applications/Xcode.app/Contents/Developer/usr/libexec/git-core/git
[16726] 07:29:41.106793 exec-cmd.c:238          trace: resolved executable dir: /Applications/Xcode.app/Contents/Developer/usr/libexec/git-core
[16726] 07:29:41.106960 git.c:460               trace: built-in: git upload-pack --strict --timeout=0 .
[16726] 07:29:41.107047 pkt-line.c:80           packet:  upload-pack> version 2
[16726] 07:29:41.107065 pkt-line.c:80           packet:  upload-pack> agent=git/2.39.5.(Apple.Git-154)
[16726] 07:29:41.107167 pkt-line.c:80           packet:  upload-pack> ls-refs=unborn
[16726] 07:29:41.107189 pkt-line.c:80           packet:  upload-pack> fetch=shallow wait-for-done ref-in-want
[16726] 07:29:41.107192 pkt-line.c:80           packet:  upload-pack> server-option
[16726] 07:29:41.107194 pkt-line.c:80           packet:  upload-pack> object-format=sha1
[16726] 07:29:41.107196 pkt-line.c:80           packet:  upload-pack> object-info
[16726] 07:29:41.107198 pkt-line.c:80           packet:  upload-pack> 0000
[16726] 07:29:41.107926 pkt-line.c:80           packet:  upload-pack< command=ls-refs
[16726] 07:29:41.107929 pkt-line.c:80           packet:  upload-pack< agent=git/oxide-0.68.0
[16726] 07:29:41.107931 pkt-line.c:80           packet:  upload-pack< 0001
[16726] 07:29:41.107933 pkt-line.c:80           packet:  upload-pack< symrefs
[16726] 07:29:41.107941 pkt-line.c:80           packet:  upload-pack< peel
[16726] 07:29:41.107943 pkt-line.c:80           packet:  upload-pack< unborn
[16726] 07:29:41.107945 pkt-line.c:80           packet:  upload-pack< ref-prefix refs/heads/
[16726] 07:29:41.107947 pkt-line.c:80           packet:  upload-pack< 0000
[16726] 07:29:41.108354 pkt-line.c:80           packet:  upload-pack> ee3c97678e89db4eab7420b04aef51758359f152 refs/heads/dev
[16726] 07:29:41.108388 pkt-line.c:80           packet:  upload-pack> 3f72b39ad1600e6dac63430c15e0d875e9d3f9d6 refs/heads/main
[16726] 07:29:41.108391 pkt-line.c:80           packet:  upload-pack> 0000
[16726] 07:29:41.108609 pkt-line.c:80           packet:  upload-pack< command=fetch
[16726] 07:29:41.108612 pkt-line.c:80           packet:  upload-pack< agent=git/oxide-0.68.0
[16726] 07:29:41.108613 pkt-line.c:80           packet:  upload-pack< 0001
[16726] 07:29:41.108692 pkt-line.c:80           packet:  upload-pack< thin-pack
[16726] 07:29:41.108694 pkt-line.c:80           packet:  upload-pack< ofs-delta
[16726] 07:29:41.108696 pkt-line.c:80           packet:  upload-pack< include-tag
[16726] 07:29:41.108698 pkt-line.c:80           packet:  upload-pack< done
[16726] 07:29:41.108699 pkt-line.c:80           packet:  upload-pack< 0000
```

The last invocation is a fetch request without any `want <object-id>` line, but the `done` line (as sent by `gix`) should indicate the negotiation ended, send me the pack. But because nothing is wanted, it seems to just hang - it's likely a deadlock - `gix` would now try to read the response, but the `upload-pack` program isn't sending *anything*, but apparently reads as well as it's also hanging there.

When `kill -9` is applied to the upload-pack process, the test just passes and it continues to the second round. There it failed - in this case - because the previous git-daemon didn't shutdown.

I will investigate this further, as `gix` should probably not submit fetch requests without a single want.
Besides that, I don't think this will fix the actual issue, as thus far the first invocation of `git-daemon` seems to work much better, still.
It remains a very illusive issue.

EliahKagan: > These days it's easy for me to reproduce the issue, it's basically not working for me anymore at all, most of the time.
>
> Something that changed is that it now errors due to connection reset, not to connection refused.

I wonder if a recently updated dependency is a factor, and what happens if `--locked` is used on a system where the problem recently began to occur more often.

I wonder also if a system (or Homebrew-provided) package is a factor. `curl` was recently updated on many systems, to fix a CVE (though I don't know how those specifi changes could affect this).

Byron: After my recent trial session I don't even feel it's worth trying those fringes of possible causes as it left me with something very clear to pursue: a hang when a fetch is performed without a single 'want'.

I also learned that `gix remote refs` works twice, but once `gix free pack receive` is involved, it starts failing and resetting the connection. If it is run first, it hangs, and if it runs second, the first connection to the daemon is reset. It also became clear that the daemon is up and operational and that it deliberately resets the connection.

This would make sense as `free pack receive` changed the most, it now uses an entirely new and much more faithful protocol implementation, and one that never got tested with the pure `git://`. It's probably a legitimate bug.

Now that I am writing this I also realise that I can turn on packetline tracing on the client so it would be clear what it says to the server before it shuts down.

I hope I get to it today.

Byron: Let's see [if this works](https://github.com/GitoxideLabs/gitoxide/pull/1731).

The culprit seems to have been that `free pack receive`, albeit relying on `gix-protocol`, could still mess it up by forgetting to properly shut down the connection (by sending and EOF/flush packet), and it was able to request a pack without ever specifying an object to receive, causing a deadlock (upload pack kept reading even though the client indicate it was done). Especially the latter sounds like an attack vector for a server, as it feels a client could cause any amount of connections to be opened and held open, which is much more expensive for the server than it would be for the client.

If it was absolutely no effort, I'd definitely feel curious enough to try if it really does work like I think it would… and… I tried it, but via HTTPS I was *only* able to go into a negotiation loop. However, thinking about it, it's still trivial to hang the server forever by implementing the protocol incorrectly, so I suppose everything server must protect against that. The bug I ran into locally I couldn't reproduce using HTTP though.

Ultimately, it was quite useful to have an inferior implementation of `fetch` as implemented by `gix free pack receive` as it helped to make the underlying `gix-protocol` implementation more sturdy. It now incorporates more of what the `gix::fetch` implementation already checks.

PR #1731, opened 2024-12-20
fix pack receive

Byron:
This also re-enabled git-daemon on CI, which technically was just a workaround as it basically also didn't work locally.

In any case, now it works as it won't leave the server in a strange state due to not indicating the end of interaction.

My guess is that somehow the server would see its slots filled up with hanging processes, leading it to stop accepting
connections and probably also to stop going down.


EliahKagan:
This worked out great--a bug is fixed and the tests are working again and reenabled for CI! 🎉

PR #1732, opened 2024-12-21
fix reflog parsing

Byron:
When the reflog message contains `>`, the previous greedy parsing of the actors signature would cause too much to be consumed.
The idea is to always parse until `\t` which naturally separates the message from the well-defined parts.

@epage maybe you could have a look at this, as I am pretty sure that I have butchered `winnow` and I don't feel good about it.
The difficulty for me is to 'limit' what the actor parser can see at the `\t` location, while leaving the actor parsing as is. Maybe it's possible to 'pre-parse' from the right to '\t' so that
everything else naturally doesn't go beyond `\t`.

Thanks for your help.

### DIFF
--- a/gix-ref/src/store/file/log/line.rs
+++ b/gix-ref/src/store/file/log/line.rs
@@ -73,6 +73,7 @@ impl<'a> From<LineRef<'a>> for Line {
 
 ///
 pub mod decode {
+    use crate::{file::log::LineRef, parse::hex_hash};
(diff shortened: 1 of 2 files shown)


Byron:
I am going to merge this to include the fix in the upcoming release. However, I hope that we will get to update the code to use `winnow` facilities eventually, right now it's a hack to workaround my ignorance.

PR #1734, opened 2024-12-22
Clarify and expand descriptions of `NonFile`s

EliahKagan:
This is a minor documentation revision, to make clearer what a `NonFile`/"non-file" is, as discussed in https://github.com/GitoxideLabs/gitoxide/pull/1730#discussion_r1894381449. At least for now, they are named the same, but the term is more specifically defined.

A couple of possible other names that it could be changed to:

- `Other` as suggested in [#1730 (comment)](https://github.com/GitoxideLabs/gitoxide/pull/1730#discussion_r1894381449). As noted, this is natural. But it is not specific. If Git uses this as a technical term, as suggested, then it should probably be used. I was unable to find it in a cursory search of `git/git` on GitHub, though. I didn't find anything for "NonFile" and everything I found for "non-file" was in phrases like "non-file mailmap," "non-file remote," "non-file backend," and "non-file argument."
- `NonTrackable`, or variations such as `Untrackable` and `NotTrackable` - This idea is from [a commit message in a test](https://github.com/GitoxideLabs/gitoxide/pull/1730/files#diff-406a24eeee8cf467af52b3054214a1d2337eb2b6647169f10de369462bbbe949R14) that refers to the condition of not being a `NonFile` as that of being a "trackable file." I think this is also intuitive when considered abstractly, and accurate. But I'm not sure if it would be intuitive in practice.

Instead of changing it to one of those here, I've just clarified and expanded the documentation. This is with the idea that, with the documentation adjusted this way, it should be possible to figure out if any other name still confers a significant improvement in clarity.

There is another conceptually related change that I made to nearby documentation that is not really covered by this. I'll post a review comment on it so that if it is considered out of scope or unwanted then it can be undone.,

### DIFF
--- a/gitoxide-core/src/repository/clean.rs
+++ b/gitoxide-core/src/repository/clean.rs
@@ -188,7 +188,7 @@ pub(crate) mod function {
             }
 
             match disk_kind {
-                Kind::NonFile => {
+                Kind::Untrackable => {
                     if debug {
                         writeln!(err, "DBG: skipped non-file at '{}'", entry.rela_path).ok();
                     }
@@ -265,7 +265,7 @@ pub(crate) mod function {
                     "WOULD remove"
                 },
                 suffix = match disk_kind {
-                    Kind::NonFile => unreachable!("always skipped earlier"),
+                    Kind::Untrackable => unreachable!("always skipped earlier"),
                     Kind::Directory if entry.property == Some(gix::dir::entry::Property::EmptyDirectory) => {
                         " empty"
                     }
--- a/gitoxide-core/src/repository/clean.rs
+++ b/gitoxide-core/src/repository/clean.rs
@@ -190,7 +190,7 @@ pub(crate) mod function {
             match disk_kind {
                 Kind::Untrackable => {
                     if debug {
-                        writeln!(err, "DBG: skipped non-file at '{}'", entry.rela_path).ok();
+                        writeln!(err, "DBG: skipped untrackable entry at '{}'", entry.rela_path).ok();
                     }
                     continue;
                 }
--- a/gix-dir/src/entry.rs
+++ b/gix-dir/src/entry.rs
@@ -26,11 +26,13 @@ pub enum Property {
 /// The kind of the entry, seated in their kinds available on disk.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Ord, PartialOrd)]
 pub enum Kind {
-    /// Something that is not a file, like a named pipe or character device.
+    /// Something that is not a regular file, directory, or symbolic link.
     ///
-    /// These can only exist in the filesystem.
+    /// These can only exist in the filesystem, because Git repositories do not support them.
+    /// They do not appear as blobs in a repository, and their type is not specifiable in a tree object.
+    /// Examples include named pipes (FIFOs), character devices, block devices, and sockets.
     NonFile,
-    /// The entry is a blob, executable or not.
+    /// The entry is a blob, representing a regular file, executable or not.
     File,
     /// The entry is a symlink.
     Symlink,
--- a/gix-dir/src/entry.rs
+++ b/gix-dir/src/entry.rs
@@ -28,10 +28,11 @@ pub enum Property {
 pub enum Kind {
     /// Something that is not a regular file, directory, or symbolic link.
     ///
-    /// These can only exist in the filesystem, because Git repositories do not support them.
-    /// They do not appear as blobs in a repository, and their type is not specifiable in a tree object.
+    /// These can only exist in the filesystem,
+    /// because Git repositories do not support them, thus they cannot be tracked.
+    /// Hence, they do not appear as blobs in a repository, and their type is not specifiable in a tree object.
     /// Examples include named pipes (FIFOs), character devices, block devices, and sockets.
-    NonFile,
+    Untrackable,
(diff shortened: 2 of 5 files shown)


gix-dir/src/entry.rs
```diff
@@ -26,11 +26,13 @@ pub enum Property {
...
-    /// Something that is not a file, like a named pipe or character device.
+    /// Something that is not a regular file, directory, or symbolic link.
     ///
-    /// These can only exist in the filesystem.
+    /// These can only exist in the filesystem, because Git repositories do not support them.
+    /// They do not appear as blobs in a repository, and their type is not specifiable in a tree object.
+    /// Examples include named pipes (FIFOs), character devices, block devices, and sockets.
     NonFile,
-    /// The entry is a blob, executable or not.
+    /// The entry is a blob, representing a regular file, executable or not.
```
EliahKagan, line 35: This is changing documentation of something other than `NonFile`s. My rationale is twofold:

1. Regular files are not the only kind of blob, because symlinks are also blobs:

   ```text
   120000 blob 965b606f331b51d566b46025f9ff311a7aad0c12    LICENSE-APACHE
   120000 blob 76219eb72e8524f15c21ec93b9b2592da49b5460    LICENSE-MIT
   ```

2. The documentation of `NonFile` now refers to not being representable as a blob, so avoiding implying that only regular files are representable as blobs seems like a benefit. In addition, in another files, `types.rs`, the documentation of `NonFile`s now explicitly distinguishes them from regular files, directories, and symlinks, so I think there's value in explicitly documenting the items here in a way that includes all three of those terms where they apply.

(Of those two points, only the second speaks to why this seems, to me, to be in scope for this PR.)

gitoxide-core/src/repository/clean.rs
```diff
@@ -188,7 +188,7 @@ pub(crate) mod function {
             }
 
             match disk_kind {
-                Kind::NonFile => {
+                Kind::Untrackable => {
                     if debug {
                         writeln!(err, "DBG: skipped non-file at '{}'", entry.rela_path).ok();
```
EliahKagan, line 193:
```suggestion
                        writeln!(err, "DBG: skipped untrackable entry at '{}'", entry.rela_path).ok();
```

gix-dir/tests/dir/walk.rs
```diff
@@ -68,7 +68,7 @@ fn one_top_level_fifo() {
 
     assert_eq!(
         entries,
-        &[entry("top", Untracked, NonFile),],
+        &[entry("top", Untracked, Untrackable),],
         "Non-files are like normal files, but with a different state"
```
EliahKagan, line 72:
```suggestion
        "Untrackable entries are like normal files, but with a different state"
```

gix-dir/tests/dir/walk.rs
```diff
@@ -103,9 +103,9 @@ fn fifo_in_traversal() {
         &[
             entry_nokind(".git", Pruned).with_property(DotGit).with_match(Always),
             entry("dir-with-file/nested-file", Untracked, File),
-            entry("dir/nested", Untracked, NonFile),
+            entry("dir/nested", Untracked, Untrackable),
             entry("file", Untracked, File),
-            entry("top", Untracked, NonFile),
+            entry("top", Untracked, Untrackable),
         ],
         "Non-files only differ by their disk-kind"
```
EliahKagan, line 110:
```suggestion
        "Untrackable entries only differ by their disk-kind"
```

gix-status/src/index_as_worktree_with_renames/mod.rs
```diff
@@ -388,7 +388,7 @@ pub(super) mod function {
         impl<T, U> gix_dir::walk::Delegate for Delegate<'_, '_, T, U> {
             fn emit(&mut self, entry: EntryRef<'_>, collapsed_directory_status: Option<Status>) -> Action {
                 // Status never shows untracked non-files
```
EliahKagan, line 390:
```suggestion
                // Status never shows untracked entries of untrackable type
```

Byron (approved):
Thanks so much for submitting this change, just in time and I think it's substantial!

Even though this PR goes for a minimal change to improve the documentation, the description also offered different variant names. Here I went for `Untrackable` which should help making clear what *not* to do when encountering such a directory entry.

EliahKagan (commented):
Going along with changing `NonFile` to `Untrackable`, I think the term "non-file" in documentation comments should also be reworded, since without `NonFile`, it's not clear what "non-file" means.

Since I don't know what other changes you plan, I don't know if the suggestions I've made here are the way to go or not.

Byron:
A great catch, thanks! I was a bit ~~sloppy~~ quick here 😅.

EliahKagan (commented):
I've reworded the documentation comments that contained "non-file" to go along with the change from `NonFile` to `Untrackable`. See https://github.com/GitoxideLabs/gitoxide/pull/1734/commits/154b21f0e9beb0e5b6615f091f5f0716df6a3f7b. This automatically cancelled auto-merge (so that you can review the further change).

EliahKagan:
I just realized the commit message in 154b21f somewhat overinflates its importance: it's not changing documentation comments, but instead a regular comment and debug and assertion messages. (Fortunately, since it's not really changing documentation, it's not a conventional commit with `doc:`, somewhat limiting the breadth of the confusion.)

PR #1747, opened 2025-01-05
Cancel tree diffing early when matching path is found

cruessler:
This PR is ready for review, but it still requires a couple of iterations before it can be considered ready to be merged.

It introduces a custom `Recorder` that is mainly a slimmed down version of `gix_diff::tree::Recorder`. `Recorder` shortens diffing by cancelling it as soon as a change to `interesting_path` is found. This is based on the assumption that there will be at most one change per file path.

The results seem promising. Compared to the existing implementation, I was able to observe a speedup of about 5 % for `gix blame STABILITY.md`.

```
gitoxide on  main [$?] is 📦 v0.40.0 via 🦀 v1.83.0 took 9s
❯ hyperfine "$HOME/github/Byron/gitoxide/target/release/gix blame STABILITY.md" "$HOME/worktrees/gitoxide/branch-3/target/release/gix blame STABILITY.md"
Benchmark 1: /home/christoph/github/Byron/gitoxide/target/release/gix blame STABILITY.md
  Time (mean ± σ):     507.0 ms ±   7.6 ms    [User: 485.4 ms, System: 19.5 ms]
  Range (min … max):   498.6 ms … 520.3 ms    10 runs

Benchmark 2: /home/christoph/worktrees/gitoxide/branch-3/target/release/gix blame STABILITY.md
  Time (mean ± σ):     483.0 ms ±   4.7 ms    [User: 458.3 ms, System: 22.4 ms]
  Range (min … max):   475.7 ms … 491.5 ms    10 runs

Summary
  '/home/christoph/worktrees/gitoxide/branch-3/target/release/gix blame STABILITY.md' ran
    1.05 ± 0.02 times faster than '/home/christoph/github/Byron/gitoxide/target/release/gix blame STABILITY.md'
```

For `gix blame Cargo.lock`, the speedup was in the range of 25 % to 30 %.

```
gitoxide on  main [$?] is 📦 v0.40.0 via 🦀 v1.83.0 took 9s
❯ hyperfine "$HOME/github/Byron/gitoxide/target/release/gix blame Cargo.lock" "$HOME/worktrees/gitoxide/branch-3/target/release/gix blame Cargo.lock"
Benchmark 1: /home/christoph/github/Byron/gitoxide/target/release/gix blame Cargo.lock
  Time (mean ± σ):      2.579 s ±  0.009 s    [User: 2.374 s, System: 0.197 s]
  Range (min … max):    2.568 s …  2.593 s    10 runs

Benchmark 2: /home/christoph/worktrees/gitoxide/branch-3/target/release/gix blame Cargo.lock
  Time (mean ± σ):      2.022 s ±  0.007 s    [User: 1.821 s, System: 0.192 s]
  Range (min … max):    2.010 s …  2.034 s    10 runs

Summary
  '/home/christoph/worktrees/gitoxide/branch-3/target/release/gix blame Cargo.lock' ran
    1.28 ± 0.01 times faster than '/home/christoph/github/Byron/gitoxide/target/release/gix blame Cargo.lock'
```

I also ran `gix blame` on all files in `gitoxide` to make sure the changed version’s output exactly matched the current version’s output.

# Open questions

- Can we ignore `gix_diff::tree`’s return value or is this a bad idea?
- Where does `Recorder` go and can we find a better name for this struct?
- Would it make sense to rename `stats.trees_diffed` to `stats.trees_partially_diffed`?

### DIFF
--- a/gix-blame/src/file/function.rs
+++ b/gix-blame/src/file/function.rs
@@ -326,7 +326,92 @@ fn tree_diff_at_file_path(
     let tree_iter = odb.find_tree_iter(&tree_id, rhs_tree_buf)?;
     stats.trees_decoded += 1;
 
-    let mut recorder = Recorder::new(file_path.into());
+    struct FindChangeToPath {
+        inner: gix_diff::tree::Recorder,
+        interesting_path: BString,
+        change: Option<gix_diff::tree::recorder::Change>,
+    }
+
+    impl FindChangeToPath {
+        fn new(interesting_path: BString) -> Self {
+            let inner =
+                gix_diff::tree::Recorder::default().track_location(Some(gix_diff::tree::recorder::Location::Path));
+
+            FindChangeToPath {
+                inner,
+                interesting_path,
+                change: None,
+            }
+        }
+    }
+
+    impl Visit for FindChangeToPath {
+        fn pop_front_tracked_path_and_set_current(&mut self) {
+            self.inner.pop_front_tracked_path_and_set_current();
+        }
+
+        fn push_back_tracked_path_component(&mut self, component: &BStr) {
+            self.inner.push_back_tracked_path_component(component);
+        }
+
+        fn push_path_component(&mut self, component: &BStr) {
+            self.inner.push_path_component(component);
+        }
+
+        fn pop_path_component(&mut self) {
+            self.inner.pop_path_component();
+        }
+
+        fn visit(&mut self, change: gix_diff::tree::visit::Change) -> gix_diff::tree::visit::Action {
+            use gix_diff::tree::visit::Action::*;
+            use gix_diff::tree::visit::Change::*;
+
+            if self.inner.path() == self.interesting_path {
+                self.change = Some(match change {
+                    Deletion {
+                        entry_mode,
+                        oid,
+                        relation,
+                    } => gix_diff::tree::recorder::Change::Deletion {
+                        entry_mode,
+                        oid,
+                        path: self.inner.path_clone(),
+                        relation,
+                    },
+                    Addition {
+                        entry_mode,
+                        oid,
+                        relation,
+                    } => gix_diff::tree::recorder::Change::Addition {
+                        entry_mode,
+                        oid,
+                        path: self.inner.path_clone(),
+                        relation,
+                    },
+                    Modification {
+                        previous_entry_mode,
+                        previous_oid,
+                        entry_mode,
+                        oid,
+                    } => gix_diff::tree::recorder::Change::Modification {
+                        previous_entry_mode,
+                        previous_oid,
+                        entry_mode,
+                        oid,
+                        path: self.inner.path_clone(),
+                    },
+                });
+
+                // When we return `Cancel`, `gix_diff::tree` will convert this `Cancel` into an
+                // `Err(...)`. Keep this in mind when using `FindChangeToPath`.
+                Cancel
+            } else {
+                Continue
+            }
+        }
+    }
+
+    let mut recorder = FindChangeToPath::new(file_path.into());
     let result = gix_diff::tree(parent_tree_iter, tree_iter, state, &odb, &mut recorder);
     stats.trees_diffed += 1;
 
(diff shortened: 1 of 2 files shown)


gix-blame/src/file/function.rs
```diff
@@ -322,15 +327,121 @@ fn tree_diff_at_file_path(
...
     stats.trees_decoded += 1;
 
-    let mut recorder = gix_diff::tree::Recorder::default();
-    gix_diff::tree(parent_tree_iter, tree_iter, state, &odb, &mut recorder)?;
+    let mut recorder = Recorder::new(file_path.into());
+    // TODO
+    // `recorder` cancels the traversal by returning `Cancel` when a change to `file_path` is
+    // found. `gix_diff::tree` converts `Cancel` into `Err(...)` which is why we ignore its return
+    // value here. I don’t know whether this has the potential to hide bugs.
+    let _ = gix_diff::tree(parent_tree_iter, tree_iter, state, &odb, &mut recorder);
```
Byron, line 335: Indeed, this is where one would have to ignore the `Cancelled` variant of the error, but fail on all others.
The reason this is an error is… that it's intended to be used for cancellation, which is when one wants this to error out easily.
Here this isn't too useful though.

gix-blame/src/file/function.rs
```diff
@@ -322,15 +327,121 @@ fn tree_diff_at_file_path(
...
-        gix_diff::tree::recorder::Change::Addition { path, .. } => path == file_path,
-        gix_diff::tree::recorder::Change::Deletion { path, .. } => path == file_path,
-    }))
+    Ok(recorder.change)
+}
+
+// TODO
+// The name is preliminary and can potentially include more context. Also, this should probably be
+// moved to its own location.
+struct Recorder {
```
Byron, line 344: I'd probably call it after what it does, like `CancelDiffOncePathTouched` (or better :)).
The implementation I'd also put inside the function that uses it.

gix-blame/src/file/function.rs
```diff
@@ -322,15 +327,121 @@ fn tree_diff_at_file_path(
...
+    fn new(interesting_path: BString) -> Self {
+        Recorder {
+            interesting_path,
+            path_deque: Default::default(),
+            path: Default::default(),
+            change: None,
+        }
+    }
+
+    fn pop_element(&mut self) {
```
Byron, line 361: These utilities can be removed if one would have an `inner: Recorder` which does the bookkeeping. Then one can focus on `visit` and implement what matters.

Byron (changes requested):
Thanks a lot, that's great!

Who would have thought that such a small change can have such large impact! It's also something I didn't even think off, so I am really glad you did!

> Would it make sense to rename stats.trees_diffed to stats.trees_partially_diffed?

I'd keep the name simple as it's typically `Debug` printed anyway. But it's probably fair to be specific in its doc-string to indicate the diff is usually partial.

gix-blame/src/file/function.rs
cruessler: Thanks for the context, I’ve adapted this part.

gix-blame/src/file/function.rs
cruessler: I’ve renamed and inlined `Recorder`.

gix-blame/src/file/function.rs
cruessler: I’ve made `Recorder` a wrapper around `inner: gix_diff::tree::Recorder` (I hope I understood your suggestion correctly). This might have come at a minimal cost to performance, but that impression might also have been noise in running `hyperfine`.

gix-blame/src/file/function.rs
```diff
@@ -322,15 +326,102 @@ fn tree_diff_at_file_path(
...
+                        oid,
+                        path: self.inner.path_clone(),
+                    },
+                });
+
+                // When we return `Cancel`, `gix_diff::tree` will convert this `Cancel` into an
+                // `Err(...)`. Keep this in mind when using `FindChangeToPath`.
+                Cancel
+            } else {
+                Continue
```
Byron, line 409: This needed some getting used to to read due to the similarity with `continue` the keyword.
If that feels like it the next time you come across is, maybe better prefix it with `Action::`.

gix-blame/src/file/function.rs
```diff
@@ -322,15 +326,102 @@ fn tree_diff_at_file_path(
...
+                    } => gix_diff::tree::recorder::Change::Modification {
+                        previous_entry_mode,
+                        previous_oid,
+                        entry_mode,
+                        oid,
+                        path: self.inner.path_clone(),
+                    },
+                });
+
+                // When we return `Cancel`, `gix_diff::tree` will convert this `Cancel` into an
```
Byron, line 405: This comment, along with the comment further down, can probably be removed as self-explanatory as part of cleanup in a future PR.

Byron (approved):
Thanks a lot, this will work!

gix-blame/src/file/function.rs
cruessler: Done in #1755.

PR #1743, opened 2024-12-27
Skip uninteresting commits for blame

cruessler:
This PR is a PoC. It is a first attempt at implementing custom graph traversal for `blame::file`. This allows us to skip commits that we know don’t contain changes to the file being blamed. The PR is rough around the edges and contains a lot of `TODO`s and `todo!()`s.

A first round of tests showed a significant reduction in the number of commits traversed for `gix blame STABILITY.md`:

```
# before
Statistics {
    commits_traversed: 9261,
    commits_to_tree: 10710,
    trees_decoded: 10710,
    trees_diffed: 37,
    blobs_diffed: 36,
}

# after
Statistics {
    commits_traversed: 3657,
    commits_to_tree: 3749,
    trees_decoded: 3749,
    trees_diffed: 37,
    blobs_diffed: 36,
}
```

Also, profiling `gix blame STABILITY.md` on my machine revealed that the command took about 20 % less time to run.

PR #1751, opened 2025-01-09
Rework how blame is passed to parents

cruessler:
Before this PR, blame would have been passed to the first parent in some of the cases where there was more than one.

This PR changes that by only removing a particular suspect from further consideration once it has been compared to all of its parents. For hunks where blame cannot be passed to any parent, we can safely assume that they were introduced by a particular suspect, so we remove those hunks from `hunks_to_blame` and create a `BlameEntry` out of them.

We can illustrate the change using the following small example history:

```text
---1----2
    \    \
     3----4---
```

Let’s now say that we’re blaming a file that has the following content:

```text
line 1 # introduced by (1), then changed by (3)
line 2 # introduced by (1)
line 3 # introduced by (1), then changed by (2)
```

The resulting blame should look like this:

```text
(3) line 1
(1) line 2
(2) line 3
```

The previous version of the algorithm would have passed blame to just (2) or (3), depending on which one came first in the list of parents.

I discovered this discrepancy while working on creating a test for #1743.

Performance-wise there doesn’t seem to be a significant difference to `main`. Where there is, I think it is due to the implementation now being more correct. I haven’t dug really deep into benchmarking the changes, though.

It is possible that there’s still cases that this version of the algorithm does not get right, but I think it is a step in the right direction. Let me know what you think!

# Open question

- Do we prefer `new_unblamed_hunk(…)` or `UnblamedHunk { … }` in tests?

  Related to that: there’s a couple of comments in tests that mention `range_in_destination` although, at this point, this is just the name of a local variable in `new_unblamed_hunk` (it used to be the name of one of `UnblamedHunk`’s fields).

  I can certainly make this more consistent, but am slightly undecided as to whether it is worth the effort.

### DIFF
--- a/gix-blame/src/file/tests.rs
+++ b/gix-blame/src/file/tests.rs
@@ -16,23 +16,24 @@ fn new_unblamed_hunk(range_in_blamed_file: Range<u32>, suspect: ObjectId, offset
 }
 
 mod process_change {
+    use std::str::FromStr;
+
     use super::*;
-    use crate::file::{force_non_zero, process_change, Change, Offset, UnblamedHunk};
-    use crate::BlameEntry;
+    use crate::file::{process_change, Change, Offset, UnblamedHunk};
     use gix_hash::ObjectId;
 
     #[test]
     fn nothing() {
-        let mut lines_blamed = Vec::new();
         let mut new_hunks_to_blame = Vec::new();
         let mut offset_in_destination: Offset = Offset::Added(0);
-        let suspect = ObjectId::null(gix_hash::Kind::Sha1);
+        let suspect = ObjectId::from_str("0000000000000000000000000000000000000000").unwrap();
+        let parent = ObjectId::from_str("1111111111111111111111111111111111111111").unwrap();
 
         let (hunk, change) = process_change(
-            &mut lines_blamed,
             &mut new_hunks_to_blame,
             &mut offset_in_destination,
             suspect,
+            parent,
             None,
             None,
         );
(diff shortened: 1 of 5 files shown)


gix-blame/src/file/tests.rs
```diff
@@ -918,368 +870,345 @@ mod process_change {
...
-        let suspect_2 = ObjectId::from_hex(b"2222222222222222222222222222222222222222").unwrap();
-        let mut lines_blamed: Vec<BlameEntry> = vec![BlameEntry {
-            start_in_blamed_file: 0,
-            start_in_source_file: 0,
-            len: force_non_zero(1),
-            commit_id: suspect,
-        }];
-        let hunks_to_blame = vec![new_unblamed_hunk(1..3, suspect_2, Offset::Added(1))];
+        let suspect = ObjectId::from_hex(b"0000000000000000000000000000000000000000").unwrap();
+        let parent = ObjectId::from_str("1111111111111111111111111111111111111111").unwrap();
```
Byron, line 1157: Here I thought that a constant would be much less noisy and not harder to read, but I leave it to your preference and maybe a next PR can adjust this if you agree.

Byron (approved):
> It is possible that there’s still cases that this version of the algorithm does not get right, but I think it is a step in the right direction.

I absolutely think so too! This is a great finding and I'd hope that this maybe already fixes blame for all multi-parent cases. And I say that clearly as someone who didn't fully understand what's going on with some parts of the algorithm, something I'd hope to correct at some point.

> * Do we prefer `new_unblamed_hunk(…)` or `UnblamedHunk { … }` in tests?
>   Related to that: there’s a couple of comments in tests that mention `range_in_destination` although, at this point, this is just the name of a local variable in `new_unblamed_hunk` (it used to be the name of one of `UnblamedHunk`’s fields).
>   I can certainly make this more consistent, but am slightly undecided as to whether it is worth the effort.

I don't think I have a preference, but vaguely remember having introduced `new_unblamed_hunk()` for … some reasons I don't entirely recall.


Regarding everything else, I definitely didn't setup my review to the point where I can say that I follow each and every change, but assume it must be right as you are the expert. What I could do is to remove one TODO, which goes along with the general suggestion to try and avoid unnecessary `collect()` calls.

Byron:
Argh, I had couple of fixes that I forgot to push, have to adjust this in a follow-up.

gix-blame/src/file/tests.rs
cruessler: I just tried, and it appears that we can’t extract into a constant as neither `from_hex` nor `from_str` work in a `const` context which would be required for `const SUSPECT: ObjectId = …`. Or did you have something else in mind?

Byron: Right, these aren't const so can't easily be used in a static context. I think then I would define functions for it that deal with the `.unwrap()`, so it looks more like `let suspsect = null_sha(); let parent = one_sha()`.

cruessler: Makes sense! Done in #1756.

PR #1752, opened 2025-01-09
git login shell

Byron:
Support for another standard path which helps executing things.


tests/tools/src/lib.rs
```diff
@@ -652,6 +652,7 @@ fn configure_command<'a, I: IntoIterator<Item = S>, S: AsRef<OsStr>>(
 }
 
 fn bash_program() -> &'static Path {
+    // TODO: use `gix_path::env::login_shell()` when available.
```
EliahKagan, line 655: For the reasons described above, the login shell is not in general a reasonable or safe choice of a fallback for running scripts without shebangs or whose shebangs were not resolvable.

(In addition, for some of the fixture scripts in the gitoxide project itself, this currently has to be `bash`; even other popular Bourne-style shells such as `zsh` and `ksh93` have relevantly different semantics. However, since their shebangs use `/usr/bin/env` to run `bash`, which works on most Unix-like systems, this specific point about the behavior of gitoxide's own fixture scripts is minor.)

gix-path/tests/path/env.rs
```diff
@@ -0,0 +1,71 @@
...
+        "it finds something as long as git is installed somewhere on the system (or a default location)"
+    );
+}
+
+#[test]
+fn login_shell() {
+    // On CI, the $SHELL variable isn't necessarily set. Maybe other ways to get the login shell should be used then.
+    if !gix_testtools::is_ci::cached() {
+        assert!(gix_path::env::login_shell()
+            .expect("There should always be the notion of a shell used by git")
```
EliahKagan, line 15: Doesn't `git` always use `sh` for automated purposes, such as running configured commands that run in a shell, and running scripts that do not have shebangs but are assumed to be shell scripts? In Git for Windows, `sh` is the `bash` it comes with, but I think it is still using it as `sh`. In general, even shells that are capable of POSIX-compatible behavior for running `sh` scripts must be invoked as `sh` (rather than their own name) or otherwise told to do so, to behave compatibly.

If the expectation expressed here is referring to something in `git` itself and is not specific to Git for Windows, then I would expect that to be a bug in `git`.

This does not necessarily mean that a `login_shell` facility should not exist and be tested in gitoxide. But I don't know of any reasonable uses for it in any code currently in gitoxide. The only use for it I am aware of is if applications that use gitoxide want to offer an interactive terminal with a reasonable choice of shell and want to make sure to pick the Git for Windows `bash.exe` on Windows.

EliahKagan (commented):
I don't think this will work as intended. The login shell is a shell that the user prefers for interactive use. There are no restrictions on its syntax. It need not be POSIX-compatible or Bourne-style. For example, `csh`/`tcsh`, `fish`, `xonsh`, and occasionally even `pwsh` or `tclsh` are used as login shells by some users on Unix-like systems.

Different popular shells have different syntax. Even where the syntax is superficially similar--which is not always the case--it tends to differ in quoting rules, affecting what requires quoting and how quoting must be done. (The quoting rules in `csh` and `tcsh` are particularly tricky.)

Furthermore, `$SHELL` need not conceptually *be* a shell. While, when set, it is usually some kind of shell--or, for users who do not log on, `/bin/false`--even that is not guaranteed.

The only good automated use I am aware of that runs or arranges to run code in the login shell is when an application has startup script logic or shell integrations. Then `$SHELL` may be used as one source of information about what shells the user may wish to have configured. Most often, this should still be accompanied by interactively checking if the user wants that to be done. More importantly, it has to be done differently for different shells, typically by inferring the shell language from the shell's name.

Using whatever the user has picked as a login shell to run Git hooks would be particularly undesirable. Hopefully hooks have working shebangs, but if they don't and are interpreted as being in a different language, then unexpected effects that are hard to reason about will occur. I think this would also make it infeasible, in many if not most situations, to write hooks securely.

That is not *necessarily* to say that there is no value in gitoxide having a notion of login shell and making this information available to applications. I've noted a hypothetical use below, for interactive applications. In addition, part of the goal here seems to be to more often manage to run scripts in the Git for Windows `bash.exe`. That seems like a good idea, though it looks like the approach taken here to do so is not reliable, as detailed below.

If the goal is to find a shell that the user may prefer or consider acceptable for running scripts or running configured commands, and if there are particular shells that are known to be reasonable to try for this purpose, then they could be named in a list and looked up via a `PATH` search. But as far as I know, `sh` (including other shells that behave as `sh` when run with the name `sh`) is the only generally safe and reasonable choice.

Byron:
Thanks a lot for sharing your incredible insights! I wholeheartedly agree and believe that I have unconditionally ported the shell-selection behaviour of `git2-hooks` here, while trying to make improvements in how to acquire `bash.exe` on Windows.

Originally the intent was to just run hooks, but another goal would be to be able to run programs in the same environment as one would have when running in a terminal. Particularly, we need programs like `ssh` and `gpg` to be able to pickup the user's configuration.

Interestingly, when checking which environment GUI applications start up, at least on MacOS, then it seemed that they do pick up the environment of the login shell, yet there are reported failures where external programs still don't work as they should. It's a bit puzzling, but also I am currently lacking good reproductions of the issue so I am poking around in the dark. That should probably stop in favor of a more distinctive approach.

> If the goal is to find a shell that the user may prefer or consider acceptable for running scripts or running configured commands, and if there are particular shells that are known to be reasonable to try for this purpose, then they could be named in a list and looked up via a `PATH` search. But as far as I know, `sh` (including other shells that behave as `sh` when run with the name `sh`) is the only generally safe and reasonable choice.

This is the behaviour currently implemented in `gix-command`, even though on Windows it will bypass the shell if possible by splitting arguments itself to run the program directly.
The reason `git2-hooks` arrived at this is probably that it also wants to assure that hooks run in the correct environment, while assuming that executable bits and shebangs are set.


> One possible motive for finding out the login shell is to actually run it as a login shell to set up a login-like environment. The code here does not seem strongly oriented toward doing that, since in most shells (I can only say most, since `$SHELL` need not even be a shell at all) it would be necessary to invoke it in a special way to cause it to behave _as_ a login shell. Although some shells recognize `-l` or `--login` for this purpose, the way a login shell is [actually started](https://unix.stackexchange.com/a/46856) in a Unix-like system is by prepending a `-` to the name that is passed to it as its own name, i.e., to the value of `argv[0]`.

I clearly didn't do my research and rushed to action, but now learned there are differences between interactive shells, interactive login shells, non-interactive login shells, and those that are none of that. I also see that `login` shell is nothing applications should ever trigger by themselves.

I will keep working on answers to the line comments as well.

gix-path/tests/path/env.rs
Byron: Now that I learned more about the purpose of login shells and also that `git2-hooks` might not be the reference implementation for how to do this right. #1758 should correct this.

gix-path/src/env/mod.rs
```diff
@@ -28,6 +28,23 @@ pub fn installation_config_prefix() -> Option<&'static Path> {
...
+/// Return the shell that Git would prefer as login shell, the shell to execute Git commands from.
+///
+/// On Windows, this is the `bash.exe` bundled with it, and on Unix it's the shell specified by `SHELL`,
+/// or `None` if it is truly unspecified.
+pub fn login_shell() -> Option<&'static Path> {
+    static PATH: Lazy<Option<PathBuf>> = Lazy::new(|| {
+        if cfg!(windows) {
+            installation_config_prefix()
+                .and_then(|p| p.parent())
+                .map(|p| p.join("usr").join("bin").join("bash.exe"))
```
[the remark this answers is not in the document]
Byron: Thanks a lot! Admittedly I just tried to find a base path which is the smallest distance away from where I wanted to go, and went with that.

#1758 should do better.

PR #1753, opened 2025-01-10
wip changes against more than one parent

Byron:
The PR #1751 as it should have been.

### DIFF
--- a/gix-blame/src/file/function.rs
+++ b/gix-blame/src/file/function.rs
@@ -172,10 +172,9 @@ where
                 if more_than_one_parent {
                     // None of the changes affected the file we’re currently blaming.
                     // Copy blame to parent.
-                    hunks_to_blame = hunks_to_blame
-                        .into_iter()
-                        .map(|unblamed_hunk| unblamed_hunk.clone_blame(suspect, parent_id))
-                        .collect();
+                    for unblamed_hunk in &mut hunks_to_blame {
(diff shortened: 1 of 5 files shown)


cruessler:
Thanks a lot for the refactor! Again, I’ve learned one or two things from it!

EliahKagan:
Something seems to be wrong with the checks here: [there are none](https://github.com/GitoxideLabs/gitoxide/pull/1753/checks). I don't know why. Maybe it is a bug in GitHub triggered by some combination of force pushes? In any case, it looks like the intent here is to merge this once tests pass, but auto-merge is unable to proceed because it is waiting on checks that are not running, nor queued, nor even enumerated.

Byron:
That is very strange indeed!

Instead of trying to fix it by maybe redoing the push after a trivial change, I thought it was easiest to just force-merge with admin overrides. 
I'd hope that `main` stays green but if not, I hope the follow-up PR to fix it will work again.

EliahKagan:
On main since the merge, it looks like there's a readily fixable [clippy error](https://github.com/GitoxideLabs/gitoxide/actions/runs/12706605241/job/35419888495), but otherwise everything is okay.

Byron:
So close :D, here is the fix https://github.com/gitbutlerapp/gitbutler/pull/5910.

cruessler:
The link goes to gitbutler, but `main` is green again. :-)

EliahKagan:
I think #1754 had the fix.

PR #1758, opened 2025-01-11
Fix `gix_path::env::login_shell()` by removing it

Byron:
Thanks to the review of @EliahKagan we can stop the proliferation of a bad idea in its tracks and replace it with what hopefully is a better idea.

Follow-up on #1752 .

### Tasks

* [x] replace `login_shell()` with an alternative
* [x] improve the heurstic by which `sh.exe` is located on windows.
* [x] add `gix env` to print out such paths more easily

### Notes for the Reviewer

* On Windows, `git --exec-path` is always called even though one might do without under some circumstances. This optimization can be done later once this function is widely enough used.

### DIFF
--- a/gix-path/src/env/mod.rs
+++ b/gix-path/src/env/mod.rs
@@ -36,17 +36,17 @@ pub fn installation_config_prefix() -> Option<&'static Path> {
 /// as it could possibly be found in `PATH`.
 /// Note that the returned path might not be a path on disk.
 pub fn shell() -> &'static OsStr {
-    static PATH: Lazy<Option<OsString>> = Lazy::new(|| {
+    static PATH: Lazy<OsString> = Lazy::new(|| {
         if cfg!(windows) {
-            installation_config_prefix()
-                .and_then(|p| p.parent())
+            core_dir()
+                .and_then(|p| p.ancestors().nth(3) /* skip mingw64/libexec/git-core */)
                 .map(|p| p.join("usr").join("bin").join("sh.exe"))
-                .map(Into::into)
+                .map_or_else(|| OsString::from("sh"), Into::into)
         } else {
-            Some("/bin/sh".into())
+            "/bin/sh".into()
         }
     });
-    PATH.as_deref().unwrap_or(OsStr::new("sh"))
+    PATH.as_ref()
 }
 
 /// Return the name of the Git executable to invoke it.
--- a/gix-path/src/env/mod.rs
+++ b/gix-path/src/env/mod.rs
@@ -1,4 +1,4 @@
-use std::ffi::OsString;
+use std::ffi::{OsStr, OsString};
 use std::path::{Path, PathBuf};
 
 use bstr::{BString, ByteSlice};
@@ -28,21 +28,25 @@ pub fn installation_config_prefix() -> Option<&'static Path> {
     installation_config().map(git::config_to_base_path)
 }
 
-/// Return the shell that Git would prefer as login shell, the shell to execute Git commands from.
+/// Return the shell that Git would use, the shell to execute commands from.
 ///
-/// On Windows, this is the `bash.exe` bundled with it, and on Unix it's the shell specified by `SHELL`,
-/// or `None` if it is truly unspecified.
-pub fn login_shell() -> Option<&'static Path> {
-    static PATH: Lazy<Option<PathBuf>> = Lazy::new(|| {
+/// On Windows, this is the full path to `sh.exe` bundled with it, and on
+/// Unix it's `/bin/sh` as posix compatible shell.
+/// If the bundled shell on Windows cannot be found, `sh` is returned as the name of a shell
+/// as it could possibly be found in `PATH`.
+/// Note that the returned path might not be a path on disk.
+pub fn shell() -> &'static OsStr {
+    static PATH: Lazy<Option<OsString>> = Lazy::new(|| {
         if cfg!(windows) {
             installation_config_prefix()
                 .and_then(|p| p.parent())
-                .map(|p| p.join("usr").join("bin").join("bash.exe"))
+                .map(|p| p.join("usr").join("bin").join("sh.exe"))
+                .map(Into::into)
         } else {
-            std::env::var_os("SHELL").map(PathBuf::from)
+            Some("/bin/sh".into())
         }
     });
-    PATH.as_deref()
+    PATH.as_deref().unwrap_or(OsStr::new("sh"))
 }
 
 /// Return the name of the Git executable to invoke it.
--- a/gix-path/src/env/mod.rs
+++ b/gix-path/src/env/mod.rs
@@ -106,6 +106,36 @@ pub fn xdg_config(file: &str, env_var: &mut dyn FnMut(&str) -> Option<OsString>)
         })
 }
 
+static GIT_CORE_DIR: Lazy<Option<PathBuf>> = Lazy::new(|| {
+    let mut cmd = std::process::Command::new(exe_invocation());
+
+    #[cfg(windows)]
+    {
+        use std::os::windows::process::CommandExt;
+        const CREATE_NO_WINDOW: u32 = 0x08000000;
+        cmd.creation_flags(CREATE_NO_WINDOW);
+    }
+    let output = cmd.arg("--exec-path").output().ok()?;
+
+    if !output.status.success() {
+        return None;
+    }
+
+    BString::new(output.stdout)
+        .trim_with(|b| b.is_ascii_whitespace())
+        .to_path()
+        .ok()?
+        .to_owned()
+        .into()
+});
+
+/// Return the directory obtained by calling `git --exec-path`.
+///
+/// Returns `None` if Git could not be found or if it returned an error.
+pub fn core_dir() -> Option<&'static Path> {
+    GIT_CORE_DIR.as_deref()
+}
+
 /// Returns the platform dependent system prefix or `None` if it cannot be found (right now only on windows).
 ///
 /// ### Performance
@@ -129,22 +159,7 @@ pub fn system_prefix() -> Option<&'static Path> {
                 }
             }
 
-            let mut cmd = std::process::Command::new(exe_invocation());
-            #[cfg(windows)]
-            {
-                use std::os::windows::process::CommandExt;
-                const CREATE_NO_WINDOW: u32 = 0x08000000;
-                cmd.creation_flags(CREATE_NO_WINDOW);
-            }
-            cmd.arg("--exec-path").stderr(std::process::Stdio::null());
-            gix_trace::debug!(cmd = ?cmd, "invoking git to get system prefix/exec path");
-            let path = cmd.output().ok()?.stdout;
-            let path = BString::new(path)
-                .trim_with(|b| b.is_ascii_whitespace())
-                .to_path()
-                .ok()?
-                .to_owned();
-
+            let path = GIT_CORE_DIR.as_deref()?;
             let one_past_prefix = path.components().enumerate().find_map(|(idx, c)| {
                 matches!(c,std::path::Component::Normal(name) if name.to_str() == Some("libexec")).then_some(idx)
             })?;
--- a/gix-path/src/env/mod.rs
+++ b/gix-path/src/env/mod.rs
@@ -30,7 +30,7 @@ pub fn installation_config_prefix() -> Option<&'static Path> {
 
 /// Return the shell that Git would use, the shell to execute commands from.
 ///
-/// On Windows, this is the full path to `sh.exe` bundled with it, and on
+/// On Windows, this is the full path to `sh.exe` bundled with Git, and on
 /// Unix it's `/bin/sh` as posix compatible shell.
 /// If the bundled shell on Windows cannot be found, `sh` is returned as the name of a shell
 /// as it could possibly be found in `PATH`.
@@ -39,7 +39,7 @@ pub fn shell() -> &'static OsStr {
     static PATH: Lazy<OsString> = Lazy::new(|| {
         if cfg!(windows) {
             core_dir()
-                .and_then(|p| p.ancestors().nth(3) /* skip mingw64/libexec/git-core */)
+                .and_then(|p| p.ancestors().nth(3)) // Skip something like mingw64/libexec/git-core.
                 .map(|p| p.join("usr").join("bin").join("sh.exe"))
                 .map_or_else(|| OsString::from("sh"), Into::into)
         } else {
@@ -122,7 +122,7 @@ static GIT_CORE_DIR: Lazy<Option<PathBuf>> = Lazy::new(|| {
     }
 
     BString::new(output.stdout)
-        .trim_with(|b| b.is_ascii_whitespace())
+        .strip_suffix(b"\n")?
         .to_path()
         .ok()?
         .to_owned()
--- a/gix-path/tests/path/env.rs
+++ b/gix-path/tests/path/env.rs
@@ -25,6 +25,14 @@ fn installation_config() {
 }
 
 #[test]
+fn core_dir() {
+    assert!(
+        gix_path::env::core_dir().expect("Git is always in PATH").is_dir(),
+        "The core directory is a valid directory"
+    );
+}
+
+#[test]
 fn system_prefix() {
     assert_ne!(
         gix_path::env::system_prefix(),
--- a/gix-path/tests/path/env.rs
+++ b/gix-path/tests/path/env.rs
@@ -27,7 +27,9 @@ fn installation_config() {
 #[test]
 fn core_dir() {
     assert!(
-        gix_path::env::core_dir().expect("Git is always in PATH").is_dir(),
+        gix_path::env::core_dir()
+            .expect("Git is always in PATH when we run tests")
+            .is_dir(),
         "The core directory is a valid directory"
     );
 }
--- a/gix-path/tests/path/env.rs
+++ b/gix-path/tests/path/env.rs
@@ -8,13 +8,11 @@ fn exe_invocation() {
 }
 
 #[test]
-fn login_shell() {
-    // On CI, the $SHELL variable isn't necessarily set. Maybe other ways to get the login shell should be used then.
-    if !gix_testtools::is_ci::cached() {
-        assert!(gix_path::env::login_shell()
-            .expect("There should always be the notion of a shell used by git")
-            .exists());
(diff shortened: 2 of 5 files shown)


gix-path/src/env/mod.rs
```diff
@@ -28,21 +28,25 @@ pub fn installation_config_prefix() -> Option<&'static Path> {
...
+/// Note that the returned path might not be a path on disk.
+pub fn shell() -> &'static OsStr {
+    static PATH: Lazy<Option<OsString>> = Lazy::new(|| {
         if cfg!(windows) {
-            installation_config_prefix()
-                .and_then(|p| p.parent())
-                .map(|p| p.join("usr").join("bin").join("bash.exe"))
+            core_dir()
+                .and_then(|p| p.ancestors().nth(3) /* skip mingw64/libexec/git-core */)
+                .map(|p| p.join("usr").join("bin").join("sh.exe"))
```
EliahKagan, line 43: In addition to this `(git-installation)\usr\bin\sh.exe`, there is also `(git installation)\bin\sh.exe`. I do not know which is better.

The `bin` directory that is not under `usr` has less in it: only `git.exe`, `bash.exe`, and `sh.exe`. I believe `usr`-less `bin` directory's `git.exe` is a shim that delegates to the "real" `(environment prefix)\bin\git.exe` executable, since I have found that its contents are identical to those of `(git installation)\cmd\git.exe`, which is documented as such as shim (https://github.com/git-for-windows/git/pull/2506).

My guess is that the situation is similar, though maybe the advantages and disadvantages that arise from using a shim versus the target executable are different for `sh.exe` and `bash.exe` than for `git.exe`. The reason I guess that the `sh.exe` in the `usr`-less `bin` is a shim is that it is much smaller than the `usr\bin` one:

```text
:\Users\ek\scoop\apps\git\2.47.1> ls bin/sh.exe

    Directory: C:\Users\ek\scoop\apps\git\2.47.1\bin

-a---          11/25/2024  7:28 AM          47496 sh.exe

C:\Users\ek\scoop\apps\git\2.47.1> ls usr/bin/sh.exe

    Directory: C:\Users\ek\scoop\apps\git\2.47.1\usr\bin

-a---          11/25/2024  7:12 AM        2553064 sh.exe
```

I point this out in case you know you wish to use one over the other. I do not know which is better, nor do I have a guess. I've listed this in #1761 so it is not forgotten and to remind me to look into it in the future. This is definitely not a blocker for the PR.

gix-path/src/env/mod.rs
```diff
@@ -102,6 +106,36 @@ pub fn xdg_config(file: &str, env_var: &mut dyn FnMut(&str) -> Option<OsString>)
...
+        cmd.creation_flags(CREATE_NO_WINDOW);
+    }
+    let output = cmd.arg("--exec-path").output().ok()?;
+
+    if !output.status.success() {
+        return None;
+    }
+
+    BString::new(output.stdout)
+        .trim_with(|b| b.is_ascii_whitespace())
```
EliahKagan, line 125: I this this change is sufficient to fix #1760:

```suggestion
        .strip_suffix(b"\n")?
```

As noted there, there may be a benefit to fixing that here, but it is definitely fine to leave if for later. I've put more details in the issue, including the (limited) testing I've done of this change.

gix-path/tests/path/env.rs
```diff
@@ -26,6 +24,14 @@ fn installation_config() {
     );
 }
 
+#[test]
+fn core_dir() {
+    assert!(
+        gix_path::env::core_dir().expect("Git is always in PATH").is_dir(),
```
EliahKagan, line 30: I did not immediately realize what the `expect` message was expressing. Assuming this understanding of mine is correct, it might be a bit clearer, but this is subjective and I am not sure:

```suggestion
        gix_path::env::core_dir().expect("Git is always in PATH when we run tests").is_dir(),
```

src/plumbing/options/mod.rs
```diff
@@ -145,6 +145,7 @@ pub enum Subcommands {
     Corpus(corpus::Platform),
     MergeBase(merge_base::Command),
     Merge(merge::Platform),
+    Env,
```
EliahKagan, line 148: This is very helpful and will make my testing on ARM64 Windows more efficient!

EliahKagan (approved):
This looks great! Nothing need change, though I've suggested a few small things. Most are not mentioned here in this comment, and the things I do mention in this comment are not necessarily any more important than those mentioned in separate comments.

#### Parsing `git --exec-path`

I think it may be reasonable to include a fix for #1760 here, but this PR does not introduce that bug (nor had #1752). One of my review comments has my proposed one-line change fix for it, but it's just as easy to do separately if it would lead to delay or you're not sure it's right. I opened it as a separate issue rather than describing it here because it is definitely not a blocker for this PR.

#### Does Git for Windows really use its `sh.exe` as `/bin/sh`? (Yes.)

I decided to actually verify that Git for Windows is using its `sh.exe` rather than its `bash.exe` to run commands that are documented to run in a shell, where no shebang is involved or even relevant in principle, because it is not running a script file but a command, such as from a variable. I am pretty sure I've verified this before, but I wanted to be certain.

(Its `sh.exe` and `bash.exe` are the same, i.e. its `sh` is `bash`, but they behave subtly differently in some circumstances because, like most POSIX-compatible shells, `bash` has [a POSIX mode](https://www.gnu.org/software/bash/manual/bash.html#Bash-POSIX-Mode) where it is more strictly POSIX compatible that must either be enabled with options or an environment variable or, more typically, by naming it like `sh` rather than like `bash`. So the distinction is meaningful.)

It does use `sh.exe` for this purpose, as I had guessed in https://github.com/GitoxideLabs/gitoxide/pull/1752#discussion_r1909440486 but was not totally certain of. So the approach in this PR is correct. The following shows how I checked, and omits the expected fetch-related error message from Git itself:

```text
parnassus@windows-arm CLANGARM64 ~/repos/gitoxide (main)
$ git -c core.sshCommand='ps | grep -E "^\s*(PID|$$)\b" >&2; :' fetch
      PID    PPID    PGID     WINPID   TTY         UID    STIME COMMAND
     2323       1    2319       9716  cons0     197108 21:50:17 /usr/bin/sh
```

The strange use of `grep` is because the `ps` in the Git Bash environment does not support filtering processes by passing a PID. Further details about that, in case they are somehow of interest, can be seen in [this gist](https://gist.github.com/EliahKagan/f68bc4539cf77d41fedd17ab81ab2741).

#### Future directions

I've opened the research-oriented feature request #1761 about other edge cases and scenarios where something deliberate besides just `/bin/sh` may benefit from being done in situations where `/bin/sh` would ordinarily be used. That overlaps, though only partially, with the goals of #1752 and this PR.

I mention that chiefly because I think it is of interest. To be clear, nothing there has to be done here, nor should be, except the already-checked box that refers to this and is already fully done here.

gix-path/src/env/mod.rs
```diff
@@ -28,21 +28,25 @@ pub fn installation_config_prefix() -> Option<&'static Path> {
...
 }
 
-/// Return the shell that Git would prefer as login shell, the shell to execute Git commands from.
+/// Return the shell that Git would use, the shell to execute commands from.
 ///
-/// On Windows, this is the `bash.exe` bundled with it, and on Unix it's the shell specified by `SHELL`,
-/// or `None` if it is truly unspecified.
-pub fn login_shell() -> Option<&'static Path> {
-    static PATH: Lazy<Option<PathBuf>> = Lazy::new(|| {
+/// On Windows, this is the full path to `sh.exe` bundled with it, and on
```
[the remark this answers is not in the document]
Byron: In this case, it might even be a typo as I try to be explicit and mindful about that as it's also not the first time you had to flag that.

gix-path/src/env/mod.rs
```diff
@@ -28,21 +28,25 @@ pub fn installation_config_prefix() -> Option<&'static Path> {
...
+/// Unix it's `/bin/sh` as posix compatible shell.
+/// If the bundled shell on Windows cannot be found, `sh` is returned as the name of a shell
+/// as it could possibly be found in `PATH`.
```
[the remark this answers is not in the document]
Byron: That's a great point and I have made the change.

gix-path/src/env/mod.rs
Byron: Generally I'd like the `bin/sh` shim more as it's closer to what's used on Linux with `/bin/sh`, but would avoid to actually doing so on Windows as the shim will have to redirect to the actual binary which probably slows things down measurably.

With that said, I didn't find the shim last time I looked, because I tested on a VM that just has the full-blown git SDK for Windows, which doesn't seem to have it.

The normal Git installation does have the shim though, but it's apparently not the most portable assumption to make about the Git installation.

Speaking of portable, I have never tested with the 'portable' version of the Git installer, if there is one.

gix-path/src/env/mod.rs
Byron: I will apply it - the `is_ascii_whitespace` I copied from somewhere and knew it would replace the version that is now re-added with the patch. It's good to know that this is the more correct version.

gix-path/tests/path/env.rs
Byron: There I was very fast and sloppy and think the patch helps to be a bit clearer, without claiming Git is installed on each machine in the world :D.

Byron (commented):
Thanks a lot for the review and the additional research!

I applied all of the suggestions and will catch up on #1760 and #1761 shortly.

EliahKagan:
It looks like `cargo fmt` has to be run due to the greater length from applying https://github.com/GitoxideLabs/gitoxide/pull/1758#discussion_r1912376486, and that this is the cause of the [new CI failure](https://github.com/GitoxideLabs/gitoxide/actions/runs/12731645715/job/35485868142?pr=1758) that is blocking merging.

Scutua (approved):

gix-path/src/env/mod.rs
EliahKagan: > With that said, I didn't find the shim last time I looked, because I tested on a VM that just has the full-blown git SDK for Windows, which doesn't seem to have it.

Thanks--in #1761 I had forgotten about how the Git for Windows SDK is another distinctive environment, which is really a different kind of Git for Windows installation, that should be tested specifically. I've added it to the list in #1761 for further research. It might be something that be treated the same as an upstream MSYS2 with `git` installed in it, but I'm not sure. (For now, the box is not checked, since I'm not clear on how many or what the locations are for `sh.exe` with it.)

> The normal Git installation does have the shim though, but it's apparently not the most portable assumption to make about the Git installation.

It sounds like `(git installation)\bin\sh.exe` shouldn't be preferred, then. I've added a note to #1761 about that too, also pointing here. Assuming the shim should neither be preferred nor used as a fallback, the only thing left to do for that item would be to add a comment to the code briefly explaining the choice, which I could do after testing locally on a system that has the Git for Windows SDK installed. (The absence of a check mark there is only because I haven't added that code comment yet.)

> Speaking of portable, I have never tested with the 'portable' version of the Git installer, if there is one.

Yes, the [downloads page](https://git-scm.com/downloads/win) lists a *Portable ("thumbdrive edition")* download.

Although it's been ages since I've used that directly, it is also used by other packagers, including the [Scoop](https://scoop.sh/) package, which is how I install Git for Windows, except when I am doing it differently to test something else. So I think the portable version is covered already. I've added it to the list in #1761. I put a check mark on it, which can be removed if someone finds I am missing something here.

One subtlety is that Scoop (and other package managers such as Chocolatey) provide their *own* shims. However, we are not attempting to use those shims, and they are in addition to what one gets with a standard full Git for Windows installation. (They point at Git for Windows's own shims.)

To be clear, we very well may be running those shims--in a Scoop installation, the `git.exe` shim is typically what is found in a `PATH` search and called, including in `git --exec-path`. But then the path `git --exec-path` provides as output is a path into the Git for Windows installation itself, and nothing we do with absolute paths based on that hits the Scoop shims anymore.

EliahKagan:
If a command is ever to be run in a shell for the environment the shell confers (rather than for its expansions or other language features), then there is one fragment of the logic for doing so that I think may make sense to have in `gix_command`: robustly quoting the command name when it is a path and quoting is requested. I've proposed this in #1799.

gix-path/src/env/mod.rs
Byron: > Thanks--in #1761 I had forgotten about how the Git for Windows SDK is another distinctive environment, which is really a different kind of Git for Windows installation, that should be tested specifically. I've added it to the list in #1761 for further research. It might be something that be treated the same as an upstream MSYS2 with `git` installed in it, but I'm not sure. (For now, the box is not checked, since I'm not clear on how many or what the locations are for `sh.exe` with it.)

I would have loved to provide that information but just yesterday that VM demolished itself after a Windows update and is inoperable. During the weekend I plan to restore it from backup, so I could help here if still useful.

> One subtlety is that Scoop (and other package managers such as Chocolatey) provide their _own_ shims. However, we are not attempting to use those shims, and they are in addition to what one gets with a standard full Git for Windows installation. (They point at Git for Windows's own shims.)

Maybe the Shim could be a fallback if it turns out to be more portable, i.e. each or most of the Git packages one can use on Windows have it under that particular path, like `<install-dir>/bin/<shim>.exe`. 
A practical solution could be to support the most common installation methods without shim for just slightly better invocation performance, but use the shim for less common packages if it makes the implementation significantly simpler.

> To be clear, we very well may be running those shims--in a Scoop installation, the `git.exe` shim is typically what is found in a `PATH` search and called, including in `git --exec-path`. But then the path `git --exec-path` provides as output is a path into the Git for Windows installation itself, and nothing we do with absolute paths based on that hits the Scoop shims anymore

I agree. Expressing the above differently I could also say that I am happy with finding the actual file locations like we do now, but also think it might be simpler to rely on shims more if these paths are more consistent across the various installation methods.

EliahKagan: > just yesterday that VM demolished itself after a Windows update and is inoperable

That must be annoying--good luck!

> Maybe the Shim could be a fallback *[...]*

Currently, for `git` itself, we almost do the opposite. We take `git` from a `PATH` search if we can find it that way. When there is an applicable shim--whether from a package manager like `scoop` or from Git for Windows itself--that is usually what this finds. Though in some environments we may find the shim target--for example, on my Windows 10 development machine with Git for Windows installed via `scoop`, when I run `git` in PowerShell it is the `scoop` shim, but when I run `git` in Git Bash it is not.

If the `PATH` turned up no `git`, we check common locations, and look in directories where it is commonly installed. The current logic for that does not use the shims--it looks in whichever of the `bin` directories in `mingw64` or `mingw32` is applicable to the particular Git for Windows installation is being examined. (I believe the current behavior is still as it was when #1456 was merged.) It needs to be updated also to look into the `clangarm64` directory when applicable. Git for Windows uses this on ARM64, which has had stable releases for a short while.

I've been meaning, and will eventually manage, to update that. Alternatively, if we use the shim then the paths are ["environment"](https://www.msys2.org/docs/environments/)-independent. One reason I've been slightly reluctant to do that is that there may sometimes be a performance benefit to bypassing the shims, for speed. I've been reluctant to give that up when it might turn out that it should instead be expanded (or kept but controlled by an environment variable--probably not a configuration variable, since we invoke `git` to find out where to look for some of those).

I'm not sure about Chocolatey, and also I haven't recently researched what other package managers besides `choco` and `scoop` are in wide use on Windows. But given the root of a `scoop` installation, we could find out if a `scoop`-managed Git for Windows is installed. We could also find Git for Windows's own directory and, in so doing, bypass all shims. Both operation seem pretty simple in ordinary setups and I believe the only subtleties are those related to edge cases, which I would guess would not pose a big problem.

Whether this makes sense to do probably depends on what it is being done. But it seems to me that it is not justified to do it if the reason is to be more likely to find `git`. I think it would be strange to use `scoop` but not put its `shims` directory in one's `PATH`, and `scoop` automatically adds its `shims` directory to the `PATH` when installed. But then again, some `scoop` users may use it mainly to install graphical programs that they invoke in ways that don't require the Scoop shims to be in their `PATH`, so maybe this is a real use case.

EliahKagan: > Generally I'd like the `bin/sh` shim more as it's closer to what's used on Linux with `/bin/sh`, but would avoid to actually doing so on Windows as the shim will have to redirect to the actual binary which probably slows things down measurably.
> 
> With that said, I didn't find the shim last time I looked, because I tested on a VM that just has the full-blown git SDK for Windows, which doesn't seem to have it.

I have checked a current Git for Windows SDK on Windows 10 x86-64 and this confirms your findings. Where `(root)` is the SDK installation root--on my system, I allowed the default `C:\git-sdk-64`--there no `(root)\bin\sh.exe`, since there is no `(root)\bin`, but `(root)\usr\bin\sh.exe` still exists.

Since `(root)\usr\bin\sh.exe` also exists in regular Git for Windows installations (i.e. full non-SDK installations created by the installer or by unpacking a PortableGit archive), and `(root)` can be found on both the same way three directories above the directory given by `git --exec-path`, we should indeed continue to use `(root)\usr\bin\sh.exe`. If it turns out that the `(root)\bin\sh.exe` shim gives a better environment, then it could be preferred if present with `(root)\usr\bin\sh.exe` as a fallback, but I have not yet found that to be necessary.

A similar situation applies to `bash.exe`, which we currently look for in the nonexistent `(root)\bin` for `gix-testtools`. So this should probably be changed to `(root)\usr\bin\bash.exe`, so long as that causes no problems. I already have a feature branch where it makes sense to include that change, and I expect it to make it into one of the PRs I open soon.

Byron: Thank you, I am looking forward to it.

EliahKagan: Interestingly, that *does* look like it causes problems. (This is alluded to in https://github.com/GitoxideLabs/gitoxide/pull/1862#issuecomment-2679599404 but it is on a separate feature branch from the PR. That PR modifies `gix-path` and `gix-command`, while this is a change in `gix-testtools` that I am testing on a separate feature branch with other such changes, which I may open a PR for soon, but it would likely be another draft PR until the problems are ironed out.)

Byron: Just as a side-note: I only loosely follow what's going on, as the shell and shebang handling across platforms, maybe particularly on Windows, is intricate. Due to mw not running Windows often it's also not very present in memory. Thus I am probably not as helpful as I'd hope in this matter.

PR #1764, opened 2025-01-12
Refine how `finalize_entry` sets executable bits

EliahKagan:
This adjusts how executable bits that are added subsequent to file creation, as occurs when `destination_is_initially_empty: true` is not included in checkout options, are computed and added. The idea is to produce an effect more similar to what users expect and may wish to rely on, as well as to make things somewhat closer to what Git does in this area while still keeping the fundamental design the same.

Further changes, which may affect the design more or in other ways end up being trickier, may still be worthwhile; I do not claim that the specific approach here will ultimately be the best one.

### DIFF
--- a/gix-worktree-state/src/checkout/entry.rs
+++ b/gix-worktree-state/src/checkout/entry.rs
@@ -1,5 +1,5 @@
-use std::borrow::Cow;
 use std::{
+    borrow::Cow,
     fs::OpenOptions,
     io::Write,
     path::{Path, PathBuf},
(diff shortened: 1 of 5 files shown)


gix-worktree-state/src/checkout/entry.rs
```diff
@@ -286,14 +286,114 @@ pub(crate) fn finalize_entry(
...
+mod tests {
+    fn pretty(maybe_mode: Option<u32>) -> String {
+        match maybe_mode {
+            Some(mode) => format!("Some({mode:04o})"),
+            None => "None".into(),
+        }
+    }
+
+    #[test]
+    fn set_mode_executable() {
```
Byron, line 324: A perfect test!

Byron (approved):
Thanks so much for this incredibly thorough PR. The thoughtful and exhaustive tests are a highlight as well.

Byron:
I will try to release this in the coming hours, hoping that I can land a feature first.

EliahKagan:
I just did a few things that are worth noting somewhere:

- Updated the PR description here to link to https://github.com/GitoxideLabs/gitoxide/security/advisories/GHSA-fqmf-w4xh-33rh.
- Fixed a minor whitespace nit in https://github.com/GitoxideLabs/gitoxide/security/advisories/GHSA-fqmf-w4xh-33rh.
- Opened [rustsec/advisory-db#2193](https://github.com/rustsec/advisory-db/pull/2193) to create the corresponding RUSTSEC advisory.

PR #1765, opened 2025-01-13
Avoid another "unused import" warning on Windows

EliahKagan:
There were two places in #1764 where I had an unconditional import that causes a warning on Windows about how it is unused. 4d5e656 fixed one. This fixes the other:

    warning: unused import: `Permissions`
     --> gix-worktree-state\src\checkout\entry.rs:3:23
      |
    3 |     fs::{OpenOptions, Permissions},
      |                       ^^^^^^^^^^^


Byron (approved):
Thanks a lot!

PR #1766, opened 2025-01-13
Add `gix blame -L start,end`

cruessler:
This enables running blame for a portion of a file. This feature will be helful for debugging `gix blame`, in particular in the context of #1743.

### DIFF
--- a/gix-blame/src/error.rs
+++ b/gix-blame/src/error.rs
@@ -27,4 +27,6 @@ pub enum Error {
     Traverse(#[source] Box<dyn std::error::Error + Send + Sync>),
     #[error(transparent)]
     DiffTree(#[from] gix_diff::tree::Error),
+    #[error("Invalid line range was given, line range is expected to be a 1-based inclusive range in the format '<start>,<end>'")]
+    InvalidLineRange,
 }
--- a/gix-blame/src/file/function.rs
+++ b/gix-blame/src/file/function.rs
@@ -25,6 +25,10 @@ use std::ops::Range;
 ///    - The first commit returned here is the first eligible commit to be responsible for parts of `file_path`.
 /// * `file_path`
 ///    - A *slash-separated* worktree-relative path to the file to blame.
+/// * `range`
+///    - A 1-based inclusive range, in order to mirror `git`’s behaviour. `Some(20..40)` represents
+///      21 lines, spanning from line 20 up to and including line 40. This will be converted to
+///      `19..40` internally as the algorithm uses 0-based ranges that are exclusive at the end.
 /// * `resource_cache`
 ///    - Used for diffing trees.
 ///
@@ -61,6 +65,7 @@ pub fn file<E>(
     traverse: impl IntoIterator<Item = Result<gix_traverse::commit::Info, E>>,
     resource_cache: &mut gix_diff::blob::Platform,
     file_path: &BStr,
+    range: Option<Range<u32>>,
 ) -> Result<Outcome, Error>
 where
     E: Into<Box<dyn std::error::Error + Send + Sync + 'static>>,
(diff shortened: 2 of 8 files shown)


gix-blame/src/file/function.rs
```diff
@@ -61,6 +61,7 @@ pub fn file<E>(
     traverse: impl IntoIterator<Item = Result<gix_traverse::commit::Info, E>>,
     resource_cache: &mut gix_diff::blob::Platform,
     file_path: &BStr,
+    range: Option<Range<u32>>,
```
Byron, line 64: `range` would need to be documented. Given how 'special' it is no details should be spared :).

gix-blame/src/error.rs
```diff
@@ -27,4 +27,6 @@ pub enum Error {
     Traverse(#[source] Box<dyn std::error::Error + Send + Sync>),
     #[error(transparent)]
     DiffTree(#[from] gix_diff::tree::Error),
+    #[error("Invalid line range was given")]
```
Byron, line 30: This could say how the range is supposed to look like. I.e. 1-based inclusive.

gix-blame/src/file/function.rs
```diff
@@ -85,19 +86,34 @@ where
...
         return Ok(Outcome::default());
     }
 
-    let mut hunks_to_blame = vec![{
-        let range_in_blamed_file = 0..num_lines_in_blamed as u32;
-        UnblamedHunk {
-            range_in_blamed_file: range_in_blamed_file.clone(),
-            suspects: [(suspect, range_in_blamed_file)].into(),
+    let range_in_blamed_file = {
+        // This block assumes that `range` has 1-based line numbers and converts it the the format
```
Byron, line 97: I could imagine a function here like `one_based_inclusive_to_zero_based_exclusive_range()` to 'codify' the comment.

src/plumbing/options/mod.rs
```diff
@@ -162,6 +164,9 @@ pub enum Subcommands {
         statistics: bool,
         /// The file to create the blame information for.
         file: std::ffi::OsString,
+        /// Only blame lines in the given range.
```
Byron, line 167: I'd mention the '1-based inclusive' part and provide an example.
Git does it, too: 

```
   -L <range>            process only line range <start>,<end> or function :<funcname>
```

src/shared.rs
```diff
@@ -441,4 +467,16 @@ mod value_parser_tests {
         let c = Cmd::parse_from(["cmd", "-a=75"]);
         assert_eq!(c.arg, Some(Some(0.75)));
     }
+
+    #[test]
+    fn range() {
+        #[derive(Debug, clap::Parser)]
+        pub struct Cmd {
```
Byron, line 474: That's really neat!

Byron (changes requested):
Thanks a lot for making `blame` even better!

Besides my comments which are more like nits to let it appear less rushed, one change I really have to ask for: Changes to `gix-blame` need to go into their own commit, and be prefixed with `feat!: ` to indicate a breaking change. A single test there probably would be a good idea (or necessity) as well.

The changes to the other two crates can be bundled as they logically are one (`core` + `gitoxide`).

Thanks a lot!

cruessler:
> Thanks a lot for making `blame` even better!
> 
> Besides my comments which are more like nits to let it appear less rushed, one change I really have to ask for: Changes to `gix-blame` need to go into their own commit, and be prefixed with `feat!: ` to indicate a breaking change. A single test there probably would be a good idea (or necessity) as well.
> 
> The changes to the other two crates can be bundled as they logically are one (`core` + `gitoxide`).
> 
> Thanks a lot!

Sure, no worries, and thank for letting me know! I’m not fully familiar with all the conventions yet, so always grateful to learn them. :-)

gix-blame/src/file/function.rs
cruessler: So many helpful comments! :-) I addressed all of them.

Byron:
Thanks a lot for the fixes!

Additionally, I have modified the second commits subject to not contain a `!` as it's not a breaking change, just a new feature.

Byron (approved):

PR #1769, opened 2025-01-15
various improvements

Byron:
Various improvements to help with  https://github.com/gitbutlerapp/gitbutler/pull/5945


EliahKagan (commented):
This unassuming PR is a substantial boon to usability. When using `gix status` newly created minimal repositories to investigate bugs, I would previously have to make a commit in some situations, so that `gix status` would not give an error. With this change, that is no longer the case.

There is one situation where it seems this is not working properly, however. I've opened #1770 about the case of files that would be shown to be in the `A` state if `HEAD` existed.

Byron:
Thank you!

I completely missed this case, but better late than never!
Also I think I know what the problem with `A` files is, it should be an easy fix.

PR #1774, opened 2025-01-17
Use `parse_spec_no_baseline` with `:/` for all 2.47.* on CI

EliahKagan:
The bug in Git that causes #1622 has been fixed since 2.48.0 and does not affect any versions prior to 2.47, but the fix is not backported to subsequent 2.47.* point releases. In particular, 2.47.2 has been released, with backported security fixes, but it does not have a backported fix for this non-security bug.

This builds on #1635 and #1719 by updating the range of Git versions where we skip the affected baseline checks on CI (for platforms this affects on our CI) from `(2, 47, 0)..(2, 47, 2)` to `(2, 47, 0)..(2, 48, 0)`, i.e., with the exclusive upper bound changed from 2.47.2 to the correct value of 2.48.0. This also revises the comments accordingly.

For further details, see https://github.com/GitoxideLabs/gitoxide/issues/1622#issuecomment-2599247175. This change is item (1) there.

### DIFF
--- a/gix/tests/gix/revision/spec/from_bytes/regex.rs
+++ b/gix/tests/gix/revision/spec/from_bytes/regex.rs
@@ -96,16 +96,17 @@ mod find_youngest_matching_commit {
         let repo = repo("complex_graph").unwrap();
 
         // The full Linux CI `test` job regenerates baselines instead of taking them from archives.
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot for taking such good care of this and following up. The sequence of actions in https://github.com/GitoxideLabs/gitoxide/issues/1622#issuecomment-2599247175 seems reasonable as well.

PR #1775, opened 2025-01-18
use NotADirectory io error kind where appropriate

dvtkrlbs:


dvtkrlbs:
hmm interestingly enough locally msrv check passed or maybe I ran it wrong

dvtkrlbs:
oh yeah my bad the actual msrv is defined at the workflow I missed that

Byron:

As it looks like it will be a while until `gitoxide` arrives at Rust v1.83, I'd like to close this PR in the hopes I will also remember once the MSRV here reaches 1.83 to bring it back.

PR #1777, opened 2025-01-18
Test GNU/Linux ARM builds on the new ARM CI runner

EliahKagan:
While ARM64 runners for macOS have been available, and not limited to enterprise or paid customers, for a while now, free Linux ARM runners for public projects were not likewise available. [They are now available](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/) in public preview, as of a couple days ago. This PR extends and modifies `cy.yml` to makes use of them, by:

- Adding an `ubuntu-24.04-arm` job to `test-fast`.
- Making `test-32bit` a matrix job definition, and adding a 32-bit ARMv7 job to it. The 32-bit x86 job uses a container on an x86-64 runner, just as before, while the new 32-bit ARM job uses a container on an ARM64 runner. (Although GitHub Actions supports 32-bit ARM runners in the sense that the runner software can run on such a system, there are no GitHub-hosted 32-bit ARM runners..)
- Removing the `test-32bit-cross` job, which was not able to run many tests; it was only testing `gix-hashtable`. Its value seems less now that `test-32bit` covers ARMv7. However, it may be worth reviewing the specific target, in case it is a goal to continue running some tests with a `androideabi` target. (If so, it should be possible to add that as a target to a non-cross 64-bit or 32-bit environment.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -204,9 +204,11 @@ jobs:
         container-arch: [ i386, arm32v7 ]
         include:
           - container-arch: i386
+            runner-arch: amd64
             runner-os: ubuntu-latest
             toolchain: stable-i686-unknown-linux-gnu
           - container-arch: arm32v7
+            runner-arch: arm64
             runner-os: ubuntu-24.04-arm
             toolchain: stable-armv7-unknown-linux-gnueabihf
       fail-fast: false  # FIXME: Probably remove this.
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot for this massive simplification *and* improvement to the CI setup. I wasn't aware that ARM runners are now available for Linux as well.

The `androideabi` target probably isn't the most important one particularly given how limited the tests were, and it seems best to wait until there is renewed demand before testing it on CI again.

EliahKagan:
I've opened #1780 to track the problem described in https://github.com/GitoxideLabs/gitoxide/pull/1778#issuecomment-2599773835 and give what information I know about it. I had seen something like that here when working on this, but had thought or hoped it had gone away.

PR #1778, opened 2025-01-18
the next release

Byron:
Just so that the auto-push will succeed when running `cargo smart-release`


Byron:
@EliahKagan It seems that the new Arm jobs *might* be more error prone. One just failed during initialization/startup. If this keeps happening, we might consider making them optional.

EliahKagan:
> It seems that the new Arm jobs _might_ be more error prone. One just failed during initialization/startup. If this keeps happening, we might consider making them optional.

I've opened #1780 to track this.

PR #1792, opened 2025-01-22
Split out ARM64 test-fast from matrix, make nonrequired

EliahKagan:
As suggested in https://github.com/GitoxideLabs/gitoxide/issues/1790#issuecomment-2606481883.

It likely won't have to be kept this way. But making it nonrequired for now makes it so that investigating what triggers the SIGSEGV (and SIGBUS) errors -- as well as other errors that were found while investigating that (d9e7fdb, e71b0cf, 5a71963) -- doesn't have to be rushed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,7 +135,6 @@ jobs:
           - windows-latest
           - macos-latest
           - ubuntu-latest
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot, much appreciated 🙏.

PR #1794, opened 2025-01-22
Fix shortcoming regarding time crate

KGrewal1:
Very minor change: was looking for easy to fix issues and saw this in shortcomings, but seems that this crate has already transitioned over to using `jiff` instead of `time` and that the linked files are no longer in the same place as they were in the issue (alternatively if I've misunderstood the changes since, I'd be happy to add the local time function as `jiff` does not have the segfault issue `time` had)


Byron:
That's a great catch, thank you!

It turns out that local time functions already exist on the `Time` type, so there is indeed nothing to do here anymore.

Issue #1799, opened 2025-01-23
`sh` quoting of program names on an opt-in basis


EliahKagan:
### Summary 💡

Possible tasks:

- [x] Document that `gix_command::Prepare` with `use_shell` does not quote (unless it will) ([#1800](https://github.com/GitoxideLabs/gitoxide/pull/1800))
- [x] Document which data in a `Prepare` instance are subject to shell expansions if unquoted ([#1800](https://github.com/GitoxideLabs/gitoxide/pull/1800))
- [ ] Clarify the semantics of [`args`](https://github.com/GitoxideLabs/gitoxide/blob/97f85f78f4026203dca2b62ffe2174e2f028941e/gix-command/src/lib.rs#L25-L26) ([#1813](https://github.com/GitoxideLabs/gitoxide/pull/1813))
- [x] *Maybe:* When `use_shell` is `true`, let caller opt into POSIX shell style quoting ([#1801](https://github.com/GitoxideLabs/gitoxide/pull/1801))

It might make sense to split this into multiple issues, in that the first three items could be done by themselves, though they become more important and possibly also different if fourth is done.

#### When using a shell, the *how* varies based on the *why*

By default, `gix_command::Prepare` rightly does not use a shell. But it supports using a shell if requested. This shell is generally `sh` and assumed to have POSIX shell semantics. There are at least three reasons one might want to use a shell:

1. To run a command in the same way `git` would run it, where `git` is guaranteed to use a shell.
2. To run a command that is expressed as a single string, with arguments split how `sh` splits them.
3. To get a different environment that a shell is hoped to confer.

#### Reasons (1) and (2) are the most important

In the absence of (1), a shell must be used very carefully, and basically must never be used for running a command `git` documents not to use a shell or that can otherwise be assumed not to be run with a shell when `git` runs it. For example, if a command that should not run in a shell has a `\` in its name due that being a path separator on Windows, and it is run in a shell anyway, then that will turn into an escape character.

When (2) is the only reason to use a shell, and neither (1) nor (3) apply, or they apply but can be "optimized away," then shell-like argument splitting can be done without using a shell. `Prepare` currently does this on Windows, if it is it is allowed but not required to use a shell, i.e. if `use_shell` and `allow_manual_arg_splitting` were both set to `true`.

[...]

PR #1800, opened 2025-01-23
various improvements

Byron:
To support https://github.com/gitbutlerapp/gitbutler/pull/6060 .

Maybe related to documentation tasks in #1799 .

### DIFF
--- a/gix-command/src/lib.rs
+++ b/gix-command/src/lib.rs
@@ -93,19 +93,33 @@ mod prepare {
 
     /// Builder
     impl Prepare {
-        /// If called, the command will not be executed directly, but with `sh`, but only if the
-        /// command passed to [`prepare`](super::prepare()) requires this.
+        /// If called, the command will be checked for characters that are typical for shell scripts, and if found will use `sh` to execute it
+        /// or whatever is set as [`with_shell_program()`](Self::with_shell_program()).
+        /// If the command isn't valid UTF-8, a shell will always be used.
         ///
-        /// This also allows to pass shell scripts as command, or use commands that contain arguments which are subsequently
-        /// parsed by `sh`.
-        pub fn with_shell(mut self) -> Self {
+        /// If a shell is used, then arguments given here with [arg()](Self::arg) or [args()](Self::args) will be substituted via `"$@"` if it's not
+        /// already present in the command.
+        ///
+        /// If this method is not called, commands are always executed verbatim, without the use of a shell.
+        pub fn command_may_be_shell_script(mut self) -> Self {
             self.use_shell = self.command.to_str().map_or(true, |cmd| {
                 cmd.as_bytes().find_byteset(b"|&;<>()$`\\\"' \t\n*?[#~=%").is_some()
             });
             self
         }
 
-        /// Set the name or path to the shell `program` to use, to avoid using the default shell which is `sh`.
+        /// If called, unconditionally use a shell to execute the command and its arguments, and `sh` to execute it,
+        /// or whatever is set as [`with_shell_program()`](Self::with_shell_program()).
+        ///
+        /// If a shell is used, then arguments given here with [arg()](Self::arg) or [args()](Self::args) will be substituted via `"$@"` if it's not
+        /// already present in the command.
+        ///
+        /// If this method is not called, commands are always executed verbatim, without the use of a shell.
+        pub fn with_shell(mut self) -> Self {
+            self.use_shell = true;
+            self
+        }
+        /// Set the name or path to the shell `program` to use if a shell is to be used, to avoid using the default shell which is `sh`.
         pub fn with_shell_program(mut self, program: impl Into<OsString>) -> Self {
             self.shell_program = Some(program.into());
             self
@@ -113,7 +127,7 @@ mod prepare {
 
         /// Unconditionally turn off using the shell when spawning the command.
         /// Note that not using the shell is the default so an effective use of this method
-        /// is some time after [`with_shell()`][Prepare::with_shell()] was called.
+        /// is some time after [`command_may_be_shell_script()`](Prepare::command_may_be_shell_script()) was called.
         pub fn without_shell(mut self) -> Self {
             self.use_shell = false;
             self
@@ -128,18 +142,20 @@ mod prepare {
             self
         }
 
-        /// Use a shell, but try to split arguments by hand if this can be safely done without a shell.
+        /// Like [`command_may_be_shell_script()`](Prepare::command_may_be_shell_script()), but try to split arguments by hand if this can be safely done without a shell.
         ///
-        /// If that's not the case, use a shell instead.
-        pub fn with_shell_allow_manual_argument_splitting(mut self) -> Self {
+        /// This is useful on platforms where spawning processes is slow, or where many processes have to be spawned in a raw which should be sped up.
+        /// Manual argument splitting is enabled by default on Windows only.
+        pub fn command_may_be_shell_script_allow_manual_argument_splitting(mut self) -> Self {
             self.allow_manual_arg_splitting = true;
-            self.with_shell()
+            self.command_may_be_shell_script()
         }
 
-        /// Use a shell, but prohibit splitting arguments by hand even if this could be safely done without a shell.
-        pub fn with_shell_disallow_manual_argument_splitting(mut self) -> Self {
+        /// Like [`command_may_be_shell_script()`](Prepare::command_may_be_shell_script()), but don't allow to bypass the shell even if manual argument splitting
+        /// can be performed safely.
+        pub fn command_may_be_shell_script_disallow_manual_argument_splitting(mut self) -> Self {
             self.allow_manual_arg_splitting = false;
-            self.with_shell()
+            self.command_may_be_shell_script()
         }
 
         /// Configure the process to use `stdio` for _stdin.
--- a/gix-command/tests/command.rs
+++ b/gix-command/tests/command.rs
@@ -246,7 +246,8 @@ mod prepare {
     #[test]
     fn multiple_arguments_in_one_line_with_auto_split() {
         let cmd = std::process::Command::from(
-            gix_command::prepare("echo first second third").with_shell_allow_manual_argument_splitting(),
+            gix_command::prepare("echo first second third")
+                .command_may_be_shell_script_allow_manual_argument_splitting(),
         );
         assert_eq!(
             format!("{cmd:?}"),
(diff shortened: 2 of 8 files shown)


gix-command/src/lib.rs
```diff
@@ -93,27 +93,41 @@ mod prepare {
...
 
-        /// Set the name or path to the shell `program` to use, to avoid using the default shell which is `sh`.
+        /// If called, unconditionally use a shell to execute the command and its arguments, and `sh` to execute it,
+        /// or whatever is set as [`with_shell_program()`](Self::with_shell_program()).
+        ///
+        /// If a shell is used, then arguments given here with [arg()](Self::arg) or [args()](Self::args) will be substituted via `"$@"` if it's not
+        /// already present in the command.
+        ///
+        /// If this method is not called, commands are always executed verbatim, without the use of a shell.
+        pub fn with_shell(mut self) -> Self {
```
EliahKagan, line 118: Will this cause code targeting the previous API and calling `with_shell` without actually requiring a shell to be used to continue to build and appear to work, but to work differently by using a shell in circumstances where it wasn't before? If so, I wonder if that might introduce bugs, or worsen existing bugs, in code that uses `gix-command`.

EliahKagan (commented):
The documentation changes look great and, as you alluded to, possibly fix the first three tasks listed in #1799.

If I understand correctly, `with_shell` has a new meaning now, subtly but importantly different from its meaning before, and it can still be called in the same way. I have a worry about that, expressed below, but I haven't thought of a use case that I specifically expect to be broken.

gix-command/src/lib.rs
Byron: I was aware of this and probably didn't even mark it as `fix:` which is what I thought about when doing so.
`with_shell()` now actually does what it says, as opposed to before when it surprisingly only used a shell when the command seemed to need it. Especially on Windows this matters so I really thought of it as a fix for that.

EliahKagan: Yes, it is true that it could fix bugs where people assumed `use_shell` had a more intuitive meaning. An alternative could be to abandon the method name `use_shell` altogether and call all the methods something different. That would break anything that should be broken. It might also break things that need not break.

On the other hand, I recognize it's possible to take that too far and accumulate lots of names that one does use in the future even when they would be the best description available. So I don't know what's best.

---

In this project itself, I think the changed methods may be useful for future bug investigation. I suspect we have some invocations that treat the command name as a shell script when, to match the expectations set by Git's behavior and the guarantees in Git's documentation, it would be necessary to treat the command as a shell script when it was obtained from some sources, while being necessary to treat it *not* as a shell script (i.e. just as a path) when it was obtained from other sources.

The specific case of this that I most strongly suspect the SSH client. When obtained from the environment `GIT_SSH`, it must not be run or parsed as a shell script (regardless of its value). When obtained from `GIT_SSH_COMMAND`, or the value of the [`core.sshCommand`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coresshCommand) configuration variable, it is a shell script and must either be run as one or (in sufficiently simple cases) parsed as one. These requirements are [explicated in git(1)](https://git-scm.com/docs/git#Documentation/git.txt-codeGITSSHcode).

I have not checked recently--and maybe not at all, in terms of running experiments to verify it--that we are doing the wrong thing here. When I have a chance to test it, I'll open an issue if we are. I recall it having seemed so when I was working on #1342, and that it seemed like a contributing factor to the bug reported in #1432 and discussed further in connection with some related topics in https://github.com/GitoxideLabs/gitoxide/pull/1585#discussion_r1759942829. This was the suspected aspect of #1432  highlighted in https://github.com/GitoxideLabs/gitoxide/issues/1432#issuecomment-2351419019.

If we don't yet have an issue for it, then I'll try to open one once I get around to checking the effect in a way that facilitates direct comparison to Git behavior. Some invocations of the SSH client occur only under some circumstances; for example, it is only invoked to figure out what kind of SSH client it is if that cannot be inferred from its name (and is not overridden by configuration), and I think that is one of the places where `GIT_SSH` may be being wrongly treated like `GIT_SSH_COMMAND`.

Byron: Thanks for bringing this up!

I was aware of the difference between `GIT_SSH` and `GIT_SSH_COMMAND` at the time and *should* have implemented it correctly. Due to me renaming `with_shell()` to its new name, in this codebase the semantics are preserved, too, so I'd think there is no issue, still.

However, there I support and value an investigation to assure there aren't any subtle bugs around that in the hopes that this also helps the API of `gix-command` to mature. After all, it's vital to make it as bullet-proof and unambiguous to use as possible.

EliahKagan: > Due to me renaming `with_shell()` to its new name, in this codebase the semantics are preserved, too, so I'd think there is no issue, still.

I very much agree that this PR introduces no *new* incorrect behavior--related to the distinction between `GIT_SSH` and `GIT_SSH_COMMAND`, or otherwise. But I think there is probably a preexisting issue, and if so, it remains intact. (One of the ways this PR is significant to that is that it calls attention to such possible things, which is a good thing about this PR, not a bad thing.)

On Windows, `\` is the primary directory separator, so it often appears in paths that are otherwise unremarkable, like `C:\Windows\System32\OpenSSH\ssh.exe`. But `\` is also taken to mean that we must run a command in a shell if we are allowed to do so (since `\` is a shell metacharacter). Specifically:

- The method now called `command_may_be_shell_script` is meant to conditionally refrain from planning to run unremarkable paths using a shell, but the presence of `\` causes it to go ahead and plan that by setting the `use_shell` field to `true`:

  https://github.com/GitoxideLabs/gitoxide/blob/f3dc83bbbadf0ff7fb2675b4d1299307c9aee63a/gix-command/src/lib.rs#L108-L110

- When the `use_shell` field is set to `true`, using a shell is planned, but it may still be averted if the `allow_manual_arg_splitting` field is set to `true`, which is its default value on Windows. However, `\` is also one of the characters that suppresses the effect of `allow_manual_arg_splitting`:

  https://github.com/GitoxideLabs/gitoxide/blob/f3dc83bbbadf0ff7fb2675b4d1299307c9aee63a/gix-command/src/lib.rs#L222-L234

  Where, matching on `split_args`:

  https://github.com/GitoxideLabs/gitoxide/blob/f3dc83bbbadf0ff7fb2675b4d1299307c9aee63a/gix-command/src/lib.rs#L243-L244

Therefore, if there are places where the function now called `command_may_be_shell_script` is called when it shouldn't be called, on Windows this will often have the effect of running a command in a shell that should not be run that way--this effect is *more likely* on Windows, due to `\` being a common character in paths.

To be clear, this *behavior* is not in any way exacerbated by the changes in this PR. My thinking was that, as far as this sort of thing goes, the changes here are a boon, because they make it easier to reason about what is going on.

However, I just now realize that there is actually a sense in which the changes in this PR exacerbate this. Before, because the method had the name `with_shell`, it conveyed the impression that a shell would be used in more circumstances than it was intended to convey. But as detailed above, it seems that a shell was also used, on Windows, in more circumstances than it was intended to be used. These partially "cancelled out": the impression it gave was, on Windows, closer to the truth than it probably is now.

I don't suggest that as a reason to undo anything from this PR. Instead, the specific effect of `\` in a command on those two methods on Windows, in light of the special significance of `\` on Windows, can modified, documented, or both; and any code in this project that calls `command_may_be_shell_script` when it shouldn't can be identified and fixed.

> I was aware of the difference between `GIT_SSH` and `GIT_SSH_COMMAND` at the time and should have implemented it correctly.

As far as I can tell, in the *main code path* used to decide how to invoke an SSH client to actually connect to a remote, this is implemented perfectly. In `gix::repository::config::ssh_connect_options`, the `fallback_active` flag keeps track of where the SSH client command is configured from:

https://github.com/GitoxideLabs/gitoxide/blob/f3dc83bbbadf0ff7fb2675b4d1299307c9aee63a/gix/src/repository/config/mod.rs#L63-L70

This is used to decide, I believe correctly, whether to disallow the use of a shell:

https://github.com/GitoxideLabs/gitoxide/blob/f3dc83bbbadf0ff7fb2675b4d1299307c9aee63a/gix/src/repository/config/mod.rs#L71-L79

These options are ultimately passed to `gix_transport::client::blocking_io::ssh::connect`. In most cases, this `connect` function arranges a scenario in which the SSH client will run once. This happens through the last thing this `connect` function does, which to unconditionally call `gix_transport::client::blocking_io::file::SpawnProcessOnDemand::new_ssh`. That call relates to actually using the SSH client to make a connection to a server and transfer data. It properly passes along the `disallow_shell` flag:

https://github.com/GitoxideLabs/gitoxide/blob/f3dc83bbbadf0ff7fb2675b4d1299307c9aee63a/gix-transport/src/client/blocking_io/ssh/mod.rs#L136-L144

However, before this `connect` function does that, it figures out *how* the SSH client will be invoked to do that. It checks if this is configured, and otherwise tries to ascertain what SSH implementation it is or behaves like. Under some conditions, it runs the SSH client in order to do this. I am unclear at this time on how often or exactly under what circumstances that happens. The logic is:

https://github.com/GitoxideLabs/gitoxide/blob/f3dc83bbbadf0ff7fb2675b4d1299307c9aee63a/gix-transport/src/client/blocking_io/ssh/mod.rs#L112-L113

But when it does it, it runs a command that it sets up like this:

https://github.com/GitoxideLabs/gitoxide/blob/f3dc83bbbadf0ff7fb2675b4d1299307c9aee63a/gix-transport/src/client/blocking_io/ssh/mod.rs#L115-L125

That code unconditionally calls the `command_may_be_shell_script` method, and does not subsequently do anything to reverse its effects.

(In #1342, I modified part of this, but not the part that is relevant to the use of a shell. I wondered why it was calling what was, back then, named `with_shell`, but this was not part of what was necessary to modify to achieve the goals of that PR.)

Consequently, when `gix-transport` pre-runs an SSH client to figure out what kind of SSH client it is, it seems to me that it runs it in a way that allows a shell to be used and, combined with the aforementioned effect of `\` suppressing the conditional shell-avoidance behaviors of both `command_may_be_shell_script` and `allow_manual_arg_splitting`, this will often run it in a (Bourne-style) shell on Windows, where it will typically fail or otherwise malfunction because the shell treats `\` as the escape character.

(Separately from that, I am unclear on why `gix_transport::client::blocking_io::file::SpawnProcessOnDemand::new_local` always passes a `false` value of `ssh_disallow_shell`. But since, if I understand correctly, that is not using SSH, my guess is that it is okay.)

Byron: Thanks so much for sharing! I see now how this PR made it easier to realize that on Windows, the shell would actually run more often than it would on Unix.

> (Separately from that, I am unclear on why `gix_transport::client::blocking_io::file::SpawnProcessOnDemand::new_local` always passes a `false` value of `ssh_disallow_shell`. But since, if I understand correctly, that is not using SSH, my guess is that it is okay.)

Memory does not serve, unfortunately, but given the importance of not messing this up (i.e. URLs could be crafted to use the shell to execute arbitrary commands) I'd hope that was intentional. However, I'd think it's better to document why this is `false` given that shells ideally are avoided entirely.

PR #1803, opened 2025-01-24
When adding +x, get and set mode through the file descriptor

### LINKED ISSUE #1786: `finalize_entry` uses path and FD that may diverge, could use just FD
### Current behavior 😯

When finishing checking out a file, we do this:

https://github.com/GitoxideLabs/gitoxide/blob/8df0db2f8fe1832a5efd86d6aba6fb12c4c855de/gix-worktree-state/src/checkout/entry.rs#L285-L296

This has both the path and an open file object, the latter of which on a Unix-like system has an associated open file descriptor. The path is used to set permissions on the file when it is being made executable (and when this was not done eagerly when creating the file anew). But the `File` object is used to access metadata and update the index. [The metadata it accesses](https://github.com/GitoxideLabs/gitoxide/blob/8df0db2f8fe1832a5efd86d6aba6fb12c4c855de/gix-index/src/entry/stat.rs#L94-L110) does not include the permissions, so this does not inherently produce any inconsistencies (and the commented note is not likely related to permissions). But this combination is nonetheless unintuitive, especially if the file is moved or replaced:

- We set permissions on a path that may, in rare cases, be the path to a different file that has replaced this one. See be0abaf ([#1764](https://github.com/GitoxideLabs/gitoxide/pull/1764)) and d674a43 ([here](https://github.com/GitoxideLabs/gitoxide/pull/1779/commits/d674a43c3c5ef81c0cc1ae59339edd505742d62d#diff-d0b4461d10af9456046d997e62ea57d71dedcefd3138a0dca55a5d1c0f10f945R54), [#1779](https://github.com/GitoxideLabs/gitoxide/pull/1779)).
- The index gets information from the open file descriptor, which if I understand correctly applies to the original open file even it is no longer where it was originally and even if something else is there now.

As noted in #1783 ("Expected behavior", *5(iii)*), we should be able to set the permissions through the open file descriptor (from the `File` object) rather than using the path:

> File permissions on a Unix-like system can be set through an open file descriptor, without using the path. (If an OS is not Unix-like then we are not setting `+x` anyway.) POSIX standardizes [`fchmod`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/fchmod.html) for this purpose.
>
> I believe `gix_index::fs`, used there, already depends on `rustix` where `cfg!(unix)`. One way to call `fchmod` is through [`rustix::fs::fchmod`](https://docs.rs/rustix/latest/rustix/fs/fn.fchmod.html). (It does not support file descriptors opened with the `O_PATH` flag, but I believe that is inapplicable here, because that's only useful when the file contents are not to be accessed in any way, whereas in contrast the file descriptor associated with our `File` object represents a file that is open in the full traditional sense and that has, presumably, just been written to.)

### Expected behavior 🤔

I am not sure what the expected behavior is, because I don't know if there is a specific reason for timestamp and size metadata to be read through the file descriptor that was presumably used for writing, while using the original path to set permissions instead of using `fchmod`.

- If this distinction supports valuable semantics, then it should be kept, and the only bug is that nothing explains that this is intentional. A comment would fix it. Depending on the reason and how important it is, a documentation comment on a function might be warranted.
- If this distinction does not support valuable semantics, then `fchmod` (possibly through a higher-level library like `rustix`) should be used. Unless there is a separate reason not to do that.

This may also eventually no longer be relevant: Setting permissions might stop being needed at all if files that don't have the target permissions are replaced instead of having their permissions changed.

### Git behavior

I'm not sure, in full. However, as noted in #1783 and #1784, Git replaces files during checkout (deleting/unlinking them and creating a new file of the same), rather than reusing them (keeping them but overwriting their contents), in more circumstances than gitoxide does. So I wouldn't expect the choice between `chmod` and `fchmod` to arise as often in the Git checkout implementation.

### Steps to reproduce 🕹

For the aspect of this that is about it being unclear why one operation uses the open file and the other uses the path, see above links.

For the aspect that speculates that this could worsen some race conditions, I don't have steps to reproduce that. I don't know how much of a problem it is and, as noted above, it may be that the current code is correct except for being under-documented with respect to why this disparity is actually useful.

EliahKagan:
Fixes #1786

This changes `finalize_entry` so it uses the open file descriptor, rather than the path, to read the old mode and write the new mode. (This is the PR mentioned in https://github.com/GitoxideLabs/gitoxide/issues/1786#issuecomment-2611744264.)

#### How it worked before

Previously:

- The old mode was read using `lstat`, called indirectly through the higher-level function `std::fs::symlink_metadata`.

  https://github.com/GitoxideLabs/gitoxide/blob/02efddd0a093295746dcb640bae1652b80daf45c/gix-worktree-state/src/checkout/entry.rs#L288

- The new mode was set using `chmod`, called indirectly through the higher-level function `std::fs::set_permissions`.

  https://github.com/GitoxideLabs/gitoxide/blob/02efddd0a093295746dcb640bae1652b80daf45c/gix-worktree-state/src/checkout/entry.rs#L290

  It is [documented](https://doc.rust-lang.org/stable/std/fs/fn.set_permissions.html) to use `chmod`, which may not be ideal, since if we are to continue to operate on a path here, `lchmod` is probably what we want.

#### How it would work now

With the changes in this PR, `gix-worktree-state` takes a direct dependency on `rustix`. I think this may be okay since `gix-worktree-state` already depends on `gix-index`, which depends on `rustix`. (I used the same version range and features as were used there.)

It uses the `fstat` and `fchmod` functions that `rustix` lightly wraps--as well as the `RawMode` and `Mode` types to avoid having to handle conversions and casts on systems like macOS where `mode_t` is 16-bit--to implement the new behavior of operating on the open file descriptor instead of the path:

- The old mode is read using `fstat`:

  https://github.com/GitoxideLabs/gitoxide/blob/5372911fa092d0bda4f9a159fb9e88ce9608d2aa/gix-worktree-state/src/checkout/entry.rs#L301

- The new is set using `fchmod`:

  https://github.com/GitoxideLabs/gitoxide/blob/5372911fa092d0bda4f9a159fb9e88ce9608d2aa/gix-worktree-state/src/checkout/entry.rs#L303

#### Caveat 1 (but maybe benefit?): We must not dereference symlinks

Because `fchmod` operates on a file descriptor, it changes the permissions of whatever file is open. If we opened this file through a symlink, then this may be the wrong thing to do. I believe that this will not happen, at least on a Unix-like system. We have, typically, opened this file to write to it. So we try hard not to open a symlink, and I think avoid this in a way that completely robust, at least on systems where we are setting executable permissions. But this is an area I recommend be considered and examined when reviewing this PR.

See https://github.com/GitoxideLabs/gitoxide/issues/1786#issuecomment-2611744264 for more information on this and related issues, including an area where I think this PR may actually improve resilience to wrongly following symlinks.

#### Caveat 2 (minor): Was the path intended to be used for more?

There is one other area I suggest giving special attention when reviewing this PR, though I very much doubt it is a problem: I was surprised that the `rela_path_as_path` variable introduced near the top of `set_executable_after_creation` was actually only used for arranging for executable permissions to be added. Given the way the code was written, I wonder if it was intended to use it for more than that.

---

There is some more information in the commit messages. In particular, 588e6b0 presents the curious `rela_path_as_path` situation.

### DIFF
--- a/gix-worktree-state/src/checkout/entry.rs
+++ b/gix-worktree-state/src/checkout/entry.rs
@@ -285,14 +285,23 @@ pub(crate) fn finalize_entry(
     // For possibly existing, overwritten files, we must change the file mode explicitly.
     #[cfg(unix)]
     if let Some(path) = set_executable_after_creation {
-        let old_raw_mode = rustix::fs::fstat(&file).map_err(std::io::Error::from)?.st_mode;
-        let new_mode = let_readers_execute(old_raw_mode);
-        rustix::fs::fchmod(&file, new_mode).map_err(std::io::Error::from)?;
+        set_executable(&file)?;
     }
     // NOTE: we don't call `file.sync_all()` here knowing that some filesystems don't handle this well.
     //       revisit this once there is a bug to fix.
     entry.stat = Stat::from_fs(&gix_index::fs::Metadata::from_file(&file)?)?;
     file.close()?;
+    Ok(())
+}
+
+/// Use `fstat` and `fchmod` on a file descriptor to make a regular file executable.
+///
+/// See `let_readers_execute` for the exact details of how the mode is transformed.
+#[cfg(unix)]
+fn set_executable(file: &std::fs::File) -> Result<(), std::io::Error> {
+    let old_raw_mode = rustix::fs::fstat(&file)?.st_mode;
+    let new_mode = let_readers_execute(old_raw_mode);
+    rustix::fs::fchmod(&file, new_mode)?;
     Ok(())
 }
 
--- a/gix-worktree-state/src/checkout/entry.rs
+++ b/gix-worktree-state/src/checkout/entry.rs
@@ -135,7 +135,7 @@ where
             };
 
             // For possibly existing, overwritten files, we must change the file mode explicitly.
-            finalize_entry(entry, file, set_executable_after_creation.then_some(dest))?;
+            finalize_entry(entry, file, set_executable_after_creation)?;
             num_bytes
         }
         gix_index::entry::Mode::SYMLINK => {
@@ -275,16 +275,15 @@ pub(crate) fn open_file(
     try_op_or_unlink(path, overwrite_existing, |p| options.open(p)).map(|f| (f, set_executable_after_creation))
 }
 
-/// Close `file` and store its stats in `entry`, possibly setting `file` executable depending on
-/// `set_executable_after_creation`.
+/// Close `file` and store its stats in `entry`, possibly setting `file` executable.
 pub(crate) fn finalize_entry(
     entry: &mut gix_index::Entry,
     file: std::fs::File,
-    #[cfg_attr(windows, allow(unused_variables))] set_executable_after_creation: Option<&Path>,
+    #[cfg_attr(windows, allow(unused_variables))] set_executable_after_creation: bool,
 ) -> Result<(), crate::checkout::Error> {
     // For possibly existing, overwritten files, we must change the file mode explicitly.
     #[cfg(unix)]
-    if let Some(path) = set_executable_after_creation {
+    if set_executable_after_creation {
         set_executable(&file)?;
     }
     // NOTE: we don't call `file.sync_all()` here knowing that some filesystems don't handle this well.
--- a/gix-worktree-state/src/checkout/entry.rs
+++ b/gix-worktree-state/src/checkout/entry.rs
@@ -275,20 +275,21 @@ pub(crate) fn open_file(
     try_op_or_unlink(path, overwrite_existing, |p| options.open(p)).map(|f| (f, set_executable_after_creation))
 }
 
-/// Close `file` and store its stats in `entry`, possibly setting `file` executable depending on `set_executable_after_creation`.
-#[cfg_attr(windows, allow(unused_variables))]
+/// Close `file` and store its stats in `entry`, possibly setting `file` executable depending on
+/// `set_executable_after_creation`.
 pub(crate) fn finalize_entry(
     entry: &mut gix_index::Entry,
     file: std::fs::File,
-    set_executable_after_creation: Option<&Path>,
+    #[cfg_attr(windows, allow(unused_variables))] set_executable_after_creation: Option<&Path>,
 ) -> Result<(), crate::checkout::Error> {
     // For possibly existing, overwritten files, we must change the file mode explicitly.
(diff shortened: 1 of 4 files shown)


gix-worktree-state/src/checkout/entry.rs
```diff
@@ -297,102 +293,104 @@ pub(crate) fn finalize_entry(
...
+}
+
+/// Given the st_mode of a regular file, compute the mode with executable bits safely added.
+///
+/// Currently this adds executable bits for whoever has read bits already. It doesn't use the umask.
+/// Set-user-ID and set-group-ID bits are unset for safety. The sticky bit is also unset.
+///
+/// This returns only mode bits, not file type. The return value can be passed to chmod or fchmod.
+#[cfg(unix)]
+fn let_readers_execute(mut raw_mode: rustix::fs::RawMode) -> rustix::fs::Mode {
```
Byron, line 314: It's a great idea to factor this into its own, documented function!

Byron (approved):
Thanks so much for the fix, this is great work!

> There is one other area I suggest giving special attention when reviewing this PR, though I very much doubt it is a problem: I was surprised that the `rela_path_as_path` variable introduced near the top of `set_executable_after_creation` was actually only used for arranging for executable permissions to be added. Given the way the code was written, I wonder if it was intended to use it for more than that.

I think this was just me struggling to do what I wanted as I was missing the `rustix` functions to work on an open file handle.

gix-worktree-state/src/checkout/entry.rs
EliahKagan: Thanks! This `let_readers_execute` function is actually not really new in *this* PR, though its documentation comment is. This function was renamed (in de939de) from `set_mode_executable`, and also changed to operate on different types. I had introduced `set_mode_executable` in bf33d2b (#1764) to be able to test this logic without actually doing any operations on the filesystem.

In contrast, the related `set_executable` function (which calls `let_readers_execute` between its `fstat` and `fchmod` calls) is new in this PR. I introduced it in fc82e78, originally to make `?` seamlessly work, rather than having to write `.map_err(std::io::Error::from)`, but my feeling is that having it separate from `finalize_entry` may improve readability slightly even beyond that.

I wasn't sure what the best name would be for `set_executable`. I also considered `set_file_executable`, but it seemed to me that would still not be clearer with respect to the important distinction between a file descriptor and its path, and I went with `set_executable`.

PR #1805, opened 2025-01-24
add `PrepareFetch::with_leave_dirty()` to avoid cleanup on drop.

ColinKinloch:
I want to be able to use `PrepareFetch` etc. on a given directory without it being deleted on failure.

Currently I'm checking out to a temporary directory and moving them to the target directory, which may generally be a better idea than this MR :shrug:.

I'm not sure about the `leave_dirty` name, `persist` is used for a similar concept. I'm open to suggestions.

### DIFF
--- a/gix/src/clone/access.rs
+++ b/gix/src/clone/access.rs
@@ -67,8 +67,10 @@ impl PrepareFetch {
 
 impl Drop for PrepareFetch {
     fn drop(&mut self) {
-        if let Some(repo) = self.repo.take() {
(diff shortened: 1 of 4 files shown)


Byron:
Thanks a lot for contributing, and sorry for the late reply!

I was wondering if [`std::mem::ManuallyDrop`](https://doc.rust-lang.org/std/mem/struct.ManuallyDrop.html) would be a suitable workaround here. As `Drop` should have no reason to ever do anything more than to delete its directory, I think this should be quite future-proof as well.

Thus I'd prefer a documentation change to make it more obvious how avoiding drop behaviour can be done.
What do you think?

PR #1807, opened 2025-01-25
docs: specify ThreadSafeRepository is not Send/Sync without "parallel"

### LINKED ISSUE #1806: `ThreadSafeRepository` is not `Send` or `Sync` with `default-features = false`
Additionally, `Repository` is not `Send`.

This may or may not be desired behavior, but it was definitely surprisingly. It would be nice if there was at least a note in the docs.

It seems that the minimal feature set required to activate thread safety is `features = ["parallel"]`.

bryceberger:
Closes #1806


Byron (approved):
Thanks a lot for your help!

PR #1809, opened 2025-01-25
update msrv to 1.79, add `rust-toolchain.toml`

### LINKED ISSUE #1808: Does not compile with MSRV
Currently, the MSRV is:
- 1.70 according to `gix/Cargo.toml`
- 1.74 (nightly!) according to `.github/workflows/msrv.yml`
- 1.76 according to a comment in `gix-commitgraph/CHANGELOG.md` stating:
  > Our MSRV follows the one of `helix`, which in turn follows Firefox
- 1.84 for all crates that are not dependents of `gix` (list?), according to `STABILITY.md`

However, `gix` can only build on 1.77+, due to c-string literals, used in https://github.com/rusqlite/rusqlite/pull/1483. The CI passes because it only runs `cargo check`, which does work on 1.74.

Additionally, tests only build on 1.79+, due to https://github.com/rust-lang/rust/pull/121346/.

`gix-macros/tests/momo/ux/error_if_ineffective.rs` fails on nightly versions. This appears to be due to differences in macro source location tracking on nightly versions.

bryceberger:
Closes #1808

Additionally, use the root `[workspace.package]` as a source of truth for all crates `repository`, `edition`, `license`, and `rust-version`. Also used for `authors` only if the only author listed was Sebastian Thiel.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ version = "0.41.0"
 default-run = "gix"
 include = ["src/**/*", "/build.rs", "LICENSE-*", "README.md"]
 resolver = "2"
+rust-version = "1.70"
 
 [[bin]]
 name = "ein"
@@ -306,6 +307,13 @@ members = [
     "gix-traverse/tests",
     "gix-shallow"
 ]
+
+[workspace.package]
+repository = "https://github.com/GitoxideLabs/gitoxide"
+authors = ["Sebastian Thiel <sebastian.thiel@icloud.com>"]
+edition = "2021"
+license = "MIT OR Apache-2.0"
+rust-version = "1.79"
 
 [workspace.dependencies]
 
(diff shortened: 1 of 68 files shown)


bryceberger:
This currently doesn't touch any of the CI --- if accepted, those might also need to change to respect `rust-toolchain.toml`

Byron (commented):
Thanks for the initiative!

I left [a comment](https://github.com/GitoxideLabs/gitoxide/issues/1808#issuecomment-2614260268) which attempts to direct it a bit.

This PR can't be happening as:

* the MSRV doesn't have to be driven by tests
* this project can't use workspace settings as it would affect changelog generation. Changelogs and versioning depend on per-crate file changes in Git, which won't happen if these are inherited.

Thanks for your understanding - I hope we can make some improvements in another way.

PR #1810, opened 2025-01-25
fix hardcoded path test on nixos

bryceberger:
nixos provides the following contents in `/bin` and `/usr/bin`:

    /bin:
    sh

    /usr/bin:
    env

That's it. From what I know, every linux distribution, as well as macos, has `/usr/bin/env`


Byron (approved):
I will try to remember that `/bin/sh` and `/usr/bin/env` are the only posix-paths one can assume exist.

PR #1812, opened 2025-01-26
Clarify description of `single` escaping

EliahKagan:
This improves how `gix_quote::single` documents how it handles the cases it treats specially.

The fix in #717 (discussed in #713) wasn't accompanied by a change to the documentation, which continued either to describe the preceding behavior or to be ambiguous. This changes it to describe the current behavior unambiguously. Only documentation is changed.


Byron (approved):

I remember having noticed this as well when using it for shell-command quotation, but then forgot to change it. So it's good this discrepancy isn't going to live on any longer.

Issue #1761, opened 2025-01-12
Research how to find the POSIX-compatible shell


EliahKagan:
### Summary 💡

It is a truism that the POSIX-compatible shell is `sh` and that it is located at `/bin/sh`. For systems where that is not so, or where it is in some important way not the whole truth, I'd like to figure out what should be done for running commands that should otherwise be run in `/bin/sh`.

- [x] Common situation on Unix-like systems: it's just `/bin/sh`!
- [ ] Solaris and illumos: Is the POSIX `sh` elsewhere than `/bin/sh`? Which do we use?
- [x] Git for Windows, full installation: `(git installation)\usr\bin\sh.exe` ([#1758](https://github.com/GitoxideLabs/gitoxide/pull/1758))
- [ ] Git for Windows, traditional [MinGit](https://github.com/git-for-windows/git/wiki/MinGit#mingit-is-git-for-windows-applications): Does it need anything special?
- [ ] Git for Windows, [BusyBox-based](https://github.com/git-for-windows/git/wiki/MinGit#experimental-busybox-based-mingit) MinGit: Does it need anything special?
- [ ] Windows with identifiable MSYS2: Do we use its `sh` as a fallback?
- [ ] Windows with an `.sh` file [association](https://github.com/git-for-windows/build-extra/blob/84ed494f4a7d9667bf3d54ed87d1da7196ec47cb/installer/install.iss#L113): Do we fall back to it if we can prove it's a shell?
- [ ] https://github.com/git-for-windows/git/pull/3301

#### Rationale

gitoxide runs some commands in shells. It must do so, for some commands, because the semantics of some environment variables and configuration variables, as defined by Git, which gitoxide wishes to be compatible with, stipulates it in some cases.

(It is also often important *not* to run commands in shells where applicable semantics make that the wrong thing to do. But that is independent of this issue.)

Whatever shell is provided as POSIX sh, and named as such, should generally be used to run shell commands in situations where no `#!` is applicable because we are running a command rather than a script. For example, `core.sshCommand` is a shell command. It must, at least in the general case, be run in some shell. (See also #1752 and #1758.)

#### Relevance

This is not a blocker for #1758. These are ideas that don't belong in the scope of #1758--except for the parts that, as noted above, are already covered there--and that I may not get to immediately. I am opening this so it is not forgotten.

[...]

PR #1815, opened 2025-01-26
Rename `Exponential` backoff to `Quadratic`

EliahKagan:
This renames `backoff::Exponential` to `backoff::Quadratic` and updates references to it, including in comments. For more details, see the commit message, and https://github.com/gitpython-developers/gitdb/pull/115#discussion_r1903215598 where this change was suggested.

---

I don't know if or how I should update the use in `gix-testtools`, which depends on a version of `gix-utils` older than what is tracked in the repository (#1510). The reference is in the `spawn_git_daemon` function:

https://github.com/GitoxideLabs/gitoxide/blob/38a0d9a083afab33e8a3cb66ab4acbc4e83d8486/tests/tools/src/lib.rs#L222

*Edit:* I added a comment to note it for later, when the dependencies of `gix-testtools` are advanced to newer versions. While I was at it, I also changed another dependency-related comment, by rewriting and moving it to refer to a way of using `shell()`, rather than `login_shell()` which was removed in #1758.

### DIFF
--- a/gix-lock/src/acquire.rs
+++ b/gix-lock/src/acquire.rs
@@ -14,7 +14,7 @@ pub enum Fail {
     /// Fail after the first unsuccessful attempt of obtaining a lock.
     #[default]
     Immediately,
-    /// Retry after failure with exponentially longer sleep times to block the current thread.
+    /// Retry after failure with quadratically longer sleep times to block the current thread.
     /// Fail once the given duration is exceeded, similar to [Fail::Immediately]
     AfterDurationWithBackoff(Duration),
 }
@@ -176,7 +176,7 @@ fn lock_with_mode<T>(
     match mode {
         Fail::Immediately => try_lock(&lock_path, directory, cleanup),
         Fail::AfterDurationWithBackoff(time) => {
-            for wait in backoff::Exponential::default_with_random().until_no_remaining(time) {
+            for wait in backoff::Quadratic::default_with_random().until_no_remaining(time) {
                 attempts += 1;
                 match try_lock(&lock_path, directory, cleanup.clone()) {
                     Ok(v) => return Ok((lock_path, v)),
--- a/tests/tools/src/lib.rs
+++ b/tests/tools/src/lib.rs
@@ -219,6 +219,7 @@ pub fn spawn_git_daemon(working_dir: impl AsRef<Path>) -> std::io::Result<GitDae
             .spawn()?;
 
     let server_addr = addr_at(free_port);
+    // TODO(deps): Upgrading dependencies will require changing `Exponential` to `Quadratic`.
     for time in gix_lock::backoff::Exponential::default_with_random() {
         std::thread::sleep(time);
         if std::net::TcpStream::connect(server_addr).is_ok() {
@@ -652,8 +653,8 @@ fn configure_command<'a, I: IntoIterator<Item = S>, S: AsRef<OsStr>>(
 }
 
 fn bash_program() -> &'static Path {
-    // TODO: use `gix_path::env::login_shell()` when available.
     if cfg!(windows) {
+        // TODO(deps): Once `gix_path::env::shell()` is available, maybe do `shell().parent()?.join("bash.exe")`
         static GIT_BASH: Lazy<Option<PathBuf>> = Lazy::new(|| {
             GIT_CORE_DIR
                 .parent()?
--- a/gix-utils/src/backoff.rs
+++ b/gix-utils/src/backoff.rs
@@ -9,17 +9,17 @@ fn randomize(backoff_ms: usize) -> usize {
     }
 }
 
-/// A utility to calculate steps for exponential backoff similar to how it's done in `git`.
-pub struct Exponential<Fn> {
+/// A utility to calculate steps for quadratic backoff similar to how it's done in `git`.
+pub struct Quadratic<Fn> {
     multiplier: usize,
     max_multiplier: usize,
     exponent: usize,
     transform: Fn,
 }
 
-impl Default for Exponential<fn(usize) -> usize> {
+impl Default for Quadratic<fn(usize) -> usize> {
     fn default() -> Self {
-        Exponential {
+        Quadratic {
             multiplier: 1,
             max_multiplier: 1000,
             exponent: 1,
@@ -28,10 +28,10 @@ impl Default for Exponential<fn(usize) -> usize> {
     }
 }
 
-impl Exponential<fn(usize) -> usize> {
-    /// Create a new exponential backoff iterator that backs off in randomized, ever increasing steps.
+impl Quadratic<fn(usize) -> usize> {
+    /// Create a new quadratic backoff iterator that backs off in randomized, ever increasing steps.
     pub fn default_with_random() -> Self {
-        Exponential {
+        Quadratic {
             multiplier: 1,
             max_multiplier: 1000,
             exponent: 1,
@@ -40,7 +40,7 @@ impl Exponential<fn(usize) -> usize> {
     }
 }
 
-impl<Transform> Exponential<Transform>
+impl<Transform> Quadratic<Transform>
 where
     Transform: Fn(usize) -> usize,
 {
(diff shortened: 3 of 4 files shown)


gix-lock/src/acquire.rs
```diff
@@ -14,7 +14,7 @@ pub enum Fail {
     /// Fail after the first unsuccessful attempt of obtaining a lock.
     #[default]
     Immediately,
-    /// Retry after failure with exponentially longer sleep times to block the current thread.
+    /// Retry after failure with quadratically longer sleep times to block the current thread.
```
Byron, line 17: Could you make it so that changes here go to a separate commit? Otherwise the release system will think that `gix-lock` had a breaking change and document it as well, even though it applies to `gix-utils` only.

In general, any commit that uses `<prefix>: <subject>`/'conventional commit' style must only touch a single crate.

tests/tools/src/lib.rs
```diff
@@ -652,8 +653,8 @@ fn configure_command<'a, I: IntoIterator<Item = S>, S: AsRef<OsStr>>(
 }
 
 fn bash_program() -> &'static Path {
-    // TODO: use `gix_path::env::login_shell()` when available.
     if cfg!(windows) {
+        // TODO(deps): Once `gix_path::env::shell()` is available, maybe do `shell().parent()?.join("bash.exe")`
```
Byron, line 657: I am looking forward to that!

Byron (changes requested):

The changes are perfectly mergeable, but I'd have to ask for a commit-refactor first to get the changelogs right. Please note that I typically have a second commit that says `adapt to changes in <crate>` to make code compile again.

gix-lock/src/acquire.rs
EliahKagan: No problem! I've just split that into two commit and force pushed. (While I was at it, I also rebased it on to the current main, though I suppose that doesn't really matter.) Please let me know if the way I did it is reasonable or if I should make further changes. I named the second commit a conventional commit as well with a `fix:` (not `fix!:`) prefix. I don't know if that should be done or not, and it would be very fast and easy for me to reword the message to remove it.

Byron: Thank you! 

Let's also remove the `fix:` from the commit that adapts to the changes in `gix-util`, as this isn't anything that is relevant to the users of the crate (presumably).

Thanks again.

EliahKagan: I've gone ahead and removed the `fix:` from the commit that adapts to the changes in `gix-util`. (This automatically disabled auto-merge, so you may want to reenable that after you have a look.)

tests/tools/src/lib.rs
EliahKagan: I'm not sure I'm looking forward to it, actually. I'm not sure if `shell()` currently works properly in all the cases listed in #1761 where it reasonably should. But one case listed there that I expect it can and should be made to work--if it doesn't work already--is [BusyBox-based MinGit](https://github.com/git-for-windows/git/wiki/MinGit#experimental-busybox-based-mingit). For that, there would not be a `bash.exe` in the same directory as the `sh` implementation `shell()` should find.

I look forward more to the possibility that `gix-path`--or some `gix-*` crate, since maybe this is actually in the ambit of `gix-command`--could report the locations of more executables.

`sh` should be preferred over `bash` for running shell scripts even when they are identical executables or one is a symlink to the other--because `bash`, and most other Bourne-style shells, behaves more POSIX-compatibly when invoked by the name `sh`. Yet finding and running `bash` is something that is often reasonable to do for specific reasons. It is a better choice in the gitoxide test suite, because the fixture scripts use `bash`-specific language features and they are not currently run in a way where their `#!` lines are automatically respected. There may even be some applications that need to use `bash` to run scripts that ideally should run in `sh`, just because they have been doing that for a long time already and users rely on it.

There are also various other executables that are often but not always available, and are relevant to Git, that are not found in the same directories as the Git for Windows `sh.exe` and `bash.exe`. An important case is `perl`. (MinGit does not have `perl`, but full Git for Windows installations do ship it, and I think it could be useful to find the one that comes with Git for Windows.)

More broadly, this could be part of a way of figuring out how to run interpreters specified in `#!` lines. Perhaps we should also have a function in `gix-path`, or somewhere. that takes a command name, to which `"bash"` could be passed.

Byron: I see, thanks for sharing! Finding binaries isn't straightforward and having read this, I'd be glad if providing such functionality in `gix-*` crates would be minimised, strictly limited to strong use-cases or demands.

> More broadly, this could be part of a way of figuring out how to run interpreters specified in `#!` lines. Perhaps we should also have a function in `gix-path`, or somewhere. that takes a command name, to which `"bash"` could be passed.

If I remember correctly, on Windows this is already done via `gix-command`. But I don't recall if anything special is done to launch the command specified in the shebang. but think the implementation closely follows the one in Git.

EliahKagan: `gix-command` attempts to interpret shebangs, but I don't think it does so using the environment modifications Git for Windows puts in place. Traditionally Git for Windows did this in the `git.exe` shim but not the real `git.exe` executable that the shim points to. But since https://github.com/git-for-windows/git/pull/2506 the real `git.exe` will detect when it looks like the customizations were not made and do them itself.

For `PATH` customizations especially, I think it might not be good to perform them or act by default as though they have been done. The kind of thing it does now, searching the preexisting `PATH`, seems better. Applications that use `gix-*` crates may run on Windows systems where Git for Windows is absent. Even when it is present, it is not clear that hooks should in general be searched--or run in interpreter found by searching--first in Git for Windows related directories. There is also the subtlety that some executables found in directories Git for Windows adds to the `PATH` might malfunction in some uses if run without that environment set.

So while I think this may be useful functionality, it should probably continue not to be the default, and be some combination of opt-in and/or enabled automatically only for interpreters that are very likely to be wanted from Git for Windows. This might just be `sh`, but `bash` (which is alongside `sh`) and `perl` (which is not) are plausible others.

Byron: Thanks for sharing, I wasn't aware that `git.exe` auto-modifies its `PATH`, something which indeed isn't done in `gix-command`. I also agree that it shouldn't be done by default, even though it might be useful to have at some point for opting in.

For now I assume that `sh.exe` used by Git will automatically allow all sibling commands to be called as well, which should make shell scripts on Windows more usable overall.

EliahKagan: As described in the note in https://github.com/GitoxideLabs/gitoxide/discussions/1840#discussioncomment-12205789, there's a separate issue that affects shebangs like `#!/bin/sh` as `gix-command` handles them on Windows: paths that start with `/` in a shebang line, at least if they start with only one `/`, are very unlikely to be meant in the way Windows interprets them--as being relative to the current drive--but `gix-command` uses them with that interpretation. Is there an existing issue for that?

Byron: I don't think so, and would definitely appreciate an issue or even a quick PR with a fix. Running absolute paths accidentally in such a way that it will pick up files relative to some directory sounds like a way to execute adversary-controlled binaries.

EliahKagan: Yes, that's true. I'll try to do something for this sometime soon. Since some of the regression tests that I wrote for #1840 and expected already to pass failed due to this problem, they might be good tests for a fix, or might be possible to adapt for that purpose. So I'll look into doing at least a partial fix for this based on that, either before a fix for #1840 or at the same time. I can then circle back to polish things further, as well as to examine this and related shebang handling behavior in greater depth.

EliahKagan: Just as a quick and incomplete update: I have not forgotten about this, but it looks like fixing this is not a quick PR, unless the fix is to disable `gix-command` shebang processing completely. A fix for this that one can be reasonably confident does not make things worse overall is one of the goals alluded to in #1862 and that I hope that PR may help with, but which I can possibly manage to do even if that PR does not work out.

PR #1818, opened 2025-01-30
Expand `from_fs` doc

### LINKED ISSUE #1817: `stat.ino()` is assumed to be small
### Current behavior 😯

`gix_index::entry::Stat::from_fs` obtains information from a `stat` structure on like this on all Unix-like systems:

https://github.com/GitoxideLabs/gitoxide/blob/ffb73b5f69dbe86ff88f1c473af65f368a6bcbe5/gix-index/src/entry/stat.rs#L94-L110

But even if that the rationale holds for Linux, it is not clear that the device number always fits in a 32-bit integer without loss of information, because enormous device numbers do exist on some notable Unix-like operating systems with some filesystems. It is furthermore not accurate that such large numbers are impractical, because devices need not be assigned sequentially. For example, [this forum discussion](https://forums.freebsd.org/threads/why-st_dev-is-so-large.91429/) pertains to huge device numbers for ZFS volumes on FreeBSD.

In addition, I am not confident this holds on Linux, since even on the same architecture there is more than one stat-related syscall. [This reference](https://chromium.googlesource.com/chromiumos/docs/+/master/constants/syscalls.md) of syscalls gives stat as an example of an entry point with multiple syscalls on the same platforms that use data types with different sizes.

### Expected behavior 🤔

I don't know if this actually causes a problem. This issue is thus incomplete in that I am not sure what happens with such devices. Maybe the truncation is okay? But it is not obvious why it would be.

However, even if the implementation is not changed, the comments should be revised, once the effect is known, to avoid implying that device numbers must be sequential or that the field fits in 32-bits on all systems, and to explain either:

- why is okay to cast it to `u32`, *or*
- what kinds of limitations or problems may arise.

### Git behavior

It looks like Git is similar. On *most* systems, `unsigned int` is 32-bit, even if `unsigned long` is 64-bit, and [Git defines](https://github.com/git/git/blob/5f8f7081f7761acdf83d0a4c6819fe3d724f01d7/statinfo.h#L16-L24):

```c
struct stat_data {
	struct cache_time sd_ctime;
	struct cache_time sd_mtime;
	unsigned int sd_dev;
	unsigned int sd_ino;
	unsigned int sd_uid;
	unsigned int sd_gid;
	unsigned int sd_size;
};
```

Which is [populated like this](https://github.com/git/git/blob/5f8f7081f7761acdf83d0a4c6819fe3d724f01d7/statinfo.c#L23-L34):

```c
void fill_stat_data(struct stat_data *sd, struct stat *st)
{
	sd->sd_ctime.sec = (unsigned int)st->st_ctime;
	sd->sd_mtime.sec = (unsigned int)st->st_mtime;
	sd->sd_ctime.nsec = ST_CTIME_NSEC(*st);
	sd->sd_mtime.nsec = ST_MTIME_NSEC(*st);
	sd->sd_dev = st->st_dev;
	sd->sd_ino = st->st_ino;
	sd->sd_uid = st->st_uid;
	sd->sd_gid = st->st_gid;
	sd->sd_size = munge_st_size(st->st_size);
}
```

It is not clear that this is intended in Git either. Other lossy conversions are identified as such.

### Steps to reproduce 🕹

I don't currently have a procedure to attempt to produce any incorrect runtime behavior based on this. I am unsure if it is only the comment that should be corrected, or if the representation and/or runtime behavior should also be changed.

EliahKagan:
Fixes #1817

This expands and slightly corrects the `gix_index::entry::from_fs` documentation:

- To be more precise about the expectation of where the metadata come from (it needn't specifically be the `symlink_metadata` function, and it is sometimes even an `fstat`-like function rather than an `lstat`-like function, though not `stat`).
- To mention that default values are used for unavailable data.
- To mention that truncation is sometimes performed on some fields.

Comments are also updated regarding some cases of truncation: why casting down to 32-bit should be okay for `st_dev` and `st_ino`. The `st_dev` case is discussed in #1817. I think this applies also to `st_ino`, but if that is less certain then further changes may be needed here.

### DIFF
--- a/gix-index/src/entry/stat.rs
+++ b/gix-index/src/entry/stat.rs
@@ -75,7 +75,14 @@ impl Stat {
         }
     }
 
-    /// Creates stat information from the result of `symlink_metadata`.
+    /// Creates stat information from file metadata.
+    ///
+    /// The information passed to this function should originate from a function like
+    /// `symlink_metadata`/`lstat` or `File::metadata`/`fstat`.
+    ///
+    /// The data are adjusted for use in the index, using default values of fields that are not
+    /// meaningful on the target operating system or that are unavailable, and truncating data
+    /// where doing so does not lose essential information for keeping track of file status.
     pub fn from_fs(stat: &crate::fs::Metadata) -> Result<Stat, SystemTimeError> {
         let mtime = stat.modified().unwrap_or(std::time::UNIX_EPOCH);
         let ctime = stat.created().unwrap_or(std::time::UNIX_EPOCH);
@@ -88,7 +95,7 @@ impl Stat {
             ino: 0,
             uid: 0,
             gid: 0,
-            // truncation to 32 bits is on purpose (git does the same).
+            // Truncation to 32 bits is on purpose (git does the same).
             size: stat.len() as u32,
         };
         #[cfg(not(windows))]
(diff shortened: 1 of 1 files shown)


Byron (approved):

PR #1822, opened 2025-01-31
Upgrade to Winnow 0.7

epage:
FYI as this doesn't use `cut_err` or `Partial`, you could switch away from `ErrMode` for some potential simplification and performance improvements.

I think it should just be
1. Remove `ErrMode`
2. Switch from `ModalParser` to `Parser`
3. Switch from `ModalResult` to `winnow::Result`

However, it'd be harder to undo it (picking apart which `Result`s need to be `ModalResult`s) so I held off on it from this PR

### DIFF
--- a/gix-actor/src/signature/decode.rs
+++ b/gix-actor/src/signature/decode.rs
@@ -3,7 +3,7 @@ pub(crate) mod function {
     use bstr::ByteSlice;
     use gix_date::{time::Sign, OffsetInSeconds, SecondsSinceUnixEpoch, Time};
     use gix_utils::btoi::to_signed;
-    use winnow::error::{ErrMode, ErrorKind};
+    use winnow::error::ErrMode;
     use winnow::stream::Stream;
     use winnow::{
         combinator::{alt, opt, separated_pair, terminated},
@@ -18,7 +18,7 @@ pub(crate) mod function {
     /// Parse a signature from the bytes input `i` using `nom`.
     pub fn decode<'a, E: ParserError<&'a [u8]> + AddContext<&'a [u8], StrContext>>(
         i: &mut &'a [u8],
-    ) -> PResult<SignatureRef<'a>, E> {
+    ) -> ModalResult<SignatureRef<'a>, E> {
         separated_pair(
             identity,
             opt(b" "),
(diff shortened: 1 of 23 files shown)


gix-actor/src/signature/decode.rs
```diff
@@ -68,31 +68,29 @@ pub(crate) mod function {
...
         let eol_idx = i.find_byte(b'\n').unwrap_or(i.len());
-        let right_delim_idx =
-            i[..eol_idx]
-                .rfind_byte(b'>')
-                .ok_or(ErrMode::Cut(E::from_error_kind(i, ErrorKind::Eof).add_context(
-                    i,
-                    &start,
-                    StrContext::Label("Closing '>' not found"),
-                )))?;
+        let right_delim_idx = i[..eol_idx]
```
Byron, line 74: This change was introduced in https://github.com/GitoxideLabs/gitoxide/pull/1439 and I wonder if this there isn't an 'packagiomatic' way of writing this. This code implements the parser by hand, but that was done merely due to lack of knowledge on how to use `winnow` primitives.

Is there anything that could be done to make it nicer?

Byron (commented):
Thanks a lot for your tremendous help with `winnow` updates, it's much appreciated!

It seems using `winnow::Result` is lighter as it doesn't include `ErrMode`, so it might be good for performance improvements. For now I don't know what that would do to the error output, but it seems that can be maintained as is.

Additionally I have left a comment with a question on how to improve a particular piece of code, hoping you find the time to take a look.

Thanks again.

Byron (approved):

tisonkun:
I happened to look into this patch and would like to share some thoughts on `ModalResult` vs `Result`.

1. Currently, the `gix_actor::identity::decode::identity` parser returns `ErrMode` inline, so `ModalResult` and `ModalParser` are still needed.
2. However, the `identity` parser doesn't used in `alt`/`opt`, so perhaps directly returning an error without the ErrMode wrapper is possible. (won't hurt error reporting, in theory)
3. But this is like a one-way door as mentioned in the description ("harder to undo it").

Back to the reason I jump into this patch: this is the final winnow 0.6 dependency in the tree of my application, so I'd expect a new release to eliminate that :D

epage:
> But this is like a one-way door as mentioned in the description ("harder to undo it").

I do see it less likely that gitoxide will need `ErrMode`.  Gitoxide is mostly focused on binary format parsing.  `cut_err` is mostly important when wanting to improve the quality of error messages which is usually for parsing text.

gix-actor/src/signature/decode.rs
epage: The main challenge is the `rfind`, right?  I've not thought of a good way to handle cases like that.

If its parsing within the `\n` delimited range, there is `and_then`, like `take_until(0.., b'\n').and_then(parser)`.

Byron: Yes, it's the `rfind()` to assure it finds the right-most `>` as delimiter, within the line boundary.

PR #1823, opened 2025-02-01
Add test for commits not ordered chronologically

cruessler:
This test covers a case that I came across working on #1743. Git doesn’t guarantee commits are ordered by commit date, so this adds a test for that specific case. In order to facilitate review, I decided to go with this small and targeted PR.

Since I keep adding tests to `make_blame_repo.sh`, I started wondering whether this was the best way of organizing the different test cases or whether it would be beneficial to spread them across multiple smaller files. What do you think?

### DIFF
--- a/gix-blame/tests/blame.rs
+++ b/gix-blame/tests/blame.rs
@@ -236,6 +236,11 @@ mktest!(
 );
 mktest!(file_only_changed_in_branch, "file-only-changed-in-branch", 2);
 mktest!(file_changed_in_two_branches, "file-changed-in-two-branches", 3);
+mktest!(
+    file_topo_order_different_than_date_order,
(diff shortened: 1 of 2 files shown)


Byron (approved):
That's wonderful, and I should really keep an eye out for the topo-order !+date-order case myself. 

It's something to keep in mind when implementing any kind of traversal for sure. The main problem about traversals for me certainly is that I forget *everything* about them very quickly.

Regarding test-organization, I think it's up to your preference. I keep finding myself working with possibly big and ever-growing 'thematic' fixtures. And since the theme here is blame, and the test-cases seem to be nicely isolated despite sharing a single repository, tree and commit-graph, I'd not change a thing.

In fact, it's probably great to have a complex commit-graph to lure out strange bugs.

PR #1802, opened 2025-01-24
Use `ubuntu-22.04-arm` for the ARM64 `test-fast` job

EliahKagan:
Fixes #1790

In the AArch64/ARM64 (64-bit, non-containerized) test-fast job, this uses the `ubuntu-22.04-arm` runner instead of the `ubuntu-24.04-arm` runner. This is to avoid the errors described in #1790, i.e., to work around rust-lang/rust#135867.

Such problems have not been observed on the 22.04 runner, including in tests intended to find them, and switching to it seems to be a complete workaround for the problem. In contrast, continuing to use the 24.04 runner, but attempting to work around the problem by switching from the stable to the beta channel, looks like it would greatly decrease the frequency of the errors but not eliminate them. A problem with `actions/checkout` failing is likewise observed on the 24.04 runner only, so using 22.04 avoids that too.

Because that seems like a complete workaround, this also reverts 50da7cb (#1792). That is to say that the ARM64 test-fast job is again in the `test-fast` matrix. It is capable of cancelling or being cancelled by the other `test-fast` checks. Code duplication in the workflow is somewhat decreased. The job will again block PR auto-merge.

Similar errors do not seem to have occurred in the `test-32bit` job that runs an arm32v7 Docker image in `ubuntu-24.04-arm`, and it is not clear that changing the runner image would help with #1780, nor even if that issue is still happening. Therefore, it is not changed there at this time.

This affects only ARM Linux runners. The x86-64 runners continue to use `ubuntu-latest`, which is currently resolved to `ubuntu-24.04`, and that does not need to be changed. Likewise, the `macos-latest` runners use ARM processors (Apple Silicon) and they are fine.

Various experiments were done in a separate workflow. Those experiments comprise all but the last commit here. I often drop or squash such things, but here they seem sufficiently valuable and interesting to keep. However, if that is not preferred, another option is to squash all but the last commit into one commit. The last experiment is the most useful and comprehensive, and thus the most important to have in the history. My preference for keeping the workflow details for all of them, rather than just that last one, is slight. So I would be pleased to squash those if desired.

The last commit includes the removal of the experiment workflow. Irrespective of what is done with its history, I don't think there is value in having that file in current commits. In addition, if it were kept, it would have to be modified to avoid running hundreds of extra checks on each and every push.

PR #1830, opened 2025-02-04
Change test-32bit arm32v7 runner to ubuntu-22.04-arm

EliahKagan:
Changed from ubuntu-24.04-arm, which seems to be having some problems in the test-32bit job (where it had still been in use), as discussed in #1828.

This can be viewed as a follow-up on #1802, which made the analogous change for the non-containerized test job only.

[This is the run](https://github.com/GitoxideLabs/gitoxide/actions/runs/13131588835/job/36637705370?pr=1830) of the changed job. It passes, though that does not demonstrate that this is better, since the failures before the change were nondeterministic and didn't happen most of the time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -209,7 +209,7 @@ jobs:
             toolchain: stable-i686-unknown-linux-gnu
           - container-arch: arm32v7
             runner-arch: arm64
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot, I hope this helps.

PR #1843, opened 2025-02-15
Fix a doc in `gix_path::convert`, revise others, and refactor slightly

EliahKagan:
This fixes a docstring that inadvertently described a different function with an opposite effect, revises others for clarity and stylistic consistency, and refactors to slightly decrease code duplication while elucidating the relationship between functions. More details are in the commit messages.

(It feels a bit silly to open this as a separate pull request from #1841, rather than including these changes there, but I think the considerations for reviewing these two pull requests are completely independent. Having this separate may also make it possible to review it even if it's not immediately clear whether #1841 should be accepted. Even though some nearby lines are changed, they aren't adjacent, so I *think* there is no merge conflict.)


Byron:
Thanks a lot, so much better than before!

PR #1844, opened 2025-02-16
Test Windows `<reserved>.middle.extension` filenames

EliahKagan:
I believe the implementation of `gix-validate` currently covers all modified cases of Windows reserved device names like `NUL`, for each name it recognizes as reserved when unmodified. But there is one common variation we didn't have test cases for: paths like `NUL.tar.xz` where the reserved device name is followed by something that is not technically an extension. The Windows filename extension in `NUL.tar.xz` is `xz` rather than `tar.xz`:

```text
> evcxr
Welcome to evcxr. For help, type :help
>> let p = std::path::Path::new("NUL.tar.xz");
>> p.file_stem()
Some("NUL.tar")
>> p.extension()
Some("xz")
```

As detailed in f5ef10d, the implementation covers this already by looking to see if the reserved name has a `.` (among other patterns) after it, regardless of what comes after the `.`, but this seems valuable to test because it is a case that is sometimes inadvertently not covered in implementations of the Windows path rules, so a regression in this area is possible if the implementation in `gix-validate` is ever substantially altered.

We did have tests already of related cases, such as paths where the reserved name is followed by multiple dots separated by spaces. But it seems to me that those preexisting cases--while important in their own right--neither express nor guard against the simpler case of a path like `NUL.tar.xz`.

Besides that, this PR includes two other changes: existing tests in the same file of paths containing `\` characters are changed to use raw literals so it is clearer if and how many times the backslashes are repeated (6daaba3) and, least importantly, more UNC-style paths are tested to avoid conveying the impression that `\\` would only be rejected in `\\?\` (eaa59a3).

### DIFF
--- a/gix-validate/tests/path/mod.rs
+++ b/gix-validate/tests/path/mod.rs
@@ -53,10 +53,10 @@ mod component {
 
         mktest!(ascii, b"ascii-only_and-that");
         mktest!(unicode, "😁👍👌".as_bytes());
-        mktest!(backslashes_on_unix, b"\\", UNIX_OPTS);
+        mktest!(backslashes_on_unix, br"\", UNIX_OPTS);
         mktest!(drive_letters_on_unix, b"c:", UNIX_OPTS);
         mktest!(virtual_drive_letters_on_unix, "֍:".as_bytes(), UNIX_OPTS);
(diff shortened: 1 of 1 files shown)


Byron:
That's wonderful, thank you!

PR #1845, opened 2025-02-17
Enable shell-related and env var tests Windows already supports

EliahKagan:
This enables two `gix-command` tests on Windows that had previously only run on Unix-like systems. One required a small change that didn't affect the robustness of the test, while the other required no change. Full details are in the commit messages.

### DIFF
--- a/gix-command/tests/command.rs
+++ b/gix-command/tests/command.rs
@@ -459,10 +459,9 @@ mod spawn {
     }
 
     #[test]
-    #[cfg(unix)]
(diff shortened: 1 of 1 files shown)


Byron (approved):

I will merge once the PR drops the draft status (in case there was anything else that's missing).

EliahKagan:
> I will merge once the PR drops the draft status (in case there was anything else that's missing).

Sorry about that, I just forgot to mark it ready for review! (My internet connection cut out for a bit and I forgot to come back to this to check the CI results.)

PR #1854, opened 2025-02-22
montly report

Byron:
## Tasks

* [x] report
* [x] update timesheets
* [x] publish on reddit

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -325,6 +325,7 @@ bin-dir = "gitoxide-max-pure-v{ version }-{ target }/{ bin }{ binary-ext }"
 pedantic = { level = "warn", priority = -1 }
 #
 # Reviewed and allowed lints
+needless_continue = "allow"
 enum_glob_use = "allow"  # x97
 missing_errors_doc = "allow"  # x1792
 missing_panics_doc = "allow"  # x447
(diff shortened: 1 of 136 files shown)


emilazy:
What causes the additional `panic!()` newline? I’m not seeing it here with Rust 1.84.1 on macOS, and the tests are failing as a result. Is it a 1.85 change?

Byron:
I don't exactly know what you are referring to but am assuming it's a failing journey test that wants to see specific line numbers in a traceback?
If so, then yes, there is a test that is prone to failure either if the underlying code changes, or if something in the panic handler changes.

emilazy:
Talking about https://github.com/GitoxideLabs/gitoxide/pull/1854/commits/cf98a159feb6a23dbe47887cd2744c9db93430aa; I'm not getting that extra newline and so the journey tests are broken for me. Doesn't seem to be a line number thing, I'm just not seeing the leading newline.

Byron:
Right - I remember that with some new Rust release, the leading newline appeared which was taken as expected value then.
CI still seems to agree which I suppose makes it a requirement. Ideally, this would be no dependency on something this 'flaky', so it's worth rewriting the test so the output of the program is filtered down to what matters. The test-suite has support for filters already, and doing so might be relevant for portability as different versions of Rustc produce difference outputs.

PR #1858, opened 2025-02-22
Add `gix blame --since`

cruessler:
This PR adds support for `--since=<date>` to `gix blame`. I’ve run a couple of manual tests on my machine and the results seem to match `git`’s results (apart from differences due to the slider problem as far as I can see).

### DIFF
--- a/gix-blame/src/types.rs
+++ b/gix-blame/src/types.rs
@@ -7,6 +7,17 @@ use std::{
     ops::{AddAssign, Range, SubAssign},
 };
 
+/// Options to be passed to [`file()`](crate::file()).
+#[derive(Default, Debug, Clone)]
+pub struct Options {
+    /// A 1-based inclusive range, in order to mirror `git`’s behaviour. `Some(20..40)` represents
+    /// 21 lines, spanning from line 20 up to and including line 40. This will be converted to
+    /// `19..40` internally as the algorithm uses 0-based ranges that are exclusive at the end.
+    pub range: Option<std::ops::Range<u32>>,
+    /// Don't consider commits before the given date.
+    pub since: Option<gix_date::Time>,
+}
+
 /// The outcome of [`file()`](crate::file()).
 #[derive(Debug, Default, Clone)]
 pub struct Outcome {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1545,6 +1545,7 @@ name = "gix-blame"
 version = "0.0.0"
 dependencies = [
  "gix-commitgraph 0.26.0",
+ "gix-date 0.9.3",
  "gix-diff",
  "gix-filter",
  "gix-fs 0.13.0",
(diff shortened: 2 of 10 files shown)


gix-blame/src/types.rs
```diff
@@ -7,6 +7,16 @@ use std::{
     ops::{AddAssign, Range, SubAssign},
 };
 
+/// Options to be passed to [`file()`](crate::file()).
+pub struct Options {
```
Byron, line 11:
```suggestion
#[derive(Default, Debug, Clone)
pub struct Options {
```

Byron (approved):
Wonderful, thanks so much!

I have one tiny nit for `Option`, and would merge right after this lands. I will wait until you have integrated it, but would also make it happen tomorrow morning if you don't get to it.

cruessler:
One thing I forgot to mention: there’s now an inconsistency between `commit_time` and other functions taking `commit: Either` as a parameter as `commit_time` borrows it as opposed to taking ownership.

gix-blame/src/types.rs
cruessler: Thanks for the suggestion, applied!

Byron:
Thanks a lot, and sorry for the long wait!

> One thing I forgot to mention: there’s now an inconsistency between `commit_time` and other functions taking `commit: Either` as a parameter as `commit_time` borrows it as opposed to taking ownership.

I don't think I follow, but if you think this is something that could be fixed, I encourage doing so in a follow-up. To me it sounds like copy-types shouldn't be borrowed by default.
I did try to look up the types you mentioned, but didn't find anything unusual.

cruessler:
Sorry, my question was lacking a lot of context! :-) I was referring to these functions:

```rust
fn commit_time(commit: &gix_traverse::commit::Either<'_, '_>)

fn collect_parents(commit: gix_traverse::commit::Either<'_, '_>, ...)

fn tree_id(commit: gix_traverse::commit::Either<'_, '_>)
```

All of them take an `Either`, but `commit_time` is the only function that takes `commit` by reference. I was wondering whether there were any reasons to remove this inconsistency.

Byron:
Yes, I think it would make sense to make it consistent. If you wanted to tackle it, it's a tough choice to go for one or the other. `Either` it's `Copy`, which means it shouldn't be passed as reference. If you'd choose to make it a reference, maybe because `std::mem::sizeof` shows it's a bit large, then `Copy` could be removed so `&` is naturally enforced.

cruessler:
> Yes, I think it would make sense to make it consistent. If you wanted to tackle it, it's a tough choice to go for one or the other. `Either` it's `Copy`, which means it shouldn't be passed as reference. If you'd choose to make it a reference, maybe because `std::mem::sizeof` shows it's a bit large, then `Copy` could be removed so `&` is naturally enforced.

I‘ll weigh the options and create a follow-up PR!

PR #1859, opened 2025-02-22
Add test for `--since`

cruessler:
I wanted to add the following test for `AsTime`, but it fails with the following message:

```
error: invalid value ''2 weeks ago'' for '--since <ARG>': Date string can not be parsed
```

Am I using `AsTime` incorrectly or is there an issue with the way the test is written?

I initially created the test to be part of #1858.

### DIFF
--- a/src/shared.rs
+++ b/src/shared.rs
@@ -441,7 +441,7 @@ pub use self::clap::{
 
 #[cfg(test)]
 mod value_parser_tests {
-    use super::{AsRange, ParseRenameFraction};
+    use super::{AsRange, AsTime, ParseRenameFraction};
     use clap::Parser;
 
     #[test]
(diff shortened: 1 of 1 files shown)


Byron:
That's interesting!

I know there are tests for `2.weeks.ago`, but I thought space should definitely work as well.

cruessler:
My first hunch was that it had to do with `'` as the error message contained 2 `''` in a row in front and after the string, but I didn’t investigate further.

cruessler:
There was an issue with the ways the test was written (now it passes). `Cmd::parse_from` does not parse quoted strings, that is handled by the shell. See https://github.com/clap-rs/clap/issues/1315.

Byron (approved):
Yes, that makes sense.
Thanks for figuring out the fix so we can have these tests.

PR #1863, opened 2025-02-25
cargo-deny maintenance

EliahKagan:
This removes [RUSTSEC-2025-0007](https://rustsec.org/advisories/RUSTSEC-2025-0007.html) as an allowed advisory since it has been withdrawn and therefore need no longer be listed, updates the versions of two `Cargo.lock` dependencies that have been yanked and whose updated versions fix what seem to be soundness bugs, and removes the old Unicode license from the allow list now that no dependencies use it. See commit messages for full details.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4652,9 +4652,9 @@ dependencies = [
 
 [[package]]
 name = "scc"
-version = "2.2.5"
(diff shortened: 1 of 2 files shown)


Byron (approved):

Issue #1832, opened 2025-02-06
Rename tracker shouldn't be order-dependent


Byron:
The same result should be produced independently of the order of input items.
If I recall correctly, it already sorts its output before returning it.

### Discussed in https://github.com/GitoxideLabs/gitoxide/discussions/1831

<div type='discussions-op-text'>

<sup>Originally posted by **EliahKagan** February  5, 2025</sup>
The i386 (32-bit x86) test-32bit job was not changed in #1830, which only changed the arm32v7 (32-bit ARM) job. But after CI passed there, the i386 job failed in the merge commit on the main branch. Although I would guess the failure might go away if the job is rerun, since it passed when it ran on the main branch of my fork when I fast-forwarded (synchronized) it, I don't know how [a failure like this](https://github.com/GitoxideLabs/gitoxide/actions/runs/13151883215/job/36700662527#step:9:2354) would plausibly happen even with low probability in the absence of a bug in gitoxide or its test suite:

```text
        FAIL [   0.010s] gix-status-tests::status index_as_worktree_with_renames::changed_and_untracked_and_renamed
──── STDOUT:             gix-status-tests::status index_as_worktree_with_renames::changed_and_untracked_and_renamed

running 1 test
test index_as_worktree_with_renames::changed_and_untracked_and_renamed ... FAILED

failures:

failures:
    index_as_worktree_with_renames::changed_and_untracked_and_renamed

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 27 filtered out; finished in 0.00s

──── STDERR:             gix-status-tests::status index_as_worktree_with_renames::changed_and_untracked_and_renamed
thread 'index_as_worktree_with_renames::changed_and_untracked_and_renamed' panicked at gix-status/tests/status/index_as_worktree_with_renames.rs:311:5:
assertion failed: `(left == right)`

Diff < left / right > :
 [
     Rewrite {
         source_rela_path: "dir/content",
         dest_rela_path: "content-copy",
         dest_dirwalk_status: Untracked,
         diff: None,
         copy: false,
     },
     DirwalkEntry {
         rela_path: "content-copy-with-rewrite",
         status: Untracked,
         disk_kind: Some(
             File,
         ),
     },
     Rewrite {
         source_rela_path: "dir/content2",
         dest_rela_path: "content-with-rewrite",
         dest_dirwalk_status: Untracked,
         diff: Some(
             DiffLineStats {
                 removals: 0,
[...]

Issue #1816, opened 2025-01-28
Pseudorandom numbers are sometimes the same across processes


EliahKagan:
#### Background

Fast pseudorandom numbers are used in a few places where they probably need not be cryptographically secure, but where the generated bytes are intended to be different across any kind of concurrent tasks, including across separate processes. Existing uses in gitoxide--but not necessarily all of their callers--are as follows:

1. [`gix_utils::backoff::Quadratic::default_with_random`](https://github.com/GitoxideLabs/gitoxide/blob/ffb73b5f69dbe86ff88f1c473af65f368a6bcbe5/gix-utils/src/backoff.rs#L33-L40) supplies a series of increasing timings that are adjusted with pseudorandom [jitter](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/), so concurrent tasks don't all retry at the same time. Its [`randomize`](https://github.com/GitoxideLabs/gitoxide/blob/ffb73b5f69dbe86ff88f1c473af65f368a6bcbe5/gix-utils/src/backoff.rs#L3-L10) helper computes:

   https://github.com/GitoxideLabs/gitoxide/blob/ffb73b5f69dbe86ff88f1c473af65f368a6bcbe5/gix-utils/src/backoff.rs#L4

   `default_with_random` is public, so there may be non-gitoxide crates that use it directly, probably in the same kind of scenarios it is used here. Here, it is used to avoid some cases of the [thundering herd](https://en.wikipedia.org/wiki/Thundering_herd_problem) problem: when accessing a network, at least [in `gix-testtools`](https://github.com/GitoxideLabs/gitoxide/blob/ffb73b5f69dbe86ff88f1c473af65f368a6bcbe5/tests/tools/src/lib.rs#L223); as well as [in `gix-lock`](https://github.com/GitoxideLabs/gitoxide/blob/ffb73b5f69dbe86ff88f1c473af65f368a6bcbe5/gix-lock/src/acquire.rs#L179), which runs in production.

   (Per [gitpython-developers/gitdb#115 (comment)](https://github.com/gitpython-developers/gitdb/pull/115#discussion_r1903215598) and [#1815](https://github.com/GitoxideLabs/gitoxide/pull/1815), this was recently renamed from `Exponential` to `Quadratic`. But that was just a name change, with no effect on this issue.)

[...]

PR #1867, opened 2025-02-27
Use ubuntu-24.04-arm again

### LINKED ISSUE #1780: Container creation sometimes fails in the 32-bit ARM test job
### Current behavior 😯

One of the changes in #1777 was to add an arm32v7 test job that runs in a container on the new arm64 runner (cbe3793, fbc27b5), analogous to the preexisting i386 test job that runs in a container on an amd64 runner. It looks like this may be brittle, with container creation failing from time to time. This is [the failure](https://github.com/GitoxideLabs/gitoxide/actions/runs/12845418981/job/35819462653) noted in https://github.com/GitoxideLabs/gitoxide/pull/1778#issuecomment-2599773835.

```text
/usr/bin/docker start 4224fb6a96d4ae28ceca367700326843715626ffe3eb995cdbe03b0aa4e0b4b2
  Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: unable to apply cgroup configuration: unable to start unit "docker-4224fb6a96d4ae28ceca3677003[26](https://github.com/GitoxideLabs/gitoxide/actions/runs/12845418981/job/35819462653#step:2:29)843715626ffe3eb995cdbe03b0aa4e0b4b2.scope" (properties [{Name:Description Value:"libcontainer container 4224fb6a96d4ae28ceca367700326843715626ffe3eb995cdbe03b0aa4e0b4b2"} {Name:Slice Value:"system.slice"} {Name:Delegate Value:true} {Name:PIDs Value:@au [4198]} {Name:MemoryAccounting Value:true} {Name:CPUAccounting Value:true} {Name:IOAccounting Value:true} {Name:TasksAccounting Value:true} {Name:DefaultDependencies Value:false}]): Failed to activate service 'org.freedesktop.systemd1': timed out (service_start_timeout=25000ms): unknown
  Error: failed to start containers: 4224fb6a96d4ae28ceca367700326843715626ffe3eb995cdbe03b0aa4e0b4b2
```

This resembles,, and is probably identical to, some failures I had seen when testing #1777, that I had erroneously assumed (or hoped) were due to a hiccup in infrastructure rather than a persistent problem. This issue tracks that in case it is a persistent problem, which seems likely. If it happens again and no fix is apparent, I can revert the parts of #1777 that are about 32-bit testing, while keeping 87387c2 from it, which does not seem to have had any problems.

<details><summary>Something that probably <em>isn't</em> the cause</summary>
<p>It is possible for a 64-bit ARM processor not to be capable of natively executing 32-bit ARM instructions--unlike 64-bit and 32-bit x86, this capability is not universal. When that happens, if <code>binfmt_misc</code> is configured to provide emulation via QEMU, a container of the incompatible architecture can still be run, but it will run much slower and some things may not work. However, while that was an early concern I had, as far as I can tell from the error that does not seem to be a factor here. Furthermore, in another repository, I checked in a reverse shell that no such architecture was enabled in <code>binfmt_misc</code> (<a href="https://github.com/EliahKagan/arm/commit/496d9c181a2cb35249ac512730d7e773ca835a5d">EliahKagan/arm@<code>496d9c1</code></a>), and also even tried turning off <code>binfmt_misc</code> (<a href="https://github.com/EliahKagan/arm/commit/efa15ff01a0d3964a704bf03a009432e5aaf13d7">EliahKagan/arm@<code>efa15ff</code></a>), and a 32-bit ARM binary was still able to run.</p>
</details>

### Expected behavior 🤔

The container specified in the `container:` key should start up at least as reliably in jobs on the ARM runner as other runners.

### Git behavior

Not directly applicable, but Git does test on various platforms. Cursory inspection of the `runs-on` keys in [this workflow](https://github.com/git/git/blob/master/.github/workflows/main.yml) suggests Git may not be using the new `ubuntu-24.04-arm` or `ubuntu-22.04-arm` GHA runners at this time.

### Steps to reproduce 🕹

I'm unsure what factors trigger this, or if it is effectively random. It seems likely that it will happen again, but I'm not certain, so I'm opening this issue rather than immediately changing the workflow.

When I was working on the PR, I think it happened most often when I had two pushes separated by a very short time. My first thought was that it might have to do with caching. That is implausible, though, at least with respect to the caching of Rust dependencies that we are doing, because the failure happens much earlier, when the GitHub Actions runner software runs Docker to set up the job, before any steps of the job have begun.

### LINKED ISSUE #1866: ubuntu-24.04-arm can probably be used again
### Summary 💡

A couple of weeks ago, https://github.com/rust-lang/rust/issues/135867 has been updated to note:

> Update (2025-02-13): GitHub has changed the new large runners to Neoverse N1 CPUs. The underlying issue is not fixed, but GHA large aarch64 runners should now work fine with Ubuntu 24.

See also https://github.com/rust-lang/rust/issues/135867#issuecomment-2660524682 and https://github.com/actions/partner-runner-images/issues/36#issuecomment-2679504033.

This was the cause of the problems we had using `ubuntu-24.04-arm` in `test-fast` (#1790), and *probably* the cause of the problems we had using it in `test-32bit` (including #1780). It is likely feasible to go back to using ubuntu-24.04-arm now, reverting #1802 and #1830, respectively.

This is a tracking issue for that. I hope to open a PR soon. This is an issue rather than starting out as a PR because fairly heavy testing is needed to determine if the problems still sometimes occur, and I don't want to run hundreds of jobs on GitoxideLabs that I can run on my personal account. (This relates to rate-limiting; in neither case would payment be involved. But it is also because I have more ability to cancel and rerun jobs that run in my fork.) There are multiple ways to achieve that, but not having a related PR at all while doing those tests is the simplest way to achieve it even if I make mistakes, and the code changes to the workflows are not themselves interesting.

### Motivation 🔦

We had preferred to use `ubuntu-24.04-arm` initially, which matches the current `ubuntu-latest` version which is currently equivalent to `ubuntu-24.04`. I think `ubuntu-24.04-arm` is intended, or intended eventually, to run on faster machines than `ubuntu-22.04-arm`, though I could have misunderstood. Because it looks like `ubuntu-24.04-arm` should work here again, there is also a benefit to trying again for it, since if problems still occur, I can try to report them.

I think switching back to 24.04 for the ARM Linux jobs is not a priority, but it also probably does not require a high amount of time or effort to attempt. Therefore, so long as I manage to test enough runs to have a reasonably good degree of belief that the problems we had are fixed, I think this is worth doing now.

*Edit:* The testing worked out, and I've opened PR #1867.

EliahKagan:
Closes #1780
Fixes #1866

This changes `ubuntu-22.04-arm` back to `ubuntu-24.04-arm` both in the `test-fast` ARM job and the `test-32bit` ARM job, because the underlying cause of the problems that we changed them to 22.04 to avoid (https://github.com/rust-lang/rust/issues/135867) have been fixed to the extent that they affect GitHub Actions runners. See #1866 for details on that, or https://github.com/actions/partner-runner-images/issues/36#issuecomment-2679504033 for a quick summary of the underlying fix.

The first commit here includes a heavy test, running hundreds of matrix jobs, to confirm the above and look for any remaining problems with `ubuntu-24.04-arm` before switching back to it. The second commit removes the separate workflow for those tests, since we should not run these extra jobs regularly, and since if they are ever useful again then future tests would still need to modify them significantly.

The heavy-test workflow runs, where the first run didn't have `fail-fast: false` on `test-32bit`, were:

- https://github.com/EliahKagan/gitoxide/actions/runs/13540509414
- https://github.com/EliahKagan/gitoxide/actions/runs/13550873274

Enough test jobs were run that failures were to be expected: even outside of ARM runners, there is a low but nonzero rate of nondeterministic failure in gitoxide CI. The failures I observed were as follows:

- [In test-fast on 24.04](https://github.com/EliahKagan/gitoxide/actions/runs/13540509414/job/37840543172#step:6:83), in one of the runs that tested a modified tool installation procedure using `cargo quickinstall`, downloading failed with HTTP 403 including after retries.

- [In test-32bit on 22.04](https://github.com/EliahKagan/gitoxide/actions/runs/13540509414/job/37840463388#step:8:3021), `gix-prompt::prompt ask::askpass_only` failed with EOF on [this sub-case](https://github.com/GitoxideLabs/gitoxide/blob/34b19d38fbf2e1cb892814ba9d8aa5d5eddf7a1f/gix-prompt/tests/prompt.rs#L46). I vaguely recall seeing similar failures in these expectrl-based tests, very rarely. But I am not sure. In any case, this is with 22.04, not 24.04.

- [In test-fast on 22.04](https://github.com/EliahKagan/gitoxide/actions/runs/13550873274/job/37874103252#step:7:3515), `gix-worktree-state-tests::worktree state::checkout::dangling_symlinks_can_be_created` failed because [the probe](https://github.com/GitoxideLabs/gitoxide/blob/34b19d38fbf2e1cb892814ba9d8aa5d5eddf7a1f/gix-worktree-state/tests/state/checkout.rs#L313) did not detect the ability to create symlinks. This is due to [#1816](https://github.com/GitoxideLabs/gitoxide/issues/1816). It makes the third known occurrence, after [#1789](https://github.com/GitoxideLabs/gitoxide/discussions/1789) and [#1816 (comment)](https://github.com/GitoxideLabs/gitoxide/issues/1816#issuecomment-2677015655). It is interesting because the first two occurrences were instead with `gix-worktree-state-tests::worktree state::checkout::overwriting_files_and_lone_directories_works` instead.

- [In test-32bit on 22.04](https://github.com/EliahKagan/gitoxide/actions/runs/13550873274/job/37874009404#step:8:3035), `gix-status-tests::status index_as_worktree_with_renames::changed_and_untracked_and_renamed` failed. Examining the actual vs. expected diffs reveals that it is due to [#1832](https://github.com/GitoxideLabs/gitoxide/issues/1832):

  ```diff
  <    DirwalkEntry {
  <        rela_path: "dir/untracked",
  <        status: Untracked,
  <        disk_kind: Some(
  <            File,
  <        ),
  >    Rewrite {
  >        source_rela_path: "empty",
  >        dest_rela_path: "dir/untracked",
  >        dest_dirwalk_status: Untracked,
  >        diff: None,
  >        copy: true,
  ```

  ```diff
  <    DirwalkEntry {
  <        rela_path: "untracked",
  <        status: Untracked,
  <        disk_kind: Some(
  <            File,
  <        ),
  >    Rewrite {
  >        source_rela_path: "empty",
  >        dest_rela_path: "untracked",
  >        dest_dirwalk_status: Untracked,
  >        diff: None,
  >        copy: true,
  ```

These are four failures in hundreds of runs. Only one of the four failures is on 24.04, with a download procedure that we do not use outside of these tests and the experiments that precede them (`ci.yml` does not use `cargo quickinstall`) and that I beleive I have seen locally a number of times on different platforms. The other three failures were on 22.04, all of them look like they are not runner-specific in any way, and none of them resemble any of the failures that had motivated us to avoid `ubuntu-24.04-arm` before.


Byron:
Thanks so much!

I love the approach taken here, 'simply' testing stability by taking a large amount of samples, quickly.
Locally I do the same, typically using `hyperfine` to run a test binary many times to trigger non-deterministic failures (the last one triggered when a file was created in second N, but was checked in second N+1, so it's a rare occurrence but not at all impossible), but I never dared to think this can be done on CI.

PR #1864, opened 2025-02-26
Make the `bash_program()` helper in `gix-testtools` a little more robust

EliahKagan:
Tasks:

- [x] Change it to prefer the shim (drop/rewind or undo 7c0af6a)
- [x] Figure out if and how is currently best to fall back to a non-shim (see outscoped)
- [x] Retest locally
- [x] Retest locally along with changes from [#1862](https://github.com/GitoxideLabs/gitoxide/pull/1862) (though they should be independent)
- [ ] If [#1862](https://github.com/GitoxideLabs/gitoxide/pull/1862) comes in first, rebase for clearer history (optional, should not be needed)

Outscoped:

- https://github.com/GitoxideLabs/gitoxide/issues/1886
- Investigate local failures with non-shim (outscoped to [#1862](https://github.com/GitoxideLabs/gitoxide/pull/1862))\*

\* The stakes here are low enough that I think that doesn't have to be investigated in detail. But the stakes in #1862 are higher, the shim vs. non-shim situation is similar, and we only surfaced one error case there, so I am going to examine the failures from *here* to become more confident in the improved solution *there*.

---

On Windows, since #1712, `gix-testtools` looks for a `bash` shell associated with Git for Windows, to run fixture scripts. This logic is broadly similar to `gix_path::env::shell()` (#1758), which it influenced (https://github.com/GitoxideLabs/gitoxide/pull/1752#discussion_r1909401210). But there are some differences, even beyond `shell()` in `gix-path` being for `sh` and `bash_program()` in `gix-testtools` being for `bash`. Some differences are stylistic, while others are behavioral.

It seems to me that they should be more similar than they are, and that there are also some improvements that should be applied to both in the same way. The changes proposed for `shell()` in #1862 include some that make it more similar to `bash_program()`, such as checking if the shell exists, as well as others such as preferring `/`, which neither `shell()` nor `bash_program()` yet do but which makes sense in both.

This PR is the `gix-testtools` counterpart to the `gix-path` changes in #1862. The topical similarity could justify including both in the same PR, but having separate feature branches is more flexible, and I think the effects are also independent. In addition, neither is ready, it is possible that one will be ready before the other, and I do not know which. Currently the changes in this PR are:

- Refactor for clarity and to reduce unnecessary dissimilarity to `gix_path::env::shell()`.
- Prefer `/` over `\` in building the path to `bash.exe`.
- Add some more tests.
[...]

PR #1862, opened 2025-02-23
Make `gix_path::env:shell()` more robust and use it in `gix-command`

EliahKagan:
Tasks since https://github.com/GitoxideLabs/gitoxide/pull/1862#issuecomment-2692158831:

- [x] Use the shim `sh.exe` when available
- [x] Fall back to the non-shim `sh.exe`
- [x] Test locally to ensure the failures still go away with the shim
- [x] Only use inferred `(git root)` if it looks like Git for Windows and looks safe
- [x] Fix unrelated `test-fixtures-windows` breakage ([#1849](https://github.com/GitoxideLabs/gitoxide/issues/1849), [#1870](https://github.com/GitoxideLabs/gitoxide/pull/1870)) and rebase
- [x] Fix unrelated `test-32bit` breakage ([#1874](https://github.com/GitoxideLabs/gitoxide/pull/1874)) and rebase
- [x] Retest locally after all nontrivial rebasing
- [x] Investigate [local failures](https://github.com/GitoxideLabs/gitoxide/pull/1864#issuecomment-2724464883) in [#1864](https://github.com/GitoxideLabs/gitoxide/pull/1864) with non-shim, in case relevant here

Outscoped:

- Report the observed `>>` problem as a bug to MSYS2 or a related project
- https://github.com/GitoxideLabs/gitoxide/issues/1868

(On all the above, see "What it means for this PR" in https://github.com/GitoxideLabs/gitoxide/pull/1862#issuecomment-2692158831.)

The original PR description follows.

---

This makes `git_path::env::shell()` more likely to work correctly on Windows, then attempts to use it in `gix_command::Prepare` when running a command with a shell due to `use_shell` being set to `true`. Full details are in the commit messages, though the changes are also summarized here.

I think this is not ready to merge yet, due to failures that happen when running the tests locally from PowerShell, but that I cannot produce otherwise. I do not know what causes these failures, and I expect I may have difficulty getting this to a point where it would be ready to merge without input. Since the failures happen in a test that was fairly recently added, I figured you might just know what is going on.

I think `shell()` can and should be used for `gix-command`. But if not then the reason will be relevant to other potential uses of `shell()`. In that case, assuming they leave enough good uses for `shell()` that it is kept, I hope to discover and articulate those reasons in the documentation for `shell()`. But I think most likely there is a bug or wrong test expectation somewhere that accounts for the local test failure, and that either the current implementation of `shell()` with the modifications included here, or something much like it, is suitable for broad use.

[...]

Issue #1849, opened 2025-02-19
3 new Windows GIX_TEST_IGNORE_ARCHIVES=1 failures with Git 2.48.1


EliahKagan:
### Current behavior 😯

When the test suite is run on Windows with `GIX_TEST_IGNORE_ARCHIVES=1`, there are [three new failures](https://gist.github.com/EliahKagan/7a088164fdf792413b47964a75bf90aa) starting in Git for Windows 2.48.1. I don't know if the problem is in Git for Windows, gitoxide, or gitoxide's test suite:

- [`gix-diff-tests::diff index::renames_by_identity`](https://gist.github.com/EliahKagan/7a088164fdf792413b47964a75bf90aa#file-output-txt-L1429-L1447)
- [`gix-diff-tests::diff tree_with_rewrites::renames_by_identity`](https://gist.github.com/EliahKagan/7a088164fdf792413b47964a75bf90aa#file-output-txt-L1482-L1500)
- [`gix-merge::merge tree::run_baseline`](https://gist.github.com/EliahKagan/7a088164fdf792413b47964a75bf90aa#file-output-txt-L1936-L2151)

Those links go to the directly relevant sections of the test output [in a gist](https://gist.github.com/EliahKagan/7a088164fdf792413b47964a75bf90aa) that [contains the full output](https://gist.github.com/EliahKagan/7a088164fdf792413b47964a75bf90aa#file-output-txt). The directly relevant sections themselves are:

#### gix-diff-tests::diff index::renames_by_identity

```text
        FAIL [  32.175s] gix-diff-tests::diff index::renames_by_identity
──── STDOUT:             gix-diff-tests::diff index::renames_by_identity

running 1 test
test index::renames_by_identity ... FAILED

failures:

failures:
    index::renames_by_identity

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 68 filtered out; finished in 32.15s

──── STDERR:             gix-diff-tests::diff index::renames_by_identity
thread 'index::renames_by_identity' panicked at gix-diff\tests\diff\index.rs:181:13:
assertion `left == right` failed: symlinks are only tracked by identity
  left: []
 right: ["link-1", "renamed-link-1"]
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

#### gix-diff-tests::diff tree_with_rewrites::renames_by_identity

```text
[...]

PR #1870, opened 2025-03-02
Use Git 2.49.0 release candidate in `test-fixtures-windows`

EliahKagan:
This is to work around #1849, while still running tests and fixture scripts that rely on Git being able to rename symlinks on Windows. [The release candidate](https://github.com/git-for-windows/git/releases/tag/v2.49.0-rc0.windows.1) has the fix from https://github.com/git-for-windows/git/pull/5437 for https://github.com/git-for-windows/git/issues/5436.

As written, this grabs the release associated with the newest tag, so roughly speaking it is "rolling" and will thus pick up subsequent release candidates. It can be undone once GitHub-hosted Windows runner images get Git for Wiindows 2.49.0 (which will likely occur a short time after it has a stable release). This only modifies the `test-fixtures-windows` job and *not* other jobs such as the `test-fast` `windows-latest` job.

This is a different approach to working around #1849 than I had suggested in https://github.com/GitoxideLabs/gitoxide/issues/1849#issuecomment-2692142316 that I would do. If this is not suitable, then the approach I mentioned there of conditionally skipping tests or suppressing assertions could be considered. A third approach could be to simply add the newly failing tests to `etc/test-fixtures-windows-expected-failures-see-issue-1358.txt`, though currently that is a list of long-standing failures that occur with all Git for Windows versions and should gradually shrink (#1358). Only one of these approaches should be used. If the one here is not preferred, then I can update this PR to use another.

I think probably any of these approaches is fine. The reason I did it this way is that current work (in #1862 and #1864), and related work that may follow it, seems like it would benefit from being fully tested on Windows on CI with `GIX_TEST_IGNORE_ARCHIVES=1`, including the three tests that rely on fixture scripts being able to run `git mv` on a symlink. This could also have been done by *downgrading* the Git for Windows installation rather than upgrading it. But as long as we are downloading and running an installer, it seems to me that it is better to catch any incompatibilities with recent versions, even at the risk of observing bugs in Git for Windows.

Observing bugs in Git for Windows that would affect the test suite might also be considered an advantage rather than a disadvantage. For now, I've indicated in comments that the new step (which adds about 30 seconds) should be removed with the Windows runners have Git 2.49.0. If this step is made permanent, then I think it would need to be refined, to make sure the sense of "newest" in which it retrieves the release associated with the newest tag is the sense that we actually want. As a temporary workaround for #1849, it seems to me that this is okay.

Ordinarily, when installing Git for Windows on CI, I would use PortableGit, or a package manager like Scoop or Chocolatey. However, the installation already present on the runners is an if by the Inno Setup installer (i.e. a non-portable non-SDK non-MinGit full installation of Git for Windows), including files that facilitate uninstallation and that other versions of the Inno Setup installer can detect and replace.

I don't want multiple Git installations on the runner, because they, or at least their bundled MSYS2 installations, can interact (https://github.com/GitoxideLabs/gitoxide/pull/1862#issuecomment-2692158831), which is not part of what `test-fixture-windows` seeks to test. I also do not want to accidentally end up only partially replacing the old installation. Therefore, I have done it by downloading and running the Inno Setup installer unattended. (If we continue upgrading Git for Windows on this CI job, then I could also look at how it is prepared for the runner images, and adjust how we do it accordingly.)

The new step precedes `actions/checkout` rather than following it, which is intentional. Using `git` less before attempting to upgrade it makes the upgrade more likely to succeed without problems, since on Windows there are difficulties replacing files that are in use. Also, if something is ever very badly wrong with the newly installed Git for Windows, then we are more likely to find out this way, because `actions/checkout` uses the new version.


EliahKagan:
I've rebased this to replace the change from cdee7ff (added in #1882) with the proposed technique here. But the approach in cdee7ff is what I would have suggested if you don't like the approach advocated here of changing the `git` version on the runner (in which case this could just not be merged).

Byron (approved):
Thanks so much, so sorry for the delay and the added chaos.

EliahKagan:
RC2 came out shortly after this was merged. Just as when RC1 came out this automatically used it instead of RC0, now that RC2 is out this automatically uses it instead of RC1 and [the tests still pass](https://github.com/EliahKagan/gitoxide/actions/runs/13786081179/job/38555777433#step:2:82).

> Thanks so much

No problem!

> so sorry for the delay and the added chaos.

I think this particular PR has actually benefited from the intervening events. This is because, although working with RC0, RC1, and RC2 suggests it will continue to work as long as we need it, if the technique it uses does turn out to be unsuitable then reverting this whole PR will give the probably next-best solution of listing the failing cases as being expected to fail.

Once Git for Windows 2.49.0 has a stable release *and* the Windows GHA runners have it, it may make sense to revert 4237e5a. Reverting the whole PR feels like it would be correct in that case but would not, since it would put the tests that shouldn't fail back in the list of tests expected to fail. But this is unlikely to cause trouble, because `test-fixtures-windows` checks for an exact match to the tests it expects to fail, so wrongly adding passing tests to the list would be caught (and would block auto-merge).

PR #1871, opened 2025-03-02
Fix Windows-related `-c` options in `match_baseline_files`

EliahKagan:
The `match_baseline_files.sh` fixture script adds entries to the index using `git update-index --add --cacheinfo`. These entries don't exist on disk, so it should be okay that some would cause problems on Windows if an attempt were made to actually create them. The tests that use the generated repository work on Windows.

It is intended that the fixture script also be able to run on Windows, and Windows-related `-c` options are passed in commands where they are expected to be needed: `core.ignoreCase=false` and `core.protectNTFS=false`.

But the fixture script nonetheless was failing on Windows because two commands that don't need `core.ignoreCase=false` but *do* need `core.protectNTFS=false` had `core.ignorecase=false` instead. This corrects that.

This fixes two of the long-standing `GIX_TEST_IGNORE_ARCHIVES=1` Windows failures reported in #1358:

- `gix-pathspec::pathspec search::files`
- `gix-pathspec::pathspec search::prefixes_are_always_case_sensitive`

This includes a regenerated archive (generated on Arch Linux with Git 2.48.1) and updates `etc/test-fixtures-windows-expected-failures-see-issue-1358.txt` to remove the two tests that now pass on Windows even with `GIX_TEST_IGNORE_ARCHIVES=1`.

The `test-fixtures-windows` job currently [fails here](https://github.com/GitoxideLabs/gitoxide/actions/runs/13612288266/job/38050999043?pr=1871) due to #1849, the same as on main if CI is rerun there, as described in https://github.com/GitoxideLabs/gitoxide/issues/1849#issuecomment-2692142316. If #1870 is merged, the failures should go away. This PR could be rebased after merging if #1870, if desired.

PR #1872, opened 2025-03-03
Clarify `pure-rust-build` cc1 wrapping

EliahKagan:
The main change here is to comment and refactor the additions made in #1682, to clarify what is going on. The fb67a78 and 04806da commit messages have further details. There is one part where I'm not sure if the way I've clarified it is the best way; I'll post something below for that so it can be reviewed.

This also includes some other comment improvement/maintenance in CI workflows: b856ad9 removes a comment that is obsolete since #1793, and ab049f6 makes clearer what `shell: bash` is for in the various places we use it, since it serves several different purposes.

(The failure here is in `test-fixtures-windows` and unrelated to the changes. That job is broken due to #1849, including on main, if rerun. As in #1871, this could be rebased after #1870 is merged, if desired. But unlike #1871, this does not make significant changes that affect other behavior of that job, so rebasing is probably not necessary to be confident that the failure isn't obscuring any separate problems from the changes here.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,16 +70,29 @@ jobs:
           ! grep -qP '(?<!\blinux-raw)-sys\b' tree.txt
       - name: Wrap cc1 (and cc1plus if present) to record calls
         run: |
+          set -o noclobber  # Catch any collisions with existing entries in /usr/local.
+
+          # Define the wrapper script for a compiler driver (for cc1 or cc1plus). This wrapper
+          # records calls, then delegates to the executable it wraps. When recording calls, writes
+          # to the log are synchronized so fragments of separate log entries aren't interleaved,
+          # even in concurrent runs. This wrapper knows what executable it is wrapping because,
+          # when deployed, this wrapper (or a symlink) replaces that executable, which will itself
+          # have been moved aside by being renamed with a `.orig` suffix, so this can call it.
           cat >/usr/local/bin/wrapper1 <<'EOF'
-          #!/bin/sh -e
+          #!/bin/sh
+          set -e
           printf '%s\n' "$0 $*" |
             flock /run/lock/wrapper1.fbd136bd-9b1b-448d-84a9-e18be53ae63c.lock \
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -68,31 +68,55 @@ jobs:
...
+          # Define a helper file that, when sourced, wires up the `~/display` symlink `wrapper1`
+          # uses to report calls as GitHub Actions step output (in addition to writing them to a
+          # log file). This is needed because stdout and stderr are both redirected elsewhere when
+          # the wrapper actually runs, and `/dev/tty` is not usable. This must be sourced in the
+          # same step as the `cargo` command that causes wrapped executables to be run, because
+          # different steps write to different pipe objects. (This also needs the shell that
+          # sourced it to remain running. But that is not the cause of the underlying limitation.)
+          cat >/usr/local/bin/set-display.sh <<'EOF'
+          ln -s -- "/proc/$$/fd/1" ~/display
+          EOF
```
EliahKagan, line 110: I'm not sure if it's clearer to present it this way, or to have the command in the subsequent step as it previously was, but just include this comment there. I lean slightly toward thinking this is clearer, considering that this step is defining other scripts anyway. But this file contains only one command, and it must be sourced rather than run. Unlike the scripts, the only reason it exists is so that it can be commented here, in this preparatory step where the comment makes the most sense.

If the comment is moved into the subsequent step, then it would be worded like this:

```sh
          # Wire up the `~/display` symlink `wrapper1` uses to report calls as GitHub Actions step
          # output (in addition to writing them to a log file). This is needed because stdout and
          # stderr are both redirected elsewhere when the wrapper actually runs, and `/dev/tty` is
          # not usable. This must occur in this step, because different steps write to different
          # pipe objects. (This also needs the shell that runs it to remain running. But that is
          # not the cause of the underlying limitation.)
```

Byron (approved):
I took a look and do believe the documentation is much clearer than it was before. Thanks a lot!

PR #1873, opened 2025-03-03
Add new feature zlib-rs

NobodyXu:
https://github.com/trifectatechfoundation/zlib-rs?tab=readme-ov-file#performance

zlib-rs is generally on par with zlib-ng and on sometimes faster https://trifectatech.org/blog/zlib-rs-is-faster-than-c/

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -246,9 +246,9 @@ dependencies = [
 
 [[package]]
 name = "async-std"
-version = "1.13.0"
+version = "1.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c634475f29802fde2b8f0b505b1bd00dfe4df7d4a000f0b36f7671197d5c3615"
+checksum = "730294c1c08c2e0f85759590518f6333f0d5a0a766a27d519c1b244c3dfd8a24"
 dependencies = [
  "async-attributes",
  "async-channel 1.9.0",
@@ -279,9 +279,9 @@ checksum = "8b75356056920673b02621b35afd0f7dda9306d03c79a30f5c56c44cf256e3de"
 
 [[package]]
 name = "async-trait"
-version = "0.1.87"
+version = "0.1.88"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d556ec1359574147ec0c4fc5eb525f3f23263a592b1a9c07e0a75b427de55c97"
+checksum = "e539d3fca749fcee5236ab05e93a52867dd549cc157c8cb7f99595f3cedffdb5"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -302,27 +302,25 @@ checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
 
 [[package]]
 name = "aws-lc-rs"
-version = "1.12.5"
+version = "1.12.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5e4e8200b9a4a5801a769d50eeabc05670fec7e959a8cb7a63a93e4e519942ae"
+checksum = "dabb68eb3a7aa08b46fddfd59a3d55c978243557a90ab804769f7e20e67d2b01"
 dependencies = [
  "aws-lc-sys",
- "paste",
  "zeroize",
 ]
 
 [[package]]
 name = "aws-lc-sys"
-version = "0.26.0"
+version = "0.27.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0f9dd2e03ee80ca2822dd6ea431163d2ef259f2066a4d6ccaca6d9dcb386aa43"
+checksum = "77926887776171ced7d662120a75998e444d3750c951abfe07f90da130514b1f"
 dependencies = [
  "bindgen",
  "cc",
  "cmake",
  "dunce",
  "fs_extra",
- "paste",
 ]
 
 [[package]]
@@ -916,9 +914,9 @@ checksum = "930c7171c8df9fb1782bdf9b918ed9ed2d33d1d22300abb754f9085bc48bf8e8"
 
 [[package]]
 name = "deranged"
-version = "0.3.11"
+version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b42b6fa04a440b495c8b04d0e71b707c585f83cb9cb28cf8cd0d976c315e31b4"
+checksum = "9c9e6a11ca8224451684bc0d7d5a7adbf8f2fd6887261a1cfc3c0432f9d4068e"
 dependencies = [
  "powerfmt",
 ]
@@ -1111,6 +1109,7 @@ checksum = "11faaf5a5236997af9848be0bef4db95824b1d534ebc64d0f0c6cf3e67bd38dc"
 dependencies = [
  "crc32fast",
  "libz-ng-sys",
+ "libz-rs-sys",
  "libz-sys",
  "miniz_oxide",
 ]
@@ -1123,9 +1122,9 @@ checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
 name = "foldhash"
-version = "0.1.4"
+version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a0d2fde1f7b3d48b8395d5f2de76c18a528bd6a9cdde438df747bfcba3e05d6f"
+checksum = "d9c4f5dac5e15c24eb999c26181a6ca40b39fe946cbe4c263c7209467bc83af2"
 
 [[package]]
 name = "foreign-types"
@@ -1281,14 +1280,16 @@ dependencies = [
 
 [[package]]
 name = "getrandom"
-version = "0.3.1"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43a49c392881ce6d5c3b8cb70f98717b7c07aabbdff06687b9030dbfbe2725f8"
+checksum = "73fea8450eea4bac3940448fb7ae50d91f034f941199fcd9d909a5a07aa455f0"
 dependencies = [
  "cfg-if",
+ "js-sys",
  "libc",
- "wasi 0.13.3+wasi-0.2.2",
- "windows-targets 0.52.6",
+ "r-efi",
+ "wasi 0.14.2+wasi-0.2.4",
+ "wasm-bindgen",
 ]
 
 [[package]]
@@ -1461,7 +1462,7 @@ dependencies = [
  "pretty_assertions",
  "serde",
  "thiserror 2.0.12",
- "winnow 0.7.3",
+ "winnow 0.7.4",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


NobodyXu:
I need some help with the windows CI, not sure why it's pulling libz-sys and libz-ng-sys when it already has libz-rs

Other CIs, except for MSRV, are not related to this PR

NobodyXu:
Tried checking using `cargo tree` for x86_64-pc-windows-msvc, but no luck

I have no clue what's going on here

emilazy:
This would be great for Jujutsu. We’ve held off on defaulting to `zlib-ng` because of packaging concerns. It could even go in `max-performance-safe`, I think, since it shouldn’t require a C compiler, or even be the default.

I’m not sure what’s up with Windows, but it looks kind of like both the `zlib` and `zlib-rs` features are getting enabled at once? If there’s appetite for this change, I can try and find the time to dig into it further.

I think the `justfile` will need adjusting, also.

Please let us know what we can do to help move this along, @Byron :)

gix-features/Cargo.toml
```diff
@@ -67,6 +66,8 @@ crc32 = ["dep:crc32fast"]
 zlib = ["dep:flate2", "flate2?/rust_backend", "dep:thiserror"]
```
emilazy, line 66: Should we be turning on `rust_backend` when using `zlib-rs`?

NobodyXu: The rust_backend is another pure rust implementation that is much much slower than other backend, it's only there as a fallback

emilazy: Right, but the `zlib-rs` feature pulls in the `zlib` feature, and enables the other Rust backend, which we probably don’t want? I’m wondering if that might somehow be part of the CI problems.

gix-archive/Cargo.toml
```diff
@@ -32,7 +32,7 @@ gix-object = { version = "^0.47.0", path = "../gix-object" }
 gix-path = { version = "^0.10.14", path = "../gix-path", optional = true }
 gix-date = { version = "^0.9.3", path = "../gix-date" }
 
-flate2 = { version = "1.0.33", optional = true }
+flate2 = { version = "1.0.33", optional = true, default-features = false }
```
Byron, line 35: I don't think that `default-features` is a problem - these, i.e. the `rust-backend` are superseded by any other choice.

gix-features/Cargo.toml
```diff
@@ -67,6 +66,8 @@ crc32 = ["dep:crc32fast"]
...
 ## Use the C-based zlib-ng backend, which can compress and decompress significantly faster.
 zlib-ng = ["zlib", "flate2?/zlib-ng"]
+## Use the rust-based zlib backend, which can compress and decompress significantly faster and is on par with zlib-ng or even faster.
+zlib-rs = ["zlib", "flate2?/zlib-rs"]
```
Byron, line 70: It's great to see this here!

gix/Cargo.toml
```diff
@@ -252,9 +252,9 @@ pack-cache-lru-static = ["gix-pack/pack-cache-lru-static"]
 ## Provide a hash-map based LRU cache whose eviction is based a memory cap calculated from object data.
 pack-cache-lru-dynamic = ["gix-pack/pack-cache-lru-dynamic"]
 
-## Activate other features that maximize performance, like usage of threads, `zlib-ng` and access to caching in object databases.
+## Activate other features that maximize performance, like usage of threads, `zlib-rs` and access to caching in object databases.
 ## Note that some platforms might suffer from compile failures, which is when `max-performance-safe` should be used.
-max-performance = ["max-performance-safe", "zlib-ng", "fast-sha1"]
+max-performance = ["max-performance-safe", "zlib-rs", "fast-sha1"]
```
Byron, line 257: I don't want to push my luck 😅, and would prefer to start by merely making the use of `zlib-rs` easy with the provided feature.

Byron (changes requested):
Thanks for making this happen, and apologies for the late response.

I am still catching up with what accumulated over just 7 days, it's going to be a while.

Meantime, I think this PR needs changes, but nothing huge or in the way of a merge.
Unfortunately, I also don't know why 32bit ARM and Windows suffer from duplicate definitions. There is [this one issue](https://github.com/rust-lang/flate2-rs/issues/478) of the same happening on Linux, but there it seems to be about pulling in multiple versions of `libz-sys`.

To get this PR going, I think the default performance backend should remain `zlib-ng`, but `jj` for instance could configure their unix builds to use `zlib-rs` more easily in future.

@emilazy is cordially invited to help shape this PR so `gitoxide` would work better for `jj`.

What do you think?

gix/Cargo.toml
NobodyXu: Maybe another feature then?

In cargo-binstall we want to get rids of zlib-ng as we already pulled in zlib-rs, having zlib-ng actually makes cross compilation harder

Byron:
And there even is [a test](https://github.com/GitoxideLabs/gitoxide/blob/2d32701165b016b371c9947d0656d3abd4056796/.github/workflows/ci.yml#L24) for that, so it should be straightforward to validate that indeed no C-compiler is needed anymore.

Also, I'd be happy to reach that goal, and if it can be achieved earlier, that's better, too. 

For now it seems the easiest way to pacify CI is to undo the default backend change. This buys time for figuring out what's actually going on there.

My thinking is that ideally, `flate2` won't have to rely on the C bindings of `zlib-rs` either, it's a Rust library and there shouldn't be conflicts on the symbol level. Even if something somewhere in the dependency graph uses the C version of Zlib, `flate2` shouldn't be affected by it. In case this is feasible, CC @folkertdev .

folkertdev:
I believe the problem is this

```
│   │   │   ├── gix-transport v0.45.0 (/tmp/gitoxide/gix-transport)
│   │   │   │   ├── base64 v0.22.1
│   │   │   │   ├── bstr v1.11.3 (*)
│   │   │   │   ├── curl v0.4.47
│   │   │   │   │   ├── curl-sys v0.4.80+curl-8.12.1
│   │   │   │   │   │   ├── libc v0.2.170
│   │   │   │   │   │   ├── libz-sys v1.1.21
│   │   │   │   │   │   │   └── libc v0.2.170
```

I suspect that is hard to work around (though using `reqwest` and `rustls` does sound like the better long-term solution than `curl`). So for the immediate future prefixing the symbols is the quickest solution.

We do want to provide a nice rust interface. The brave rustls folks already use our rust interface, but we basically break it on every release, so it's not something I'd recommend. As it stands, creating a good rust interface would require some funding.

Cargo.lock
```diff
@@ -246,9 +246,9 @@ dependencies = [
 
 [[package]]
 name = "async-std"
-version = "1.13.0"
+version = "1.13.1"
```
NobodyXu, line 249: @Byron there was conflicts in the lockfile, I ended up deleting it and recreating it using `cargo update` and that bumps a lot of dependencies (which I think is a good thing to do periodically)

gix-archive/Cargo.toml
```diff
@@ -33,7 +33,9 @@ gix-path = { version = "^0.10.14", path = "../gix-path", optional = true }
 gix-date = { version = "^0.9.3", path = "../gix-date" }
 
 flate2 = { version = "1.0.33", optional = true }
-zip = { version = "2.3.0", optional = true, default-features = false, features = ["deflate"] }
+zip = { version = "2.3.0", optional = true, default-features = false, features = [
```
NobodyXu, line 36: toml formatting from zed, I can remove it if you don't like it @Byron

gix/Cargo.toml
NobodyXu: @Byron reverted this change

erikjee:
@Byron I'm the person at Trifecta Tech Foundation (where zlib-rs lives) blessed with fundraising duties. 
Last summer we applied to Sovereign Tech for zlib-rs (and zstd) but that got rejected. 
Applying together might have better chances, either at STA or another fund. Please let me know if you want to talk about that.

So far, funding zlib-rs has been difficult and only partly successful because ~~everyone seems to like the free stuff better~~ I haven't explained its relevance well enough.

gix-archive/Cargo.toml
Byron: I like it and wish my toolchain would do that consistently. If `cargo fmt` would also deal with manifests I'd be very happy.

Byron (approved):
Thanks for the updates, it's ready to go now!

@erikjee Will you be at RustWeek, because I am considering to go. If I go, I think with the right amount of support ("You can do it" [patting shoulder]) I could finally go through with the application for `gitoxide` and just get it done while there. Applications might be closed already, but I'd hope they reopen in a couple of months.
Strangely, despite being in a good spot, the more successful `gitoxide` is, the less time I have to actually work on it, and funded outside help could definitely fix that. And yes, why not make `zlib-rs` a sub-project of that application.

erikjee:
@Byron 

> Will you be at RustWeek, because I am considering to go. 

I'm one of the organizers :) So, yes!
Ready to pat you on the back etc. The unconf would be best, because I'll actually have some time. I think Folkert already sent you an email inviting you to join. I'll reply there.

emilazy:
> We do want to provide a nice rust interface. The brave rustls folks already use our rust interface, but we basically break it on every release, so it's not something I'd recommend. As it stands, creating a good rust interface would require some funding.

Just a thought, but maybe you could provide a Rust API that is 1:1 with the C API first? That is, define the C API with non-`#[extern]` Rust definitions, and then have `extern "C"`s with the same signatures that just forward on the calls – or even just have one set of definitions with `#[cfg_attr]`s. That way, a Rust package like `flate2` that wants to avoid polluting the symbol table can use the C API directly from Rust, without anyone getting blocked on API design concerns.

NobodyXu:
> Just a thought, but maybe you could provide a Rust API that is 1:1 with the C API first?

I think this is the status quo?

folkertdev:
No, we don't do this part

>  define the C API with non-`#[extern]` Rust definitions

That's an interesting idea, I'll have to think on that a bit

emilazy:
Not quite. If you look at https://docs.rs/libz-rs-sys/latest/libz_rs_sys/fn.deflate.html, it’s `extern "C-unwind"` and has `#[export_name = "deflate"]`. That is what results in the symbol conflict we see here. It could instead work like this:

```rust
pub unsafe extern "C-unwind" fn deflate(strm: *mut z_stream, flush: i32) -> c_int {
    …
}

#[cfg(feature = "c-api")]
#[export_name = "deflate"]
pub unsafe extern "C-unwind" fn export_deflate(strm: *mut z_stream, flush: i32) -> c_int {
  deflate(strm, flush)
}
```

(Actually I guess this *very slightly* changes semantics, because `&deflate` won’t necessarily be the same function pointer seen from C. But it probably doesn’t matter too much. You could also keep the `extern` but just `#[cfg_attr(not(feature = "c-api"), export_name = "zlib_rs_deflate")]`, though that’s a bit of a hack.)

emilazy:
Hmm, actually it looks like there’s already a `custom-prefix` feature! But of course that has the problem of breaking crates which *do* need the C API because of features being additive, and needing wiring up externally by packaging. Still, that means it should be technically possible to build something that uses both `curl` and `zlib-rs` already. But I think my proposal of splitting the exports (or just using a macro to conditionalize the `extern` on a crate feature) is probably best.

folkertdev:
we've added a `export-symbols` feature flag now (https://github.com/trifectatechfoundation/zlib-rs/pull/322), which will be part of the next release. Thanks @emilazy for that suggestion!

PR #1875, opened 2025-03-04
Minor clarity fixes in `gix-commitgraph`

EliahKagan:
This fixes a misspelled test name, and simplifies a documentation URL that had a long `#:~:text` anchor that no longer worked even in browsers that support them and that (fortunately) is not needed for the page to be useful.

This is pretty minor, but it doesn't seem to naturally go together with any other PRs I have open or plan to open anytime soon. So it is its own PR.

(As elsewhere, the `test-fixtures-windows` failure here is unrelated to this PR; see #1849 and #1870. The `test-32bit` failure is likewise unrelated; see #1874.)


Byron (approved):
Thanks a lot!

PR #1876, opened 2025-03-04
Fix tests when `GIT_AUTHOR_NAME` or `GIT_COMMITTER_NAME` are set

joshtriplett:
Some of the tests failed to unset these variables, so they'd fail in an
environment with these variables set (such as on a developer system that
sets them).


Byron (approved):
Thanks a lot for the fix, and sorry for the huge delay in getting this reviewed!

Still catching up with a pile of emails that appeared after not touching a computer for 7 days. Lesson thoroughly learned :D.

Issue #1881, opened 2025-03-10
Lossy configuration on release builds is a bad default


emilazy:
### Current behavior 😯

https://github.com/GitoxideLabs/gitoxide/blob/d0ef276302aad10fda222236c69e78e580215f31/gix/src/config/cache/util.rs#L19-L25

We just ran into our tests failing in Jujutsu only on release builds, because our snapshots of the generated configuration suddenly lost all their whitespace. I was briefly wondering if some kind of bizarre compiler bug might be involved.

### Expected behavior 🤔

Please reconsider this behaviour!

Release builds should not affect functionality, only performance and the presence of debug checks. Any behavioural change should be handled through feature flags. (From a quick search, https://github.com/GitoxideLabs/gitoxide/blob/d0ef276302aad10fda222236c69e78e580215f31/gix-transport/src/client/blocking_io/http/mod.rs#L300-L305 also looks questionable from this perspective.)

But really, I think modifying repository configuration with `gix` and writing it back out again should not silently throw away user data in general, even with a feature flag. If it affects lookup performance I think using a different data structure is the correct approach, not stripping out whitespace and comments.

### Git behavior

Git doesn’t clobber configuration in this way.

### Steps to reproduce 🕹

https://gist.github.com/demize/8a468be11b3b19069efca37323c352a2#file-minimum-reproducible-example-rs is not a minimal reproducer (just taking a snapshot and writing it back out suffices), but it works to demonstrate the problem.

Byron:
Thanks for bringing this to my attention. It's one of these late-night ideas that somehow stuck and survived till now.

Would you want to remove this special case or get its removal started in a PR? If it turns out to be more work, I can take it from there.

*Please beware of slow(er) responses these days, I am looking at a massive pile of email + backlog from before that after 7 days "off"*.

emilazy:
I’ve opened https://github.com/GitoxideLabs/gitoxide/pull/1882. Thanks for being open to a change here and sorry for my somewhat provocative report; I was very surprised when we tracked down this failure and felt the need to put some of my strength of feeling into it :)

Byron:
No worries at all, I kind of failed to recognise any provocation, or else I would have been more sympathetic in my reply. After all, I am quite aware of the pain merely by observing some of it first hand when threading on new grounds as a user of the crate. Some portions are a delight, many are not, unfortunately.

And now I hope there isn't any more gems like this hidden in the codebase 😅.

PR #1882, opened 2025-03-10
Don’t change semantic behaviour based on `debug_assertions`

emilazy:
I believe these are the only two cases where release builds are conditioned on for something other than actual debug checks. Let me know if you want to add more exclamation marks to these commit messages; compiler‐flag‐dependent things are kind of borderline for breaking‐change‐ness in my view.

Closes: #1881

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -390,6 +390,7 @@ jobs:
         cargo-deny-advisories
         wasm
         tests-pass
+        test-32bit
 
     defaults:
       run:
@@ -424,7 +425,6 @@ jobs:
       - test-journey
       - test-fast
       - test-fixtures-windows
-      - test-32bit
       - lint
       - cargo-deny
       - check-packetline
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,10 +24,9 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
- "getrandom",
  "once_cell",
  "version_check",
- "zerocopy",
+ "zerocopy 0.7.35",
 ]
 
 [[package]]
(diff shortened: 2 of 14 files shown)


Byron (approved):
Thanks so much for your help with this! 

I hope CI will pass now so the fix will merge automatically.

Byron:
@EliahKagan It seems the test-32bit job is failing at least temporarily so I removed it from the 'needed to auto-merge' list. For now I didn't investigate what's going on there and probably won't get to it anytime soon.

EliahKagan:
@Byron The `test-32bit` failure is due to changes in `rustup`. Merging #1874 would fix it.

The `test-fixtures-windows` failure is due to #1849. Merging #1870 (either as is or, if you don't prefer the approach there, then modified with a different technique) would fix it.

But I think it's fine to put that off until some point after this PR is merged. The jobs can be put back as dependencies of `tests-pass` later. I can update those PRs to do that too.

Also, #1878 and #1879 were for two of the RUSTSEC advisories that caused `cargo deny advisories` to fail here. But unlike #1870 and #1874, I suspect that #1878 and #1879 have no advantages over the relevant changes here, and that #1878 and #1879 could thus be closed as superseded after this PR is merged.

EliahKagan:
> Edit2: Apologies, @EliahKagan , I realize that all of these issues are already fixed with PRs waiting, and I should rather have processed my emails in order a little more. Even though this will merge, I will merge all of your pending PRs next and fix merge-conflicts should they arise after this PR merges.

@Byron Now that this is merged, I can rebase and update #1849 and #1874 right now, which I think might make things easier and lead to a more readable history. My guess is that this would be considered valuable, so please ping me if you do *not* want me to do that (so I would know not to).

EliahKagan:
> Now that this repository is in an organization, permissions are much more fine-grained. If you have suggestions on how to make your life easier, I'd be happy to make adjustments.

Thanks--I've opened #1883 to ask if the ability to interact with Dependabot is something that can and should be delegated. (That is, I have moved this aspect of the conversation from here to that new question, in case it becomes complex.)

PR #1884, opened 2025-03-12
various improvements

Byron:
Various improvements in support of https://github.com/gitbutlerapp/gitbutler/pull/7597.


gix-fs/tests/fs/snapshot.rs
```diff
@@ -0,0 +1,54 @@
...
+    let snap = smut.recent_snapshot(check, open)?;
+    assert!(snap.is_none());
+
+    std::fs::write(&file_path, "content")?;
+    let snap = smut.recent_snapshot(check, open)?.expect("content read");
+    assert_eq!(&**snap, "content", "it read the file for the first time");
+
+    std::fs::write(&file_path, "change")?;
+    let snap = smut.recent_snapshot(check, open)?.expect("content read");
+    assert_eq!(&**snap, "change", "it picks up the change");
```
EliahKagan, line 31: Although the test *usually* passes, this assertion fails occasionally. This has occurred a number of times on CI. It has not (yet?) occurred on macOS, though I wonder if the reason is just that we have fewer macOS jobs on CI. See #1896 for details.

PR #1888, opened 2025-03-16
Respect `diff.algorithm` in `gix blame`

cruessler:
Currently, the diff algorithm used for diffing while running blame is hard-coded to `Histogram` in `blame::file()` (or, more specifically, in `blob_changes()`). With this PR, `diff_algorithm` becomes an option to be provided to `blame::file()`. For `gix blame`, the value of `diff.algorithm` as returned by `Repository::diff_algorithm()` is used.

- feat!: add `diff_algorithm` to `blame::file()`
- Adapt to changes in `gix-blame`


Byron (approved):
That makes sense, thanks a lot!

Issue #1894, opened 2025-03-18
armv7-linux-androideabi cross test "attempt to subtract with overflow" in miniz_oxide 0.8.5


EliahKagan:
#### Background

cf7f34d (#1882) bumped `prodash` from 29.0.0 to 29.0.1 in `Cargo.toml`, as well as various resolved dependency versions in `Cargo.lock`, which included bumping `miniz_oxide` from 0.8.0 to 0.8.5. 

Around that time and starting somewhat earlier, I was experimenting with running tests via [`cross`](https://crates.io/crates/cross) in the hope of finding a way to reproduce some `s390x-unknown-linux-gnu` failures without s390x hardware. As noted in #1890, I was unable to reproduce those failing s390x tests that way, and I still have not managed to do so.

However, I also experimented with another target in `cross`, with the intention that it would help me understand what failures were target-specific. I didn't get failures in s390x, but I expected I would and that the other target could serve as a kind of control. It turned out that s390x was the control and the other target eventually had unexpected failures.

Specifically, for the other target, I chose `armv7-linux-androideabi`. I chose this because it had been partially tested via `cross` on CI before the test was replaced with a more comprehensive non-Android arm32v7 containerized (but non-`cross`) test on ARM hardware in #1777.

In its past testing on CI, `armv7-linux-androideabi` had problems using the external `git` command. Upgrading `git` in the `cross` container and patching the runner script to avoid preloading an unusable shared library overcame this, and I was able to run all tests with `cross` using that target.

All tests all passed--aside from one fuzzing-related performance test that had to be skipped due to slowdown from QEMU emulation of an unrelated architecture without hardware acceleration, and a couple of cases where tests were marked with the wrong `#[cfg(...)]`, which was easily fixed. But...

#### Then `prodash`, and various locked dependencies, were upgraded

Starting after a rebase that included cf7f34d in the history, several test cases failed in `cross` runs on the `armv7-linux-androideabi` target. In the recent run presented here, this is eight tests. Running `git revert --no-commit cf7f34d` and rerunning the tests confirmed that this change is what triggered the failures.

[...]

Issue #1890, opened 2025-03-17
4 pack-related tests fail on s390x


EliahKagan:
### Current behavior 😯

On an s390x system running Ubuntu 22.04 LTS with git 2.48.1 installed from the git-core PPA, two tests fail when using committed fixture archives, and those tests as well as two others fail when ignoring archives.

Full test output of the most recent runs can be viewed [in this gist](https://gist.github.com/EliahKagan/b5ba74fcc3b2658cf88f09a8aaac0e96). The same failures have been happening for a while; that gist links to a few older gists. This is not new as of git 2.48.1; it is the same problem as observed with git 2.47.0 (also installed through the git-core PPA) as mentioned in #1687 and discusses in comments in #1698, starting at https://github.com/GitoxideLabs/gitoxide/pull/1698#issuecomment-2495663132.

Without `GIX_TEST_IGNORE_ARCHIVES cargo nextest run --workspace --no-fail-fast`, the failing tests are:

```text
gix::gix remote::fetch::blocking_and_async_io::fetch_pack
gix::gix remote::fetch::blocking_and_async_io::fetch_pack_without_local_destination
```

With `GIX_TEST_IGNORE_ARCHIVES=1 cargo nextest run --workspace --no-fail-fast`, the failing tests are:

```text
gix::gix remote::fetch::blocking_and_async_io::fetch_pack
gix::gix remote::fetch::blocking_and_async_io::fetch_pack_without_local_destination
gix-pack-tests::pack pack::multi_index::access::general
gix-pack-tests::pack pack::multi_index::verify::integrity
```

Notably, all these failing tests relate to packs. But various other tests related to packs do not fail.

Rerunning the first two failing tests (those that occur even when using committed archives) by themselves and enabling a backtrace shows:

```text
ubuntu@gull:~/repos/gitoxide (main %=)$ git restore .
ubuntu@gull:~/repos/gitoxide (main %=)$ gix clean -xd -m *generated* -e
removing gix/tests/fixtures/generated-archives/make_fetch_repos__home_ubuntu_repos_gitoxide_gix_tests_fixtures_generated-do-not-edit_make_remote_repos_1456912905-unix_base.tar
removing gix/tests/fixtures/generated-do-not-edit/ (🗑️)
ubuntu@gull:~/repos/gitoxide (main =)$ RUST_BACKTRACE=1 cargo nextest run -p gix --features blocking-network-client --no-fail-fast fetch_pack
[...]

PR #1895, opened 2025-03-18
Set up `cross` for limited but substantial local testing of s390x and Android

EliahKagan:
This allows local testing of some targets that are interesting but may not be justified to test--or at least not very extensively--on CI, given the tests we already have. The targets are `s390x-unknown-linux-gnu` and `armv7-linux-androideabi`, though it's hoped that the same `cross` configuration and `Dockerfile` will work with various other targets too.

#### Why this does extend CI, for now

Right now, it does attempt some CI checks, but this is mostly just as part of the process of working on the PR. I am not sure if new CI checks are part of what I will ultimately propose here once this is ready to review. Even if so, they might be pared down, so as to build and test less, and therefore take less time.

#### Relationship to #1890

I had originally hoped that this would allow #1890 to be reproduced without s390x hardware. Currently this does not manage that. I have another feature branch on this feature branch, which I shall eventually get to the point of merging into this feature branch and making part of this PR (whereupon this PR might be ready, I am not sure). But even those forthcoming commits, which include s390x-related improvements such as using the s390x build of `git` and most of its dependencies, has not produced the failures.

It seems to me that this will eventually be worth merging even if it never manages to reproduce that, but it might not be worth having any CI for in that case.

#### Relationship to #1894

This did find a regression, which may be a bug in gitoxide, brought about by dependency updates in cf7f34d (#1882), in the `armv7-linux-androideabi`. I've reported this with extensive details in #1894.

The hope that this could help showcase that with CI even if the CI jobs are ultimately removed before merging is really what led me to add CI jobs here. But another reason is for them to look for ways I might be depending on my local configuration inadvertently.

#### Ways QEMU is invoked

One way I am depending on my local configuration inadvertently is that `binfmt_misc` needs to be configured in the local kernel to run QEMU automatically. I have this set up, but it seems not to be set up on CI.

This is not needed for running test binaries, because `cross` uses a runner script in the container that conditionally runs an appropriate QEMU emulator explicitly. But it is needed to run target-architecture binaries directly inside the container, which is something the tests *also* need to be able to do:

- With the configuration set up here, they need to be able to do it so they can run the `arrow` process filter. For s390x, in the commits I plan to merge into here, they also need to be able to do it so they can run an s390x build of `git`.
- With the commits currently in this PR, an x86-64 build of `git` is used in both the Android `cross` container, where no native `git` is readily available, and in the s390x `cross` container, where an s390x `git` is readily available but involves some subtleties to set up.

So that's something I intend to improve. Then even if the new CI jobs are removed, I can document the issue precisely. (We can set up `binfmt_misc` on CI, but we likely should not attempt to do it locally through the `justfile`, since it is a privileged operation.)

#### Improvements to the actual test suite

In addition to attempting to provide further ways of testing locally, this also does include some actual bug fixes, albeit only in the test suite, not in the code under test. For Android, one test has the wrong `#[cfg(...)]`, which had to be fixed to let it pass. Another had an analogous problem, which was useful to fix since it let it run rather than wrongly be skipped. Some of the error reporting I had added for a test helper in a previous PR was also limited in ways that made it hard to understand test output in some situations, which is improved.

Also, there are considerable subtleties in what environment variables get through `cross`, because they are filtered or modified at multiple points, so this adds an `internal-tools` subcommand to view them in a way that can easily be run with `cargo` or `cross`, which may occasionally be helpful to see how `cargo` modifies the environment, but is something I found especially helpful to see how `cross` modifies the environment.


EliahKagan:
This isn't ready yet. At minimum, even if nothing else is to be accomplished here, the CI jobs should be removed (or made much more minimal?).

But it might be valuable even if it doesn't give any more insight into #1890 or #1894 than it already has. It is already usable locally.

For me to decide what more to do on this and when, I'm curious if it would be considered of value to merge, with the CI parts removed. (The non-obvious CI things, like enabling `binfmt_misc` and why, can be moved to comments here, as well as remaining in the history unless you prefer that the CI commits themselves be dropped.)

If so, I could do that while keeping a branch with CI to reproduce any demonstrate further locally obtained insights on #1890 or #1894, in the event that any arise. If not, then I would want to consider something far more limited than this with CI removed: there are minor bugfixes to the test suite included here--discovered when running `just cross-test-android` locally--that I think should come in.

Byron (approved):
Thanks for reeling me in, and thanks for your incredible help in making this codebase more portable!

This PR represents considerable research and I'd want it all preserved, so let's keep the commits as is and put alterations, like the removal of the CI job on top.
From there it can be merge when deemed ready.

PR #1899, opened 2025-03-20
Improve filesystem timestamp granularity check

### LINKED ISSUE #1896: `FileSnapshot::into_owned_or_cloned` intermittently fails according to its `journey` test
### Current behavior 😯

The [`gix-fs::fs snapshot::journey`](https://github.com/GitoxideLabs/gitoxide/blob/0bf1d5b9f0b0971b9f25a8e44b7818e37c78d68e/gix-fs/tests/fs/snapshot.rs#L4-L42) test case was introduced in f93aa61 (#1884) along with the [`into_owned_or_cloned`](https://github.com/GitoxideLabs/gitoxide/blob/0bf1d5b9f0b0971b9f25a8e44b7818e37c78d68e/gix-fs/src/snapshot.rs#L43-L49) method that it tests. That test case fails intermittently.

This was first observed in #1892, but it is not related to the changes in that PR. It seems to happen occasionally on all platforms, though I have not found it to fail on macOS, only on Ubuntu and Windows. I suspect that this is merely because we have fewer macOS jobs than Ubuntu or Windows jobs, but I do not know that to be why it hasn't happened on macOS. All failures observed so far have been on CI. I have not experimented locally with it.

Although #1892 makes a change to a Windows job that is one of the jobs where the failure was observed, the change is made for its future effect, and it does not change what software currently gets installed. More importantly, the observed failure in that job on Windows (as well as in another separate Windows job) is from *before* that PR. The failures observed in that PR have been on Ubuntu (and also in multiple jobs).

Besides being observed in that PR, I have observed failures on the main branch of my fork, where I have rerun jobs to check for it, and unintentionally in #1864 after a rebase that causes it to have f93aa61 in its history. It does not appear that failure is more or less likely on any branch compared to any other branch (for branches that have the test in their history).

The failures look like:

```text
        FAIL [   0.005s] gix-fs::fs snapshot::journey
──── STDOUT:             gix-fs::fs snapshot::journey

running 1 test
test snapshot::journey ... FAILED

failures:

failures:
    snapshot::journey

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 26 filtered out; finished in 0.00s

──── STDERR:             gix-fs::fs snapshot::journey

thread 'snapshot::journey' panicked at gix-fs/tests/fs/snapshot.rs:31:5:
assertion `left == right` failed: it picks up the change
  left: "content"
 right: "change"
```

The failing part is:

https://github.com/GitoxideLabs/gitoxide/blob/0bf1d5b9f0b0971b9f25a8e44b7818e37c78d68e/gix-fs/tests/fs/snapshot.rs#L29-L31

Specifically, that `assert_eq!` fails, due to the test observing the old contents `content`, rather than the new contents `change`.

### Expected behavior 🤔

The test should pass. I don't know if the problem is the test or the code under test though. It seems like the test straightforwardly asserts that any write that is completed from the perspective of Rust code that runs in sequence with a subsequent access to the snapshot should be observed in that subsequent access. If so, then that suggests the bug could be in the code under test.

### Git behavior

Not applicable.

### Steps to reproduce 🕹

Running the test repeatedly seems to work to reproduce it, at least on CI. Usually the test passes but sometimes it fails.

Observed failures on CI so far, in the order they have been observed, have been:

- https://github.com/EliahKagan/gitoxide/actions/runs/13918804232/job/38947009079#step:10:1326 \
  in `test-32bit` on `arm32v7`

- https://github.com/GitoxideLabs/gitoxide/actions/runs/13918804658/job/38948596656?pr=1892#step:8:1283 \
  in `test-fast` on `ubuntu-latest`

- https://github.com/GitoxideLabs/gitoxide/actions/runs/13918804658/job/38954967812?pr=1892#step:10:1326 \
  in `test-32bit` on `i386`

- https://github.com/EliahKagan/gitoxide/actions/runs/13913874365/job/38956318291#step:7:1373 \
  in `test-fixtures-windows`

- https://github.com/EliahKagan/gitoxide/actions/runs/13913874365/job/38963529809#step:10:1324 \
  in `test-32bit` on `arm32v7`

- https://github.com/EliahKagan/gitoxide/actions/runs/13913874365/job/38970448502#step:8:1333 \
  in `test-fast` on `windows-latest`

- https://github.com/GitoxideLabs/gitoxide/actions/runs/13927061201/job/38974254636?pr=1864#step:8:1284 \
  in `test-fast` on `ubuntu-24.04-arm`

I originally "reported" this in https://github.com/GitoxideLabs/gitoxide/pull/1892#issuecomment-2732206834 but I figured it should have an issue, since its connection to that PR is coincidental.

EliahKagan:
Fixes #1896

This avoids false positives for the check of whether we can proceed to run the filesystem snapshot `journey` test, by making a hundred files about as fast as we can using ordinary non-parallelized techniques, then checking if all their timestamps are distinct.

This carries a small risk of false negatives, but experiments on macOS, where the `journey` test is typically able to run, suggest that the rate of false negatives will be low. (False positives are also, in theory, possible, but should be extremely rare or never happen.)

#### Distinct timestamp experiment

First, I experimented outside of CI on Ubuntu, Windows, and macOS, initially with the code shown in https://github.com/GitoxideLabs/gitoxide/pull/1897#pullrequestreview-2699690027, but after that with this test program:

```rust
use std::io::Result;
use std::path::Path;
use std::time::{Instant, SystemTime};

fn main() -> Result<()> {
    let reps = 50;
    let n = 100;

    let mut uniques = Vec::new();
    for _ in 0..reps {
        uniques.push(run_experiment(n)?);
    }
    println!("{:?}", uniques);
    println!("{}, {}", uniques.contains(&1), uniques.contains(&n));
    Ok(())
}

fn run_experiment(n: u32) -> Result<u32> {
    let start = Instant::now();
    let tmp = tempfile::tempdir().unwrap();
    let times = unique_times(n, tmp.path())?;
    let finish = Instant::now();
    println!("{} / {}", times.len(), n);
    println!("{times:?}");
    println!("Experiment took {:?}.", finish - start);
    Ok(times.len().try_into().expect("it's out of n, which is u32"))
}

fn unique_times(n: u32, root: &Path) -> Result<Vec<SystemTime>> {
    let names: Vec<_> = (0..n).map(|i| format!("{i:03}")).collect();
    for name in &names {
        std::fs::write(root.join(name), name)?;
    }
    let mut times = Vec::new();
    for name in names {
        times.push(root.join(name).symlink_metadata()?.modified()?);
    }
    times.sort();
    times.dedup();
    Ok(times)
}
```

Running that locally, the last two lines of output on Ubuntu were:

```text
[4, 4, 4, 3, 3, 3, 3, 3, 4, 4, 3, 3, 3, 3, 3, 4, 4, 5, 4, 3, 4, 3, 3, 3, 3, 3, 3, 3, 3, 3, 4, 3, 3, 3, 4, 3, 4, 4, 3, 3, 4, 4, 3, 3, 4, 4, 4, 4, 4, 3]
false, false
```

And on Windows (which has 100 ns representation granularity, and which uses a temporary directory that is not on a separate `tmpfs`-like filesystem):

```text
[25, 25, 25, 24, 24, 25, 26, 25, 24, 24, 25, 26, 26, 24, 26, 25, 25, 24, 24, 23, 25, 25, 24, 23, 23, 24, 24, 26, 28, 26, 24, 23, 24, 23, 23, 23, 24, 23, 24, 25, 23, 23, 24, 24, 24, 30, 25, 26, 24, 23]                                        false, false
```

And on macOS, which has 1 ns granularity at least on the current default filesystem:

```text
[100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100]                                                                                                              false, true
```

For full details, including full output, see [this repository](https://github.com/EliahKagan/granularity).

This is not guaranteed to be that way on all such systems. The use of a different filesystem type or device type could potentially confer high granularity on a system that ordinarily doesn't have it, or result in diminished granularity on a system where it ordinarily is high. If it were guaranteed, then actually checking this would be overkill because it would be sufficient to enable the test on macOS but not other operating systems. That is expected to be what happens on CI as it is currently set up, as well as in most local environments, but not necessarily in all local environments. Therefore:

- I think this is better than assuming which platforms work, though it might make sense to have CI fail if the check returns `false` on macOS.

- Before considering this ready, I'm going to do that at least temporarily, to check that my non-CI macOS experiments are similar to CI (i.e. to make sure the test it least sometimes actually running). That's why this is still a draft.

  *Edit:* I've done that. The passing test with this macOS+CI specific assertion when repeating the whole test 250 times can be seen [in this run](https://github.com/EliahKagan/gitoxide/actions/runs/13959179656/job/39077179279).

- Although I've marked this as fixing the issue, it might be better--even if merging this--to view it as a workaround in the hope of developing a better fix later, more along the lines of what you suggested in [#1897 (comment)](https://github.com/GitoxideLabs/gitoxide/pull/1897#discussion_r2004448543). That way, the test could run on all platforms we test on CI, and most other platforms anyone would want to run it on, too.

  On the other hand, even then, it might be considered valuable to avoid sleeping during the test if the granularity is high enough that it isn't necessary. (I don't mean to avoid the delay, but rather with the idea that the test might be slightly more robust.) Also, the technique used here should be possible to adapt to infer the timestamp granularity.

#### CI experiment

The actual fix here is in d4a8d79, but I also changed the test so it runs 250 times instead of once (including separate calls to the helper function each time) before that commit, and changed it back afterwards. This was to make sure that the reason I didn't see failures was that this really does eliminate the failures (or at least make them extremely rare), rather than by chance, since the failures did not always happen before.

[This workflow output on my fork](https://github.com/EliahKagan/gitoxide/actions/runs/13958357773/job/39074728541) shows that the tests passed even before it was changed back from running 250 times to one time. Please note also that this 250-times repetition is of the whole test--it is not related to the "inner" repetition of creating 100 files.

### DIFF
--- a/gix-fs/tests/fs/snapshot.rs
+++ b/gix-fs/tests/fs/snapshot.rs
@@ -3,13 +3,6 @@ use std::path::Path;
 
 #[test]
 fn journey() -> Result<(), Box<dyn std::error::Error>> {
-    for _ in 0..250 {
-        do_journey()?;
-    }
-    Ok(())
-}
-
-fn do_journey() -> Result<(), Box<dyn std::error::Error>> {
     let tmp = tempfile::tempdir().unwrap();
     if !has_granular_times(tmp.path())? {
         return Ok(());
--- a/gix-fs/tests/fs/snapshot.rs
+++ b/gix-fs/tests/fs/snapshot.rs
@@ -64,5 +64,9 @@ fn has_granular_times(root: &Path) -> std::io::Result<bool> {
 
     // This could be wrongly false if a filesystem has very precise timings yet is ridiculously
     // fast. Then the `journey` test wouldn't run, though it could. But that's OK, and unlikely.
+    // However, for now, on CI, on macOS only, we assert the expectation of high granularity.
+    if cfg!(target_os = "macos") && is_ci::cached() {
+        assert_eq!(times.len(), n);
+    }
     Ok(times.len() == n)
 }
--- a/gix-fs/tests/fs/snapshot.rs
+++ b/gix-fs/tests/fs/snapshot.rs
@@ -49,11 +49,11 @@ fn do_journey() -> Result<(), Box<dyn std::error::Error>> {
 }
 
 fn has_granular_times(root: &Path) -> std::io::Result<bool> {
-    let n = 100;
+    let n = 50;
 
-    let names: Vec<_> = (0..n).map(|i| format!("{i:03}")).collect();
-    for name in &names {
-        std::fs::write(root.join(name), name)?;
+    let names = (0..n).map(|i| format!("{i:03}"));
+    for name in names.clone() {
+        std::fs::write(root.join(&name), name)?;
     }
     let mut times = Vec::new();
     for name in names {
@@ -64,9 +64,12 @@ fn has_granular_times(root: &Path) -> std::io::Result<bool> {
 
     // This could be wrongly false if a filesystem has very precise timings yet is ridiculously
     // fast. Then the `journey` test wouldn't run, though it could. But that's OK, and unlikely.
-    // However, for now, on CI, on macOS only, we assert the expectation of high granularity.
     if cfg!(target_os = "macos") && is_ci::cached() {
-        assert_eq!(times.len(), n);
+        assert_eq!(
+            times.len(),
+            n,
+            "should have very granular timestamps at least on macOS on CI"
+        );
     }
     Ok(times.len() == n)
 }
--- a/gix-fs/tests/fs/snapshot.rs
+++ b/gix-fs/tests/fs/snapshot.rs
@@ -11,7 +11,7 @@ fn journey() -> Result<(), Box<dyn std::error::Error>> {
 
 fn do_journey() -> Result<(), Box<dyn std::error::Error>> {
     let tmp = tempfile::tempdir().unwrap();
-    if !has_nanosecond_times(tmp.path())? {
+    if !has_granular_times(tmp.path())? {
         return Ok(());
     }
 
@@ -48,17 +48,21 @@ fn do_journey() -> Result<(), Box<dyn std::error::Error>> {
     Ok(())
 }
 
-fn has_nanosecond_times(root: &Path) -> std::io::Result<bool> {
-    let test_file = root.join("nanosecond-test");
+fn has_granular_times(root: &Path) -> std::io::Result<bool> {
+    let n = 100;
 
-    std::fs::write(&test_file, "a")?;
-    let first_time = test_file.metadata()?.modified()?;
+    let names: Vec<_> = (0..n).map(|i| format!("{i:03}")).collect();
+    for name in &names {
+        std::fs::write(root.join(name), name)?;
+    }
+    let mut times = Vec::new();
+    for name in names {
+        times.push(root.join(name).symlink_metadata()?.modified()?);
+    }
+    times.sort();
+    times.dedup();
 
-    std::fs::write(&test_file, "b")?;
-    let second_time = test_file.metadata()?.modified()?;
-
-    Ok(second_time.duration_since(first_time).is_ok_and(|d|
-            // This can be falsely false if a filesystem would be ridiculously fast,
-            // which means a test won't run even though it could. But that's OK, and unlikely.
-            d.subsec_nanos() != 0))
+    // This could be wrongly false if a filesystem has very precise timings yet is ridiculously
+    // fast. Then the `journey` test wouldn't run, though it could. But that's OK, and unlikely.
+    Ok(times.len() == n)
 }
--- a/gix-fs/tests/fs/snapshot.rs
+++ b/gix-fs/tests/fs/snapshot.rs
@@ -3,6 +3,13 @@ use std::path::Path;
 
 #[test]
 fn journey() -> Result<(), Box<dyn std::error::Error>> {
+    for _ in 0..250 {
+        do_journey()?;
+    }
+    Ok(())
+}
+
+fn do_journey() -> Result<(), Box<dyn std::error::Error>> {
     let tmp = tempfile::tempdir().unwrap();
     if !has_nanosecond_times(tmp.path())? {
         return Ok(());
--- a/gix-fs/tests/fs/snapshot.rs
+++ b/gix-fs/tests/fs/snapshot.rs
@@ -51,13 +51,13 @@ fn do_journey() -> Result<(), Box<dyn std::error::Error>> {
 fn has_granular_times(root: &Path) -> std::io::Result<bool> {
     let n = 50;
 
-    let names = (0..n).map(|i| format!("{i:03}"));
-    for name in names.clone() {
-        std::fs::write(root.join(&name), name)?;
+    let paths = (0..n).map(|i| root.join(format!("{i:03}")));
+    for (index, path) in paths.clone().enumerate() {
+        std::fs::write(&path, index.to_string().as_bytes())?;
     }
     let mut times = Vec::new();
-    for name in names {
-        times.push(root.join(name).symlink_metadata()?.modified()?);
+    for path in paths {
+        times.push(path.symlink_metadata()?.modified()?);
     }
     times.sort();
     times.dedup();
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -162,6 +162,7 @@ jobs:
           - macos-latest
           - ubuntu-latest
           - ubuntu-24.04-arm
+      fail-fast: false
 
     runs-on: ${{ matrix.os }}
 
@@ -265,6 +266,7 @@ jobs:
             runner-arch: arm64
             runner-os: ubuntu-24.04-arm
             host-triple: armv7-unknown-linux-gnueabihf
+      fail-fast: false
 
     runs-on: ${{ matrix.runner-os }}
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -162,7 +162,6 @@ jobs:
           - macos-latest
           - ubuntu-latest
           - ubuntu-24.04-arm
-      fail-fast: false
 
     runs-on: ${{ matrix.os }}
 
@@ -266,7 +265,6 @@ jobs:
             runner-arch: arm64
             runner-os: ubuntu-24.04-arm
             host-triple: armv7-unknown-linux-gnueabihf
-      fail-fast: false
 
     runs-on: ${{ matrix.runner-os }}
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1903,6 +1903,7 @@ dependencies = [
  "fastrand",
  "gix-features 0.40.0",
  "gix-utils 0.1.14",
+ "is_ci",
  "serde",
  "tempfile",
 ]
--- a/gix-fs/Cargo.toml
+++ b/gix-fs/Cargo.toml
@@ -27,5 +27,6 @@ serde = { version = "1.0.114", optional = true, default-features = false, featur
 fastrand = { version = "2.1.0", default-features = false, features = ["std"] }
 
 [dev-dependencies]
-tempfile = "3.5.0"
 crossbeam-channel = "0.5.0"
+is_ci = "1.1.1"
+tempfile = "3.5.0"


gix-fs/tests/fs/snapshot.rs
```diff
@@ -41,17 +41,25 @@ fn journey() -> Result<(), Box<dyn std::error::Error>> {
...
 
-    Ok(second_time.duration_since(first_time).is_ok_and(|d|
-            // This can be falsely false if a filesystem would be ridiculously fast,
-            // which means a test won't run even though it could. But that's OK, and unlikely.
-            d.subsec_nanos() != 0))
+    // This could be wrongly false if a filesystem has very precise timings yet is ridiculously
+    // fast. Then the `journey` test wouldn't run, though it could. But that's OK, and unlikely.
+    // However, for now, on CI, on macOS only, we assert the expectation of high granularity.
+    if cfg!(target_os = "macos") && is_ci::cached() {
+        assert_eq!(times.len(), n);
```
Byron, line 62: As a third argument, I'd hope there can be an explanation why this is a failure. When the assertion hits, I am sure I would be puzzled without it. It might even be possible to condense the comment above into the assertion message.

gix-fs/tests/fs/snapshot.rs
```diff
@@ -41,17 +41,25 @@ fn journey() -> Result<(), Box<dyn std::error::Error>> {
     Ok(())
 }
 
-fn has_nanosecond_times(root: &Path) -> std::io::Result<bool> {
-    let test_file = root.join("nanosecond-test");
+fn has_granular_times(root: &Path) -> std::io::Result<bool> {
+    let n = 100;
```
Byron, line 45: This might be an unnecessary optimisation, but I'd hope that the number can be reduced to the lowest value (and some margin) that still works for Windows. From what I saw, I think this number could be 50 and still safely have the same effect, 30 might also be possible.
Of course, this might be very wrong as well 😅.

gix-fs/tests/fs/snapshot.rs
```diff
@@ -41,17 +41,25 @@ fn journey() -> Result<(), Box<dyn std::error::Error>> {
...
 
-    std::fs::write(&test_file, "a")?;
-    let first_time = test_file.metadata()?.modified()?;
-
-    std::fs::write(&test_file, "b")?;
-    let second_time = test_file.metadata()?.modified()?;
+    let names: Vec<_> = (0..n).map(|i| format!("{i:03}")).collect();
```
Byron, line 47: It looks like what really wants to be stored here is `paths`.
The `collect()` could also be avoided in favor of cloning the iterator in the first loop - doing so might be good for less 'noise'.

Byron (approved):
Thanks a million, this is great!

I have a few nits with one suggestion related to performance that may be worth more consideration.
Besides that, please feel free to merge when you think it's good to go.

gix-fs/tests/fs/snapshot.rs
EliahKagan: 50 is one of the values I had tested with [in that experiment](https://github.com/EliahKagan/granularity) before changing it to 100 and then making the patch here, and I *think* based on this that 50should be fine. However, I picked 100 because it still seemed fast to me. It is by far the slowest on Windows, where it hardly ever takes more than 30 milliseconds with `n=100` and is usually closer to 25 milliseconds. Or at least that's how long it took in that experiment. [This gist has the timings](https://gist.github.com/EliahKagan/f1052de63d1205e632caab3083207cf7) for all three platforms.

gix-fs/tests/fs/snapshot.rs
EliahKagan: Yes, I'll add that.

gix-fs/tests/fs/snapshot.rs
EliahKagan: Oh, I remember why I didn't write the files inside the iterator. It was because that would prevent me from using `?`.

gix-fs/tests/fs/snapshot.rs
Byron: That's fair - I also had the time it takes to delete the tempdir in mind. Maybe we can stick to 50 then and change it if there is an indication for a need.

gix-fs/tests/fs/snapshot.rs
EliahKagan: As far as I can manage, making it an iterator over paths and cloning the iterator doesn't make things simpler unless I give up on using the filename (or the number in any form) as the file contents. I can iterate over a pair of the name and path, but it's not simpler if I use a `for` loop, and it's not simpler if I use something other than `?` to not have to use a `for` loop.

I can extract the filename, but the type system doesn't keep track of how the path is known to have a filename or how the filename once extracted from it is known to be valid Unicode. Checking, or `expect`ing (or `unwrap`ping), more than eliminates any simplicity gained.

Since the file contents need not be `0`-padded anyway, it feels like maybe I should use `enumerate()` on the cloned `paths` iterator. But then I am formatting a number again to write it to the file, and the gain in simplicity is gone, as is the effective decrease in repetition. Or I can use the full path as the contents of the file. But then I still have to turn it from a `PathBuf`/`OsString`-like thing into something I can write to the file.

So for storing `paths`, I'm not sure how you were thinking I should do that (assuming simplicity and clarity is the goal). Cloning the iterator instead of materializing the names into a vector is fine though. By itself, that change is:

```diff
-    let names: Vec<_> = (0..n).map(|i| format!("{i:03}")).collect();
-    for name in &names {
-        std::fs::write(root.join(name), name)?;
+    let names = (0..n).map(|i| format!("{i:03}"));
+    for name in names.clone() {
+        std::fs::write(root.join(&name), name)?;
     }
```

Or I can collect paths, by using another `for` loop instead of an iterator, but I don't think that's what you have in mind, and it's longer:

```diff
-    let names: Vec<_> = (0..n).map(|i| format!("{i:03}")).collect();
-    for name in &names {
-        std::fs::write(root.join(name), name)?;
+    let mut paths = Vec::new();
+    for i in (0..n) {
+        let name = format!("{i:03}");
+        let path = root.join(&name);
+        std::fs::write(&path, name)?;
+        paths.push(path);
     }
     let mut times = Vec::new();
-    for name in names {
-        times.push(root.join(name).symlink_metadata()?.modified()?);
+    for path in paths {
+        times.push(path.symlink_metadata()?.modified()?);
     }
```

The code would be simpler if I checked each file's timestamp right after creating it. But I don't want to do that, because it is an actual filesystem access operation, and I can't rule out that it could make the delay between creating the files longer by enough to matter.

There might be a very obvious elegant improvement here, but it is not obvious to me.

Byron: Sorry for the hassle - I will submit a commit here for your review, and you can change it as you see fit.
Then please feel free to merge yourself.

Byron: It seems like the `assert_eq!` change hasn't landed yet, so I don't want to push into the branch.

Instead, here are my changes that you are welcome to apply.

```patch
commit d5940238661a2325164cec0e61da945a256a896a
Author: Sebastian Thiel <sebastian.thiel@icloud.com>
Date:   Thu Mar 20 09:36:16 2025 +0800

    refactor
    
    - also use a smaller `n` as half already seemed to work.

diff --git a/gix-fs/tests/fs/snapshot.rs b/gix-fs/tests/fs/snapshot.rs
index 94c851fbd..82b2be835 100644
--- a/gix-fs/tests/fs/snapshot.rs
+++ b/gix-fs/tests/fs/snapshot.rs
@@ -42,15 +42,15 @@ fn journey() -> Result<(), Box<dyn std::error::Error>> {
 }
 
 fn has_granular_times(root: &Path) -> std::io::Result<bool> {
-    let n = 100;
+    let n = 50;
 
-    let names: Vec<_> = (0..n).map(|i| format!("{i:03}")).collect();
-    for name in &names {
-        std::fs::write(root.join(name), name)?;
+    let paths = (0..n).map(|i| root.join(format!("{i:03}")));
+    for (index, path) in paths.clone().enumerate() {
+        std::fs::write(&path, index.to_string().as_bytes())?;
     }
     let mut times = Vec::new();
-    for name in names {
-        times.push(root.join(name).symlink_metadata()?.modified()?);
+    for path in paths {
+        times.push(path.symlink_metadata()?.modified()?);
     }
     times.sort();
     times.dedup();
```

Please feel free to merge when you think it's ready - I think my review is mostly nits after all.

PS: The content of the file might not be relevant even though the code suggests this - maybe better to just write empty files then.

gix-fs/tests/fs/snapshot.rs
EliahKagan: > That's fair - I also had the time it takes to delete the tempdir in mind.

I think on Windows the extra files shouldn't make that take too much longer--but I don't know if that's the case and it depends on how the recursive delete operation `tempdir` performs is implemented. Unfortunately I don't remember the details, and thus I don't know if I am right in my expectation that the deletion delay is not great.

On a branch I shall push shortly, I have changed it to 50.

gix-fs/tests/fs/snapshot.rs
EliahKagan: > Instead, here are my changes that you are welcome to apply.

That does seems slightly clearer than what I just pushed. I'll apply the parts of that change that are not part of what I already have.

> The content of the file might not be relevant even though the code suggests this - maybe better to just write empty files then.

It is arguably better for performance, and even for decreasing the risk of returning `true` wrongly, to write no bytes. Originally I had not done that because I wanted to keep things like `a` with the content `a`, and `b` with the content `b`, as before. But then I decided not to write empty files because I worried it would end up getting "optimized" or "refactored" into something like `touch` that might not, on all systems, work in the same way, but might instead, even when creating the file anew, do something extra to set the timestamp (since the "real" purpose of `touch` is to update a timestamp). Since having the files nonempty and with predictable distinct contents also helps manual debugging, I figured I might as well just do that.

However, if you'd prefer the files be empty, I'd be happy to change it to that.

EliahKagan:
I'm waiting on tests in https://github.com/EliahKagan/gitoxide/actions/runs/13960418689, which verify that the technique still works under the changes (via outer repetition of the whole test, something we don't want to do at the tip of main and which is thus reverted in the next commit). Assuming that still goes well, I'll merge or enable auto-merge.

Byron:
Thanks again for all your help with this! It's huge issue that I introduced and couldn't easily fix, but you resolved it in the shortest possible time 🙏.

gix-fs/tests/fs/snapshot.rs
EliahKagan: To investigate how the number of files affects the time it takes for the temporary directory to be removed, I modified [that related experiment](https://github.com/EliahKagan/granularity) to include that in its timing. It looks like about half the total time is in the removal. This makes sense since the data written to the files are minimal so most of the time measured before may be in creating/opening/closing the files rather than in writing to them.

[That file](https://github.com/EliahKagan/granularity/blob/main/output-windows.txt) is updated with the new Windows timings. This is still for `n=100`, and it's a little over 40 milliseconds (for the earlier runs, which are the ones most relevant here, since what happens here is partially analogous to *one* run of that experiment). I didn't put `n=50` results in that repo, but I did test them as well; they take about half as long, i.e. slightly more than 20 milliseconds. Based on that, I think `n=50` is low enough here, even if the approach from this PR remains in place for a while.

EliahKagan:
Thanks! I hope this will not dissuade you from replacing the approach added here with something that would let the test run everywhere, if that can be done without making it work any less well where it already can run. Or from encouraging me to do so, if you have thoughts about how, beyond (or in more detail on) the idea in https://github.com/GitoxideLabs/gitoxide/pull/1897#discussion_r2004448543.

Byron:
The test was added only because I strangely was missing one, but younger me might have predicted the difficulty in doing so and skipped it. Now that we went through the trouble, I am happy there is a test running at all, even though it won't do so right now on all platforms.
Thus, for now I think it's good enough and a clear improvement to what was (or wasn't) there before, so it's unlikely I get to work in it anytime soon.

Here is my `gitoxide` backlog, and that's just what I started but didn't finish. There is more bugs hidden in issues that I have to fix with priority.
Screenshot 2025-03-20 at 12 53 20

PR #1901, opened 2025-03-20
Derive Clone and Copy for Either

cruessler:
This is a small PR that adds `#[derive(Clone, Copy)]` to `Either`. `std::mem::size_of::<Either<'_, '_>>()` is 56, but I don’t know whether that qualifies as “a bit large”. :-) If you think it does, I would adapt this PR accordingly.


cruessler:
I’m so happy the code is more consistent now. :-)

PR #1903, opened 2025-03-21
[gix-object] Support empty tags with or without trailing NL

pierrechevalier83:
When representing an annotated tag with an empty commit message, we used to only support a tag ending with two newlines (one after the tagger line + one after the empty commit message).

This was due to a misconception that annotated tags ending with a single NL shouldn't exist in the wild since there isn't an obvious way to create them with `git`.
This misconception shows up in the discussion thread for [issue 603](https://github.com/GitoxideLabs/gitoxide/issues/603)

It turns out that both encodings of empty annotated tags appear in the wild.

We must be able to parse either and roundtrip for either.

Before [PR 604](https://github.com/GitoxideLabs/gitoxide/pull/604), we used to special case the empty tag msg case and not add a NL. To be able to represent `b""`, `b"\n"`, `b"\n\n"`..., we special case any `message` that is a pure sequence of `b'\n'` and actually parse the `b'\n'` into the tag's message.

This allows us to calculate the correct size that matches the number of bytes that git would produce, as well as round-trip to and from the commit encoding.

The existing tests (in particular `round_trip` for `empty.txt` and `empty_missing_nl.txt`) convince me that the logic is sound. Also, the size being `139` bytes for `empty_missing_nl.txt` and `140` bytes for `empty.txt` matches the output of `cat file | wc -l` for each file.

### DIFF
--- a/gix-object/src/tag/decode.rs
+++ b/gix-object/src/tag/decode.rs
@@ -40,8 +40,8 @@ pub fn message<'a, E: ParserError<&'a [u8]>>(i: &mut &'a [u8]) -> ModalResult<(&
     const PGP_SIGNATURE_BEGIN: &[u8] = b"\n-----BEGIN PGP SIGNATURE-----";
     const PGP_SIGNATURE_END: &[u8] = b"-----END PGP SIGNATURE-----";
 
-    if i.is_empty() {
-        return Ok((b"".as_bstr(), None));
+    if i.iter().all(|b| *b == b'\n') {
+        return i.map(|message: &[u8]| (message.as_bstr(), None)).parse_next(i);
     }
     delimited(
         NL,
--- a/gix-object/src/tag/write.rs
+++ b/gix-object/src/tag/write.rs
@@ -29,10 +29,10 @@ impl crate::WriteTo for Tag {
             encode::trusted_header_signature(b"tagger", &tagger.to_ref(), out)?;
         }
 
-        out.write_all(NL)?;
-        if !self.message.is_empty() {
-            out.write_all(self.message.as_ref())?;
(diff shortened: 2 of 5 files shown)


Byron (changes requested):
Thanks a lot, that's great!
I definitely had problems getting the parsing right and by now tests are all we have :D.

Regarding the commit I force-pushed, it's only changing the subject, even though it looks like I am the author now.
This is due to the author being strangely formatted:

```sh
❯ gix cat @
tree 12ca5ffc6ce6ad1682ecafdd31bb2e223f5ad0b3
parent 6d5f77413e63cad45415d0f674c58d41091bffa1
author Pierre Chevalier pierrec@meta.com <> 1742556630 +0000
committer Sebastian Thiel <sebastian.thiel@icloud.com> 1742613444 +0800
gpgsig -----BEGIN PGP SIGNATURE-----
```

This already was the case in the original.

```sh
❯ gix cat 2435710b4
tree 12ca5ffc6ce6ad1682ecafdd31bb2e223f5ad0b3
parent 6d5f77413e63cad45415d0f674c58d41091bffa1
author Pierre Chevalier pierrec@meta.com <> 1742556630 +0000
committer Pierre Chevalier <pierrechevalier83@gmail.com> 1742565757 +0000
```

Thus I'd leave it to you to change the subject and force-push again so the authorship comes out correctly. And maybe there is a way to fix the author line while at it :).

pierrechevalier83:
> Thus I'd leave it to you to change the subject and force-push again so the authorship comes out correctly. And maybe there is a way to fix the author line while at it :).

Thanks for the fast review. I've fixed the author/committer fields and rebased the change.

Byron (approved):
Fantastic, thanks!

Issue #1886, opened 2025-03-14
Using the latest `shell()` or related facility in `bash_program()`


EliahKagan:
### Summary 💡

The limitation described in https://github.com/GitoxideLabs/gitoxide/pull/1510#issuecomment-2558193445, where `gix-testtools` cannot depend on current versions of `gix-*` crates, requires that `gix-testtools` depend on an old `gix-path` that may not have all the same facilities and bug fixes. In some cases, this is no problem:

https://github.com/GitoxideLabs/gitoxide/blob/6d5f77413e63cad45415d0f674c58d41091bffa1/tests/tools/src/lib.rs#L222

In other cases, however, it prevents a feature from being used in `gix-testtools` at the time it is being developed:

https://github.com/GitoxideLabs/gitoxide/blob/6d5f77413e63cad45415d0f674c58d41091bffa1/tests/tools/src/lib.rs#L657

This gets worse if the facility becomes available, `gix-testtools` is changed to use it, and the facility's behavior changes. Then the effect is not obvious, and also bugs remain that would intuitively be thought of a fixed within this repository.

But right now I am more concerned about the missed opportunity, as applied to the shell-finding functionality of `gix-path` specifically. This is something I suspect will, and *should*, be iterated on with small gradual changes. Being able to use it immediately in `gix-testtools` would "test drive" the changes using gitoxide's test fixtures, which I expect would be very helpful. In particular, it would make me a lot less nervous about trying to make improvements to this functionality, including as may be needed to fix significant bugs in the shorter term, and for implementing other improvements such as environment customization for subprocesses (discussed in #1868) for the longer term. I expound on the anticipated benefits, and how they relate to some other ongoing work, in the "Motivation" section below.

Even before eventually alleviating the limitation in cargo-smart-release--which I assume should eventually be done, but which I don't know how to do, and which seems nontrivial to implement and nontrivial to test--I think it might be possible to use some form of dependency injection to give `gix-testtools` the result of `shell()`.

Here are some ideas. They are grouped by relatedness, not in the order of how good or bad they are. (Unfortunately none of them seem especially good to me.)

#### Having test modules call a function in `gix-testtools`

`gix_testtools` could provide an initialization function which tests are expected to call to pass a path or a `Fn` that can be called to get one. If not called, a fallback value found in a simpler way would be used, unless an environment variable is present that signifies that the absence of initialization is to be treated as a hard error and result in a panic.

[...]

PR #1904, opened 2025-03-21
Make the justfile more portable and slightly more resilient

EliahKagan:
See the commit messages (and diff) for full details. Some highlights are described below.

#### When `cargo metadata` fails and it's not a problem

Before these changes:

```text
ek in 🌐 catenary in gitoxide on  main is 📦 v0.41.0 via 🦀 v1.85.1
❯ rm -r gix-packetline-blocking/src

ek in 🌐 catenary in gitoxide on  main [✘] is 📦 v0.41.0 via 🦀 v1.85.1
❯ just copy-packetline
error: failed to load manifest for workspace member `/home/ek/repos/gitoxide/gix-ref`
referenced by workspace at `/home/ek/repos/gitoxide/Cargo.toml`

Caused by:
  failed to load manifest for dependency `gix-object`

Caused by:
  failed to load manifest for dependency `gix-odb`

Caused by:
  failed to load manifest for dependency `gix-pack`

Caused by:
  failed to load manifest for dependency `gix-diff`

Caused by:
  failed to load manifest for dependency `gix-filter`

Caused by:
  failed to load manifest for dependency `gix-packetline-blocking`

Caused by:
  failed to parse manifest at `/home/ek/repos/gitoxide/gix-packetline-blocking/Cargo.toml`

Caused by:
  can't find library `gix_packetline_blocking`, rename file to `src/lib.rs` or specify lib.path
./etc/copy-packetline.sh
```

The `copy-packetline.sh` script actually succeeded, because it didn't need the `cargo manifest` output. But `cargo manifest` was called anytime `just` was used to run any recipe, or even just to list the recipes. Besides occasionally making the first invocation take noticeably longer, this is misleading, because it looks like something relevant to the requested operation failed. (See also https://github.com/GitoxideLabs/gitoxide/issues/1310#issuecomment-2040299934.)

After these changes:

```text
ek in 🌐 catenary in gitoxide on  run-ci/just is 📦 v0.41.0 via 🦀 v1.85.1
❯ rm -r gix-packetline-blocking/src

ek in 🌐 catenary in gitoxide on  run-ci/just [✘] is 📦 v0.41.0 via 🦀 v1.85.1
❯ just copy-packetline
etc/copy-packetline.sh
```

#### Making future additions more discoverable

Related to the above, there are various approaches we can use to have Rust code that builds and runs even when the workspace is broken. This is relevant if we decide to introduce Rust code that generates Rust code that, if absent, may cause the workspace to be considered broken. One of the approaches for this, which requires no extra tools, is to have a project that opts out of being in any workspace by having an empty `[workspace]` section in its `Cargo.toml`.

Such a project is then less discoverable. But it can be made discoverable by including commands that use it in the `justfile`. That is only feasible if the `justfile` avoids eagerly failing when the workspace is broken.

That's actually may main motivation for doing this PR: to open up the ability to do things like `copy-packetline.sh` in Rust, as I fear may be needed for #1886 (depending on what the answer to https://github.com/GitoxideLabs/gitoxide/issues/1886#issuecomment-2727490062 turns out to be).

#### When `cargo metadata` fails and it *is* a problem

Recipes that actually use the output of `cargo metadata`, such as `journey-tests`, showed the same errors twice. In principle, they could in the future come to have been written in a way that would cause them to behave even more strangely, if they did something first that did not use `cargo`. Now that we only use `cargo metadata` when we need the output, this is made into a hard error, and thus is only run once.

It may be that this was originally intended--though then, since it was run eagerly before all recipes, it would have kept anything from working even when failures was not a problem. The reason it was not a hard error before is that the failing command is on the left side of a pipe, where it does not cause the whole pipeline to fail, and the `jq` parsing command on the right side of the pipe produced empty output but still reported success even when its input was empty.

#### Quoting

Before these changes, on Windows:

```text
C:\Users\ek\source\repos\gitoxide [main ≡]> just
C:\Users\ek\.cargo\bin\just.exe --list
/usr/bin/bash: line 1: C:Usersek.cargobinjust.exe: command not found
error: Recipe `default` failed on line 6 with exit code 127
```

After these changes, on Windows:

```text
C:\Users\ek\source\repos\gitoxide [run-ci/just ≡]> just
'C:\Users\ek\.cargo\bin\just.exe' --list
Available recipes:
    audit                           # Run various auditing tools to help us stay legal and safe
    check                           # Build all code in suitable configurations [alias: c]
    check-mode                      # Find shell scripts whose +x/-x bits and magic bytes (e.g. `#!`) disagree
    check-size                      # Run `cargo diet` on all crates to see that they are still in bounds
    ci-check-msrv                   # Check the minimal support Rust version, with the currently installed Rust version
    ci-journey-tests                # Run all journey tests - should be run in a fresh clone or after `cargo clean`
    ci-test                         # Run all tests, without clippy, and try building docs
    clear-target                    # Clean the `target` directory
    clippy *clippy-args             # Run `cargo clippy` on all crates
    clippy-fix                      # Run `cargo clippy` on all crates, fixing what can be fixed, and format all code
    copy-packetline                 # Delete `gix-packetline-blocking/src` and regenerate from `gix-packetline/src`
    default                         # List available recipes
    doc $RUSTDOCFLAGS='-D warnings' # Run `cargo doc` on all crates
    find-yanked                     # Cancel this after the first few seconds, as yanked crates will appear in warnings
    fmt                             # Run nightly `rustfmt` for its extra features, but check that it won't upset stable `rustfmt`
    journey-tests                   # Run journey tests (`max`)
    journey-tests-async             # Run journey tests (`lean-async`)
    journey-tests-pure              # Run journey tests (`max-pure`)
    journey-tests-small             # Run journey tests (`small`)
    nextest *FLAGS='--workspace'    # Run tests with `cargo nextest` (all unit-tests, no doc-tests, faster) [alias: nt]
    nix-shell-macos                 # Enter a nix-shell able to build on macOS
    summarize EXPRESSION='all()'    # Run tests with `cargo nextest`, skipping none except as filtered, omitting status reports
    test                            # Run all tests, clippy, including journey tests, try building docs [alias: t]
    unit-tests                      # Run all unit tests
    unit-tests-flaky                # These tests aren't run by default as they are flaky (even locally)
```

Some other quoting improvements are included. For example, `summarize` no longer pastes single quotes around the expression, but instead quotes it robustly.

Although this mainly benefits Windows, in principle it can be needed on any platform, such as when a path used in a shell has spaces in one of its components.

#### Using the same `just` implementation

`just_executable()` was already used in the default recipe to make sure we run the same `just` for `just --list` rather than a different `just`. (It was used in a shell without quoting, but that is fixed here as described above.) But one recipe had previously hard-coded `just` and used it for formatting. This fixes that.

#### Descriptions

All recipes have `#` descriptions that show when the recipes are listed, and they are revised to slightly improve clarity and to eliminate unnecessary stylistic differences. (As shown above.)

---

For some of the changes here, there is more than one way they could be done, and whether this way is clearer or less clear than alternatives is subjective. Accordingly, even though this is kind of thing that, along the lines of the discussion in #1883, I might consider merging myself without a review, in this case I think it's better that I check.

### DIFF
--- a/justfile
+++ b/justfile
@@ -243,7 +243,9 @@ audit:
 nextest *FLAGS='--workspace':
     cargo nextest run {{ FLAGS }}
 
-summarize EXPRESSION='all()': (nextest '--workspace --run-ignored all --no-fail-fast --status-level none --final-status-level none -E' quote(EXPRESSION))
+summarize EXPRESSION='all()':
+    cargo nextest run --workspace --run-ignored all --no-fail-fast \
+        --status-level none --final-status-level none -E {{ quote(EXPRESSION) }}
 
 # run nightly rustfmt for its extra features, but check that it won't upset stable rustfmt
 fmt:
--- a/justfile
+++ b/justfile
@@ -144,7 +144,7 @@ check:
     cargo check --no-default-features --features max-control
 
 # Run cargo doc on all crates
-doc $RUSTDOCFLAGS="-D warnings":
+doc $RUSTDOCFLAGS='-D warnings':
     cargo doc --workspace --no-deps --features need-more-recent-msrv
     cargo doc --features=max,lean,small --workspace --no-deps --features need-more-recent-msrv
 
@@ -158,9 +158,9 @@ unit-tests:
     cargo nextest run -p gix-archive --features tar
     cargo nextest run -p gix-archive --features tar_gz
     cargo nextest run -p gix-archive --features zip
-    cargo nextest run -p gix-status-tests --features "gix-features-parallel"
-    cargo nextest run -p gix-worktree-state-tests --features "gix-features-parallel"
-    cargo nextest run -p gix-worktree-tests --features "gix-features-parallel"
+    cargo nextest run -p gix-status-tests --features gix-features-parallel
+    cargo nextest run -p gix-worktree-state-tests --features gix-features-parallel
+    cargo nextest run -p gix-worktree-tests --features gix-features-parallel
     cd gix-object; \
         set -ex; \
         cargo nextest run; \
@@ -172,10 +172,10 @@ unit-tests:
     cargo nextest run -p gix-odb-tests --features gix-features-parallel
     cargo nextest run -p gix-pack --all-features
     cargo nextest run -p gix-pack-tests --features all-features
-    cargo nextest run -p gix-pack-tests --features "gix-features-parallel"
-    cargo nextest run -p gix-index-tests --features "gix-features-parallel"
+    cargo nextest run -p gix-pack-tests --features gix-features-parallel
+    cargo nextest run -p gix-index-tests --features gix-features-parallel
     cargo nextest run -p gix-packetline --features blocking-io,maybe-async/is_sync --test blocking-packetline
-    cargo nextest run -p gix-packetline --features "async-io" --test async-packetline
+    cargo nextest run -p gix-packetline --features async-io --test async-packetline
     cargo nextest run -p gix-transport --features http-client-curl,maybe-async/is_sync
     cargo nextest run -p gix-transport --features http-client-reqwest,maybe-async/is_sync
     cargo nextest run -p gix-transport --features async-client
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot - this is great and much appreciated.

Indeed, I only gave it a quick look and generally only want these things to work, without maintaining them ideally 😁. So if you feel there is a delay or it's really nothing you'd think I could object to, then please do feel free to merge these kind of PRs yourself.

(*I will always look at them anyway and will stay in the loop that way, should there be something to change, it can always be a follow-up as well*)

PR #1905, opened 2025-03-22
fix: use platform path separator instead of `/`

nik-rev:
I'm working on Git Blame for Helix and some people on Windows experienced issues where they don't see the blame (https://github.com/helix-editor/helix/pull/13133?notification_referrer_id=NT_kwDOCTrtqLUxNTM1OTU2ODI4OToxNTQ4NTY4NzI#issuecomment-2744734878)

I found the reason is because Gix uses `/` for path separator always, instead of using the platform's separator. This PR fixes that

note: the way I'm turning `MAIN_SEPARATOR_STR` into a byte is a bit awkward here, I wish there was some kind of `MAIN_SEPARATOR_BYTE`

### DIFF
--- a/gix-blame/src/file/function.rs
+++ b/gix-blame/src/file/function.rs
@@ -661,7 +661,16 @@ fn find_path_entry_in_commit(
     let res = tree_iter.lookup_entry(
         odb,
         buf2,
-        file_path.split(|b| *b == b'/').inspect(|_| stats.trees_decoded += 1),
+        file_path
+            .split(|b| {
+                b == const {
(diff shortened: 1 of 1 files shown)


Byron:
Hi @nik-rev, thanks a lot for contributing!

I left a comment in the [original helix PR](https://github.com/helix-editor/helix/pull/13133#pullrequestreview-2707733443) explaining why this PR here has to be closed.
In short, `BString` relative paths are always slash-separated, and I think it's lacking its own type to make that clear.

CC @cruessler

PR #1909, opened 2025-03-23
Accept `ToComponents` in `gix_fs::Stack`

cruessler:
This is a draft PR, intended to make sure I got the task right before making any larger changes that I would need to make now that I’ve changed `gix_fs::Stack`’s API. It is expected that it does not even compile at this stage.

A few observations:

- There’s places that currently call `Stack::new` with a `PathBuf`. Would it make sense to `impl ToComponents for PathBuf`?
- Having to add `bstr` as a dependency made me wonder: is `gix-fs` the right crate for `ToComponents` to live in or should it be moved somewhere else, e. g. into `gix-path`? `gix-path` also has `os_string_into_bstring` which should probably replace the chain `.as_os_str().as_bytes().into()` that I had to resort to on one line because `gix-fs` does not depend on `gix-path` currently.

Let me know what you think!

### DIFF
--- a/gix-fs/src/stack.rs
+++ b/gix-fs/src/stack.rs
@@ -1,6 +1,42 @@
+use std::ffi::OsStr;
 use std::path::{Component, Path, PathBuf};
 
 use crate::Stack;
+
+///
+pub mod to_normal_path_components {
+    use std::ffi::OsString;
+
+    /// The error used in [`ToNormalPathComponents::to_normal_path_components()`](super::ToNormalPathComponents::to_normal_path_components()).
+    #[derive(Debug, thiserror::Error)]
+    #[allow(missing_docs)]
+    pub enum Error {
+        #[error("Input path \"{path}\" contains relative or absolute components", path = std::path::Path::new(.0.as_os_str()).display())]
+        NotANormalComponent(OsString),
+    }
+}
+
+/// Obtain an iterator over `OsStr`-components which are normal, none-relative and not absolute.
+pub trait ToNormalPathComponents {
+    /// Return an iterator over the normal components of a path, without the separator.
+    // TODO(MSRV): turn this into `impl Iterator` once MSRV is 1.75 or higher
+    fn to_normal_path_components(
+        &self,
+    ) -> Box<dyn Iterator<Item = Result<&OsStr, to_normal_path_components::Error>> + '_>;
+}
+
+impl ToNormalPathComponents for &Path {
+    fn to_normal_path_components(
+        &self,
+    ) -> Box<dyn Iterator<Item = Result<&OsStr, to_normal_path_components::Error>> + '_> {
+        Box::new(self.components().map(|component| match component {
+            Component::Normal(os_str) => Ok(os_str),
+            _ => Err(to_normal_path_components::Error::NotANormalComponent(
+                self.as_os_str().to_owned(),
+            )),
+        }))
+    }
+}
 
 /// Access
 impl Stack {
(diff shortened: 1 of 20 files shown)


gix-fs/src/lib.rs
```diff
@@ -65,9 +65,9 @@ pub fn current_dir(precompose_unicode: bool) -> std::io::Result<PathBuf> {
 #[derive(Clone)]
 pub struct Stack {
     /// The prefix/root for all paths we handle.
-    root: PathBuf,
-    /// the most recent known cached that we know is valid.
-    current: PathBuf,
```
Byron, line 70: These have to stay `PathBuf` as the type is a link between relative paths and the filesystem location they represent.
It's also in the interest of performance to keep it close to what it ought to represent.

gix-fs/src/stack.rs
```diff
@@ -41,15 +46,62 @@ pub trait Delegate {
...
+    fn to_components(&self) -> impl Iterator<Item = &BStr> {
+        self.split(|b| *b == b'/').map(Into::into)
+    }
+}
+
 impl Stack {
     /// Create a new instance with `root` being the base for all future paths we handle, assuming it to be valid which includes
     /// symbolic links to be included in it as well.
-    pub fn new(root: PathBuf) -> Self {
+    pub fn new(root: impl ToComponents) -> Self {
```
Byron, line 97: Since this is the root and it is a filesystem location, it must remain a `PathBuf`.

`ToComponents` is really only for the relative paths that are passed in.
It's interesting that `ToComponents` currently doesn't convey information about whether the component is coming from a relative path, which kind of is a problem, but maybe one that can be tackled later.

Byron (commented):
Thanks for getting started on this, it's much appreciated!

> * There’s places that currently call `Stack::new` with a `PathBuf`. Would it make sense to `impl ToComponents for PathBuf`?
No - I have made some more comments in code with details.

> * Having to add `bstr` as a dependency made me wonder: is `gix-fs` the right crate for `ToComponents` to live in or should it be moved somewhere else, e. g. into `gix-path`? `gix-path` also has `os_string_into_bstring` which should probably replace the chain `.as_os_str().as_bytes().into()` that I had to resort to on one line because `gix-fs` does not depend on `gix-path` currently.

Indeed, I think `gix-path` is the better spot for it. The reason is also that I think it will harbour the `RepositoryPathBuf/RepositoryPath` one day, which will be the typed version of a `BString` as we use it today.

The place where `ToComponents` needs to go is `make_relative_path_current()` and its callers. Then the callers have to be adjusted to avoid calls to `gix::path::<conversion>` as these shouldn't be needed anymore.

I hope that helps.

cruessler:
I’m running into an issue when trying to turn `relative: &Path` into `relative: impl gix_path::ToComponents`. The `!matches!(comp, Component::Normal(_))` in the code below does not work anymore when `omponents` is `Vec<&BStr>` instead of `Components<'_>`. It seems we would already need something like `RepositoryPath` for that `matches!` to work properly. Or am I missing anything? Removing the check does not seem to be an option as it would alter the function’s behaviour.

https://github.com/cruessler/gitoxide/blob/c8c42b4b86e8bf7d8f0f7130d2da98dfed246be9/gix-fs/src/stack.rs#L103-L111

Byron:
I think the problem here is twofold:

* the Stack wants to (and has to) build an actual `Path, so each component of the iterator needs to be an `OsStr`, which is fallible at least on Windows.
* The conversion (and component) check could be put into the iterator itself (especially if it is fallible)

To me this means the trait would have to be owned by `gix-fs` for now as it needs to assure there are no non-normal components. Maybe that would be a start?

Another conclusion here probably is that there will be many `ToComponents` implementations, with encoding different requirements, and that's probably fine as well. We might consider naming them accordingly, this one here could be a `ToPathComponent` for instance.

cruessler:
I hope what I just pushed is closer to your suggestions, even though the new trait `ToPathComponents` lives in `gix-path` instead of `gix-fs`. So far, I’ve tried to change as little as possible of the existing code. That’s the reason I’ve added `.collect()` because that way `.is_empty()` can still be used.

Byron:
Thanks a lot - this looks like it should work!
Something I don't understand is how the MSRV check can fail while `ToComponents` in `gix` does the very same - it is using a feature that ought to be available from Rust 1.75, just one version above the current MSRV which has to be held until `helix` upgrades their MSRV. Let's check :).
As a next step, all the callers of `make_relative_path_current()` would have to be adjusted to use the trait too - in theory this won't be a breaking change as the trait will then be extended to be more compatible.
Please feel free to merge my commit down into yours and adjust the subject lines accordingly, as I changed names once again 😅.

@pascalkuthe Would helix be able to update their MSRV to something newer?

Byron:
Yes, that’s the one.
Great!

cruessler:
I’ve rebased and merged the commits, so now there’s only 2 left which definitely is a simplification. Before I randomly start changing callers, up to which level do we want to accept `relative_path: ToNormalPathComponents`?

cruessler:
Feel free to take over! I’ve just markes this PR as ready and will have a look at the next item on your list of suggestions!

PR #1910, opened 2025-03-23
Add `Either::tree_id`

cruessler:
- Add `Either::tree_id`
- Use `Either::tree_id` in `gix-blame`

This is based on a suggestion in this comment: https://github.com/GitoxideLabs/gitoxide/pull/1824#discussion_r1940597689.

### DIFF
--- a/gix-blame/src/file/function.rs
+++ b/gix-blame/src/file/function.rs
@@ -452,12 +452,12 @@ fn tree_diff_at_file_path(
     lhs_tree_buf: &mut Vec<u8>,
     rhs_tree_buf: &mut Vec<u8>,
 ) -> Result<Option<gix_diff::tree::recorder::Change>, Error> {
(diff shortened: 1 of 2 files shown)


Byron:
Thanks a lot, a great catch!

While rushing over it I kind of caused a blunder: I essentially ended up editing your commit messages (to assure it gets into the changelog) without really thinking about the implication: I changed your words.
This could be an issue in principle and is nothing I really want to do, but I guess I just wanted to be done with the PR without having to send it into a review iteration.

Please let me know if these kinds of small changes are in your interest or not - if not I will not do that again but rather ask you to change commit messages.

PS: To never have me change your commits, you could sign them - naturally I don't destroy peoples signatures 😅.

cruessler:
> Please let me know if these kinds of small changes are in your interest or not - if not I will not do that again but rather ask you to change commit messages.

I’m completely fine with you also changing commit messages! I’m always trying to match existing conventions, so this is helpful feedback, but I prefer if it does not block progress on a PR.

PR #1918, opened 2025-04-04
Fix clippy

EliahKagan:
`clippy` has recently begun to fail with:

    error: unnecessary semicolon
      --> gix-transport/src/client/blocking_io/http/traits.rs:30:18
       |
    30 |                 };
       |                  ^ help: remove
       |
       = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_semicolon
       = note: `-D clippy::unnecessary-semicolon` implied by `-D warnings`
       = help: to override `-D warnings` add `#[allow(clippy::unnecessary_semicolon)]`

While it looks like this might first have been observed in #1917, it is unrelated to any change there. It happens when the current tip of main (4660f7a) is rerun, as observed in:
https://github.com/EliahKagan/gitoxide/actions/runs/14254079128/job/39958292846

---

This is initially empty as I'm double checking that it does happen. I'll amend with a fix shortly.

*Edit:* Yes, [it reproduces here](https://github.com/GitoxideLabs/gitoxide/actions/runs/14255961692/job/39958584453?pr=1918#step:5:1066). I've amended to add the fix.

PR #1919, opened 2025-04-04
the next release

Byron:
### Tasks

* [x] update changelogs
* [x] publish all crates
* [x] trigger a new release build once tags are pushed


EliahKagan:
Now that #1918 has gone through, I think it may decrease confusion if this is rebased onto main.

Byron:
Thanks for the hint. 
If CI succeeds I will just merge, otherwise I can merge `main` in. The commits here are pinned as tags have been created for them.

PR #1922, opened 2025-04-04
fix: Make email with spaces round-trip

pierrechevalier83:
Before this change, a commit containing `"<email@provider.com >"` looses the final space when round-tripping, changing the size of the gix-object and it hash.

This was due to the spaces getting trimmed at parse time.

Fix it and show a few examples with tests that now round-trip as expected.


Byron (approved):

Note that I fixed the subject, and decided that the first commit isn't a breaking change as the API doesn't break. Further, I added documentation that makes clear that spaces can be included in the email.

Lastly, I brought back an assertions that was made as part of another fix.

PR #1927, opened 2025-04-05
More cargo-deny maintenance

EliahKagan:
Changes:

- Remove the old `ring` license from the allowlist, since we no longer depend on any versions of `ring` that use it.

- Convert the comment explanation for why we ignore [RUSTSEC-2024-0436](https://rustsec.org/advisories/RUSTSEC-2024-0436.html) into reified form as the value of a `reason` key, so tools that process the file don't lose it (even if only to make debugging slightly easier when using `tomlq`).

- Split the `cargo-deny-advisories` scan into two scans (still in that one job) that cover what we covered before, *except* ignoring [RUSTSEC-2025-0021](https://rustsec.org/advisories/RUSTSEC-2025-0021.html) when the affected version is *only* reachable as a dependency of `gix-testtools`. This is as discussed in [#1924 (comment)](https://github.com/GitoxideLabs/gitoxide/pull/1924#issuecomment-2780161660), and it makes `cargo-deny-advisories` pass again.

See the commit messages for further details.

---

Regarding the license allowlist, disused licenses do seem to accumulate gradually; see also 9c708db (#1863). `cargo deny` supports configuring this as an error. But I have refrained from doing so, because:

- We don't list that many unusual licenses anymore.
- Which licenses produce the warning differs based on how `cargo deny check licenses` is run, showing another occurrence when `--all-features` is omitted, which might occasionally make sense to do when running it locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -346,8 +346,8 @@ jobs:
       - uses: actions/checkout@v4
       - uses: EmbarkStudios/cargo-deny-action@v2
         with:
-          arguments: --workspace --all-features
           command: check advisories
+          arguments: --workspace --all-features
 
   cargo-deny:
     runs-on: ubuntu-latest
@@ -356,8 +356,8 @@ jobs:
       - uses: actions/checkout@v4
       - uses: EmbarkStudios/cargo-deny-action@v2
         with:
-          arguments: --workspace --all-features
           command: check bans licenses sources
+          arguments: --workspace --all-features
 
   wasm:
     name: WebAssembly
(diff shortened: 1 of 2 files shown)


EliahKagan (commented):
Since `cargo-deny-advisories` is modified here and that job does not block auto-merge even when it fails, I want to double-check that it still works the way I have described here. I plan to enable auto-merge after verifying that.

*Edit:* [It is working](https://github.com/GitoxideLabs/gitoxide/actions/runs/14279263338/job/40026751934?pr=1927) (as it did when run in my fork).

Byron (approved):
Thanks a lot! I can't wait for *not* getting these failure emails anymore 😅.

Issue #1928, opened 2025-04-05
`repo.references()` fails to deduplicate packed/loose refs


yuja:
### Current behavior 😯

With the following setup, `repo.references()` emits both loose and packed refs.

```sh
git init -q

mkdir -p .git/refs/heads/a
echo 4b825dc642cb6eb9a060e54bf8d69288fbee4904 > .git/refs/heads/a-
echo 4b825dc642cb6eb9a060e54bf8d69288fbee4904 > .git/refs/heads/a/b
echo 4b825dc642cb6eb9a060e54bf8d69288fbee4904 > .git/refs/heads/a0
git pack-refs --all
```

This generates the following packed-refs:
```
# pack-refs with: peeled fully-peeled sorted
4b825dc642cb6eb9a060e54bf8d69288fbee4904 refs/heads/a-
4b825dc642cb6eb9a060e54bf8d69288fbee4904 refs/heads/a/b
4b825dc642cb6eb9a060e54bf8d69288fbee4904 refs/heads/a0
EOF
```

Create loose refs:
```sh
mkdir -p .git/refs/heads/a
echo 0000000000000000000000000000000000000000 > .git/refs/heads/a-
echo 0000000000000000000000000000000000000000 > .git/refs/heads/a/b
echo 0000000000000000000000000000000000000000 > .git/refs/heads/a0
```

List all refs:
```rust
fn main() {
    let path = env::args().nth(1).unwrap();
    let repo = gix::open(path).unwrap();
    for git_ref in repo.references().unwrap().all().unwrap() {
        let git_ref = git_ref.unwrap();
        println!("{} {:?}", git_ref.name().as_bstr(), git_ref.target());
    }
}
```

`refs/heads/a/b` in packed-refs should be omitted:
```
refs/heads/a- Object(Sha1(0000000000000000000000000000000000000000))
refs/heads/a/b Object(Sha1(4b825dc642cb6eb9a060e54bf8d69288fbee4904))
refs/heads/a0 Object(Sha1(0000000000000000000000000000000000000000))
refs/heads/a/b Object(Sha1(0000000000000000000000000000000000000000))
```

I think this is regression caused by the fix for https://github.com/GitoxideLabs/gitoxide/issues/1850. Maybe we'll need to sort file-system entries in the same way as `tree::EntryRef`?

Byron:
Thanks a lot for reporting and for figuring out a reproduction.

[...]

PR #1931, opened 2025-04-05
make `fs::walkdir_sorted_new()` sort entries by paths literally (#1928)

yuja:
This follows up 7b1b5bf864e74706aefeb1213e8bdb0545d5464a. Since packed-refs
appears to be sorted by full ref name, loose-refs should also be emitted in
that order.

The comparison function is copied from gix::diff::object::tree::EntryRef.
Non-utf8 file names are simply mapped to "" on Windows. We could add some
fallback, but callers can't handle such file names anyway.

#1928


Byron:
Thanks a lot.
I am merging this and will create a new patch release of `gix-features` shortly after.

Byron:
And here it is: https://github.com/GitoxideLabs/gitoxide/releases/tag/gix-features-v0.41.1

PR #1935, opened 2025-04-06
Make author and committer date roundtrip

pierrechevalier83:
Store the bytes in raw Git format in the Signature, allowing round-tripping even for creatively formatted git data.

Whenever we need to interpret the time, parse it on demand.

Note that the most common need for interpreting the time is limited to its `seconds` part.
We expose a way to parse the `seconds` without parsing the rest to avoid unnecessary costs.

Fixes issue 1923

### Tasks Byron

* [x] final line-by-line review and general cleanup of `gix-date` maybe
* [x] Double-check commits properly indicate breakage
* [x] Remove `sign` field, while we are breaking things
* [ ] create a new top-level release


pierrechevalier83:
> I also realized that turning time: gix_date::Time into a BString feels wrong, besides being a worst-case change for the downstream which at this point we have to consider.

> I also doubt that this will be it, i.e. there is already some special casing of user names and emails, and flexible parsing of whitespace, which also wouldn't be restored correctly.
> Hence I'd think that if it was parsed, there should be some sort of raw: Option<BString> field that contains the entire line. Only then round-tripping will work for sure, at least with the signature field.
> Doing so will open up gates to hell when there is mutable fields and a raw field that go out of sync, when present.

While I understand in theory, in practice, I think this may be enough for the signature field.
In the `> 7000` repos we host, we currently have only two exceptions to what gitoxide can handle without this change:
* The date being of form: "+123400"
* The date being of form: "+123456" (non-zero seconds)
So it looks like we should be OK with 3 `BString`s.

> This makes me question the whole approach, and I'd prefer to say that once the user has a non *Ref version of something, it won't round-trip. Or said differently, once the user has potentially changed an object, it will be written 'cleanly' without extra spaces and whatever else was leniently parsed.

Just to be sure I understand your suggestion precisely, are you advocating for having `ObjectRef` round-trip with `Git`, but `Object` not round-trip with `ObjectRef` as a preferable situation?

I see two potential issues with this:
* It feels unintuitive to loose the equivalence between the owned and the non-owned version. The contract being that they can round trip, and the details of ownership being a pure performance consideration makes more sense to me as a user.
* In terms of ownership, say `Object` has a sane typed representation of time but `ObjectRef` needs a `BStr`. Where do the `BStr`'s owned bytes live?

> This makes me want to understand the use-case more. When does one decode and re-encode an object, and expect it to be the same byte for byte? During FSCK? Fair, but then it should actually be interesting that it's non-conforming and one could show a diff with the lines that don't match anymore.

In Mononoke, we implemented a "Git server" that's not backed by the file system, but by a "blobstore" (conceptually, db storage of bytes). It speaks the Git protocol. There are two places we get Git bytes:
* From http via the Git protocol when the Git client communicates with us
* We store the Git representation of objects in the blobstore
For everything else internally, we use `gix-object` types to represent the actual structure and decide what to do.

It's a choice that may not be absolutely set-in-stone, but we tend to use the owned version of `gix-object` types by default mainly for convenience, to avoid having to worry about lifetimes when we do largely parallel things. Arguably, we could store the raw bytes and the ref objects in the same struct and stick it all in an Arc whenever we need to do parallel things, but that's not currently what we do.
Our code is not quite as performance sensitive as something like a Git client as we can make trade-offs like throwing hardware at it and the architecture of Mononoke tends to focus on bounding the worst-case computations and mem usage (by sharding unbounded things like files) and trading off storage for compute by pre-computing everything that may be needed at runtime (for instance, the blame asynchronously outside of the write path), so most of what the `mononoke_git_server` does is fetch bytes from storage and stream them to clients at a high level.
We also clearly make trade-offs of the kind: better anything now than the best thing later as the current system that is backed by Git is much more of a hassle to maintain than even a naive implementation of the Mononoke Git server (and we can optimize once we hit bottlenecks in prod).
For us, blobstore storage is indexed by the Git hash, so if the hash of the gix-object that we are manipulating doesn't match the hash from Git (the one the client asks us about), we get issues.

Anyway, to get back to the issue at hand, I think we should maintain round-tripping between `Object` and `ObjectRef` if possible.
We could do it in a stricter way than with `BString` `BStr`, for instance by only adding an optional seconds field to `gix-date::Time`; but it feels like a decision that we may regret if at some point we find another slight variation on how Git may encode the time.
A mix of `Object` using `Time` and `ObjectRef` using `BStr` would only work with a backing of an array of bytes and that doesn't feel great to me, but I guess it's an option.
Or there is the way in this PR, which I think is practical.
For one `Option<BString>` for the entire signature, it may turn out to be necessary ultimately, but I haven't seen evidence for it in practice, so I'd rather close that bridge if and when we get there.

FYI, I'm currently tackling the long tail of repos that have caused issues (failure for us to import them in Mononoke or to clone them from Mononoke) out of >7000 repos, only ~30 couldn't be imported yet, and with the other PRs I made and this one, there is a single issue left that I'm aware of and hasn't been adressed yet (a strange way to encode the PGP signature. PR soon).

So at least experimentally, we're not too far from handling every edge case. Once all this code is upstream, I'll make sure we can clone each repo again, to catch any regression or omission.

Byron:
Sorry for the late reply, I feel off a cliff there, unrelated to this issue which, admittedly, is among the big 'small' problems to solve. However, I think it will also help to define a boundary for what round-tripping means, and to what extend that's possible.

Just as a model I have in my head, we have

* **decompressed object buffer**
    - a `&[u8]` of bytes that represent the object, it's what is returned by the object database
    - It's also what is stored in `gix::Object`, it's entirely unparsed.
* **decoded object** (`*Ref`)
   - These are the `gix_object::Object` variants, where everything is parsed into fields. These fields still refer to the original bytes buffer.
   - it goes as far as having references to the original `hex` version of object ids for instance, merely as matter of convenience.

If we break downstream to support more cases of round-tripping, I'd want to do it so that it's done for good.

Looking at `SignatureRef` I also think that `time: &BStr` would be the way to go, then downstream would call `time()` to parse it on demand, similar to how it's already done everywhere else.

However, `Signature` would still contain the fully parsed types.

> Just to be sure I understand your suggestion precisely, are you advocating for having `ObjectRef` round-trip with `Git`, but `Object` not round-trip with `ObjectRef` as a preferable situation?
> 
> I see two potential issues with this:
> 
> * It feels unintuitive to loose the equivalence between the owned and the non-owned version. The contract being that they can round trip, and the details of ownership being a pure performance consideration makes more sense to me as a user.

It's true, even though I think it's more subtle as the difference is already in the types themselves. `CommitRef` for instance uses a `BStr` for the `tree` header field, but `Commit` uses `ObjectId`. They are equivalent, but the `BStr` can technically represent anything you would want. Practically, when writing, there is validation, but in case of `SignatureRef` there most characters are actually allowed.
Probably not a strong argument, but in short, it's already the case and I think it's natural (enough).

> * In terms of ownership, say `Object` has a sane typed representation of time but `ObjectRef` needs a `BStr`. Where do the `BStr`'s owned bytes live?

They live in the original byte buffer.

> It's a choice that may not be absolutely set-in-stone, but we tend to use the owned version of `gix-object` types by default mainly for convenience, to avoid having to worry about lifetimes when we do largely parallel things. Arguably, we could store the raw bytes and the ref objects in the same struct and stick it all in an Arc whenever we need to do parallel things, but that's not currently what we do.

I see, it's really valuable to see where the requirement is coming from, and I get why it currently is done as it is.
`ein t hours` has a similar problem, as it has to do parallel computation on commits. The way it does it is to produce byte buffers of commits, which are sent verbatim to the consumers where they are decoded. It's very natural, too, and there are no lifetime problems.
Of course, no matter what, if there would be modifications to an object, it will be written as sanitized version at least if one passes it through an owned decoded object.

> In Mononoke, we implemented a "Git server" that's not backed by the file system, but by a "blobstore" (conceptually, db storage of bytes). It speaks the Git protocol. There are two places we get Git bytes:
> 
> * From http via the Git protocol when the Git client communicates with us
> * We store the Git representation of objects in the blobstore
>   For everything else internally, we use `gix-object` types to represent the actual structure and decide what to do.
> 
> It's a choice that may not be absolutely set-in-stone, but we tend to use the owned version of `gix-object` types by default mainly for convenience, to avoid having to worry about lifetimes when we do largely parallel things. Arguably, we could store the raw bytes and the ref objects in the same struct and stick it all in an Arc whenever we need to do parallel things, but that's not currently what we do. Our code is not quite as performance sensitive as something like a Git client as we can make trade-offs like throwing hardware at it and the architecture of Mononoke tends to focus on bounding the worst-case computations and mem usage (by sharding unbounded things like files) and trading off storage for compute by pre-computing everything that may be needed at runtime (for instance, the blame asynchronously outside of the write path), so most of what the `mononoke_git_server` does is fetch bytes from storage and stream them to clients at a high level. We also clearly make trade-offs of the kind: better anything now than the best thing later as the current system that is backed by Git is much more of a hassle to maintain than even a naive implementation of the Mononoke Git server (and we can optimize once we hit bottlenecks in prod).

Thanks, that's cool :)! And it's even open-source last time I checked.

> For us, blobstore storage is indexed by the Git hash, so if the hash of the gix-object that we are manipulating doesn't match the hash from Git (the one the client asks us about), we get issues.

Here I don't understand is why objects are decoded, without mutating them, to then be re-serialized in order to obtain their hash.
In my thinking, when getting a mutable/owned version of an object, one wants to mutate it, so getting a different hash and writing clean values are both expected.


> We could do it in a stricter way than with `BString` `BStr`, for instance by only adding an optional seconds field to `gix-date::Time`; but it feels like a decision that we may regret if at some point we find another slight variation on how Git may encode the time.

I agree, and right now it seems that `SignatureRef` would be fine with `time` being a `BStr` to be able to capture everything.

> For one `Option<BString>` for the entire signature, it may turn out to be necessary ultimately, but I haven't seen evidence for it in practice, so I'd rather close that bridge if and when we get there.

I agree.

> FYI, I'm currently tackling the long tail of repos that have caused issues (failure for us to import them in Mononoke or to clone them from Mononoke) out of >7000 repos, only ~30 couldn't be imported yet, and with the other PRs I made and this one, there is a single issue left that I'm aware of and hasn't been adressed yet (a strange way to encode the PGP signature. PR soon).
> 
> So at least experimentally, we're not too far from handling every edge case. Once all this code is upstream, I'll make sure we can clone each repo again, to catch any regression or omission.

I am excited about this!

For this PR, I think the problem is that it feels too wrong to have `time: BString` in `Signature`, but it's OK to have in `SignatureRef`. Doing *may* require you to change some parts of the server implementation which pass around owned versions of these types, but I think that changing it will not make it less clear overall. In the simplest case, one would hand full byte buffers to the consumer, which decodes the object there.

I did my best but apologize if some of what I wrote is off or incoherent, I had like 10 interruptions 😅👧.

> A mix of `Object` using `Time` and `ObjectRef` using `BStr` would only work with a backing of an array of bytes and that doesn't feel great to me, but I guess it's an option.

This makes me hopeful. Maybe you can spike a case of mononoke parallelism which doesn't pass `Object` across threads, but plain buffers - an owned version of `gix_object::Data<'buf>` should be all you need.

pierrechevalier83:
> Sorry for the late reply, I feel off a cliff there, unrelated to this issue which, admittedly, is among the big 'small' problems to solve.

No worries at all. I really appreciate the time you are taking to very thoroughly weigh our options and to explain the wider vision so I can fumble in the dark a little less :)
I think we are getting to a common understanding of this problem space.

Let me suggest 3 possible way forwards, and I will follow your lead on which one to implement:

## Option 1: `gix_date::Time` is made to roundtrip and backs both `SignatureRef` and `Signature`.

This would consist of making an enum: `OffsetFormat` with maybe `HHMM` and `HHMMSS` variants and stick it in `gix-date::Time`. It would be analogous to the `Sign` enum which currently exists to distinguish between `+0000` and `-0000` offsets.
There may be a tiny performance loss compared to main, but I think it would be acceptable for the improved "correctness".

## Option 2: Heterogeneous Signature with roundtrip

We implement the fix described in Option 1, which applies to `Signature`, but we back `SignatureRef` with a `&BStr` for "future-proofing" of handling new weird cases that may show up. We get roundtripping today, but no guarantee of roundtripping without further modification for the owned type in case we hit more exotic scenarios in the wild.

## Option 3: Heterogenous Signature without roundtrip

* `SignatureRef` keeps having a `&BStr` and `.time()` for parsing on demand as in this PR so far
* `Signature` goes back to having a `gix-date::Time` which is unmodified

In all cases, Mononoke would need to tend towards a state where we use the `Ref` version of objects backed by the raw bytes, but options 1 and 2 would have the advantage of giving us more flexibility on when exactly to make that transition as we would get round-tripping initially, so both would be preferable from my point of view.
From my understanding of the vision for gitoxide that you've laid out above and the analogy with `Commit` and `ObjectId`, I think that options 2 and 3 would both fit well within the wider vision.

What do you think? Which one should I implement?

Byron:
Thanks for the analysis and for deriving options! Some of these I implicitly discarded, but it's great to see them written out.

Bringing up the `sign` field in `gix_date::Time` is great because I think it's a mistake, created in a time when the purpose of `*Ref` and the meaning for round-tripping weren't sketched out yet. The idea is that the fully-owned versions would always serialize to a 'standard' version of an object, whereas buffer-backed decoded objects should round-trip. Right now, this most definitely isn't the case for some variants of name/emails (and there are tests for these which are sourced from the real world), but dealing with this can be done once there is demand.

Thus, let's go ahead with **Option 3**. I'd still expect that `SignatureRef::time()` is still infallible if the object parsed in the first place, but it seems there is no way to know since a user could set `myref.time = "foo"` and call `time()` on it. I will go with your intuition for that one, which should also be relevant for `SignatureRef::to_owned()`.

Lastly, if you wanted to truly gift `gitoxide`, I'd love it if `Time::sign` could go away in favor of round-tripping through `SignatureRef` - I am sure Git has a choice on the sign it uses for `0` and the `Time` serialization would just stick to that.

Thanks again, it's my pleasure to work with you.

pierrechevalier83:
> Thus, let's go ahead with **Option 3**. I'd still expect that `SignatureRef::time()` is still infallible if the object parsed in the first place, but it seems there is no way to know since a user could set `myref.time = "foo"` and call `time()` on it. I will go with your intuition for that one, which should also be relevant for `SignatureRef::to_owned()`.

I went ahead and implemented it in an extra commit on top for ease of review. I can go ahead and improve the history later.
I'm mostly sharing for feedback at this time. I'm uneasy with the `.expect("Time must be valid")` lines, but not 100% sure what's the best way to deal with invalid times wrt converting between `Signature` and `SignatureRef`.

> Lastly, if you wanted to truly gift `gitoxide`, I'd love it if `Time::sign` could go away in favor of round-tripping through `SignatureRef` - I am sure Git has a choice on the sign it uses for `0` and the `Time` serialization would just stick to that.

I can follow-up with this in a separate PR once we've got this one finalized. Would you like me to open an issue so we don't forget?

pierrechevalier83:
It seems the fuzzer like me is not a fan of:
```
Time must be valid: InvalidDateString { input: "0 -217" }
```

gix-actor/src/signature/mod.rs
```diff
@@ -61,16 +63,10 @@ mod convert {
             Signature {
                 name: name.to_owned(),
                 email: email.to_owned(),
-                time: time.to_owned(),
+                time: Time::from_bytes(time).expect("Time must be valid"),
```
Byron, line 66: This `From` implementation would also have to be removed now, but should probably be replaced with `TryFrom`.

Byron (commented):
Thanks for sketching this! And yes, let's keep committing on top.

> I can follow-up with this in a separate PR once we've got this one finalized. Would you like me to open an issue so we don't forget?

I added some tasks for myself in the PR text and will just take this on myself before merging.

I think everything looks good and `ein t hours` as well as `gix-mailmap` are good examples for how the downstream could adjust to these changes without losing anything.

Lastly, the only change needed here is to make `to_owned()` fallible - that should also fix the fuzzer. Once done, I will take it from there.
If you won't get to it within a day or so please let me know and I can finish the PR from here.

Looking forward to merging! This PR affected `gitoxide` a lot as it now finally became clear how to handle real-world garbage (`*Ref`) while also assuring users will produce only proper objects when creating them (`*[NonRef]`).

Byron:
Just FYI, I make good progress and plan to merge this PR today.

Byron (approved):
Alright, I think this will be it!

I recommend taking a look at the changes to `gix-actor` and `gix-date` just to catch up with my changes. `gix-date` should be more conformant with Git now, and overall I am very happy with how it turned out.

pierrechevalier83:
> I recommend taking a look at the changes to gix-actor and gix-date just to catch up with my changes. gix-date should be more conformant with Git now, and overall I am very happy with how it turned out.

Awesome. I had a look and it makes sense based on our discussion. Thanks a lot for carrying this PR over the line :)

Byron:
I also hope it works for you in Mononoke!
Ultimately I think the code will not be harder to maintain, but I am definitely interested in hearing your verdict once you have one.

ilyagr:
I'm adapting `jj` to these changes, and I got a bit confused. I think I mostly figured things out, but:

**Question:** Is the medium-term plan to keep `Signature` as is, or will it also have time turned into a String?

**Suggestion:** If `Signature` stays the same, I find it really confusing that `SignatureRef` is no longer, semantically, a Ref to a Signature. Perhaps rename it to `UnparsedSignatureRef`? I'm not sure whether there's a need for a real `SignatureRef`. (And if `Signature` changes to track the unparsed timestamp, I think there might be a need for `ParsedSignature`).

~~**TODO:** I'll try to remember to update the docs a bit...~~

**Update:** I collected some of my guesses in https://github.com/GitoxideLabs/gitoxide/pull/2038.

Byron:
> **Question:** Is the medium-term plan to keep `Signature` as is, or will it also have time turned into a String?

Yes, `Signature` is going to stay as is, `time` will not turn into a `String`.

And I think you caught something that adds to confusion: `SignatureRef` isn't a ref to `Signature`, but a ref to the underlying buffer. All `*Ref` types are, but previously they allowed themselves to parse more fields eagerly.
In the name of round-tripping, this can't be done anymore so is done lazily.

The non-`Ref` versions are for creation or manipulation by the user, so everything is strongly typed there.

In short, as `SignatureRef` never was a reference to `Signature`, I think its name is still appropriate.
Maybe I am missing something though.

ilyagr:
Thanks!

This makes a lot more sense now. My first reaction is that, if I was starting from scratch, I'd name `SignatureRef` something else, e.g. `SignatureView`, and I wouldn't use `.as_ref` as a function to create it.

(It's an unfortunate property of Rust that one might, in fact, want a `SignatureRef` type that's semantically the same as `&Signature` but not quite the same, like `&str` vs `&String` and `Path` vs `PathBuf`.)

However, this might be arguable. After our combined efforts, https://github.com/GitoxideLabs/gitoxide/pull/2038 makes me optimistic that a lot of this confusion can be solved with just docs.

Byron:
That's great to hear, and let's keep watching this!

I see that `SignatureRef` may now be named counter-intuitively as one will easily think it's a Ref to a `Signature`. `SignatureView` could be better, but I'd hope there is an even better name to indicate this structure is more of an index into a buffer.

On top of that, I'd want to avoid leaving this new name as special case, and instead see if `CommitRef` should now also be named differently. In other words, a new name for these Ref types should be validated holistically.

Suggestions are definitely welcome, and I will also see if something comes to mind.

ilyagr:
Now I'm on the other side of this discussion and want to round-trip a Commit object, even if it has a malformed signature. (Thanks everyone for working towards making this possible, whether or not it is already!)

However, it seems that `gix_object::Commit` can only accept valid signatures. Is there a way to pass a string or `Vec<u8>` to something like that to generate a commit object?

ilyagr:
Ah, maybe I figured it out. I guess I have to create a https://docs.rs/gix-object/latest/gix_object/struct.CommitRef.html.

It looks like I'll have to finagle all the types like `Cow<BStr>` manually, there doesn't seem to be many helper functions, e.g. no `Commit::to_ref`.

Byron:
Yes, that's it. The NonRef versions of objects must always be valid while the ones that still point to a backing buffer may also represent something invalid (or anything parseable that the backing buffer happens to contain).

One can probably start out with a `Commit`, serialize that, deserialise it into a `CommitRef` and make the final modifications there.

PR #1936, opened 2025-04-06
Handle trailing slash in ref list prefix filtering

### LINKED ISSUE #1934: refs.references().prefixed() lets too many refs through
### Current behavior 😯

Some refs reported by `repo.references()?.prefixed(...)` are outside the prefix scope. The refs that are reported are loose refs which are not present in `.git/packed-refs`.

This has been tested with `gix=0.71.0` and overriding with `gix-features=0.41.1`.

Relates to #1928 with the PR #1931.

### Expected behavior 🤔

Only refs that starts with the prefix are reported.

### Git behavior

_No response_

### Steps to reproduce 🕹

Create `.git/refs/heads/foo/bar`. That ref will be listed with `repo.references()?.prefixed("refs/heads/b")`.

moroten:
Previously, `refs/heads/foo/bar` would be listed when running `repo.references()?.prefixed("refs/heads/b")`. The code identified that the last component was not a directory and started to match it as a filename prefix for all files in all recursive directories, effectively matching `refs/heads/**/b*`.

This commit fixes that bug but also allows to use a trailing `/` in the prefix, allowing to filter for `refs/heads/foo/` and not get `refs/heads/foo-bar` as a result.

Fixes #1934.


gix-ref/src/store/file/loose/iter.rs
```diff
@@ -14,17 +14,22 @@ pub(in crate::store_impl::file) struct SortedLoosePaths {
...
                 // serial iteration as we expect most refs in packed-refs anyway.
-                gix_features::fs::walkdir_sorted_new(
+                let mut walkdir = gix_features::fs::walkdir_sorted_new(
                     path,
                     gix_features::fs::walkdir::Parallelism::Serial,
                     precompose_unicode,
-                )
-                .into_iter()
+                );
+                // With a filename prefix, only files within the base are of interest.
```
Byron, line 28: I don't think this is the case though. The idea behind a prefix is to match everything within a prefix, which can include arbitrarily deeply nested refs.

Byron (changes requested):

I think what would be needed here is a test that shows the problem this is solving, i.e. the iteration failing to properly deduplicate results from packed-refs and loose refs when a prefix is used.

Such an example should include deep nesting as well.

Prefixed search is necessary to help to quickly bypass a possibly large list of results, and it's essential to do quick listing (without reftable) during fetches in repositories where a lot of references exist.

EliahKagan (commented):
The current test failure is due to the need to commit a regenerated `gix-ref/tests/fixtures/generated-archives/make_ref_repository.tar`. This is needed because the script that generates it, `gix-ref/tests/fixtures/make_ref_repository.sh`, is modified here. To verify that this was the only cause of failure so far, I've regenerated that archive in [`ek/run-ci/regenerate-archives-for-pr-1936`](https://github.com/GitoxideLabs/gitoxide/commit/b6ab162fe95de0c282decf57415f4f2708c3f436) (b6ab162).

If you like, you can fast-forward this PR to include that commit, though it might not be useful to do so if the further changes for https://github.com/GitoxideLabs/gitoxide/pull/1936#pullrequestreview-2745325065 will include modifying `make_ref_repository.sh` again.

```sh
git fetch --all
git cherry-pick upstream/ek/run-ci/regenerate-archives-for-pr-1936
```

Please disregard this if it is not convenient or if `make_ref_repository.sh` shall be changed again.

gix-ref/src/store/file/loose/iter.rs
moroten: Went for a totally different solution.

moroten:
Thank you @EliahKagan. It turned out that I missed the binary files when using `git add -p`.

moroten:
I've rewritten the PR to use a `BStr` as prefix instead of `Path`. This allows for trailing slash.

Is there any use case where Windows backslash need to be translated to forward slash that I may have missed?

gix-ref/src/store/file/overlay_iter.rs
```diff
@@ -195,10 +195,9 @@ impl Platform<'_> {
         self.store.iter_packed(self.packed.as_ref().map(|b| &***b))
     }
 
-    /// As [`iter(…)`][file::Store::iter()], but filters by `prefix`, i.e. "refs/heads".
-    ///
-    /// Please note that "refs/heads" or "refs\\heads" is equivalent to "refs/heads/"
-    pub fn prefixed(&self, prefix: &Path) -> std::io::Result<LooseThenPacked<'_, '_>> {
+    /// As [`iter(…)`][file::Store::iter()], but filters by `prefix`, i.e. "refs/heads/" or
+    /// "refs/heads/feature-".
+    pub fn prefixed(&self, prefix: Cow<'_, BStr>) -> std::io::Result<LooseThenPacked<'_, '_>> {
```
Byron, line 200: I think it's fair to have this breaking change, as this prefix is essentially a `RepositoryPath`, i.e. a slash-separated relative path.
Something I think that should come with it is tests that show that behaviour.

Byron, line 200: This should be `&BStr` if i twas `&Path` before.

gix-ref/src/store/file/loose/iter.rs
```diff
@@ -93,7 +87,7 @@ impl file::Store {
     /// Return an iterator over all loose references that start with the given `prefix`.
     ///
     /// Otherwise it's similar to [`loose_iter()`][file::Store::loose_iter()].
-    pub fn loose_iter_prefixed(&self, prefix: &Path) -> std::io::Result<LooseThenPacked<'_, '_>> {
+    pub fn loose_iter_prefixed(&self, prefix: Cow<'_, BStr>) -> std::io::Result<LooseThenPacked<'_, '_>> {
```
Byron, line 90: This should be `&BStr` if it was `&Path` before.
It could also be `impl Into<…>` if that makes it more convenient to call, even for the test-suite.

gix-ref/tests/fixtures/make_packed_ref_repository_for_overlay.sh
```diff
@@ -10,10 +10,13 @@ git branch A
 git tag -m "tag object" tag-object
 
 mkdir -p .git/refs/remotes/origin
+mkdir -p .git/refs/heads/sub/dir
 
 cp .git/refs/heads/main .git/refs/remotes/origin/
 
 echo "ref: refs/remotes/origin/main" > .git/refs/remotes/origin/HEAD
+echo "ref: refs/heads/main" > .git/refs/heads-packed
+echo "ref: refs/heads/main" > .git/refs/heads/sub/dir/packed
```
Byron, line 19: I think these names should be changed so `packed` isn't included. This is because symbolic refs are never packed.

Byron (commented):
Thanks a lot for taking this PR further, it looks great.

I'd have to take it into another round to improve a couple of things that are mentioned below. Most importantly, a test is needed to show that `\\` isn't special anymore. Unfortunately, there isn't a breaking that that would have tested for it being treated specifically either, so it might be that the claim was never true.

From there I can promise no more rounds, I'd take it from there with an in-IDE review, fixing whatever small things that come up myself.

If you feel like it's enough, please let me know and I will happily pick it up as soon as I can to get it merged.

gix-ref/src/store/file/overlay_iter.rs
moroten: The tests are now more explicitly named to test that.

gix-ref/tests/fixtures/make_packed_ref_repository_for_overlay.sh
moroten: I just wanted to make sure that both the loose and packed refs are handled correctly and not accidentally testing the wrong thing. Changed the names to `refs/feature-suffix` and `refs/feature/sub/dir/algo` to make the test more explicit.

gix-ref/src/store/file/loose/iter.rs
moroten: I put `impl Into<Cow<...>>` in most places. Not sure if some should just be `&BStr` or not.

Byron:
Thanks again for tackling this!

I am closing this PR as I wasn't able to push my changes into it. Instead, I opened #1954.

PR #1938, opened 2025-04-07
Bump tokio to 1.44.2 to fix RUSTSEC-2025-0023

EliahKagan:
This upgrades `tokio` from 1.44.1 to 1.44.2 in `Cargo.lock` to fix the advisory [RUSTSEC-2025-0023](https://rustsec.org/advisories/RUSTSEC-2025-0023.html) ("Broadcast channel calls clone in parallel, but does not require `Sync`").

This was done with `cargo update -p tokio`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3654,7 +3654,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fc2f4eb4bc735547cfed7c0a4922cbd04a4655978c09b54f1f7b228750664c34"
 dependencies = [
  "cfg-if",
- "windows-targets 0.52.6",
+ "windows-targets 0.48.5",
 ]
 
 [[package]]
@@ -5292,9 +5292,9 @@ checksum = "1f3ccbac311fea05f86f61904b462b55fb3df8837a366dfc601a0161d0532f20"
 
 [[package]]
 name = "tokio"
-version = "1.44.1"
+version = "1.44.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f382da615b842244d4b8738c82ed1275e6c5dd90c459a30941cd07080b06c91a"
+checksum = "e6b88822cbe49de4185e3a4cbf8321dd487cf5fe0c5c65695fef6346371e9c48"
 dependencies = [
  "backtrace",
  "bytes",


Cargo.lock
```diff
@@ -3654,7 +3654,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fc2f4eb4bc735547cfed7c0a4922cbd04a4655978c09b54f1f7b228750664c34"
 dependencies = [
  "cfg-if",
- "windows-targets 0.52.6",
+ "windows-targets 0.48.5",
```
EliahKagan, line 3657: I'm not at all clear on why the `windows-targets` dependency of `libloading` is now specified as 0.48.5 instead of 0.52.6, without changing the `libloading` version or any of its other metadata. All I ran to bump the version of `tokio` was `cargo update -p tokio`. Nothing in [tokio-1.44.1...tokio-1.44.2](https://github.com/tokio-rs/tokio/compare/tokio-1.44.1...tokio-1.44.2) seems to show any obvious reason this would relate to `tokio`. Anyway, we depend on lots of different versions of `windows-targets` separately in `Cargo.lock` and I suppose this is fine.

Skgland: While this looks like a valid dependency graph resolution based on `libloading@0.8.6`s version requirement for windows-target which is [`">=0.48, <0.53"`](https://docs.rs/crate/libloading/latest/source/Cargo.toml.orig#19).
Not choosing the latest version already in the dependency graph is unexpected.
Maybe this should be filed as a cargo bug?

Skgland: Does it work to just drop this change or does cargo re-generate this change when some command is run?

EliahKagan: If I make this change to undo it:

```diff
diff --git a/Cargo.lock b/Cargo.lock
index 3a54ce742..d32caad04 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3654,7 +3654,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fc2f4eb4bc735547cfed7c0a4922cbd04a4655978c09b54f1f7b228750664c34"
 dependencies = [
  "cfg-if",
- "windows-targets 0.48.5",
+ "windows-targets 0.52.6",
 ]

 [[package]]
```

Then `cargo build --locked`, `cargo build -r --locked`, and `cargo nextest run --locked --workspace --no-fail-fast` all still succeed and make no changes to `Cargo.lock` nor to any other tracked files. But running `cargo update -p tokio` again puts it back to 0.48.5, even while making no change to the version of `tokio` itself.

Because `cargo update -p tokio` would change it again, combined with my general reluctance to manually edit `Cargo.lock`, I am reluctant to put it back unless manually editing `Cargo.lock` is considered a best practice in this kind of situation (is it?), or unless we can be pretty sure that this `cargo update -p tokio` behavior is a bug (can we?).

In case it's useful, a commit on top of these with that change made is pushed to [`run-ci/update-tokio-next`](https://github.com/EliahKagan/gitoxide/commits/run-ci/update-tokio-next/) on my fork (https://github.com/EliahKagan/gitoxide/commit/6aeba88b4c3a2190d0337db1f72c58e9709ce481). I expect that it will pass CI, but even if so, I am still not sure if it's better or worse than what's here currently.

EliahKagan:
I'll let the Dependabot PR #1941 supersede this. It makes the same change to the version, listed for `libloading,` of `windows-targets` as discussed in https://github.com/GitoxideLabs/gitoxide/pull/1938#pullrequestreview-2747939195. But for the reasons given there I'm reluctant to manually change that back. That Dependabot also produces it indicates that at least it wasn't specific to anything broken or unusual on my local machine.

PR #1940, opened 2025-04-08
Have `gix-filter` depend on `gix-packetline-blocking` unaliased

EliahKagan:
This removes the `gix-packetline-blocking/futures-lite` "ghost feature" added in d5dd239 (#1939), and instead:

- Changes `gix-filter` to depends on `gix-packetline-blocking` without aliasing it `gix-packetline`.
- Replaces uses of `gix_packetline` with `gix_packetline_blocking` in the code of `gix-filter`.

Thus, this replaces the solution in #1939 for the problem discussed in https://github.com/GitoxideLabs/gitoxide/pull/1929#issuecomment-2783852929 with a different solution that avoids carrying a "ghost feature."

In contrast, the long-standing `gix-packetline-blocking/async-io` feature added in be4de0d (#1123), though it should not be used, is *not* removed, because:

- It is referenced in attributes in `gix-packetline-blocking` code (because `gix-packetline-blocking/src` is copied from `gix-packetline/src` via `etc/copy-packetline.sh`).
- For that reason, removing it would cause the `clippy` run in the CI `lint` job to fail, as well as likely making various other reasonable operations fail.

---

This PR proposes the alternative alluded to in #1939. As they are described in https://github.com/GitoxideLabs/gitoxide/pull/1929#issuecomment-2783852929, this replaces way (1) with way (2). At the time it seemed like combining them, i.e. way (3), could make sense, but I think the phantom nature of the feature, together with how `gix-filter` seems to be the only user of `gix-packetline-blocking`, indicates that it shouldn't be carried if it's not actually being used.

The reason I didn't go with this initially is that I am not sure I understand why `gix-packetline-blocking` was aliased `gix-packetline` when given as a dependency of `gix-features`. If it was solely for the convenience of being able to write `gix_packetline` instead of `gix_packetline_blocking`, then I think the change here is an improvement. If there are deeper reasons, however, then I am not sure.

I think this could be done with `use gix_packetline_blocking as gix_packetline;`, but it didn't look like it would be better: it would have to be written in multiple places in the same file, due to the scope of `use` declarations in files that define multiple nested modules. (And maybe also related to special semantics for `#[from]`? But I am not sure about that.)

### DIFF
--- a/gix-filter/Cargo.toml
+++ b/gix-filter/Cargo.toml
@@ -22,7 +22,7 @@ gix-command = { version = "^0.5.0", path = "../gix-command" }
 gix-quote = { version = "^0.5.0", path = "../gix-quote" }
 gix-utils = { version = "^0.2.0", path = "../gix-utils" }
 gix-path = { version = "^0.10.15", path = "../gix-path" }
-gix-packetline = { package = "gix-packetline-blocking", version = "^0.18.3", path = "../gix-packetline-blocking" }
+gix-packetline-blocking = { version = "^0.18.3", path = "../gix-packetline-blocking" }
 gix-attributes = { version = "^0.25.0", path = "../gix-attributes" }
 
 encoding_rs = "0.8.32"
--- a/gix-filter/src/driver/process/client.rs
+++ b/gix-filter/src/driver/process/client.rs
@@ -41,7 +41,7 @@ pub mod invoke {
             #[error("Failed to read or write to the process")]
             Io(#[from] std::io::Error),
             #[error(transparent)]
-            PacketlineDecode(#[from] gix_packetline::decode::Error),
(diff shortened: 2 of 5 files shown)


Byron:
Thanks so much, you are on fire!

> The reason I didn't go with this initially is that I am not sure I understand why `gix-packetline-blocking` was aliased `gix-packetline` when given as a dependency of `gix-filter`. If it was solely for the convenience of being able to write `gix_packetline` instead of `gix_packetline_blocking`, then I think the change here is an improvement. If there are deeper reasons, however, then I am not sure.

Indeed, this was only for convenience, but given that it needs the blocking implementation, the code may as well reflect it.
Thus, this is certainly the preferred solution.

PR #1950, opened 2025-04-11
refactor: replace `futures_lite::ready!` with `std::task::ready!`

paolobarbolini:


Byron:
Thanks a lot, a great catch!

PR #1957, opened 2025-04-14
Advance `gix-features` version, as it republishes `bytesize` version 2

EliahKagan:
From https://github.com/GitoxideLabs/gitoxide/discussions/1952#discussioncomment-12827132, you had explained:

> In spite of a commit that would indicate breakage, one can always manually increment the crate's version number in the cargo manifest. This will also do the right thing when using CSR.

Per https://github.com/GitoxideLabs/gitoxide/discussions/1952#discussioncomment-12835206, I don't know if that needs to be done at the time of publication and not before. If it should not be done now, then this PR can simply be closed. I figured that, rather than try to exhaustively describe this idea in the discussion in #1952, I would open a PR for it and let the PR be reviewed.

This attempts to increase the `gix-features` version number now so that, regardless of the circumstances, any later `gix-features` release that republishes `bytesize` version 2 will carry a version number consistent with a breaking change.

Even if this is the right thing to do, I am not sure if I have done it in the right way. In case this would cause other problems, I will refrain from merging this without a review.

- The commit that increases the version number in `gix-features` is itself a conventional commit. The idea is to get this into the changelog too. I don't know if that makes sense, though. Will that result in `cargo-smart-release` bumping the version *again* so as to skip this version? (I had originally thought to use `feat:` instead of `feat!:`, but if `feat!:` causes a problem, then I think `feat:` causes an only slightly less severe variant of that problem: `cargo-smart-release` would still change the version number.)
- A number of references to `gix-features` give its path but also a version number. Since the version number at that path is changed, they also have to be updated. Thus the second commit here changes the version of the `gix-features` dependency in various crates that depend on it. But the other parts of those crates' `Cargo.toml` files are the same. For example, I have not advanced their version numbers--this is with the idea that `cargo-smart-release` will do so.

Information relevant to people who may be reading the commit history or changelog are in the commit messages. In particular, the first commit message contains a description of the situation and the effect of the changes that have already been made but that motivate the change in version number.


Byron (approved):
Thanks a lot, this will work!

> * The commit that increases the version number in `gix-features` is itself a conventional commit. The idea is to get this into the changelog too. I don't know if that makes sense, though. Will that result in `cargo-smart-release` bumping the version _again_ so as to skip this version? (I had originally thought to use `feat:` instead of `feat!:`, but if `feat!:` causes a problem, then I think `feat:` causes an only slightly less severe variant of that problem: `cargo-smart-release` would still change the version number.)

Indeed, it would want to have a breaking change compared to the last published version, increment the current (already incremented one), but then notice that the current version is already enough to indicate a breaking change, and do nothing effectively.

> * A number of references to `gix-features` give its path but also a version number. Since the version number at that path is changed, they also have to be updated. Thus the second commit here changes the version of the `gix-features` dependency in various crates that depend on it. But the other parts of those crates' `Cargo.toml` files are the same. For example, I have not advanced their version numbers--this is with the idea that `cargo-smart-release` will do so.

I don't know what these other parts would be, but as long as Cargo can build the project, the version update was done correctly.

For safety I checked that there indeed is no unchanged `gix-features` reference in any Cargo manifest.

EliahKagan:
> Indeed, it would want to have a breaking change compared to the last published version, increment the current (already incremented one), but then notice that the current version is already enough to indicate a breaking change, and do nothing effectively.

Ah, sounds good. (I take this to mean that the published version will not, in effect, be any higher, unless there is a reason for it to be.)

> I don't know what these other parts would be

Probably just those dependent crates' versions. And I'm not really sure what I was worried about there--crates get new and changed dependencies all the time prior to their own version numbers being bumped immediately before their next release.

> but as long as Cargo can build the project, the version update was done correctly.
> 
> For safety I checked that there indeed is no unchanged `gix-features` reference in any Cargo manifest.

Thanks!

Byron:
> Probably just those dependent crates' versions. And I'm not really sure what I was worried about there--crates get new and changed dependencies all the time prior to their own version numbers being bumped immediately before their next release.

Yes, their version number will be bumped automatically. CSR has many flaws, but this one thing is does pretty well 😅.

PR #1960, opened 2025-04-20
Fix documentation of zlib-rs feature: no longer conflicts with other zlibs

joshtriplett:
As of zlib-rs 0.5.0 (depended on by flate2 1.1.1, which all dependency versions in gitoxide already specify as a minimum), zlib-rs no longer exports C symbols by default, so it doesn't conflict with any other zlib that might be loaded into the address space. Update the documentation accordingly.


Byron (approved):
Thanks a lot for your help!

`zlib-rs` is an incredible achievement and I love to see it spread.

PR #1963, opened 2025-04-21
Switch to zlib-rs by default and drop other zlib backends

joshtriplett:
As of zlib-rs 0.5.0 (depended on by flate2 1.1.1), zlib-rs no longer
exports C symbols by default, so it doesn't conflict with any other zlib
that might be loaded into the address space.

This removed the primary issue that made zlib selection a challenge
that needed to be exposed to users: users may already have some other
particular preference on zlib implementations, or want to use a system
library, or want to ensure C dependencies use a particular zlib. Since
zlib-rs 0.5.0 no longer conflicts with other zlib implementations, gix
can make this choice independently and not create issues for the user.

Given that, use zlib-rs by default, for performance out of the box with
no C compiler requirement, and deprecate all the feature flags for other
variations. A future major version bump can drop all of these features.

This also means that (for instance) max-performance becomes the same as
max-performance-safe, so such features are deprecated as well.

Depend on flate2 with `default-features = false`, which brings us closer
to eliminating the `miniz_oxide` dependency, which will improve build
time for everyone. Currently, the `gix-archive` dependency on `zip`
still activates the `zip/deflate` feature which enables
`flate2/rust_backend`; this can be fixed to use `deflate-flate2` as soon
as https://github.com/zip-rs/zip2/pull/340 is fixed upstream.

Fixes: https://github.com/GitoxideLabs/gitoxide/issues/1961

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
         continue-on-error: true
       - name: Check for unrecognized *-sys dependencies
         run: |
-          ! grep -qP '(?<!\blinux-raw)-sys\b' tree.txt
+          ! grep -qP '(?<!\b(linux-raw|libz-rs))-sys\b' tree.txt
       - name: Wrap cc1 (and cc1plus if present) to record calls
         run: |
           set -o noclobber  # Catch any collisions with existing entries in /usr/local.
@@ -402,7 +402,7 @@ jobs:
       - name: features of gix-features
         run: |
           set +x
-          for feature in progress parallel io-pipe crc32 zlib zlib-rust-backend cache-efficiency-debug; do
+          for feature in progress parallel io-pipe crc32 zlib cache-efficiency-debug; do
             (cd gix-features && cargo build --features "$feature" --target "$TARGET")
           done
       - name: crates with 'wasm' feature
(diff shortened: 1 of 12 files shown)


joshtriplett:
OK, CI is passing now.

joshtriplett:
@Byron Thanks for merging this!

Regarding the test fix for `cargo test -r`, I *think* it'd be fine to use `cfg(test)` to make the API not exist when not testing. `cfg(test)` should work any time you're doing `cargo test`, whether you're using `-r` or not.

Byron:
I am thanking you!

> Regarding the test fix for `cargo test -r`, I _think_ it'd be fine to use `cfg(test)` to make the API not exist when not testing. `cfg(test)` should work any time you're doing `cargo test`, whether you're using `-r` or not.

That didn't work, unfortunately, as integration tests don't build the crate under test with `test`, just the test-binary itself if built with `test`. I tried that first and settled with leaving it in, but making it hard to discover while discouraging its usage.

Byron:
Indeed, that would work and it's nothing I even considered, admittedly. Moving tests from integration to unit-level can still happen, of course, but I think it's fine to do it once there is a trigger, like code in the wild that somehow abuses the presence of this test-trait.

PR #1972, opened 2025-04-27
prepare new testtools release

Byron:
With this release downstream will finally have a fix for audit-failures,
as it now uses a fixed version of `gix-features`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -353,7 +353,7 @@ dependencies = [
  "bitflags 2.9.0",
  "cexpr",
  "clang-sys",
- "itertools 0.10.5",
+ "itertools 0.12.1",
  "lazy_static",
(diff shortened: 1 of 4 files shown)


Byron:
@EliahKagan This PR also updates the `gix-testtools` dependencies to the very latest version. This is supposed to create an incentive to fix CSR on the next regular `gitoxide` release, or alternatively, temporarily downgrade the gix-testtools dependencies.
The idea is that in the worst case, one will release `gix-testtools` with `gix` so there is no duplication, and then downgrade the dependencies locally afterwards so the next `gitoxide` release can go through unimpeded.
I somehow think that CSR should be able to deal with the case where `gix-testtools` refers to the latest published version of these crates, and if not it should be very possible to teach it.

PR #1974, opened 2025-04-29
Move `commit_time` to `gix_traverse::commit::Either`

cruessler:
- feat: Add `commit::Either::commit_time()`
- Make use of `gix_traverse::commit::Either::commit_time()`


Byron (approved):
Thanks a lot!

PR #1976, opened 2025-04-30
Improve support for multiple blame ranges

holodorum:
This PR implements various improvements suggested in PR #1973.

The main change is converting the `BlameRanges` struct, into an enum to support both `OneBasedInclusive` and `ZeroBasedExclusive` range formats. The `FullFile` variant, denotes that the entire file is to be blamed, serves as a more explicit substitute for a previously used "empty" range.

### DIFF
--- a/gix-blame/src/error.rs
+++ b/gix-blame/src/error.rs
@@ -28,7 +28,9 @@ pub enum Error {
     #[error(transparent)]
     DiffTree(#[from] gix_diff::tree::Error),
     #[error("Invalid line range was given, line range is expected to be a 1-based inclusive range in the format '<start>,<end>'")]
-    InvalidLineRange,
+    InvalidOneBasedLineRange,
+    #[error("Invalid line range was given, line range is expected to be a 0-based inclusive range in the format '<start>,<end>'")]
+    InvalidZeroBasedLineRange,
     #[error("Failure to decode commit during traversal")]
     DecodeCommit(#[from] gix_object::decode::Error),
     #[error("Failed to get parent from commitgraph during traversal")]
--- a/gix-blame/src/error.rs
+++ b/gix-blame/src/error.rs
@@ -29,8 +29,8 @@ pub enum Error {
     DiffTree(#[from] gix_diff::tree::Error),
     #[error("Invalid line range was given, line range is expected to be a 1-based inclusive range in the format '<start>,<end>'")]
     InvalidOneBasedLineRange,
-    #[error("Invalid line range was given, line range is expected to be a 0-based inclusive range in the format '<start>,<end>'")]
-    InvalidZeroBasedLineRange,
+    #[error("Invalid range addition was attempted. Cannot add a range to a whole file.")]
+    InvalidRangeAddition,
     #[error("Failure to decode commit during traversal")]
     DecodeCommit(#[from] gix_object::decode::Error),
     #[error("Failed to get parent from commitgraph during traversal")]
--- a/gix-blame/src/file/tests.rs
+++ b/gix-blame/src/file/tests.rs
@@ -1338,3 +1338,100 @@ mod process_changes {
         );
     }
 }
+
+mod blame_ranges {
+    use crate::BlameRanges;
+
+    #[test]
+    fn create_from_single_range() {
+        let range = BlameRanges::from_one_based_inclusive_range(20..=40);
+        match range {
+            BlameRanges::PartialFile(ranges) => {
+                assert_eq!(ranges.len(), 1);
+                assert_eq!(ranges[0], 19..40);
+            }
+            _ => panic!("Expected PartialFile variant"),
+        }
+    }
+
+    #[test]
+    fn create_from_multiple_ranges() {
+        let ranges = BlameRanges::from_one_based_inclusive_ranges(vec![1..=4, 10..=14]);
+        match ranges {
+            BlameRanges::PartialFile(ranges) => {
+                assert_eq!(ranges.len(), 2);
+                assert_eq!(ranges[0], 0..4);
+                assert_eq!(ranges[1], 9..14);
+            }
+            _ => panic!("Expected PartialFile variant"),
+        }
+    }
+
+    #[test]
+    fn add_range_merges_overlapping() {
+        let mut ranges = BlameRanges::from_one_based_inclusive_range(1..=5);
+        ranges.add_range(3..=7).unwrap();
+
+        match ranges {
+            BlameRanges::PartialFile(ranges) => {
+                assert_eq!(ranges.len(), 1);
+                assert_eq!(ranges[0], 0..7);
+            }
+            _ => panic!("Expected PartialFile variant"),
+        }
+    }
+
+    #[test]
+    fn add_range_merges_adjacent() {
+        let mut ranges = BlameRanges::from_one_based_inclusive_range(1..=5);
+        ranges.add_range(6..=10).unwrap();
+
+        match ranges {
+            BlameRanges::PartialFile(ranges) => {
+                assert_eq!(ranges.len(), 1);
+                assert_eq!(ranges[0], 0..10);
+            }
+            _ => panic!("Expected PartialFile variant"),
+        }
+    }
+
+    #[test]
+    fn add_range_keeps_separate() {
+        let mut ranges = BlameRanges::from_one_based_inclusive_range(1..=5);
+        ranges.add_range(6..=10).unwrap();
+
+        match ranges {
+            BlameRanges::PartialFile(ranges) => {
+                assert_eq!(ranges.len(), 1);
+                assert_eq!(ranges[0], 0..10);
+            }
+            _ => panic!("Expected PartialFile variant"),
+        }
+    }
+
+    #[test]
+    fn convert_to_zero_based_exclusive() {
+        let ranges = BlameRanges::from_one_based_inclusive_ranges(vec![1..=5, 10..=15]);
+        let ranges = ranges.to_ranges(20);
+        assert_eq!(ranges.len(), 2);
+        assert_eq!(ranges[0], 0..5);
+        assert_eq!(ranges[1], 9..15);
+    }
+
+    #[test]
+    fn convert_full_file_to_zero_based() {
+        let ranges = BlameRanges::WholeFile;
+        let ranges = ranges.to_ranges(100);
+        assert_eq!(ranges.len(), 1);
+        assert_eq!(ranges[0], 0..100);
+    }
+
+    #[test]
+    fn default_is_full_file() {
+        let ranges = BlameRanges::default();
+        match ranges {
+            BlameRanges::WholeFile => (),
+            _ => panic!("Expected WholeFile variant"),
+        }
+    }
+}
--- a/gix-blame/src/file/tests.rs
+++ b/gix-blame/src/file/tests.rs
@@ -1340,79 +1340,74 @@ mod process_changes {
 }
 
 mod blame_ranges {
-    use crate::BlameRanges;
+    use crate::{BlameRanges, Error};
+
+    #[test]
+    fn create_with_invalid_ranges() {
+        let range = BlameRanges::from_one_based_inclusive_range(0..=10);
+
+        assert!(matches!(range, Err(Error::InvalidOneBasedLineRange)));
+    }
 
     #[test]
     fn create_from_single_range() {
-        let range = BlameRanges::from_one_based_inclusive_range(20..=40);
-        match range {
-            BlameRanges::PartialFile(ranges) => {
-                assert_eq!(ranges.len(), 1);
-                assert_eq!(ranges[0], 19..40);
-            }
-            _ => panic!("Expected PartialFile variant"),
+        let range = BlameRanges::from_one_based_inclusive_range(20..=40).unwrap();
+        if let BlameRanges::PartialFile(ranges) = range {
+            assert_eq!(ranges, vec![19..40]);
         }
     }
 
     #[test]
     fn create_from_multiple_ranges() {
-        let ranges = BlameRanges::from_one_based_inclusive_ranges(vec![1..=4, 10..=14]);
-        match ranges {
-            BlameRanges::PartialFile(ranges) => {
-                assert_eq!(ranges.len(), 2);
-                assert_eq!(ranges[0], 0..4);
-                assert_eq!(ranges[1], 9..14);
-            }
-            _ => panic!("Expected PartialFile variant"),
+        let range = BlameRanges::from_one_based_inclusive_ranges(vec![1..=4, 10..=14]).unwrap();
+        if let BlameRanges::PartialFile(ranges) = range {
+            assert_eq!(ranges, vec![0..4, 9..14]);
         }
     }
 
     #[test]
     fn add_range_merges_overlapping() {
-        let mut ranges = BlameRanges::from_one_based_inclusive_range(1..=5);
-        ranges.add_range(3..=7).unwrap();
-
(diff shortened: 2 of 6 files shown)


Byron:

@cruessler would probably be the one to do the first review, and I will get it merged right after.
Thanks everyone

cruessler:

Before I start reviewing this PR in detail, I have a few questions regarding the high-level API. In particular, it seems as if the decision to change `to_zero_based_exclusive(&self, max_lines: u32) -> Result<Vec<Range<u32>>, Error>` to `to_zero_based_exclusive(&self, max_lines: u32) -> Result<BlameRanges, Error>` brings a couple of downsides with it. My main concern is that it leaks implementation details: the caller now has to know about `BlameRanges`’ internals while that was previously not the case. Also, two of `BlameRanges`’ methods are now fallible, forcing the caller to deal with errors that are coupled to the struct’s internals.

I think the proposed API can be simplified and stay more insulated from calling code by reducing the number of enum variants to `WholeFile` and `PartialFile` (or something more appropriately named) and staying with `to_zero_based_exclusive(&self, max_lines: u32) -> Result<Vec<Range<u32>>, Error>`. We could even go so far as to stay with the existing design that treats an empty `ranges` as covering the whole file.

What do you think?

Byron:
I restarted the job, and I'd expect it to go through. There is some known flakiness with tests that deal with concurrent IO, and even though it's quite rare, it happens, unfortunately.

EliahKagan:
The [original failure](https://github.com/GitoxideLabs/gitoxide/actions/runs/15003323526/job/42155583292?pr=1976#step:7:2819) in the `test-fast` job on `macos-latest` was due to #1816. That issue is the only current source of *regular* flakiness that I am aware of. I have not been keeping track of most occurrences of it, but lately I have observed such a failure every couple of days. (The other source of flakiness I am aware of is #2006, but that is not a *regular* source of flakiness--it seems to happen once or twice a year when not deliberately induced, and in any case it is not what happened here.)

Because the `test-fast` jobs are defined by a matrix with an implicit `fail-fast: true` strategy (i.e., `fail-fast: false` is not specified), any `test-fast` job failure will tell whatever other `test-fast` jobs are still running to cancel. It looks like you reran only the failed `macos-latest` job. This automatically also reruns the dependent `tests-pass` job, but not the sibling jobs that had been canceled. So although the `macos-latest` job [passed when rerun](https://github.com/GitoxideLabs/gitoxide/actions/runs/15003323526/attempts/2?pr=1976), but the three canceled jobs did not rerun. `tests-pass` saw those jobs still had canceled status, and reported failure again.

I've rerun the workflow as a whole, and [all jobs passed](https://github.com/GitoxideLabs/gitoxide/actions/runs/15003323526?pr=1976). From context, it looks like the failing tests might have been all that were still blocking this PR form being merged. However, since auto-merge wasn't enabled here and I haven't been following this PR closely, I have refrained from merging it, in case my understanding is not correct.

EliahKagan:
> Thanks also for rerunning CI properly, I will keep that in mind.

The way I reran it was arguably overkill--when rerunning a whole workflow, one can select "Re-run failed jobs" instead of "Re-run all jobs", which in spite of its name, I *think* does also rerun canceled jobs from the same matrix as a failed job that were canceled because of it.

("Re-run all jobs" and "Re-run failed jobs" are available at the workflow level, and should not be confused with re-running a single job, which doesn't automatically re-run failed or canceled sibling jobs.)

> Thank you, that's the right call.

Thanks--I wasn't sure if the requested review was covered by https://github.com/GitoxideLabs/gitoxide/pull/1976#issuecomment-2847048347 or not. I'm glad I held off from merging.

Byron:
> The way I reran it was arguably overkill--when rerunning a whole workflow, one can select "Re-run failed jobs" instead of "Re-run all jobs", which in spite of its name, I _think_ does also rerun canceled jobs from the same matrix as a failed job that were canceled because of it.

That's a great pointer - I definitely *thought* "Re-run failed jobs" does the trick even for cancelled ones, which means I must have hit a rerun button on the level of the individual failed job.

cruessler:
@holodorum Now that RustWeek is over, I’ll get to the review in the next couple of days!

gix-blame/src/types.rs
```diff
@@ -36,43 +38,61 @@ use crate::Error;
...
+        let mut result = Self::PartialFile(vec![]);
+        for range in zero_based_ranges {
+            let _ = result.merge_range(range);
         }
         result
     }
+
+    /// Convert a 1-based inclusive range to a 0-based exclusive range.
+    fn inclusive_to_zero_based_exclusive(range: RangeInclusive<u32>) -> Range<u32> {
+        let start = range.start() - 1;
```
cruessler, line 92: What if `range.start()` is `0`? We should probably error in that case.

gix-blame/src/types.rs
```diff
@@ -81,60 +101,51 @@ impl BlameRanges {
...
     /// they will be merged into a single range.
-    pub fn add_range(&mut self, new_range: RangeInclusive<u32>) {
-        self.merge_range(new_range);
+    pub fn add_range(&mut self, new_range: RangeInclusive<u32>) -> Result<(), Error> {
+        match self {
+            Self::PartialFile(_) => {
+                let zero_based_range = Self::inclusive_to_zero_based_exclusive(new_range);
+                self.merge_range(zero_based_range)
+            }
+            _ => Err(Error::InvalidOneBasedLineRange),
```
cruessler, line 110: Instead of erroring here, I think I prefer turning `self` into `PartialFile` and adding `new_range`. I assume (and hope) that that is what most people would expect.

gix-blame/src/types.rs
```diff
@@ -81,60 +101,51 @@ impl BlameRanges {
...
+                    if new_range.start <= range.end && range.start <= new_range.end {
+                        *range = range.start.min(new_range.start)..range.end.max(new_range.end);
+                        return Ok(());
+                    }
+                    // Check if ranges are adjacent
+                    if new_range.start == range.end || range.start == new_range.end {
+                        *range = range.start.min(new_range.start)..range.end.max(new_range.end);
+                        return Ok(());
+                    }
+                }
```
cruessler, line 131: Currently, this does not prevent incorrect use in this edge case: `gix blame -L 1,3 -L 5,7 -L 2,6`. Maybe this can be turned into a `.fold` call?

gix-blame/src/types.rs
```diff
@@ -81,60 +101,51 @@ impl BlameRanges {
...
+                // If no overlap or adjacency found, add it as a new range
+                ranges.push(new_range);
+                Ok(())
             }
+            _ => Err(Error::InvalidOneBasedLineRange),
```
cruessler, line 136: Same as above: what about turning `self` into `PartialFile`?

gix-blame/src/types.rs
```diff
@@ -81,60 +101,51 @@ impl BlameRanges {
     /// The range should be 1-based inclusive.
     /// If the new range overlaps with or is adjacent to an existing range,
     /// they will be merged into a single range.
-    pub fn add_range(&mut self, new_range: RangeInclusive<u32>) {
-        self.merge_range(new_range);
+    pub fn add_range(&mut self, new_range: RangeInclusive<u32>) -> Result<(), Error> {
```
cruessler, line 104: Can we make the expectation part of the function name, like so: `add_one_based_inclusive_range`?

Same for all other methods that take any kind of `Range`.

gix-blame/src/types.rs
```diff
@@ -81,60 +101,51 @@ impl BlameRanges {
...
-        for range in &self.ranges {
-            if *range.start() == 0 {
-                return Err(Error::InvalidLineRange);
-            }
-            let start = range.start() - 1;
-            let end = *range.end();
-            if start >= max_lines || end > max_lines || start == end {
-                return Err(Error::InvalidLineRange);
+    /// Convert the ranges to a vector of `Range<u32>`.
+    pub fn to_ranges(&self, max_lines: u32) -> Vec<Range<u32>> {
```
cruessler, line 141: Just to be explicit, what do you think about calling this `to_one_based_exclusive_ranges`?

gix-blame/src/types.rs
```diff
@@ -334,6 +345,17 @@ pub struct UnblamedHunk {
 }
 
 impl UnblamedHunk {
+    /// Create a new instance
```
cruessler, line 348: The comment doesn’t seem to add a lot of context. :-)

gix-blame/src/types.rs
```diff
@@ -334,6 +345,17 @@ pub struct UnblamedHunk {
...
+    pub fn new(range: Range<u32>, suspect: ObjectId) -> Self {
```
cruessler, line 349: Since, in the past, we often got confused by the distinction between `blamed_file` and `source_file`, what about naming this `from_range_in_blamed_file`?

gix-blame/src/error.rs
```diff
@@ -28,7 +28,9 @@ pub enum Error {
     #[error(transparent)]
     DiffTree(#[from] gix_diff::tree::Error),
     #[error("Invalid line range was given, line range is expected to be a 1-based inclusive range in the format '<start>,<end>'")]
-    InvalidLineRange,
+    InvalidOneBasedLineRange,
+    #[error("Invalid line range was given, line range is expected to be a 0-based inclusive range in the format '<start>,<end>'")]
+    InvalidZeroBasedLineRange,
```
cruessler, line 33: Is this variant used anywhere? If not, can it be removed?

gix-blame/src/file/tests.rs
```diff
@@ -1338,3 +1338,100 @@ mod process_changes {
...
+    #[test]
+    fn create_from_single_range() {
+        let range = BlameRanges::from_one_based_inclusive_range(20..=40);
+        match range {
+            BlameRanges::PartialFile(ranges) => {
+                assert_eq!(ranges.len(), 1);
+                assert_eq!(ranges[0], 19..40);
+            }
+            _ => panic!("Expected PartialFile variant"),
+        }
```
cruessler, line 1354: What about `assert!(matches!(range, BlameRanges::PartialFile(…)))`? This is a bit more dense, but it avoids some of the boilerplate. (Same in the other tests.)

gix-blame/src/file/tests.rs
```diff
@@ -1338,3 +1338,100 @@ mod process_changes {
...
+        assert_eq!(ranges[0], 0..5);
+        assert_eq!(ranges[1], 9..15);
+    }
+
+    #[test]
+    fn convert_full_file_to_zero_based() {
+        let ranges = BlameRanges::WholeFile;
+        let ranges = ranges.to_ranges(100);
+        assert_eq!(ranges.len(), 1);
+        assert_eq!(ranges[0], 0..100);
```
cruessler, line 1426: What about `assert_eq!(ranges, …)`?

gix-blame/src/file/tests.rs
```diff
@@ -1338,3 +1338,100 @@ mod process_changes {
...
+    }
+
+    #[test]
+    fn default_is_full_file() {
+        let ranges = BlameRanges::default();
+        match ranges {
+            BlameRanges::WholeFile => (),
+            _ => panic!("Expected WholeFile variant"),
+        }
```
cruessler, line 1435: What about `assert!(matches!(ranges, BlameRanges::WholeFile))`?

gix-blame/src/file/tests.rs
```diff
@@ -1338,3 +1338,100 @@ mod process_changes {
         );
     }
 }
+
+mod blame_ranges {
```
cruessler, line 1342: Can we add tests for the following cases:

- `vec![0..=x]`
- `vec![1..=3, 5..=7 ,2..=6]` (this is the one also mentioned in another comment)

cruessler (changes requested):
I finally got to the review! I think the PR is generally solid, but needs a bit of polish before it is ready to go. Let me know if you’ve got any questions!

gix-blame/src/file/tests.rs
holodorum: I've updated it using an `if let` looks a lot cleaner indeed.

gix-blame/src/types.rs
cruessler: My idea was to use `new_range` as the only range for `PartialFile`. My thinking was that turning `self` into `PartialFile` would make the API more ergonomic and easy to use as a user would not have to know anything about the internal state of `BlameRanges` to be able to call `add_one_based_inclusive_range`. In my view, this would mirror how `git blame` vs. `git blame -L 1,5 -L 7,9` works. But I also think that it’s acceptable to be more explicit even if that makes the API harder to use.

cruessler: (What I have in mind, seems similar to the logic in `from_one_based_inclusive_ranges`.)

gix-blame/src/types.rs
```diff
@@ -82,16 +82,19 @@ impl BlameRanges {
             .collect::<Vec<_>>();
         let mut result = Self::PartialFile(vec![]);
         for range in zero_based_ranges {
-            let _ = result.merge_range(range);
+            let _ = result.merge_zero_based_exclusive_range(range?);
```
cruessler, line 85: The fact that the possibility of an error is explicitly ignored here, also seems to nudge toward making `merge_zero_based_exclusive_range` never fail.

cruessler (changes requested):
Thanks for your response, and sorry that mine is so delayed!

What would be the benefit of introducing `OneBasedInclusiveRange` and `ZeroBasedExclusiveRange` as a type and where would it be used: inside of `BlameRanges` only or also as part of the public API? I see the risk of making the API more heavy than it needs to be. In my opinion, the current method names already sufficiently convey the constraints for all the publicy exposed methods.

Once my last concern has been addressed, I’d also approve this PR, so that @Byron can have a look! I plan on prioritizing this PR now in order to get it merged as quickly as possible.

cruessler:
@holodorum I know it’s been some time (and it’s totally okay if you’ve moved on), but do you want to keep working on this PR? Otherwise I’m thinking of taking over and finishing myself. Let me know what your preferences are and if there’s anything I can do to help!

holodorum:
@cruessler got a bit distracted indeed, my bad. Will update the PR with the changes this week. Main thing outstanding is not erroring, but turning `WholeFile` into a `PartialFile` as discussed in [this comment](https://github.com/GitoxideLabs/gitoxide/pull/1976#discussion_r2104791914) right?

cruessler:
@holodorum Sounds good! If you want, you can also have a look at https://github.com/GitoxideLabs/gitoxide/compare/main...cruessler:gitoxide:improve-blame-ranges for inspiration where I already incorporated most of my feedback into your changes. I also split the commits so that each commit only contains changes related to a single crate (something that `gitoxide` does). There’s a few more edge cases that `BlameRanges` should handle, some of which are already part of my branch. What’s missing is code for handling the following situation:

```
#[test]
fn to_zero_based_exclusive_ranges_doesnt_exceed_max_lines() {
    let mut ranges = BlameRanges::from_one_based_inclusive_range(1..=5).unwrap();
    ranges.add_one_based_inclusive_range(6..=10).unwrap();

    assert_eq!(ranges.to_zero_based_exclusive_ranges(7), vec![0..5, 6..7]);
}
```

holodorum:
Ahh okay, @cruessler if you already did most of the work I'd think it makes most sense if you finish it.

cruessler:
@holodorum This is the follow-up PR: https://github.com/GitoxideLabs/gitoxide/pull/2204.

Byron:
Thanks everyone!

Closing this PR as it was superseded.

PR #1996, opened 2025-05-06
Regroup CI doctests and adjust `unit-tests` recipe

EliahKagan:
This distributes work across CI jobs more evenly. In particular, it speeds up `test-fast` on `windows-latest` (which had become the slowest of all CI jobs, including other Windows jobs). To do so, it leverages the idea that the doctests are important enough that we should probably continue automatically running them on Windows, but not so important that it particularly matters which of reasonably normal Windows environments and targets they run for.

Those changes entail modifying the `justfile`, so I've included the further overlapping change of adding `--no-fail-fast` to the commands in the `unit-tests` recipe there. This benefits CI, and I think does not hurt local runs either.

This also makes it so that the doctest-running command in `just unit-tests` runs more than zero doctests, which I think is the more intuitive effect. (That is, it runs all doctests in the workspace, rather than just all the doctests in the `gitoxide` crate, of which there currently aren't any.)

Substantial further details, including about what this is trying to improve, and rationale, are in the commit message.

The diff, as shown on GitHub, is somewhat confusing in `justfile`. Except for one line that is moved and changed, all changed lines are actually changed only by having `--no-fail-fast` added to the end.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -181,8 +181,6 @@ jobs:
         env:
           GIX_TEST_CREATE_ARCHIVES_EVEN_ON_CI: '1'
         run: cargo nextest run --workspace --no-fail-fast
-      - name: Doctest
-        run: cargo test --workspace --doc --no-fail-fast
       - name: Check that tracked archives are up to date
         run: git diff --exit-code  # If this fails, the fix is usually to commit a regenerated archive.
 
@@ -277,7 +275,7 @@ jobs:
           GIX_TEST_IGNORE_ARCHIVES: '1'
         run: cargo nextest run --workspace --no-fail-fast
 
-  test-32bit-windows-size:
+  test-32bit-windows-size-doc:
     runs-on: windows-latest
 
     env:
@@ -292,8 +290,10 @@ jobs:
       - uses: taiki-e/install-action@v2
         with:
           tool: nextest
-      - name: Test (nextest)
+      - name: Test data structure sizes (nextest)
         run: cargo nextest run --target $env:TARGET --workspace --no-fail-fast size
+      - name: Doctest
+        run: cargo test --workspace --doc --no-fail-fast
 
   lint:
     runs-on: ubuntu-latest
@@ -497,7 +497,7 @@ jobs:
       - test-fast
       - test-fixtures-windows
       - test-32bit
-      - test-32bit-windows-size
+      - test-32bit-windows-size-doc
       - lint
       - cargo-deny
       - check-packetline
--- a/justfile
+++ b/justfile
@@ -144,41 +144,41 @@ doc $RUSTDOCFLAGS='-D warnings':
 
 # Run all unit tests
 unit-tests:
-    cargo nextest run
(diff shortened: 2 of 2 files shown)


EliahKagan:
It's hard to predict the effect of caching dependencies on the performance of future runs before merging. I think this makes workflow runs overall faster, since it did in multiple runs in my fork, but we'll find out. Further adjustments might be in order.

Byron:
Thanks so much!

The CI run took 17 minutes, which is less than the 30 minutes I thought it would take by now. Overall wallclock time is 1h27m, which seems close to what I remember though.

EliahKagan:
> Thanks so much!

No problem!

> The CI run took 17 minutes

Hopefully it will be even less in most runs. I think most runs took about that long in from the start of the first job to the end of the last job, before the changes here.

> which is less than the 30 minutes I thought it would take by now.

Even though Windows is currently the slowest for running the tests, and the overall trend is to do more things on Windows, as well as to have more test cases and more fixtures across all platforms... nonetheless I think the overall trend of total time from the first job starting to the fast job ending should be a gradual decrease, because the runner machines get gradually faster, and because we try not to pack too much into an individual job. For an example of the latter that is unrelated to Windows, the `test` job was split into `test` and `test-journey` in 711c2f5 (#1674). So if it looks like the time is approaching 30 minutes, I would consider that to be a bug, unless there is a strong reason for it. (Please feel free to ping me anytime about performance degradations on CI, as well as anything else you think may be of interest.)

> Overall wallclock time is 1h27m, which seems close to what I remember though.

I think by this you mean the sum of the wall-clock durations of all the jobs. If so, it makes sense to me that this has not much improved, at least not by anything in this PR. This PR redistributes the work across jobs more evenly. But it doesn't decrease the work much. It does stop running doctests on macOS in `test-fast`. But it starts running all of them on Ubuntu in `test`. Also, moving the doctest run for Windows from the `test-fast` job to the 32-bit Windows job might actually make things take longer in total, depending on various small factors.

I think it may make sense to try to decrease this total time--especially since the likely addition of new jobs, such as Windows jobs along the "various other Windows jobs" examples given in ee55ded, may increase it if their contribution is not in some way offset by speed improvements elsewhere--but that's not part of what this PR specifically attempted. (That is also something I think should be considered together with other resources. I think the total combined running time may be less important than total energy usage; queuing delays and runner availability; GitHub Actions cache storage; workflow maintainability; and the ability to adapt the workflow to changes such as new runners or, in case it should ever be needed, different CI solutions.)

Byron:
> (Please feel free to ping me anytime about performance degradations on CI, as well as anything else you think may be of interest.)

Thank you, appreciated!

> I think by this you mean the sum of the wall-clock durations of all the jobs, i.e. the time they would take if they were not run concurrently, which is longer than the time they take when measured on a single clock.

Yes, this is what I meant - I should have written total runtime. And even though seeing that decrease is good for the environment, it's nothing we should necessarily optimise for as wallclock time is the thing that should be reduced (just not at all cost). And it's true that the total runtime doesn't directly correlate to energy usage either.

PR #1997, opened 2025-05-06
Bring back `#[momo]`

EliahKagan:
This experiment relates to #1576 and is unfinished.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1396,6 +1396,7 @@ dependencies = [
  "gix-ignore",
  "gix-index",
  "gix-lock",
+ "gix-macros",
  "gix-mailmap",
  "gix-merge",
  "gix-negotiate",
--- a/gix/Cargo.toml
+++ b/gix/Cargo.toml
@@ -304,6 +304,7 @@ cache-efficiency-debug = ["gix-features/cache-efficiency-debug"]
 
 
 [dependencies]
+gix-macros = { version = "^0.1.5", path = "../gix-macros" }
 gix-utils = { version = "^0.3.0", path = "../gix-utils" }
 gix-fs = { version = "^0.15.0", path = "../gix-fs" }
 gix-ref = { version = "^0.52.1", path = "../gix-ref" }
--- a/gix/src/config/snapshot/access.rs
+++ b/gix/src/config/snapshot/access.rs
@@ -2,6 +2,7 @@
 use std::{borrow::Cow, ffi::OsStr};
 
 use gix_features::threading::OwnShared;
+use gix_macros::momo;
 
 use crate::{
     bstr::{BStr, BString, ByteSlice},
@@ -25,6 +26,7 @@ impl<'repo> Snapshot<'repo> {
     }
 
     /// Like [`boolean()`][Self::boolean()], but it will report an error if the value couldn't be interpreted as boolean.
+    #[momo]
     pub fn try_boolean(&self, key: impl gix_config::AsKey) -> Option<Result<bool, gix_config::value::Error>> {
         self.repo.config.resolved.boolean(key)
     }
@@ -40,6 +42,7 @@ impl<'repo> Snapshot<'repo> {
     }
 
     /// Like [`integer()`][Self::integer()], but it will report an error if the value couldn't be interpreted as boolean.
+    #[momo]
     pub fn try_integer(&self, key: impl gix_config::AsKey) -> Option<Result<i64, gix_config::value::Error>> {
         self.repo.config.resolved.integer(key)
     }
@@ -47,6 +50,7 @@ impl<'repo> Snapshot<'repo> {
     /// Return the string at `key`, or `None` if there is no such value.
     ///
     /// Note that this method takes the most recent value at `key` even if it is from a file with reduced trust.
+    #[momo]
     pub fn string(&self, key: impl gix_config::AsKey) -> Option<Cow<'repo, BStr>> {
         self.repo.config.resolved.string(key)
     }
@@ -54,6 +58,7 @@ impl<'repo> Snapshot<'repo> {
     /// Return the trusted and fully interpolated path at `key`, or `None` if there is no such value
     /// or if no value was found in a trusted file.
     /// An error occurs if the path could not be interpolated to its final value.
+    #[momo]
     pub fn trusted_path(
         &self,
         key: impl gix_config::AsKey,
@@ -63,6 +68,7 @@ impl<'repo> Snapshot<'repo> {
 
     /// Return the trusted string at `key` for launching using [command::prepare()](gix_command::prepare()),
     /// or `None` if there is no such value or if no value was found in a trusted file.
+    #[momo]
     pub fn trusted_program(&self, key: impl gix_config::AsKey) -> Option<Cow<'repo, OsStr>> {
         let value = self
             .repo
@@ -112,6 +118,7 @@ impl<'repo> SnapshotMut<'repo> {
 
     /// Set the value at `key` to `new_value`, possibly creating the section if it doesn't exist yet, or overriding the most recent existing
     /// value, which will be returned.
+    #[momo]
     pub fn set_value<'b>(
         &mut self,
         key: &'static dyn crate::config::tree::Key,
@@ -135,6 +142,7 @@ impl<'repo> SnapshotMut<'repo> {
 
     /// Set the value at `key` to `new_value` in the given `subsection`, possibly creating the section and sub-section if it doesn't exist yet,
     /// or overriding the most recent existing value, which will be returned.
+    #[momo]
     pub fn set_subsection_value<'a, 'b>(
         &mut self,
         key: &'static dyn crate::config::tree::Key,
--- a/gix/src/config/snapshot/access.rs
+++ b/gix/src/config/snapshot/access.rs
@@ -26,7 +26,6 @@ impl<'repo> Snapshot<'repo> {
     }
 
     /// Like [`boolean()`][Self::boolean()], but it will report an error if the value couldn't be interpreted as boolean.
-    #[momo]
     pub fn try_boolean(&self, key: impl gix_config::AsKey) -> Option<Result<bool, gix_config::value::Error>> {
         self.repo.config.resolved.boolean(key)
     }
@@ -42,7 +41,6 @@ impl<'repo> Snapshot<'repo> {
     }
 
     /// Like [`integer()`][Self::integer()], but it will report an error if the value couldn't be interpreted as boolean.
-    #[momo]
     pub fn try_integer(&self, key: impl gix_config::AsKey) -> Option<Result<i64, gix_config::value::Error>> {
         self.repo.config.resolved.integer(key)
     }
@@ -50,7 +48,6 @@ impl<'repo> Snapshot<'repo> {
     /// Return the string at `key`, or `None` if there is no such value.
     ///
     /// Note that this method takes the most recent value at `key` even if it is from a file with reduced trust.
-    #[momo]
     pub fn string(&self, key: impl gix_config::AsKey) -> Option<Cow<'repo, BStr>> {
         self.repo.config.resolved.string(key)
     }
@@ -58,7 +55,6 @@ impl<'repo> Snapshot<'repo> {
     /// Return the trusted and fully interpolated path at `key`, or `None` if there is no such value
     /// or if no value was found in a trusted file.
     /// An error occurs if the path could not be interpolated to its final value.
-    #[momo]
     pub fn trusted_path(
         &self,
         key: impl gix_config::AsKey,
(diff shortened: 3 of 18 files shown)


Byron:
I may also add that it didn't shave off more than a marginal number of kB off the final binaries, so I felt that it wasn't really worth it. `prodash` was also affected, which has more complex, but maybe also `dyn` friendly traits now.

Lastly, working with `dyn` closures feels more cumbersome overall as callers have to take a reference on them.

If binaries would have been half-size and/or compiled much faster, I think it would be worth the effort, but in reality that didn't pan out like I would have hoped.

EliahKagan:
Both the latest stable Rust toolchain and code in `gix` have changed somewhat since `#[momo]` was last used. Should this be measured again? I am not sure how best to measure the effect.

If the effect remains small, or can be extrapolated as probably small, then maybe bringing back the use of `#[momo]` in `gix` would not be worthwhile even if IDE/editor code analysis and code completion were unaffected. In that case, no further experimentation is required at this time (and this could be closed). It might still be interesting to revisit it again sometime in the future to see if the benefit is greater.

Otherwise, the following may be of interest.

Based on the very limited experiment so far of just reverting the `#[momo]` attribute removals--in which I initially resolved conflicts by allowing `#[momo]` to come back even on functions whose signatures changed substantially--it seems to me that my suggestion in https://github.com/GitoxideLabs/gitoxide/discussions/1576#discussioncomment-10566627 to try to get editors/IDEs to treat `#[momo]` as if it were a no-op is not the right approach. Sometimes `#[momo]` cannot be applied to a function. When that happens, there is a useful and descriptive compile error. It is useful for IDEs to show that error.

On the other hand, it looks to me like the effect in VS Code and RustRover may be less severe than in the past. I'm able to see function signatures on hover, and auto-completion seems to be working fine, including in RustRover. It may be that there are problems I haven't observed, but so far the only problem I know of is when the signature is declared with pattern matching (argument destrucutring), such as in:

https://github.com/GitoxideLabs/gitoxide/blob/035af58ba5a5bcc2f398c1b0e66a70dc951bfe14/gix/src/create.rs#L128-L136

When hovering over the name of this function, the signature shown in RustRover would *ideally* be this, which is what is shown when `#[momo]` is omitted:

```rust
gix::create
pub fn into(directory: impl Into<PathBuf>, kind: Kind, Options {
        fs_capabilities,
        destination_must_be_empty,
    }: Options) -> Result<gix_discover::repository::Path, Error>
```

But when `#[momo]` is present, the signature is shown like this:

```rust
gix::create
pub fn into(directory: impl Into<PathBuf>, kind: Kind, arg_2_gen_by_momo_: Options) -> Result<gix_discover::repository::Path, Error>
```

That is not wrong, but it is less useful. However, the type is still correct, and code analysis and auto-completion seem to work fine for calls to the function, such as in:

https://github.com/GitoxideLabs/gitoxide/blob/035af58ba5a5bcc2f398c1b0e66a70dc951bfe14/gix/src/init.rs#L68

Here's a screenshot showing what RustRover shows on hover there:

Screenshot of RustRover, overing the mouse pointer over crate::create::into

Although this is not ideal, it seems to me that the impact of this problem would most often be small. However, it may be that the problems you experienced would occur the same today. Do you recall what had happened?

(An intermediate approach could be to add a `gix` feature and use `#[momo]` conditionally only when that feature is enabled. But this is only worthwhile if we should seriously consider using `#[momo]`.)

Byron:
I actually realize now that I was mis-remembering or conflating something - `#[momo]` allows to actually use `impl Fn(x) -> y` while changing it to `&dyn Fn(x) -> y` under the hood. This yields the best of both worlds.

> Although this is not ideal, it seems to me that the impact of this problem would most often be small. However, it may be that the problems you experienced would occur the same today. Do you recall what had happened?

I think function with `#[momo]` on it wouldn't autocomplete, i.e. `repo.momomethod()` wouldn't autocomplete when typing `repo.momom`. Looking at how it obviously runs the plugin to see the actual code, I'd think that by now autocompletion should work, but you would certainly be able to quickly validate that.

It could probably be taught to re-use the original parameter names so I'd not consider that an issue for now.

The issue may be that despite `gix` bing an avid user of a large part of the API-surface, as an application it didn't accrue enough monomorphisation costs to make more than a marginal difference in binary size. It's just that the average application doesn't seem to trigger too much duplication in code (*It's something you would be able to test by comparing release builds, I think*).
And those who do would probably debug this and have their own landing-pads and manually de-monomophise.

To me it turned out to be a burden with dubious value, so I'd still be hesitant to bring it back. And that's despite me liking the name, and liking the tool itself, so it didn't make me happy to remove it. I'd definitely have preferred if it could live on in a project where it makes more of a difference, or maybe even a general purpose crate that can develop on its own. A lot of time went in there and it does what it should as well, so ideally it can find applications where it make a difference.

EliahKagan:
> I think function with `#[momo]` on it wouldn't autocomplete, i.e. `repo.momomethod()` wouldn't autocomplete when typing `repo.momom`. Looking at how it obviously runs the plugin to see the actual code, I'd think that by now autocompletion should work, but you would certainly be able to quickly validate that.

I had also tested autocompletion, which did work. When I tested it, it was within the same crate. I just checked that autocompletion also works across crates. This works both to list the function and to autocomplete to it when one begins to type its name.

Screenshot of testing code completion of gix::create::into when referenced in the separate gitoxide-core crate


> It could probably be taught to re-use the original parameter names so I'd not consider that an issue for now.

It does use original names when they name function parameters. But when the function declares no parameter name but instead destructures, so the names are variable names for the destructuring, then it synthesizes an opaque function parameter name for that. The type is correct, so the editor experience at the site of use is not too bad.

Ideally it would work through the match/destructuring as well. I don't know how to modify it to do that, but this might be due to my weak understanding of procedural Rust macros; perhaps this is feasible to do.

This might come down to exactly what one means by parameter, but to clarify:

- Parameter names are preserved and presented in the IDE when matching/destructuring is not used in the parameter declaration.
- Neither parameter names *nor which parts of a `struct` are extracted* are presented in the IDE when matching/destructuring is used in the parameter declaration.

> The issue may be that despite `gix` bing an avid user of a large part of the API-surface, as an application it didn't accrue enough monomorphisation costs to make more than a marginal difference in binary size. It's just that the average application doesn't seem to trigger too much duplication in code (_It's something you would be able to test by comparing release builds, I think_).

I've checked `gix` and `ein` on GNU/Linux, macOS, and Windows. There is usually a small decrease in size, but occasionally the size increases. The most relevant size differences are when debug symbols are stripped (except on Windows, where the symbols are placed in separate files originally), but I've included the numbers both with and without symbols. Builds were done with `cargo build -r` on the `main` and `run-ci/momo` branches.

GNU/Linux:

```text
Arch Linux - main:

ek in 🌐 catenary in ~/tmp/main
❯ ls -l
total 39136
-rwxr-xr-x 1 ek ek 12978632 May  6 17:10 ein
-rwxr-xr-x 1 ek ek 27091928 May  6 17:10 gix
ek in 🌐 catenary in ~/tmp/main
❯ strip ein
ek in 🌐 catenary in ~/tmp/main
❯ strip gix
ek in 🌐 catenary in ~/tmp/main
❯ ls -l
total 32416
-rwxr-xr-x 1 ek ek 10899184 May  6 17:10 ein
-rwxr-xr-x 1 ek ek 22294448 May  6 17:10 gix
```

```text
Arch Linux - run-ci/momo:

ek in 🌐 catenary in ~/tmp/momo
❯ ls -l
total 39064
-rwxr-xr-x 1 ek ek 12979232 May  6 17:18 ein
-rwxr-xr-x 1 ek ek 27017656 May  6 17:18 gix
ek in 🌐 catenary in ~/tmp/momo
❯ strip ein
ek in 🌐 catenary in ~/tmp/momo
❯ strip gix
ek in 🌐 catenary in ~/tmp/momo
❯ ls -l
total 32360
-rwxr-xr-x 1 ek ek 10907376 May  6 17:18 ein
-rwxr-xr-x 1 ek ek 22228912 May  6 17:18 gix
```

macOS:

```text
macOS - main:

~/t/main> ls -l
total 67088
-rwxr-xr-x  1 ek  staff  11208672 May  6 17:07 ein*
-rwxr-xr-x  1 ek  staff  23138304 May  6 17:07 gix*
~/t/main> strip ein
~/t/main> strip gix
~/t/main> ls -l
total 56248
-rwxr-xr-x  1 ek  staff   9556576 May  6 17:07 ein*
-rwxr-xr-x  1 ek  staff  19238016 May  6 17:08 gix*
```

```text
macOS - run-ci/momo:

~/t/momo> ls -l
total 66920
-rwxr-xr-x  1 ek  staff  11189392 May  6 17:15 ein*
-rwxr-xr-x  1 ek  staff  23072064 May  6 17:15 gix*
~/t/momo> strip ein
~/t/momo> strip gix
~/t/momo> ls -l
total 56120
-rwxr-xr-x  1 ek  staff   9539904 May  6 17:15 ein*
-rwxr-xr-x  1 ek  staff  19188400 May  6 17:15 gix*
```

Windows:

```text
Windows - main:

-a---            5/6/2025  5:00 PM       11184640 ein.exe
-a---            5/6/2025  5:00 PM        8835072 ein.pdb
-a---            5/6/2025  5:01 PM       23184384 gix.exe
-a---            5/6/2025  5:01 PM       12890112 gix.pdb
```

```text
Windows - run-ci/momo:

-a---            5/6/2025  5:06 PM       11206144 ein.exe
-a---            5/6/2025  5:06 PM        8843264 ein.pdb
-a---            5/6/2025  5:07 PM       23182336 gix.exe
-a---            5/6/2025  5:07 PM       12873728 gix.pdb
```

> And those who do would probably debug this and have their own landing-pads and manually de-monomophise.
> 
> To me it turned out to be a burden with dubious value, so I'd still be hesitant to bring it back. And that's despite me liking the name, and liking the tool itself, so it didn't make me happy to remove it. I'd definitely have preferred if it could live on in a project where it makes more of a difference, or maybe even a general purpose crate that can develop on its own. A lot of time went in there and it does what it should as well, so ideally it can find applications where it make a difference.

If we are not planning to start using it again in `gix` or other gitoxide crates, couldn' it be moved from `gix-macros` to its own separate crate and published? (If that were done, then the implementation in `gix-macros` could be marked as deprecated with a reference to the new separate non-gitoxide crate, and in a subsequent release removed.)

cc @NobodyXu

Byron:
Thanks for the exhaustive testing and for gathering data on all platforms. There maybe another factor for explaining this still marginal impact: (too) incomplete usage.
However, looking at these numbers I doubt that doubling down on Momo will have yields significant enough.

Thus I think we should not bring `momo` back, but I as well would love to see it preserved. Marking `gix-macros` as deprecated might be too much as it would still be the place to contain proc-macros (there are a couple that would make sense to write), but giving `momo`  a life on its own would be good.

When looking at crates.io, I did find that [`momo`](https://crates.io/crates/momo) already exists, and it seems to have the same purpose. So it appears that we could remove our version of it as well, unless there are differences that are worth preserving in its own crate. A prerequisite for this is probably @NobodyXu as original author having interest in doing so.

NobodyXu:
> When looking at crates.io, I did find that [momo](https://crates.io/crates/momo) already exists, and it seems to have the same purpose.

IIRC momo hasn't been updated for a while and did not upgrade to syn v2, which was why I copied and implemented it in gix-macros

putting it in a separate crate sounds reasonable, though a more ideal solution would be that rustc adopts this itself to deduplicate code, instead of pushing the responsibility to the crate maintainer, which often doesn't justify it, plus it introduces proc-macro that could slow down the build and introduce attack vectors.

I also heard about talking of mir-only compilation and the new linker wild, that takes the mir, merge them, perform LTO on mir level to eliminate code and deduplicate, before actually passing down to LLVM.

It could improve compile time as less code is passed down to LLVM and reduces binary size.

Byron:
Thanks for sharing! This would indeed sound like it could be useful in a separate crate, a crate that you certainly would have to be the owner of as main (only?) author of the crate. Maybe llogiq could also be enticed to accept the updated `momo` in a PR.

EliahKagan:
> Thus I think we should not bring `momo` back, but I as well would love to see it preserved.

I'm happy to close this PR. I've kept it open so far in order to have something open that tracks the future of `momo`. This could be closed and an issue opened, though.

> Marking `gix-macros` as deprecated might be too much as it would still be the place to contain proc-macros (there are a couple that would make sense to write)

Couldn't `momo` itself be marked as deprecated without marking `gix-macros` as a whole deprecated?

But I think that should only be done if and when `momo` has become available elsewhere in non-deprecated form for people to use it in their own projects, either as its own new crate or as an update to the old other `momo` available on crates.io.

NobodyXu:
> But I think that should only be done if and when momo has become available elsewhere in non-deprecated form for people to use it in their own projects, either as its own new crate or as an update to the old other momo available on crates.io. (I would guess that one of these could probably happen earlier than rustc gaining functionality that makes it obsolete.)

about that, I was just thinking that would it be possible to implement it with `macro_rules!`, since building proc-macro is costly

i think it'd be hard, so maybe simply putting it in a non-deprecated crate is the best we can do now.

Byron:
> I'm happy to close this PR. I've kept it open so far in order to have something open that tracks the future of `momo`. This could be closed and an issue opened, though.

I think that would be good as well, thank you.

EliahKagan:
Thanks for bearing with the delay. I've opened #2010 for this. I hope my summary is reasonable; if not, it should be revised.

That issue is to allow this to be closed while still keeping track of the future of `#[momo]`. Discussion could move there but does not have to; I hope everyone does whatever seems most convenient.

PR #2003, opened 2025-05-09
Improve MSRV check and fix `aria-label` in MSRV badge

EliahKagan:
This makes several improvements to the MSRV check. Most are to make clearer what is going on, in the `msrv.yml` workflow and elsewhere, by modifying and expanding comments, and by slightly reorganizing the commands. But this also improves the robustness of the check, by:

- Fixing a bug where jobs could report the wrong status, which was analogous to [#1559](https://github.com/GitoxideLabs/gitoxide/pull/1559) but much less likely to occur. If both the `rustup` commands were to fail, but only on Windows, then the third command that was in that same step with them could cause the whole step to fail--since `pwsh` is used by default on Windows and it doesn't have anything analogous to `-e`--whereupon the usually more recent toolchain preinstalled in the `windows-2022` runner would be tested with, instead of the MSRV. I don't think this has actually happened in any runs so far, but I think it's worth preventing entirely.
- Passing `--locked` to the `cargo` commands in the `ci-check-msrv` recipe, so they fail if they would have to further modify `Cargo.lock`, rather than modifying it and continuing. This is because the effects of `-Zminimal-versions` are non-obvious and subject to change (`--minimal-versions` has not been stabilized), and also to try to account at least partially for possible disagreement in what the minimal dependency versions are between the `nightly` toolchain that is used to downgrade the locked dependencies (because only `nightly` has `-Zminimal-versions`) and the old stable toolchain versioned at the MSRV.
- Using `cargo build` instead of `cargo check` in the `ci-check-msrv` recipe, as suggested in [#1808 (comment)](https://github.com/GitoxideLabs/gitoxide/issues/1808#issuecomment-2614260268). (Please note, however, that [#1808](https://github.com/GitoxideLabs/gitoxide/issues/1808) is still probably not completely fixed. My guess is that the build problems noted there occurred, and could still occur, in some features, or some crates, that are not covered by the commands in the recipe.)

This also fixes an accessibility bug in the MSRV badge, where the [`aria-label`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label) attribute of the top-level `svg` element had the value `rustc: 1.70.0+`, instead of the correct and intended value `rustc: 1.75.0+` that was already present in the other places in the SVG code where the version appears (dc1d271).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,8 +126,8 @@ bench-gix-config:
 
 check-msrv-on-ci: ## Check the minimal support rust version for currently installed Rust version
 	rustc --version
-	cargo check --locked --package gix
-	cargo check --locked --package gix --no-default-features --features async-network-client,max-performance
+	cargo build --locked --package gix
+	cargo build --locked --package gix --no-default-features --features async-network-client,max-performance
 
 ##@ Maintenance
 
--- a/Makefile
+++ b/Makefile
@@ -126,8 +126,8 @@ bench-gix-config:
 
 check-msrv-on-ci: ## Check the minimal support rust version for currently installed Rust version
 	rustc --version
-	cargo check --package gix
-	cargo check --package gix --no-default-features --features async-network-client,max-performance
+	cargo check --locked --package gix
+	cargo check --locked --package gix --no-default-features --features async-network-client,max-performance
 
 ##@ Maintenance
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -433,7 +433,8 @@ jobs:
 
     defaults:
       run:
-        shell: bash  # Use bash even on Windows, if we ever reenable windows-latest for testing.
+        # Use `bash` even on Windows, if we ever reenable `windows-latest` for testing.
+        shell: bash
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -33,20 +33,16 @@ jobs:
       # IMPORTANT: adjust etc/msrv-badge.svg as well
       RUST_VERSION: 1.75.0
 
-    defaults:
-      run:
-        # Use `bash` even in the Windows job, so any failing command fails its step (due to `-e`).
-        shell: bash
-
     steps:
       - uses: actions/checkout@v4
       - uses: extractions/setup-just@v3
       - name: Set up ${{ env.RUST_VERSION }} (MSRV) and nightly toolchains
-        run: |
-          rustup toolchain install ${{ env.RUST_VERSION }} nightly --profile minimal --no-self-update
-          rustup default ${{ env.RUST_VERSION }}
+        run: rustup toolchain install ${{ env.RUST_VERSION }} nightly --profile minimal --no-self-update
+      - name: Set ${{ env.RUST_VERSION }} (MSRV) as default
+        run: rustup default ${{ env.RUST_VERSION }}
       - name: Downgrade locked dependencies to lowest allowed versions
         run: |
-          cargo +nightly update -Zminimal-versions  # TODO(msrv): Use non-`-Z` way when available.
+          # TODO(msrv): Use `cargo update --minimal-versions` when `--minimal-versions` is available.
+          cargo +nightly update -Zminimal-versions
       - name: Run some `cargo check` commands on `gix`
         run: just ci-check-msrv
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -33,6 +33,11 @@ jobs:
       # IMPORTANT: adjust etc/msrv-badge.svg as well
       rust_version: 1.75.0
 
+    defaults:
+      run:
+        # Use `bash` even in the Windows job, so any failing command fails its step (due to `-e`).
+        shell: bash
+
     steps:
       - uses: actions/checkout@v4
       - uses: extractions/setup-just@v3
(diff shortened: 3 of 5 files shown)


Makefile
```diff
@@ -126,8 +126,8 @@ bench-gix-config:
 
 check-msrv-on-ci: ## Check the minimal support rust version for currently installed Rust version
 	rustc --version
-	cargo check --package gix
-	cargo check --package gix --no-default-features --features async-network-client,max-performance
+	cargo build --locked --package gix
+	cargo build --locked --package gix --no-default-features --features async-network-client,max-performance
```
EliahKagan, line 130: I don't know why we have a `check-msrv-on-ci` rule in the `Makefile` as well as an equivalent `ci-check-msrv` recipe in the `justfile` (that the `msrv.yml` workflow uses). I've only expanded the comment in the `justfile`, but I've updated the command in the `Makefile` as well so that their behavior is equivalent as long as the functionality is duplicated across both.

Byron: Thanks for bringing this up. It also seems like this target isn't used at all, so I think it's better to remove it.

Byron (commented):
Thanks a lot!

Makefile
EliahKagan: Thanks. I'll remove it from `Makefile` in a subsequent PR.

More broadly, I'm not clear on what is supposed to go in the `Makefile` and what is supposed to go in the `justfile`. Is the `Makefile` just left over from before the `justfile` was introduced, such that everything there is there is for historical reasons only? Should its contents be migtated to the `justfile`? Or are there reason to prefer that some things be in the `Makefile` instead of the `justfile`?

Byron: It's a left-over, and the port to `justfile` was never completed.
`make stress` is still called, and it could be that it leverages the natural parallelism of `make` which is harder to do in a `justfile`.

EliahKagan: Ah, I see. I'll remember to consider the effect on parallelism if and when moving things, especially if they are not already duplicated, from the `Makefile` to the `justfile`. The `check-msrv-on-ci` rule/target probably doesn't benefit from that, though.

EliahKagan: Or maybe https://github.com/casey/just/issues/626 will be completed--for example, maybe https://github.com/casey/just/pull/1562 will be picked back up--and then they could all be moved over to the `justfile`.

EliahKagan: I just noticed that https://github.com/casey/just/issues/626 has very recently been implemented in https://github.com/casey/just/pull/2803 in the form of a new `[parallel]` attribute. My understanding is that, if a recipe $Y$ declares recipes $X_1$, $X_2$, and $X_3$ as dependencies and $Y$ is marked `[parallel]`, then running $Y$ causes `just` to run $X_1$, $X_2$, and $X_3$ in parallel rather than sequentially.

This is not equivalent to `-j` for `make`. However, at least for `make stress`, it seems to me that `[parallel]` does exactly what we want, and that we are currently using `make -j` as a workaround for not having such a feature. The [`stress`](https://github.com/GitoxideLabs/gitoxide/blob/c7af04db9b6bb1204e0f4c436d1db8f48a491e86/Makefile#L97-L115) recipe runs various commands with `time`, but the first command it runs is only to set things up for the subsequent commands, and it does not use `time`, nor any other shell features.

https://github.com/GitoxideLabs/gitoxide/blob/c7af04db9b6bb1204e0f4c436d1db8f48a491e86/Makefile#L98

That first command runs three other recipes with `$(MAKE) -j3` so they run in parallel, and the recipe declares no dependencies. Therefore, conceptually it has the three recipes identified in the first command as dependencies, but it runs them through a command with `$(MAKE)` instead, to be able to pass `-j3`.

Furthermore, as far as I can tell, while `stress` runs various recipes directly and indirectly, this is the only place where `-j` parallelism is involved. Even if one runs `make -jN stress` with one's chosen `N`, I don't think one gets greater parallelism, nor do I think greater parallelism is intended when running the stress test.

Per https://github.com/casey/just/issues/626#issuecomment-3046497841, `[parallel]` will be available starting in the next release of `just`. It seems to me that whether this will let us move everything from the `Makefile` into the `justfile` depends on how accurate the note at the top of the [`make help`](https://github.com/GitoxideLabs/gitoxide/blob/c7af04db9b6bb1204e0f4c436d1db8f48a491e86/Makefile#L2) text is:

> Note: Make is only for specific functionality used by CI - run `just` for developer targets

Assuming that is (still) true, it seems to me that `[parallel]` will let us move everything into the `justfile` (or, if preferred, a separate `justfile`, though I don't think that would be preferred). This is because, at least going by what I see when I inspect the output of `git grep -Fwn make`, the only use of `make` on CI--looking only at uses that that involve our `Makefile`--is to run `make stress` in `cron.yml`.

I believe there are other indirect implicit uses of `make` in building native code dependencies, as occurs when some features of some crates are enabled, and for this reason we do install `make` in containers that may not have it, in some other CI jobs. But these would not have anything to do with the `Makefile` in this project.

However, if `make -jN` is also being used manually for local operations, possibly explicitly specifying multiple recipes on the command line, then that is a different story. There may also be other relevant subtleties.

Although I don't know that we should do it regularly, I think it would be useful to be able to run the stress test on Windows. I expect that eliminating the `Makefile` would ease portability, since [various interesting things](https://github.com/gitpython-developers/GitPython/commit/729372f6f87639f0c4d8211ee7d173100117a257) can happen with `Makefile`s on Windows. But it may not be necessary.

(As an aside, `torvalds/linux` is one of the test repositories used there, and just cloning that on Windows is very slow, even when it is a bare clone. However, since the test uses a bare clone, it should at least be able to run, since the usual problems such as case clashes and conflicts with the reserved DOS device name `AUX` should not apply.)

Byron: Thanks a lot for sharing!

After reading this I arrive at the same conclusion and can confirm that with `[parallel]` our `Makefile` wouldn't be needed anymore. I also think it's valuable to not have a `Makefile` when `just` would now do in every way.

Thus I very much welcome this transition to happen whenever possible, and whenever you have some time.

PR #2005, opened 2025-05-09
Fix build on NetBSD.

### LINKED ISSUE #2004: no field `st_mtimensec` on type `rustix::fs::Stat`
Hi,

This has started to happen recently,

```
   Compiling gix-index v0.40.0
error[E0609]: no field `st_mtimensec` on type `rustix::fs::Stat`
  --> /tmp/shells/starship/work/vendor/gix-index-0.40.0/src/fs.rs:65:38
   |
65 |             let nanoseconds = self.0.st_mtimensec;
   |                                      ^^^^^^^^^^^^ unknown field
   |
help: a field with a similar name exists
   |
65 |             let nanoseconds = self.0.st_mtime_nsec;
   |                                      ~~~~~~~~~~~~~

error[E0609]: no field `st_ctimensec` on type `rustix::fs::Stat`
  --> /tmp/shells/starship/work/vendor/gix-index-0.40.0/src/fs.rs:94:38
   |
94 |             let nanoseconds = self.0.st_ctimensec;
   |                                      ^^^^^^^^^^^^ unknown field
   |
help: a field with a similar name exists
   |
94 |             let nanoseconds = self.0.st_ctime_nsec;
   |                                      ~~~~~~~~~~~~~

For more information about this error, try `rustc --explain E0609`.
error: could not compile `gix-index` (lib) due to 2 previous errors
*** Error code 101
```

See for example, http://www.ki.nu/pkgsrc/reports/current/NetBSD-9.0/20250505.0626/starship-1.23.0/build.log or, http://www.ki.nu/pkgsrc/reports/current/NetBSD-9.0/20250505.0626/mise-2025.5.0/build.log

This seems to be related to https://github.com/GitoxideLabs/gitoxide/pull/1170

But, I don't know exactly what has changed recently.

Any help would be appreciated, thank you.

0-wiz-0:
rustix::fs 1.x has unified the nanonseconds member across operating systems.

Fixes #2004


0323pin:
@0-wiz-0 thanks for this 👍

EliahKagan (changes requested):
Thanks! Per https://github.com/GitoxideLabs/gitoxide/issues/2004#issuecomment-2866152446, I accidentally introduced this bug in #1949 by merging that update PR without adapting `gix-index` to the changes in `rustix`.

I have verified that the changes in this PR restore the ability to build on NetBSD, and that I am able to run commands like `gix clone` with the resulting binary.

#### A small change, to the commit message only

Although commits that adapt to changes in updated dependencies [are most often](https://github.com/GitoxideLabs/gitoxide/discussions/1946) not [conventional](https://www.conventionalcommits.org/en/v1.0.0/) (as [only some commits](https://github.com/GitoxideLabs/gitoxide/blob/main/DEVELOPMENT.md) need be), in this case the commit is fixing a bug that has appeared in a published version of the `gix-index` crate.

So I think the commit message, which is currently "Fix build on NetBSD.", will have to start with `fix:` in order to facilitate proper versioning and changelog entry generation. (It could be anything like "fix: Build on NetBSD." or "fix: Fix build on NetBSD" or some other variation if preferred.)

Since you've checked the box to allow maintainers to push to your branch, I can make such a change and force-push it. I'd be happy to do that, but I figured I'd check first, since you might prefer to do it.

Byron:
Thanks a lot for contributing the fix, and thanks @EliahKagan for reviewing and validating it!

On another note, now that I read the feedback about the commit message, I just want to share that I have been operating on the implicit rules that

* if I can push into the branch and
* the commit isn't verified/signed

then I can rewrite the commits by rebasing and/or make changes to assure changelog and version generation will be desirable.

Maybe this is something we should make explicit in the contribution guidelines so that in future you can assume consent in that regard.

EliahKagan (approved):
Thanks--looks good!

0-wiz-0:
Thank you for merging this so quickly!
Do you know when the next release is planned?
We'll have to convince some projects to switch to a fixed version of gitoxide to fix their NetBSD build.

EliahKagan:
Since there have been no SemVer-breaking changes to `gix-index` between its most recent v0.40.0 release and here, and this change is non-SemVer-breaking, it seems to me that it may be enough to release `gix-index` v0.40.1. Projects that specify `gix-index` v0.40.0 in the usual way should then automatically use v0.40.1 unless `--locked` is passed.

@Byron Is this workable?

Byron:
Thanks for the ping - it's indeed no problem to publish the fix, so it should arrive automatically downstream.

Issue #2006, opened 2025-05-09
Rare `writes_through_symlinks_are_prevented_even_if_overwriting_is_allowed` failure


EliahKagan:
#### The test failure

Yesterday, I [got a failure](https://github.com/EliahKagan/gitoxide/actions/runs/14916836574/job/41904276186#step:7:2894) of [`writes_through_symlinks_are_prevented_even_if_overwriting_is_allowed`](https://github.com/GitoxideLabs/gitoxide/blob/9d0c80917b56cdfe685682b80acae74126138c19/gix-worktree-state/tests/state/checkout.rs#L106-L138) on CI in a macOS job run in my fork:

```text
        FAIL [   0.010s] gix-worktree-state-tests::worktree state::checkout::writes_through_symlinks_are_prevented_even_if_overwriting_is_allowed
  stdout ───

    running 1 test
    test state::checkout::writes_through_symlinks_are_prevented_even_if_overwriting_is_allowed ... FAILED

    failures:

    failures:
        state::checkout::writes_through_symlinks_are_prevented_even_if_overwriting_is_allowed

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 14 filtered out; finished in 0.00s
    
  stderr ───

    thread 'state::checkout::writes_through_symlinks_are_prevented_even_if_overwriting_is_allowed' panicked at gix-worktree-state/tests/state/checkout.rs:122:9:
    assertion `left == right` failed
      left: ["A-dir/a", "A-file", "FAKE-DIR", "fake-file"]
     right: ["A-dir/a", "A-file", "FAKE-DIR", "FAKE-FILE"]
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

#### Changes on the branch

This job was run on e0a8dfe9c48072cef286ad88ad6a4160e5f30cd5 in https://github.com/EliahKagan/gitoxide/pull/31, which is a Dependabot PR. (I sometimes have open Dependabot PRs in my fork, to pre-test possible dependency versions.)

But it seems to me that the failure is probably not related to any upgraded dependencies. I ran the workflow four more times, and neither that test, nor any others in that job, nor any other  job in the workflow, failed in any other runs. Also...

#### Vague recollections

[...]

PR #2008, opened 2025-05-10
Test `overwrite_existing` apparent symlink dereferencing

EliahKagan:
To investigate #2006, this explodes the occasionally failing test `writes_through_symlinks_are_prevented_even_if_overwriting_is_allowed` into 1000 identical tests, using `test_case::test_matrix`.

When run in series, they all pass, but when run in parallel with `cargo nextest`, some failures always occur, at least when this is done on macOS. This is as predicted in:
https://github.com/GitoxideLabs/gitoxide/issues/2006#issuecomment-2866226601

A more specific result in local testing on macOS 15 is that, while the number of failures is 0 with `--test-threads=1`, with 2 or more threads, the number of failures tends to *decrease* with the number of threads. The most failures occur with `--test-threads=2`, about 30 on the system tested, while `--test-threads=16` has about 10 failures.

The tests were run with this command, with any `--test-threads=<N>` argument added:

```sh
cargo nextest run -p gix-worktree-state-tests writes_through_symlinks_are_prevented_even_if_overwriting_is_allowed --no-fail-fast
```

See https://github.com/GitoxideLabs/gitoxide/issues/2006#issuecomment-2868766118 for substantial further details and other relevant information and ideas.

(Maybe this PR can turn into a fix, once the cause is figured out. If not, it could be superseded by the fix and closed.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2599,6 +2599,7 @@ dependencies = [
  "gix-worktree-state",
  "once_cell",
  "symlink",
+ "test-case",
  "walkdir",
 ]
 
@@ -4755,6 +4756,39 @@ name = "termtree"
 version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8f50febec83f5ee1df3015341d8bd429f2d1cc62bcba7ea2076759d315084683"
+
+[[package]]
+name = "test-case"
+version = "3.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "eb2550dd13afcd286853192af8601920d959b14c401fcece38071d53bf0768a8"
+dependencies = [
+ "test-case-macros",
+]
+
+[[package]]
+name = "test-case-core"
+version = "3.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "adcb7fd841cd518e279be3d5a3eb0636409487998a4aff22f3de87b81e88384f"
+dependencies = [
+ "cfg-if",
+ "proc-macro2",
+ "quote",
+ "syn 2.0.101",
+]
+
+[[package]]
+name = "test-case-macros"
+version = "3.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5c89e72a01ed4c579669add59014b9a524d609c0c88c6a585ce37485879f6ffb"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.101",
+ "test-case-core",
+]
 
 [[package]]
 name = "thiserror"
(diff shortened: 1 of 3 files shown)


scriptforfivem:
Interesting solution to this problem! I've been working on similar challenges recently.

EliahKagan:
Regarding 16171b46994001354274bc148cc08c3bf1affb5e, don't worry: I do not plan to merge a new 28-job CI test matrix to main. See EliahKagan#36 for details on these jobs. They surface the failure on GNU/Linux; see https://github.com/GitoxideLabs/gitoxide/issues/2006#issuecomment-2878817675.

Note that 16171b46994001354274bc148cc08c3bf1affb5e (EliahKagan#36) doesn't include any changes to how the test uses `gix-worktree-state` functionality. In particular, it doesn't adjust whether `gix-features/parallel` is used or the value of the `num_threads` parameter. That's the next step (https://github.com/GitoxideLabs/gitoxide/issues/2006#issuecomment-2881392455). I wanted to add the GNU/Linux repro on CI first, to have greater confidence about the meaning of whatever is observed due to subsequent changes.

EliahKagan:
Changes related to `num_threads`, to cause the checkout operation under test to be single-threaded, is now done here in b333ef349b24affdeac17cd728ce1adf246829fe (squashed from EliahKagan#37). I believe this is sufficient to test the scenario anticipated in https://github.com/GitoxideLabs/gitoxide/issues/2006#issuecomment-2881392455. However, as can be seen on CI, this does not make the failure go away.

Byron:
> However, as can be seen on CI, this does not make the failure go away.

This is great!

That means the Rust side of things *should* be fully deterministic. Either it is not, or there is some filesystem involvement. The only of such reasons (of filesystem involvement) I know is the reliance on traversal order, or possibly relying on mtimes being 'in a certain way'. None of that *should* matter here though.

Byron:
Would you be able to re-open this in your own fork if this particular CI isn't needed?
(I am trying to bring PRs down to the ones that are actually in progress)

EliahKagan:
> Would you be able to re-open this in your own fork if this particular CI isn't needed?

Yes, that's no problem. Also, I'd say the most important thing is that we have #2006 to track the issue, so it's unlikely to be forgotten. I'm also likely to be reminded of it when working on some other stuff I hope to pick back up soon.

I think GitoxideLabs has more CI resources than my personal account does, and can therefore complete many concurrent jobs faster than my fork. I don't actually know that to be the case with the way we currently have things set up. Anyway, if that's needed, then this could later be reopened here or a new PR opened, or an upstream branch could be created that doesn't have a PR. None of those things are prevented by closing this PR now.

Byron:
Thanks.

And, right, beefed up resources! Pleas feel free to reopen here when you are picking this up again.

PR #2023, opened 2025-05-22
Add test for blame in sub-directory

cruessler:
This PR adds a test for blaming a file in a sub-directory. This already works, but was not tested before. It seems that one test is not much, but I couldn’t think of specific scenarios that I would have wanted to additionally add to this PR.

This addition could be an opportunity to refactor the `mktest` macro as it currently uses `$case` for both `$case.txt` and `$case.baseline`, forcing the latter to mirror the test cases’ directory structure. We could add `$baseline` in addition to `$case`, for example. What do you think?

### DIFF
--- a/gix-blame/tests/blame.rs
+++ b/gix-blame/tests/blame.rs
@@ -229,6 +229,8 @@ mktest!(added_line_before_changed_line, "added-line-before-changed-line", 3);
 mktest!(same_line_changed_twice, "same-line-changed-twice", 2);
 mktest!(coalesce_adjacent_hunks, "coalesce-adjacent-hunks", 1);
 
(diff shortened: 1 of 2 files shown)


Byron (approved):
Thanks a lot, much appreciated :)!

> We could add `$baseline` in addition to `$case`, for example. What do you think?

I definitely appreciate highlighting baseline tests specifically, and anything that makes the tests more maintainable and extensible.

cruessler:
> I definitely appreciate highlighting baseline tests specifically, and anything that makes the tests more maintainable and extensible.

Alright! I think I’m going to wait until more duplication accumulates before I make any changes, but it’s on my list of potential improvements.

PR #2026, opened 2025-05-25
Make MSRV checks and updates easier and more thorough

EliahKagan:
This PR could be regarded as a sequel to #2003. Prior to this PR, the MSRV had to be updated manually in `Cargo.toml` for the `gix` crate, in the `msrv.yml` workflow, and in the four occurrences appearing in `msrv-badge.svg`. No mechanism was in place to automate or verify that this had been done correctly.

The changes in this PR make the `rust-version` in `Cargo.toml` for the `gix` crate the single source of truth for the MSRV. This pertains to how the MSRV is validated and the actions that are undertaken to change it, so no change to `STABILITY.md` is made or required; this also does not change how modifying the MSRV will sometimes involve adjusting `rust-version` in other `gix-*` crates.

In the `justfile`, besides makes some minor adjustments for robustness and refactoring for clarity, this makes three external-facing recipe changes:

1. A new `msrv` recipe uses `cargo metadata` and `jq` to retrieve the `rust-version` of `gix` and report it in *major*.*minor*.*patch* format.

2. A new `msrv-badge` recipe calls the `msrv` recipe and writes a badge with the current MSRV to `etc/msrv-badge.svg`.

   This is supported by a new MSRV badge *template* with the literal text `{MSRV}` in place of the MSRV added in `etc`; the `msrv-badge` recipe substitutes those placeholders with the actual MSRV.

3. The old `ci-check-msrv` recipe was confusing, because it didn't do anything MSRV-specific: it checked that `gix` could build in a couple of ways using the current default toolchain. (Comment changes in [#2003](https://github.com/GitoxideLabs/gitoxide/pull/2003) attempted to make this clearer, but I think they managed to lessen the confusion only slightly, if at all.)

   This confusion is mostly fixed here, by renaming the recipe to `check-rust-version` and having it require a `rust-version` argument. Regardless of whatever toolchain is configured as the default, it uses the toolchain specified by the `rust-version` argument (which in ordinary use will be a version number) for the commands it runs.

This facilitates local use, such as running `just msrv-badge` to update the MSRV badge after modifying `gix/Cargo.toml`, but the more important effect is to support changes in the `msrv.yml` workflow:

- The hard-coded `MSRV` environment variable in `msrv.yml` is removed. The `check-msrv` job definition in that workflow instead uses `just msrv` to find out what MSRV toolchain to preinstall, and what argument to pass in `just check-rust-version`. Passing this argument to the `check-rust-version` recipe eliminates the need to set a default toolchain.

  (The main reason the entire MSRV check is not just added as a `check-msrv` recipe in the `justfile`, with the idea that the `check-msrv` jobs could call that, is that `check-msrv` in `msrv.yml` has an intermediate step that downgrades versions in `Cargo.lock`. It might be possible to take a different approach, or to add a `justfile` recipe for that as well, but it's not something that should be done locally by `just` commands that don't clearly advertise it, and it also feels like it doesn't belong in the scope of this PR.)

- A new `check-msrv-badge` job is added to `msrv.yml`. This verifies that `etc/msrv-badge.svg` is correct for the current MSRV, by calling `just msrv-badge`, checking if there are any changes, and failing if so.

  This should make it so that the badge cannot get out of date from forgetting to update it, as well as avoid problems like the `aria-label` bug fixed in dc1d271 ([#2003](https://github.com/GitoxideLabs/gitoxide/pull/2003)) where the badge was in an inconsistent state presenting different MSRVs depending on how it was examined.

  (This is its own job rather than a step in `check-msrv` because it is conceptually separate from that, and also because there are two `check-msrv` jobs--it uses a matrix strategy--but there should only be one `check-msrv-badge` job.)

---

I believe the `check-msrv-badge` job ought to block PR auto-merge--as the `check-msrv` jobs do--since changes to the MSRV may be made in PRs and any resulting inconsistency is most useful to catch immediately when it arises. Currently, the `check-msrv` jobs are listed as required checks (which is not something I can change, via a PR or otherwise).

To support doing that without making branch protection rules more complicated, I've also added an `msrv-pass` job to `msrv.yml` that depends on the jobs in that workflow in the same way that the `tests-pass` job in `ci.yml` depends on most of the other jobs there. (I think it's possible to do this *across* workflows, but more complicated, and it doesn't seem like it would be worthwhile to do for gitoxide's workflows at this time. That is something I am hoping to set up for GitPython, though.) So one option is to remove the current required check(s) for `msrv.yml` and add `msrv-pass` as a required check.

But an alternative is to get rid of the new `msrv-pass` job, and instead merge the `msrv.yml` workflow into `ci.yml`--that is, to put the `check-msrv` and `check-msrv-badge` job definitions in `ci.yml`. I think the workflows have had different triggers in the past, with `msrv.yml` having run on more branches than `ci.yml`. But even if so, that is no longer the case. The `ci.yml` workflow is about 500 lines long, so even if it doubles in size, I don't think it will be too long to read and work in easily. It also contains a wide variety of checks, not limited to running the test suite. Accordingly, it's not obvious that the MSRV checks are more different from the checks that in `ci.yml` than the various checks already in `ci.yml` are from each other.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -487,7 +487,7 @@ jobs:
           sort -m blocking-jobs.txt expected-nonblocking-jobs.txt |
             diff --color=always -U1000 - all-jobs.txt
 
-  # Dummy job to have a stable name for the "all tests pass" requirement
+  # Dummy job to have a stable name for the "all tests pass" requirement.
   tests-pass:
     name: Tests pass
 
@@ -504,7 +504,7 @@ jobs:
       - check-packetline
       - check-blocking
 
-    if: always() # always run even if dependencies fail
+    if: always()  # Always run even if dependencies fail.
 
     runs-on: ubuntu-latest
 
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -25,7 +25,6 @@ jobs:
         os:
           - windows-2022
           - ubuntu-latest
-          - macos-15  # FIXME(portability): Remove after testing new justfile changes.
 
     runs-on: ${{ matrix.os }}
 
@@ -39,7 +38,7 @@ jobs:
       - name: Read the MSRV
         run: |
           msrv="$(just msrv)"
-          echo "MSRV=$msrv" >> "$GITHUB_ENV"
+          tee -a "$GITHUB_ENV" <<<"MSRV=$msrv"
       - name: Set up MSRV and nightly toolchains
         run: |
           rustup toolchain install "$MSRV" nightly --profile minimal --no-self-update
@@ -58,12 +57,14 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: extractions/setup-just@v3
+      - name: Ensure we start out clean
+        run: git diff --exit-code
       - name: Regenerate the MSRV badge
         run: just msrv-badge
       - name: Check for changes
         run: git diff --exit-code
 
-  # Dummy job to have a stable name for the requirement that all MSRV tests pass.
+  # Dummy job to have a stable name for the requirement that all MSRV checks pass.
   msrv-pass:
     name: MSRV checks pass
 
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -25,25 +25,57 @@ jobs:
         os:
           - windows-2022
           - ubuntu-latest
+          - macos-15  # FIXME(portability): Remove after testing new justfile changes.
 
     runs-on: ${{ matrix.os }}
 
-    env:
-      # This is dictated by `firefox` to support the `helix` editor, but now probably effectively
-      # be controlled by `jiff`, which also aligns with `regex`.
-      # IMPORTANT: When adjusting, change all occurrences in `etc/msrv-badge.svg` as well.
-      RUST_VERSION: 1.75.0
+    defaults:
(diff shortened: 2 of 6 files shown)


Byron:
Thanks a lot for cleaning this up and particularly for setting up a single source of truth for the MSRV.

> To support doing that without making branch protection rules more complicated, I've also added an `msrv-pass` job to `msrv.yml` that depends on the jobs in that workflow in the same way that the `tests-pass` job in `ci.yml` depends on most of the other jobs there.

Screenshot 2025-05-26 at 04 50 58

I have adjusted the requirements accordingly, thank you.

> But an alternative is to get rid of the new `msrv-pass` job, and instead merge the `msrv.yml` workflow into `ci.yml`

I see that this could be useful, with a strong argument due to `ci.yml` containing all kinds of tests by now. I'd like to leave it to your judgement if and when to do that, since you are the one dealing with the files more than me.

EliahKagan:
> > But an alternative is to get rid of the new `msrv-pass` job, and instead merge the `msrv.yml` workflow into `ci.yml`
> 
> I see that this could be useful, with a strong argument due to `ci.yml` containing all kinds of tests by now. I'd like to leave it to your judgement if and when to do that, since you are the one dealing with the files more than me.

I do lean slightly toward moving the MSRV checks into the `ci.yml` workflow, and I've opened #2027 to propose that.

PR #2038, opened 2025-06-05
gix-actor docs: document conversions between `Signature` and `SignatureRef`

ilyagr:
These are some docs that could have been helpful as I am adapting `jj` to https://github.com/GitoxideLabs/gitoxide/pull/1935. Please feel free to edit them further.

I guess a part of this PR is a question: did I figure out correctly how I'm supposed to use these?

See also https://github.com/GitoxideLabs/gitoxide/pull/1935#issuecomment-2946631136. I wonder if some of these classes and methods should be renamed. (Perhaps it's possible to make the old name a deprecated type alias?)

### DIFF
--- a/gix-actor/src/lib.rs
+++ b/gix-actor/src/lib.rs
@@ -71,6 +71,8 @@ pub struct Signature {
 
 /// An immutable signature that is created by an actor at a certain time.
 ///
+/// Not fully parsed.
+///
 /// Note that this is not a cryptographical signature.
 #[derive(Default, PartialEq, Eq, Debug, Hash, Ord, PartialOrd, Clone, Copy)]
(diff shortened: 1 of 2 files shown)


ilyagr:
I also noticed that there exists `git_mailmap::Signature`, but I'm guessing I should not be using that.

Then there's `    impl From<SignatureRef<'_>> for Signature {`, which I perhaps should use instead of `SignatureRef::to_owned()` (perhaps that should be called `parse`?). I thought this was a `TryFrom` (**Update:** this is what the changelog for gix-actor said), but I guess that changed (or my IDE failed to find the TryFrom).

I'm also not sure whether the fact that the `From` implementation is lossy should be documented or not.

ilyagr:
Thank you very much for the review! Your edits are also great; the final result seems clear and helpful.

Issue #2020, opened 2025-05-21
Diff Options has same behaviour for `no_locations` and `track_filename`


uberroot4:
### Current behavior 😯

Hi,

I just saw today that the Setters for the Diff-Options in `gix/src/diff.rs` has the same implementation for the `no_locations` and `track_filename`:

https://github.com/GitoxideLabs/gitoxide/blob/5f9de52cf286163b503047b1ab3b51dfa093b4d4/gix/src/diff.rs#L67-L78

The doc for the `no_locations` method says `Do not keep track of filepaths at all [...]`, for the `track_filename` it says `Keep track of [...]`.

### Expected behavior 🤔

Based on the documentation of the `no_locations` method I would expect an impl like this:
```rust
/// Do not keep track of filepaths at all, which will leave all `location` fields empty.
pub fn no_locations(&mut self) -> &mut Self {
    self.location = None;
    self
}
```

### Git behavior

/

Byron:
That's a great catch and definitely looks like an… urm… copy-paste error. I wonder what the purpose of this API was, but it's obviously not tested either as more of a translation layer or API wrap.

A PR would definitely welcome if you'd want to contribute that.

Is that something you are interested in?

uberroot4:
> Is that something you are interested in?

Yes would be nice. What would you expect to be included? I'll try to find some time next week

uberroot4:
Hi @Byron, it took me way too long to create the PR but I finally managed to get some time. PR is created as #2045

cruessler:
@Byron The PR has been merged, can this issue be closed?

Byron:
Thanks!

PR #2045, opened 2025-06-13
#2020: Fixed no_locations options for diffing

uberroot4:


Byron (approved):
Thanks for making it happen, it's great to welcome you as contributor.

PR #2047, opened 2025-06-15
Update hashbrown to the latest version

blinxen:
Update hashbrown to version 0.15.4. Should also fix https://github.com/GitoxideLabs/gitoxide/issues/2046.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,18 +18,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
 
 [[package]]
-name = "ahash"
-version = "0.8.11"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
-dependencies = [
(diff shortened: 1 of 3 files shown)


Byron (approved):

It is awesome to see that no change is necessary to perform the upgrade, I was absolutely convinced that raw APIs are in use that are now forbidden. Probably I just confused this with `crates-index-diff` though, where the upgrade was indeed more involved.

EliahKagan (commented):
This updates what `gix-hashtable` re-exports, so I think it probably does qualify as a breaking change to the `gix-hashtable` API, even though nothing else in the workspace needed to be adjusted to accommodate it.

Should something be done to ensure this precipitates an appropriate version bump and generates a changelog entry of the correct type? (Assuming I'm right that this effect is needed, amending the commit to start with `change!:` would achieve it, but it could be achieved subsequently in other ways.)

Byron:
Indeed, `chore` is discouraged even though I would expect it to work. It just felt so much more fitting, while I didn't remember `DEVELOPMENT.md` at all 😅.
Maybe it's best to wait for the next release to see if `chore!` does indeed work, and then `DEVELOPMENT.md` could be rephrased. Generally, I still wouldn't use `chore` as these changes aren't interesting for changelogs. But when they are breaking, that's a different matter and maybe that's what's currently missing in `DEVELOPMENT.md`.

PR #2048, opened 2025-06-16
add committer fallback for fetch

### LINKED ISSUE #2040: gix fetch requires a configured user while git fetch does not
### Current behavior 😯

My test environment is a completely empty chroot:
```sh
$ gix clone https://github.com/gentoo/gentoo
$ cd gentoo
$ gix fetch
```
It produces
```
Error: Failed to update references to their new position to match their remote locations

Caused by:
    0: The reflog could not be created or updated
    1: reflog messages need a committer which isn't set
```

### Expected behavior 🤔

It would be nice to either provide the same reference to how to setup the configuration as git does

### Git behavior

`git fetch` works w/out requiring a user/email configured.

### Steps to reproduce 🕹

_No response_

ralphmodales:
Fixes #2040

This PR extends the existing committer fallback logic from clone operation to fetch operation. When no committer is configured, fetch now uses the same dummy values as the clone.


Byron:
Thanks a lot!

Issue #2056, opened 2025-06-23
`gix_url::Url` `%`-encodes username in scp-like syntax


yuja:
### Current behavior 😯

This code generates the following output:
```rust
fn main() {
    let url = gix_url::parse("_git@source.example.tld:org/repo.git".as_ref()).unwrap();
    dbg!(&url);
    println!("{}", url.to_bstring());
}
```

```
[src/main.rs:3:5] &url = Url {
    scheme: Ssh,
    user: Some(
        "_git",
    ),
    password: None,
    host: Some(
        "source.example.tld",
    ),
    serialize_alternative_form: true,
    port: None,
    path: "org/repo.git",
}
%5Fgit@source.example.tld:org/repo.git
```

### Expected behavior 🤔

I'm not pretty sure, but Git appears not to decode `%5Fgit` back to `_git`.

This problem was originally reported as https://github.com/jj-vcs/jj/issues/6810.

ilyagr:
In the linked issue, Yuya also suggested a workaround: `ssh://_git@source.example.tld/org/repo.git` seems to work already.

Byron:
Thanks a lot for reporting!

This is a perfect issue for contributing as a unit test should be straightforward to produce, and a fix can't be far then either.
That could also capture the `ssh://` as a workaround, just to be sure `_` in usernames there is covered and it won't break later.

Byron:
I think this was fixed via the linked commit.

PR #2058, opened 2025-06-24
fix underscore encoding in usernames for scp-like urls

ralphmodales:
This PR fixes issue #2056 where usernames in SCP-like SSH URLs (e.g., _git@source.example.tld:org/repo.git) were incorrectly percent-encoded, resulting in outputs like %5Fgit@source.example.tld:org/repo.git. Git does not percent-encode underscores in usernames for SCP-like URLs, and this change aligns with that behavior.

### DIFF
--- a/gix-url/src/lib.rs
+++ b/gix-url/src/lib.rs
@@ -312,6 +312,16 @@ fn percent_encode(s: &str) -> Cow<'_, str> {
     percent_encoding::utf8_percent_encode(s, percent_encoding::NON_ALPHANUMERIC).into()
 }
 
+fn percent_encode_for_scp_user(s: &str) -> Cow<'_, str> {
+    // For SCP-like URLs, we should be more conservative about what we encode
+    use percent_encoding::{AsciiSet, CONTROLS};
+
+    // Define characters that MUST be encoded in SCP user portion
+    const SCP_USER_ENCODE_SET: &AsciiSet = &CONTROLS.add(b' ').add(b'@').add(b':').add(b'\t').add(b'\n').add(b'\r');
+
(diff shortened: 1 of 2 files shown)


gix-url/src/lib.rs
```diff
@@ -322,7 +332,13 @@ impl Url {
         }
         match (&self.user, &self.host) {
             (Some(user), Some(host)) => {
-                out.write_all(percent_encode(user).as_bytes())?;
+                // For SCP-like SSH URLs, don't percent-encode underscores in usernames
+                // For regular SSH URLs, use standard encoding but allow underscores
+                let encoded_user = match self.scheme {
+                    Scheme::Ssh => percent_encode_for_scp_user(user),
```
yuja, line 338: I've implemented this separately: https://github.com/GitoxideLabs/gitoxide/pull/2060

I'm not sure which version is better, but I don't think we should encode `@`, etc. because Git doesn't decode these characters.

https://github.com/git/git/blob/v2.50.0/connect.c#L1081

Byron: Thank you for researching this!

Byron (commented):

Who would have thought that this is going to be fixed by two PRs at the same time - that never happened before in a project that doesn't see 100PRs per day!

With that said, I am about to merge #2060 as I really liked the additional refactoring. Unfortunately I wasn't able to see that it's more correct, i.e. how @ should be handled, as our baseline tests currently don't capture that.

Maybe this is something you want to look into - improving the comparison accuracy of the baseline tests to ideally be able to let Git tell us how `@` should be handled.

Byron:
Closing has duplicate of #2060 - thanks for your understanding.

PR #2059, opened 2025-06-24
gix-pack: delta application is a fallible operation

bradlarsen:
`gix_pack::data::delta::apply` is actually a fallible operation when given corrupted or crafted data. Previously, the function would panic in these situations; now a `Result` value is returned instead.

See #2044 for more context.

### DIFF
--- a/gix-pack/src/cache/delta/traverse/mod.rs
+++ b/gix-pack/src/cache/delta/traverse/mod.rs
@@ -42,7 +42,7 @@ pub enum Error {
     #[error("Failed to spawn thread when switching to work-stealing mode")]
     SpawnThread(#[from] std::io::Error),
     #[error(transparent)]
-    Delta(#[from] crate::data::delta::Error),
+    Delta(#[from] crate::data::delta::apply::Error),
 }
 
 /// Additional context passed to the `inspect_object(…)` function of the [`Tree::traverse()`] method.
--- a/gix-pack/src/data/delta.rs
+++ b/gix-pack/src/data/delta.rs
@@ -1,18 +1,21 @@
-/// Returned by [`Tree::traverse()`]
-#[derive(thiserror::Error, Debug)]
-#[allow(missing_docs)]
-pub enum Error {
-    #[error("Encountered unsupported command code: 0")]
-    UnsupportedCommandCode,
-    #[error("Delta copy from base: byte slices must match")]
-    DeltaCopyBaseSliceMismatch,
-    #[error("Delta copy data: byte slices must match")]
-    DeltaCopyDataSliceMismatch,
+///
+pub mod apply {
+    /// Returned when failing to apply deltas.
+    #[derive(thiserror::Error, Debug)]
+    #[allow(missing_docs)]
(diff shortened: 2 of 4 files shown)


bradlarsen:
Note that this PR doesn't fix the underlying problem from #2044 (presumably a miscompilation or some undefined behavior from zlib on x86_64?). However, it _does_ turn the crashes there into garden-variety `Result` values that can be handled normally.

Byron (approved):
Thanks a lot, this looks great!


Next to some minor refactors I also validated that performance isn't negatively affected by this change.

<details>

Before

```
❯ gix --verbose no-repo pack verify ./tests/fixtures/repos/linux.git/objects/pack/pack-3ee05b0f4e4c2cb59757c95c68e2d13c0a491289.idx
 09:59:08 Hash of index 'pack-3ee05b0f4e4c2cb59757c95c68e2d13c0a491289.idx' done 212.8MB in 0.47s (452.6MB/s)
 09:59:08                                           collecting sorted index done 7.6M entries in 0.87s (8.7M entries/s)
 09:59:10                                                          indexing done 7.6M objects in 2.56s (3.0M objects/s)
 09:59:11 Hash of pack 'pack-3ee05b0f4e4c2cb59757c95c68e2d13c0a491289.pack' done 1.4GB in 3.62s (376.0MB/s)
 09:59:34                                                         Resolving done 7.6M objects in 23.78s (319.6K objects/s)
 09:59:34
```

After
```
 cargo run --bin gix --release --  --verbose no-repo pack verify ./tests/fixtures/repos/linux.git/objects/pack/pack-3ee05b0f4e4c2cb59757c95c68e2d13c0a491289.idx
     Running `target/release/gix --verbose no-repo pack verify ./tests/fixtures/repos/linux.git/objects/pack/pack-3ee05b0f4e4c2cb59757c95c68e2d13c0a491289.idx`
 10:01:19 Hash of index 'pack-3ee05b0f4e4c2cb59757c95c68e2d13c0a491289.idx' done 212.8MB in 0.44s (483.1MB/s)
 10:01:19                                           collecting sorted index done 7.6M entries in 0.48s (15.7M entries/s)
 10:01:21                                                          indexing done 7.6M objects in 1.77s (4.3M objects/s)
 10:01:21 Hash of pack 'pack-3ee05b0f4e4c2cb59757c95c68e2d13c0a491289.pack' done 1.4GB in 3.02s (450.0MB/s)
 10:01:45                                                         Resolving done 7.6M objects in 24.29s (312.9K objects/s)
 10:01:45                                                          Decoding done 96.0GB in 24.29s (4.0GB/s)
```

</details>

### Next Steps

And agreed, it's clear that this won't fix the compilation issue, but would at least allow to work around it. I wonder if ignoring this error now shows that everything else also fails, or that it really is only one object that can't be decoded.
If it's only one object or a small amount of them, I think we can dig in deeper and create a connection between the zlib compressed input buffer, and the desired and actual output. With that we should be able to setup a unit test for it, and then a fix either in this codebase, or upstream.

What do you think?

PR #2060, opened 2025-06-25
fix: username in scp-like url is no longer percent-encoded (#2056)

yuja:
Since Git doesn't percent-decode characters in scp-like URL, we shouldn't encode username at all.

https://github.com/git/git/blob/v2.50.0/connect.c#L1081

I've split write_to() function to clarify that any non-path components that should be separated by ":" cannot be serialized in alternative form. I've made it fall back to the URL syntax if password or port number was set. Maybe we can also check if the user or host includes ":", but I'm not sure how much foolproof we should add here.

### DIFF
--- a/gix-url/src/lib.rs
+++ b/gix-url/src/lib.rs
@@ -315,11 +315,23 @@ fn percent_encode(s: &str) -> Cow<'_, str> {
 /// Serialization
 impl Url {
     /// Write this URL losslessly to `out`, ready to be parsed again.
-    pub fn write_to(&self, mut out: &mut dyn std::io::Write) -> std::io::Result<()> {
-        if !(self.serialize_alternative_form && (self.scheme == Scheme::File || self.scheme == Scheme::Ssh)) {
-            out.write_all(self.scheme.as_str().as_bytes())?;
-            out.write_all(b"://")?;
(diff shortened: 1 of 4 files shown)


Byron:
Thanks so much, I am loving the refactoring *and* the tests for the additional logic improvements.

Since I dug into the baseline test a little bit I noticed that we are still not parsing everything that Git can correctly. If the baseline could be brought to greater compatibility, maybe then it's just another small step to assure we can serialize everything correctly as well. Maybe something for @rickprice - taking many small steps would lead to a positive outcome here.

rickprice:
I can go over the git tests and ensure ours cover the same points if that
helps?

At ActiveState I did a lot of CVE remediation, and ensuring all the tests
existed and passed was one of the things that was needed.

Frederick

Byron:
Thanks for offering!

However, I don't think there were new tests added to the URL related Git test-suite, and I believe there is a lot to work on before trying to find more. The baseline tests have over 100 failures right now, i.e. Git parsing URLs that we can't parse or parse incorrectly. Thus, in `gix-url` there might be a lot of low-hanging fruit to pick by getting it to be more compliant.

rickprice:
Okay, I will try running the tests and see if I can start fixing things.  For the moment, that is what I will focus on to try and help the project out.

PR #2062, opened 2025-06-25
Small documentation fixes

rickprice:
Small grammar and spelling changes to make the codebase better.

### DIFF
--- a/gix-features/src/parallel/reduce.rs
+++ b/gix-features/src/parallel/reduce.rs
@@ -232,7 +232,7 @@ pub trait Reduce {
     ///
     /// If an `Error` is returned, the entire operation will be stopped.
     fn feed(&mut self, item: Self::Input) -> Result<Self::FeedProduce, Self::Error>;
-    /// Called once once all items where passed to `feed()`, producing the final `Output` of the operation or an `Error`.
+    /// Called once for all items that were passed to `feed()`, producing the final `Output` of the operation or an `Error`.
     fn finalize(self) -> Result<Self::Output, Self::Error>;
 }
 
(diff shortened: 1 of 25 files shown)


gix-features/src/parallel/reduce.rs
```diff
@@ -232,7 +232,7 @@ pub trait Reduce {
     ///
     /// If an `Error` is returned, the entire operation will be stopped.
     fn feed(&mut self, item: Self::Input) -> Result<Self::FeedProduce, Self::Error>;
-    /// Called once once all items where passed to `feed()`, producing the final `Output` of the operation or an `Error`.
+    /// Called once for all items that were passed to `feed()`, producing the final `Output` of the operation or an `Error`.
```
EliahKagan, line 235: Does this mean that it is called one time for each item, and thus as many times as there are items, or that it is called only once no matter how many items there were? (The ambiguity isn't really new, because "once once" in the old wording might've meant "once" or might've meant "one time, immediately after," and because the old "where" might've meant "were"/"have been" or might've meant some other thing. But it occurs to me to ask about it, in case it can be further clarified.)

Byron: Thanks for bringing this up!

It will be called once to digest all items that were passed to `feed()` into a new output. Do you think this comes across? Since it's consuming at least when actually calling `finish()` it will be clear what's happening.
If it's not clear, I'd definitely appreciate improved wording.

EliahKagan: Thanks. I think it is probably clear in context. I commented so as not to forget. But next time I look over ways to clarify documentation, I'll try to return to this and see if there is any clearer wording.

PR #2066, opened 2025-06-27
`gix-blame`: add test for file with two roots

cruessler:
This PR adds a test for when a file can be traced to two different root commits in two different branches. This is to make sure the algorithm does not terminate when it gets to the first `TreeDiffChange::Addition`.


Byron (approved):
Thanks a lot!

I really love how Git can be asked to create a baseline, which is then turned into a comparable format by `Baseline::collect()` so it can be asserted against the actual output.

PR #2071, opened 2025-07-02
Add explicit accessors to `gix_diff::index::ChangeRef``

cruessler:
- feat: add explicit accessors for common fields
- Adapt to changes in `gix-diff`

There don’t seem to be any explicit tests specifically for `.fields()` yet, so I extended an existing test that already used `.fields().0`. Let me know if there’s a better way to test this!

This PR is based on an idea in this comment: https://github.com/gitui-org/gitui/pull/2673#discussion_r2174461979


Byron (approved):
I think that's great, thanks so much :)!

PR #2073, opened 2025-07-02
feat: add first debug version of `gix tag list`

cruessler:
This PR adds a new command: `gix tag list`.

`gix tag list` outputs a list of tags, adding information with respect to whether a tag is lightweight or annotated (the examples are taken from the `gitoxide` repository):

- for a lightweight tag, it will show the name it has in `.git/refs/tags/`: `v0.8.4`.
- for an annotated tag where the name is identical to the filename in `.git/refs/tags/`, it will show an asterisk next to the name: `gix-v0.72.1 *`.
- for an annotated tag where the name is *not* identical to the filename in `.git/refs/tags/`, it will show the tag name in square brackets: `gitoxide-v0.8.1 [tag name: v0.8.1]`.

Feel free to modify this in any way you like, though, in particular if there are already established patterns in other parts of the codebase.

### DIFF
--- a/gitoxide-core/src/repository/mod.rs
+++ b/gitoxide-core/src/repository/mod.rs
@@ -44,6 +44,7 @@ pub mod remote;
 pub mod revision;
 pub mod status;
 pub mod submodule;
+pub mod tag;
 pub mod tree;
 pub mod verify;
 pub mod worktree;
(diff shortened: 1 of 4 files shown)


EliahKagan:
> And finally, a suggestion for a rainy day: I'd love it if _by default_ it could produce the same order as `git tag --sort='version:refname'`, effectively making it semver aware.

Is this to say that it the [`versionsort.suffix`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-versionsortsuffix) configuration variable should be recognized as well?

Byron:
Right, that's another tweak one could apply for even better sorting. To me being semver aware would probably automatically entail handling this. Ultimately what matters is that the default sort order should be something that works for `gitoxide` and maybe most projects even.

PR #2075, opened 2025-07-04
various improvements

### LINKED ISSUE #2074: Misleading example in gix_path::normalize
### Current behavior 😯

The example given is

> this turns `a/./b/c/.././..` into `a`

which suggests that it removes `.` as well. However it does not affect `.` (or repeated separators), as the documentation later states; the `.` is only removed as part of the normalization done by `components()`. In particular, if there is no `..` component, then the original `Cow<Path>` is returned without having been parsed and re-assembled. So e.g. `a/./b` is returned unchanged.

### Expected behavior 🤔

This confusion should be either avoided or drawn attention to, maybe:

- don't include an example with `.`
- add an explanation like "(`.` is only incidentally removed because the Path is rewritten)"
- move the examples after the "Single `.` components" line
- include `a/./b` as an example (maybe after the "Single `.` components" line)

This is just to help someone else avoid a frustrating misunderstanding due to reading just the example, it is probably clear enough if one reads carefully and understands `std::path` better than I did.

### Git behavior

_No response_

### Steps to reproduce 🕹

_No response_

Byron:
### Tasks

* [x] better docs for `gix-fs::normalize()` (fixes #2074)
* [x] ~~`submodules()` fails in unborn repositories, needs head commit.~~ - this was already fixed two months ago.
* a new release


Byron:
### Notes about CVE-2025-48384

After `gix clone https://github.com/acheong08/CVE-2025-48384` and running `gix submodule` in that directory with the fix here it becomes clear that `gitoxide` doesn't implement configuration value stripping like Git does.

Hence the path of the submodule is what it is supposed to be as per configuration.

```
 ⅹ "sub\r" config:no head:bf04e15e4e948cd90b613dac982d6fb52c538a55 index:bf04e15e4e948cd90b613dac982d6fb52c538a55 (no worktree) [https://github.com/acheong08/totallynotsuspicious.git]
```

This means that if it was to check it out into its path, it wouldn't by any chance run into the symlink in `sub`.

```
❯ l
drwxr-xr-x@     - byron staff  9 Jul 07:57 .git
.rw-r--r--@    94 byron staff  9 Jul 07:57 .gitmodules
.rw-r--r--@ 2.3Ki byron staff  9 Jul 07:57 README.md
lrwxr-xr-x@     - byron staff  9 Jul 07:57 sub -> .git/modules/sub/hooks
drwxr-xr-x@     - byron staff  9 Jul 07:57 sub\r
```

It's notable that the path of the submodule checkout can be configured at will and I'd think that if `gitoxide` was to implement recursive submodule cloning, it would naturally check for the destination to be suitable.

CC @EliahKagan

PR #2086, opened 2025-07-25
Sort tags as versions by default

cruessler:
This PR changes `gix tag list`, sorting tags by default as version numbers in a naive, but hopefully sufficiently robust way. In order to do that, it adds `struct Version` with a method to parse strings into version parts. It implements `Ord` for `Version` and `VersionPart` to do the actual work of comparing two versions.

I’ve tested this implementation in the `gitoxide` repo where sorting was identical to `git tag --list --sort=version:refname`.

### DIFF
--- a/gitoxide-core/src/repository/revision/resolve.rs
+++ b/gitoxide-core/src/repository/revision/resolve.rs
@@ -25,8 +25,6 @@ pub enum BlobFormat {
 pub(crate) mod function {
     use std::ffi::OsString;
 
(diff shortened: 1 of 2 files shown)


Byron (approved):
Thanks a lot, the sorting is pretty much what I had in mind.

To me the chunking is the 'core-achievement' here, and I think it's a cool solution to the problem as it requires minimal assumptions about how the tag is actually named.

PR #2091, opened 2025-08-01
WIP: gix-date: handling bad offsets

elmarco:
This commit from edk2 has an invalid author date offset: https://github.com/tianocore/edk2/commit/630cb8507b2f1d7d7af3ac0f992d40f209dc1cee

```
$ gix cat 630cb8507b2f1d7d7af3ac0f992d40f209dc1cee ...
author Ruiyu Ni <ruiyu.ni@intel.com> 1516956202 -3407

But git show can still display it:
commit 630cb8507b2f1d7d7af3ac0f992d40f209dc1cee
Author: Ray Ni <ray.ni@intel.com>
Date:   Wed Jan 24 22:36:22 2018 -3407
```

How to handle it with gix-date?

### DIFF
--- a/gix-date/tests/time/format.rs
+++ b/gix-date/tests/time/format.rs
@@ -81,6 +81,14 @@ fn git_default() {
     );
 }
 
+#[test]
+fn bad_offset() {
+    assert_eq!(
+        time_bad_offset().format(gix_date::time::format::DEFAULT),
+        "Wed Jan 24 22:36:22 2018 -3407"
+    );
+}
+
 fn time() -> Time {
     Time {
         seconds: 123456789,
@@ -92,5 +100,12 @@ fn time_dec1() -> Time {
     Time {
         seconds: 123543189,
         offset: 9000,
+    }
+}
+
+fn time_bad_offset() -> Time {
+    Time {
+        seconds: 1516956202,
+        offset: -122820,
     }
 }


Byron:
Thanks for contributing a test!

I see now that `to_time()` can panic, and I think that's to be fixed right away by making it fallible.
Is this something you want to contribute?

Besides that, do you have something in particular that is failing in `gitoxide` thanks to that that Git handles better?
Particularly, round-tripping should be supported by `gix` as well as it stores the original bytes for the author and time (https://github.com/GitoxideLabs/gitoxide/blob/aff23d65a1a44e5356fb362a857d736280d3a580/gix-actor/src/lib.rs#L90-L94) so it won't fail when decoding invalid times or authors.

elmarco:
Hi @Byron and cool project! I would be happy to contribute to if I had more time ;) I am a bit unsure how to handle this issue. Decoding &str to Time is not the issue, it's the formatting back to  human string that is problematic, and I understand it's not really gix business at this point. At the same time, handling badly formatted time representations may also be not jiff responsability.

Yet, imho gitoxide should provide a way to display the same result a `git show` in this case. Do you think we should work on solving this issue in jiff by having some support for bad representations? Or do you disagree with the above?

Byron:
I think I understand now what you mean - it's to be able to decode the date at least, and leave the offset alone. Git can decode just the date string, and displays the original offset.
My feeling is that it just has no limits in terms of offset, so for it there is no problem.

As `jiff` is used for formatting, and as `jiff` is requiring correct inputs, I don't think `Time::to_time()` can ever be used. What I can imagine is to have a lenient version of the function that catches such an error and instead retries without an offset, possibly providing an extra method to format the offset just like Git does. Then the caller can explicitly handle this case.

Looking at the code…

https://github.com/GitoxideLabs/gitoxide/blob/57366d3ebd622af8927bb0e199ab8a3c0eafee99/gix-date/src/time/format.rs#L53-L58


…it would already be trivial to do as a workaround, so maybe what `gitoxide` should provide is a lenient function that does it all.

Maybe `Time::to_lenient_time() -> (jiff::Zoned, Option<gix_time::parse::TimeBuf>)`, where the `TimeBuf` holds the offset part in case it couldn't be handled by `jiff`.

Application code still has to handle more, but I see no other way. Maybe you do though.

Byron:

PR #2093, opened 2025-08-02
Upgrade the `getrandom` dependency of `gix-diff`

### LINKED ISSUE #2092: Upgrading `getrandom`
### Summary 💡

The current version of `getrandom` is 0.3.3. One use in gitoxide of `getrandom` is in `gix-diff`, where it is a direct dependency when the `wasm` feature is enabled, currently requiring a 0.2.\* version:

https://github.com/GitoxideLabs/gitoxide/blob/473fe522e84569f77bf38294a412f0d13fa54d63/gix-diff/Cargo.toml#L23-L24

https://github.com/GitoxideLabs/gitoxide/blob/473fe522e84569f77bf38294a412f0d13fa54d63/gix-diff/Cargo.toml#L47

#### The error

When Dependabot attempts to upgrade `getrandom` to a 0.3.\* version, it [fails](https://github.com/GitoxideLabs/gitoxide/network/updates/1066005040):

> Dependabot failed to update your dependencies because there was an error resolving your Rust dependency files.
> 
> Dependabot encountered the following error:
> 
> ```text
> Updating crates.io index
> error: failed to select a version for `getrandom`.
>     ... required by package `gix-diff v0.53.0 (dependabot_tmp_dir/gix-diff)`
>     ... which satisfies path dependency `gix-diff` (locked to 0.53.0) of package `gix-pack v0.60.0 (dependabot_tmp_dir/gix-pack)`
>     ... which satisfies path dependency `gix-pack` (locked to 0.60.0) of package `gix-odb v0.70.0 (dependabot_tmp_dir/gix-odb)`
>     ... which satisfies path dependency `gix-odb` (locked to 0.70.0) of package `gix-object v0.50.0 (dependabot_tmp_dir/gix-object)`
>     ... which satisfies path dependency `gix-object` (locked to 0.50.0) of package `gix-ref v0.53.0 (dependabot_tmp_dir/gix-ref)`
>     ... which satisfies path dependency `gix-ref` (locked to 0.53.0) of package `gix-discover v0.41.0 (dependabot_tmp_dir/gix-discover)`
>     ... which satisfies path dependency `gix-discover` (locked to 0.41.0) of package `gix-testtools v0.17.0 (dependabot_tmp_dir/tests/tools)`
>     ... which satisfies path dependency `gix-testtools` (locked to 0.17.0) of package `gix-path v0.10.19 (dependabot_tmp_dir/gix-path)`
>     ... which satisfies path dependency `gix-path` (locked to 0.10.19) of package `gix-fs v0.16.0 (dependabot_tmp_dir/gix-fs)`
>     ... which satisfies path dependency `gix-fs` (locked to 0.16.0) of package `gix-tempfile v18.0.0 (dependabot_tmp_dir/gix-tempfile)`
>     ... which satisfies path dependency `gix-tempfile` (locked to 18.0.0) of package `gix-lock v18.0.0 (dependabot_tmp_dir/gix-lock)`
> versions that meet the requirements `>=0.2.15, <=0.3.3` (locked to 0.3.2) are: 0.3.2
> 
> package `gix-diff` depends on `getrandom` with feature `js` but `getrandom` does not have that feature.
>  package `getrandom` does have feature `std`
> 
> 
> failed to select a version for `getrandom` which could resolve this conflict
> ```

This is not a Dependabot problem. That is, the error and the condition it reports are not specific to Dependabot, and the underlying limitation is not something we can fix by using Dependabot differently or substituting a different update procedure.

- The cause is an actual incompatibility, arising from the removal of a feature we are depending on, as described below. (Although we can't fix the underlying incompatibility by using Dependabot differently, there is a way to make Dependabot not attempt the upgrade that causes the error, which I note below.)
- Dependabot seems to be able to upgrade other packages without problems even when this error occurs. In particular, Dependabot encountered this error in the run associated with [#2090](https://github.com/GitoxideLabs/gitoxide/pull/2090), and that PR was generated without problems and seems to have had no trouble upgrading crates other than `gitrandom`.

#### The cause

`gix-diff` specifies the `js` feature of `getrandom`. The `js` feature was [removed](https://github.com/rust-random/getrandom/blob/master/CHANGELOG.md#removed-2) as of `getrandom` [0.3.0](https://github.com/rust-random/getrandom/blob/master/CHANGELOG.md#030---2025-01-25). The change was made in https://github.com/rust-random/getrandom/pull/504, which notes:

> This PR removes `linux_disable_fallback`, `rdrand`, `js`, `test-in-browser`, and `custom` crate features. As their replacement two new configuration flags are introduced: `getrandom_browser_test` and `getrandom_backend`. The latter can have the following values: `custom`, `rdrand`, `linux_getrandom`, `wasm_js`, `esp_idf`.

Further details and context are in the changelog and PR description there.

My guess is that we would not encounter problems making `gix-diff` compatible with that, but I haven't tried. One thing I might want to do first is to figure out if the current WASM tests are sufficient to reveal a breakage.

#### General motivations

It would be nice to let users of the `gix-diff` crate use the newer version of `getrandom`, which I believe has various compatibility and robustness improvements, even separate from this change in its features. Some new functionality has been backported to 0.2.\* versions released since 0.3.\* came out, but I don't think all is.

More broadly, I think it's generally a good idea to use newer versions of dependencies in the absence of a particular reason not to, unless the older version of the dependency is as well supported as the newer version.

#### Specific motivation - are we using `js` correctly?

The bigger motivation is that I am not sure our use of the `js` feature is correct. Versions of `getrandom` that offer the `js` feature [include this](https://github.com/rust-random/getrandom/blob/ce4144b2c16fe1422037c93e267e6a52336e0834/src/lib.rs#L95-L98) in their documentation of it:

```rust
//! This feature should only be enabled for binary, test, or benchmark crates.
//! Library crates should generally not enable this feature, leaving such a
//! decision to *users* of their library. Also, libraries should not introduce
//! their own `js` features *just* to enable `getrandom`'s `js` feature.
```

Contrary to that, we currently enable the `js` feature unconditionally for the `getrandom` dependency of `gix-diff`. That dependency is itself conditional, being present only when the `wasm` feature is enabled. The `js` feature is also implemented in `getrandom` in such a way that it only has an effect when `wasm32-unknown-unknown` is the target, if I understand correctly. But I am not sure that `wasm32-unknown-unknown` being the target implies that JavaScript is available, or that JavaScript ought to be used if available, or that we know this better than users of `gix-diff` would know it. Maybe this is a case where it's reasonable for a library crate to enable the `js` feature, but it's not clear to me that this is so.

It looks like one of the reasons for removing the feature and using configuration flags instead is to keep library crates from imposing the associated assumptions on crates that use them, and that this has been a problem. The discussion in https://github.com/rust-random/getrandom/issues/675, particularly at https://github.com/rust-random/getrandom/issues/675#issuecomment-2941738862, seems relevant.

#### Noisy Dependabot runs

It's useful to be able to look at Dependabot logs to see what happened. The presence of the prominent error is potentially distracting. This is the case even if one wishes only to check the status and does not look at the actual log--it looks like the update operation has failed, when actually it has succeeded other than for `getrandom`:

> Screenshot of the dependency graph feature on the Dependabot tab showing a recent version update job, indicating a 'Dependabot can't resolve your Rust dependency files' error

This could also be solved by editing `dependabot.yml` to only accept patch version updates to `getrandom`. This would probably be okay to do until the `getrandom` dependency of Dependabot is updated, though I worry that there might be some circumstance where some other `getrandom` dependency wouldn't be upgraded.

#### Via other crates

We already have other transitive dependencies on `getrandom` 0.3.\*, [through](https://gist.github.com/EliahKagan/3a7adbc0d1ea691db01c216af8dd5990#file-output-txt-L307) `tempfile`. (There is also a build-only one, [through](https://gist.github.com/EliahKagan/3a7adbc0d1ea691db01c216af8dd5990#file-output-txt-L704) `xz2`'s `cc` build dependency, via `jobserver`.) Nonetheless, at this time, upgrading the `gix-diff` dependency on `getrandom` from 0.2.\* to 0.3.\* might often not decrease build size, because we also have a separate transitive dependency on `getrandom` 0.2.\*, [through](https://gist.github.com/EliahKagan/3a7adbc0d1ea691db01c216af8dd5990#file-output-txt-L1150) `ring`.

If upgrading it as a dependency of `gix-diff` is easy, then it's worth doing even without the benefit of improving deduplication. If it's not easy, then that might make it worthwhile to stay back at 0.2.\*. But if it turns out to be hard in a way that overlaps with the difficulties upgrading it in `ring`, then maybe work on that in `ring` could show the way for doing it here, or vice versa.

My guess is that it should be feasible to do here because we don't yet have robust support for `wasm32-unknown-unknown` and because it's typically okay to introduce breaking changes in `gix-diff` (so long as SemVer reflects them), since the crate is not yet stabilized. But I mention `ring` anyway just in case there are any insights to be had. The relevant `ring` issue is https://github.com/briansmith/ring/issues/2341.

#### Near future use in `gix-utils`

For #1816, `gix-utils` will probably get `getrandom` as a direct dependency, to use to seed thread-local `fastrand::Rng` instances, for the technique described in https://github.com/GitoxideLabs/gitoxide/issues/1816#issuecomment-2620815045.

The point in the earlier comment https://github.com/GitoxideLabs/gitoxide/issues/1816#issuecomment-2618066550 about `getrandom` already being in the dependency tree is correct regardless of whether 0.2.\* or 0.3.\* is used, since both are already in the dependency tree, even separately from `gix-diff`, as described above. However, I'd prefer to use the newer version when adding it as a new dependency. I also want to avoid introducing any new wrong usage, or new occurrences of existing wrong usage, when doing so. Thus, I'd be somewhat reluctant to depend on the `js` feature when introducing `getrandom` as a dependency of `gix-utils`.

That's relevant because if `gix-diff` and `git-utils` can use the same version of `getrandom`, and either declare them the same way or declare them differently in way that is relevant to differences in how they use `getrandom`, then that's ideal.

EliahKagan:
Would fix #2092

I don't know if I'll manage to make this work. But even if not, maybe showing my attempts in this PR will make the situation in #2092 clearer.


.github/workflows/ci.yml
```diff
@@ -460,7 +460,7 @@ jobs:
       - name: crates with 'wasm' feature
         run: |
           set +x
-          for crate in gix-pack; do
+          for crate in gix-diff gix-pack; do
```
EliahKagan: I actually expect this not to work, even once the intentional breakage is removed, because `gix-diff` has default features that are incompatible with at least one of the WASM targets.

*Edit:* [Yes](https://github.com/GitoxideLabs/gitoxide/actions/runs/16697578071/job/47263939516#step:8:32), that happens.

gix-diff/Cargo.toml
```diff
@@ -44,7 +44,7 @@ gix-traverse = { version = "^0.47.0", path = "../gix-traverse", optional = true
 thiserror = "2.0.0"
 imara-diff = { version = "0.1.8", optional = true }
 serde = { version = "1.0.114", optional = true, default-features = false, features = ["derive"] }
-getrandom = { version = "0.2.8", optional = true, default-features = false, features = ["js"] }
+getrandom = { version = "0.2.8", optional = true, default-features = false }
```
EliahKagan: I've got multiple commits introducing and removing this temporary breakage to test various ways of writing the CI step. Don't worry--I plan to squash several of them into one commit when I'm done.

.github/workflows/ci.yml
```diff
@@ -460,9 +461,8 @@ jobs:
       - name: crates with 'wasm' feature
         run: |
           set +x
-          for crate in gix-pack; do
-            cargo build -p "$crate" --features wasm --target "$TARGET"
-          done
+          cargo build -p gix-diff --no-default-features --features wasm --target "$TARGET"
+          cargo build -p gix-pack --features wasm --target "$TARGET"
```
EliahKagan: This step, and most others, has `set +x`. But it seems they should have `set -x` instead, since that's what enables the `xtrace` mode, which shows what commands were run, and thus helps identify what output came from what command.

It looks like the modified step to explicitly test `gix-diff` without default features as well as `gix-pack` with default features works. But it's hard to tell for sure. That problem should go away once `set +x` in these steps is changed to `set -x`.

But I might do that in a separate PR because it has benefits separate from the other changes in this PR, and the latter may or may not work out.

*Edit:* I've done that in #2094, and I'll rebase this after that PR auto-merges.

gix-diff/build.rs
```diff
@@ -0,0 +1,14 @@
...
+    if target != "wasm32-unknown-unknown" {
+        return;
+    }
+
+    if let Some(feature_wasm) = var_os("CARGO_FEATURE_WASM") {
+        if feature_wasm == "1" {
+            println!("cargo:rustc-cfg=getrandom_backend=\"wasm_js\"");
+        }
+    }
+}
```
EliahKagan: This doesn't work. It seems the reason is that, when the `getrandom` dependency of `gix-diff` is built, this hasn't run. It's probably good that `build.rs` scripts don't have that kind of reach.

Thus, this should be removed by dropping f4549a1842a8b0cd61e9d246a3a1c8b2911e130b (and d5c5e57352ebe9fd50019b33db91b9582301c6ee), and 037cb8cad198a26b61be8dc3c002b18ae8a113ae should be amended to avoid making it sound like this would work, by clarifying that the suggested code would go in a `config.toml` file (as the documentation says), or by removing the part that mentions that (where I had for some reason thought I could put it in `Config.toml`, and then had the idea of using `build.rs` based on that).

It would be fine to configure this in `config.toml`, but it wouldn't help any users of the crate for us to do it--they would have to configure it, even in addition to requesting the `wasm` feature, which is what I was trying to avoid. Also, we probably shouldn't set it in `config.toml`, since we would want to find out if more of the CI steps needs this than we think do.

.github/workflows/ci.yml
```diff
@@ -457,12 +457,14 @@ jobs:
           for feature in progress parallel io-pipe crc32 zlib cache-efficiency-debug; do
             cargo build -p gix-features --features "$feature" --target "$TARGET"
           done
+      - name: enable wasm_js backend
+        if: matrix.target == 'wasm32-unknown-unknown'
+        run: echo RUSTFLAGS='--cfg getrandom_backend="wasm_js"' >> "$GITHUB_ENV"
```
EliahKagan: I don't believe we can cause enabling the `wasm` feature of `gix-diff` to achieve this effect, nor otherwise achieve this effect for users of `gix-diff` in a way that spares them the need to do an extra step.

That is, whoever does the build must do something to achieve the effect of passing `--cfg getrandom_backend="wasm_js"` to `rustc`, in addition to whatever features are enabled. This can be done in a `.cargo/config.toml` file, but having such a configuration here will not affect other projects that are not hosted in this repository and that host `gix-diff`.

I believe placing the onus on the downstream developer, rather than library crate developers, to enable the `wasm_js` backend, is one of the reasons for the breaking changes in `getrandom` 0.3. Reasons for this may include security--the JavaScript backend may not always be adequately secure, if I understand correctly. If all possible uses of `getrandom` through `gix-diff` do not require unpredictable random numbers, then security may not be an issue. I suspect that is the case, but I don't know for sure. Anyway, I still don't know of a way to make it work like it did in 0.2.

Options include:

1. Documenting that this is so and going ahead with the upgrade, *or*
2. Holding off on upgrading (or even planning not to upgrade) `getrandom`.

There may be other options I haven't thought of, and there may be gaps or inaccuracies in my understanding.

EliahKagan (commented):
See https://github.com/GitoxideLabs/gitoxide/pull/2093#discussion_r2249482206. If this is to be merged in something like its current form, then in addition to documenting the additional actions that would have to be taken to set the Rust flag to select the optional `getrandom` backend, I think we would want to rebase this to make one of the commits that modifies `gix-diff/Cargo.toml` conventional.

Per https://github.com/GitoxideLabs/gitoxide/issues/2092#issuecomment-3146902098, even if this should just be closed, it's my hope that helps make the situation clearer than before.

Byron (commented):
Maybe this could be as 'easy' as going with this - it works like it worked before.

Without the `wasm_js` Cargo feature I suppose it wouldn't work, but would it work without that feature and without the RUSTFLAGS setting? If that was the case, that might be preferred as it leaves everything to the caller.

Otherwise, I'd be OK merging exactly what we have here.

.github/workflows/ci.yml
Byron: Additional documentation might help - users now have to know how to configure `getrandom` if they want to build for these targets, and if I am not mistaken the compiler error will also be quite specific about what has to be done. 

Thus, maybe additional documentation in a plumbing crate of `gitoxide` wouldn't be read anyway.

EliahKagan:
> Maybe this could be as 'easy' as going with this - it works like it worked before.

I don't view this as working the same as it did before, because:

- Without the changes in this PR, whoever builds something with `gix-diff` and targets `wasm32-unknown-unknown` need only enable the `wasm` feature of `gix-diff` to get the `getrandom` dependency to be present and attempt to use the JavaScript backend.
- In contrast, with the changes in this PR, whoever does that must both enable the `wasm` feature of `gix-diff` and also cause `--cfg getrandom_backend="wasm_js"` to be included in rustflags.

However, it is true that I've avoided changing how it works as much as possible (or at least as much as I know how). If you think that's sufficient, I'd be happy to merge this. It seems to me that something should probably also be done to indicate the change. To elaborate on my previous thinking:

- This is technically breaking. Most users of `gix-diff` won't be broken, simply because most aren't on `wasm32-unknown-unknown`, but perhaps a few are, so we should probably make sure this doesn't just bump the patch version number. Although we rarely use it in this repository, I recall that `chore!:` was recognized properly by csr when used fairly recently. I could amend and reword one of the messages to include that. (`change!:` or `feat!:` would also be reasonable.)
- Documenting it probably doesn't need to be done in this PR, though, since if a release ends up being made without a documentation change (and users don't see the changelog entry), then the error messages will still guide the way to some extent: they describe the need for the configuration flag and include the URL of the `getrandom` documentation that covers it.

> Without the `wasm_js` Cargo feature I suppose it wouldn't work, but would it work without that feature and without the RUSTFLAGS setting? If that was the case, that might be preferred as it leaves everything to the caller.

Omitting both--like omitting one or the other--will break the build on `wasm32-unknown-unknown`, unless the `getrandom` dependency is removed (or removed for that target). However, if the dependency is removed, then I am not sure what the `wasm` feature of `gix-diff` would represent.

.github/workflows/ci.yml
EliahKagan: The `gix-diff` docstring in `lib.rs` is not visible when looking at the `diff` module of `gix` (`diff` pulls in the contents of `gix-diff` with `pub use gix_diff::*;` but also has its own items). So maybe often not? But people would still search for such documentation. I can try to document it in some way, if we go forward with the approach currently implemented here.

Byron:
I noticed something:

```
❯ cargo tree -i getrandom -p gix-diff
getrandom v0.3.2
└── tempfile v3.20.0
    └── gix-tempfile v18.0.0 (/Users/byron/dev/github.com/GitoxideLabs/gitoxide/gix-tempfile)
        ├── gix-diff v0.53.0 (/Users/byron/dev/github.com/GitoxideLabs/gitoxide/gix-diff)
        └── gix-lock v18.0.0 (/Users/byron/dev/github.com/GitoxideLabs/gitoxide/gix-lock)
            └── gix-index v0.41.0 (/Users/byron/dev/github.com/GitoxideLabs/gitoxide/gix-index)
                ├── gix-diff v0.53.0 (/Users/byron/dev/github.com/GitoxideLabs/gitoxide/gix-diff)
                └── gix-worktree v0.42.0 (/Users/byron/dev/github.com/GitoxideLabs/gitoxide/gix-worktree)
                    └── gix-diff v0.53.0 (/Users/byron/dev/github.com/GitoxideLabs/gitoxide/gix-diff)
```

I wonder why `gix-diff` got the `wasm` feature just to configure `getrandom`. Why not any of the other crates in the dependency chain?

It feels like I tried to make something work, and it did work because it was only a Cargo feature. Now `getrandom` explicitly discourages it, which will probably tell us as much as: don't offer `wasm` here.

If this makes sense, this should be a breaking change to indicate the dependency gets removed, and that applications need to configure it entirely themselves. It's a hurdle, but there seems to be no way around it, while also I think what we have now is obviously confusing to us, and more like a hack to make it easier for downstream. But the motivation for the hack, to make it easy, can't be fulfilled anymore.

.github/workflows/ci.yml
Byron: If you find a good spot for documentation, I think it's absolutely fair to add it. From what I wrote below it seems though that `getrandom` has nothing to do with this crate, they are only tangentially related.

EliahKagan:
Thanks--I had sort of assumed that we did need `getrandom`, but I did wonder what the reason was, given that there are no direct references to `getrandom`--nor to `wasm`--within the code of `gix-diff` itself. This relates to something I mentioned in https://github.com/GitoxideLabs/gitoxide/issues/2092#issuecomment-3146267544 but then never followed up on any further:

> I'm not clear on what is using `getrandom` when it is a dependency of `gix-diff`, and what happens if it is absent. After all, building for `wasm32-unknown-unknown` *does* succeed with the `wasm` feature and thus without `getrandom`. (This doesn't just omit the `js` feature. It omits `getrandom`, as far as I can tell.)
> 
> My guess is that something is degrading to using a lower quality random number generator in its absence. (Or is it a separately provided `getrandom` dependency, somehow?) I would want to understand that before forming a confident opinion about what specific change should be made when adapting to `getrandom` 0.3.\*, and also whether it's important or not to document anything associated with that.

I had meant to say:

> After all, building for `wasm32-unknown-unknown` *does* succeed ***without*** the `wasm` feature and thus without `getrandom`.

Is the `wasm` feature of `gix-diff`, and its dependency on `getrandom`, just [for `tempfile`](https://github.com/Stebalien/tempfile/blob/0657fdf24925287e6d5544d9e321a48d7ceaafdc/Cargo.toml#L24-L25)? If so, what bad things happen when the `wasm` feature is not enabled on a WASM target? Do operations that involve generating names for temporary files fail at runtime when attempted, even after a successful build? Is temporary file name generation more brittle?

> If this makes sense, this should be a breaking change to indicate the dependency gets removed, and that applications need to configure it entirely themselves.

I don't think I really understand how it has to be configured. If I remove the `wasm` feature of `gix-diff` and its associated direct dependency on `tempfile`, then the current WASM jobs on CI will succeed just by no longer specifying the removed `wasm` feature.

So I will need some way of testing for the functionality that requires configuring a JavaScript backend for `getrandom` on some WASM targets. Of course, I will also need that if I am to write any useful documentation about what people are supposed to do.

---

Taking a look at:

```text
C:\Users\ek\source\repos\gitoxide\gix-diff [main ≡]> git grep -Fn tempfile -- '*.rs'
src/blob/pipeline.rs:228:    /// As these files are ultimately named tempfiles, they will be leaked unless the [gix_tempfile] is configured with
src/blob/pipeline.rs:425:                                        gix_tempfile::new(
src/blob/pipeline.rs:427:                                            gix_tempfile::ContainingDirectory::Exists,
src/blob/pipeline.rs:428:                                            gix_tempfile::AutoRemove::Tempfile,
src/blob/platform.rs:310:        #[error("Could not write content of '{rela_path}' to tempfile for passing to external diff command")]
src/blob/platform.rs:320:        /// Possibly a tempfile to be removed after the run, or `None` if there is no old version.
src/blob/platform.rs:321:        pub(crate) old: Option<gix_tempfile::Handle<gix_tempfile::handle::Closed>>,
src/blob/platform.rs:322:        /// Possibly a tempfile to be removed after the run, or `None` if there is no new version.
src/blob/platform.rs:323:        pub(crate) new: Option<gix_tempfile::Handle<gix_tempfile::handle::Closed>>,
src/blob/platform.rs:431:    // TODO: fix this - the diff shouldn't fail if binary (or large) files are used, just copy them into tempfiles.
src/blob/platform.rs:442:        ) -> Result<Option<gix_tempfile::Handle<gix_tempfile::handle::Closed>>, prepare_diff_command::Error> {
src/blob/platform.rs:449:                    let mut tmp = gix_tempfile::new(
src/blob/platform.rs:451:                        gix_tempfile::ContainingDirectory::Exists,
src/blob/platform.rs:452:                        gix_tempfile::AutoRemove::Tempfile,
tests/diff/blob/pipeline.rs:21:            let tmp = gix_testtools::tempfile::TempDir::new()?;
tests/diff/blob/pipeline.rs:98:        let tmp = gix_testtools::tempfile::TempDir::new()?;
tests/diff/blob/pipeline.rs:153:        let tmp = gix_testtools::tempfile::TempDir::new()?;
tests/diff/blob/pipeline.rs:236:        let tmp = gix_testtools::tempfile::TempDir::new()?;
tests/diff/blob/pipeline.rs:306:        let tmp = gix_testtools::tempfile::TempDir::new()?;
tests/diff/blob/pipeline.rs:440:        let tmp = gix_testtools::tempfile::TempDir::new()?;
```

`gix-diff` does not use `tempfile` directly, though it *almost* uses it directly in its test suite, which uses `tempfile` as [republished](https://github.com/GitoxideLabs/gitoxide/blob/202bc6da79854d1fb6bb32b9c6bb2a6f882c77f5/tests/tools/src/lib.rs#L28) by `gix-testtools`. I think the far more important usage by `gix-diff` of `tempfile` is through `gix_tempfile`.

`gix-tempfile` itself depends on `tempfile` without omitting default features:

https://github.com/GitoxideLabs/gitoxide/blob/202bc6da79854d1fb6bb32b9c6bb2a6f882c77f5/gix-tempfile/Cargo.toml#L38

That transitively depends on `getrandom` on most targets. I say this based on [this dependency](https://github.com/Stebalien/tempfile/blob/0657fdf24925287e6d5544d9e321a48d7ceaafdc/Cargo.toml#L24-L25) and [the feature list](https://github.com/Stebalien/tempfile/blob/0657fdf24925287e6d5544d9e321a48d7ceaafdc/Cargo.toml#L40-L44) of `tempfile`.

Byron:
> Is the `wasm` feature of `gix-diff`, and its dependency on `getrandom`, just [for `tempfile`](https://github.com/Stebalien/tempfile/blob/0657fdf24925287e6d5544d9e321a48d7ceaafdc/Cargo.toml#L24-L25)? If so, what bad things happen when the `wasm` feature is not enabled on a WASM target? Do operations that involve generating names for temporary files fail at runtime when attempted, even after a successful build? Is temporary file name generation more brittle?

I don't know actually, I'd expect it to fail building entirely. It's worth noting that `getrandom` has no mention whatsoever in `gix-diff` and really is only used in `gix-tempfile`

> I don't think I really understand how it has to be configured. If I remove the `wasm` feature of `gix-diff` and its associated direct dependency on `tempfile`, then the current WASM jobs on CI will succeed just by no longer specifying the removed `wasm` feature.

That's great, I wasn't even aware that it could build without the `wasm` feature and always thought it was just providing it for convenience to users. In that case, I definitely think it's OK to remove `wasm` and leave the configuration of `getrandom` to the downstream entirely, particularly now that v0.3 requires setting rustc-flags.
This definitely would be a breaking change, but one worth making.

If I had better memory I'd probably know why `wasm` was added in the first place, but from this investigation I don't see why it would have been more than a convenience.

EliahKagan:
> That's great, I wasn't even aware that it could build without the `wasm` feature and always thought it was just providing it for convenience to users. In that case, I definitely think it's OK to remove `wasm` and leave the configuration of `getrandom` to the downstream entirely, particularly now that v0.3 requires setting rustc-flags. This definitely would be a breaking change, but one worth making.

> If I had better memory I'd probably know why `wasm` was added in the first place, but from this investigation I don't see why it would have been more than a convenience.

This looks to be entirely for the benefit of `gix-pack`, though in practice a crate outside of this workspace might make use of `gix-diff` in the same way `gix-pack` does. The most relevant changes are from f0e40ec and 6c4c196, both in #735.

The default features of `gix-pack` include `streaming-input`:

https://github.com/GitoxideLabs/gitoxide/blob/1dfc31287812dce25e182c5d79944973d830210f/gix-pack/Cargo.toml#L18-L19

The `streaming-input` enables the `gix-tempfile` dependency:

https://github.com/GitoxideLabs/gitoxide/blob/1dfc31287812dce25e182c5d79944973d830210f/gix-pack/Cargo.toml#L22-L23

`gix-tempfile` is an optional dependency. Also, it is not a dependency at all on targets whose architecture is `wasm32` (which includes `wasm32-unknown-unknown`):

https://github.com/GitoxideLabs/gitoxide/blob/1dfc31287812dce25e182c5d79944973d830210f/gix-pack/Cargo.toml#L62-L63

`gix-pack` also has a `wasm` feature, which enables the `wasm` feature of the `gix-diff` dependency:

https://github.com/GitoxideLabs/gitoxide/blob/1dfc31287812dce25e182c5d79944973d830210f/gix-pack/Cargo.toml#L33-L34

Attempting to build `gix-pack` with default features enabled, and with its `wasm` feature not enabled, fails due to the unavailability of `gix-tempfile` (this is on the `main` branch, not this feature branch, because my goal here is to figure out what this feature is for):

```text
ek in 🌐 catenary in gitoxide on  main is 📦 v0.45.0 via 🦀 v1.89.0
❯ cargo build -p gix-pack --target wasm32-unknown-unknown
   Compiling gix-pack v0.60.0 (/home/ek/repos/gitoxide/gix-pack)
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `gix_tempfile`
 --> gix-pack/src/bundle/write/error.rs:3:5
  |
3 | use gix_tempfile::handle::Writable;
  |     ^^^^^^^^^^^^ use of unresolved module or unlinked crate `gix_tempfile`
  |
  = help: if you wanted to use a crate named `gix_tempfile`, use `cargo add gix_tempfile` to add it to your `Cargo.toml`

error[E0433]: failed to resolve: use of unresolved module or unlinked crate `gix_tempfile`
 --> gix-pack/src/bundle/write/types.rs:3:5
  |
3 | use gix_tempfile::handle::Writable;
  |     ^^^^^^^^^^^^ use of unresolved module or unlinked crate `gix_tempfile`
  |
  = help: if you wanted to use a crate named `gix_tempfile`, use `cargo add gix_tempfile` to add it to your `Cargo.toml`

error[E0432]: unresolved import `gix_tempfile`
  --> gix-pack/src/bundle/write/mod.rs:10:5
   |
10 | use gix_tempfile::{AutoRemove, ContainingDirectory};
   |     ^^^^^^^^^^^^ use of unresolved module or unlinked crate `gix_tempfile`
   |
   = help: if you wanted to use a crate named `gix_tempfile`, use `cargo add gix_tempfile` to add it to your `Cargo.toml`

error[E0433]: failed to resolve: use of unresolved module or unlinked crate `gix_tempfile`
  --> gix-pack/src/bundle/write/error.rs:14:21
   |
14 |     Persist(#[from] gix_tempfile::handle::persist::Error<Writable>),
   |                     ^^^^^^^^^^^^ use of unresolved module or unlinked crate `gix_tempfile`
   |
   = help: if you wanted to use a crate named `gix_tempfile`, use `cargo add gix_tempfile` to add it to your `Cargo.toml`

error[E0433]: failed to resolve: use of unresolved module or unlinked crate `gix_tempfile`
  --> gix-pack/src/bundle/write/mod.rs:82:36
   |
82 | ...   Some(directory) => gix_tempfile::new(directory, ContainingDirectory::Exists, AutoRemove::Tempfile)?,
   |                          ^^^^^^^^^^^^ use of unresolved module or unlinked crate `gix_tempfile`
   |
   = help: if you wanted to use a crate named `gix_tempfile`, use `cargo add gix_tempfile` to add it to your `Cargo.toml`

error[E0433]: failed to resolve: use of unresolved module or unlinked crate `gix_tempfile`
  --> gix-pack/src/bundle/write/mod.rs:83:25
   |
83 | ...   None => gix_tempfile::new(std::env::temp_dir(), ContainingDirectory::Exists, AutoRemove::Tempfile)?,
   |               ^^^^^^^^^^^^ use of unresolved module or unlinked crate `gix_tempfile`
   |
   = help: if you wanted to use a crate named `gix_tempfile`, use `cargo add gix_tempfile` to add it to your `Cargo.toml`

error[E0433]: failed to resolve: use of unresolved module or unlinked crate `gix_tempfile`
   --> gix-pack/src/bundle/write/mod.rs:189:32
    |
189 |             Some(directory) => gix_tempfile::new(directory, ContainingDirectory::Exists, AutoRemove::Tempfile)?,
    |                                ^^^^^^^^^^^^ use of unresolved module or unlinked crate `gix_tempfile`
    |
    = help: if you wanted to use a crate named `gix_tempfile`, use `cargo add gix_tempfile` to add it to your `Cargo.toml`

error[E0433]: failed to resolve: use of unresolved module or unlinked crate `gix_tempfile`
   --> gix-pack/src/bundle/write/mod.rs:190:21
    |
190 |             None => gix_tempfile::new(std::env::temp_dir(), ContainingDirectory::Exists, AutoRemove::Tempfile)?,
    |                     ^^^^^^^^^^^^ use of unresolved module or unlinked crate `gix_tempfile`
    |
    = help: if you wanted to use a crate named `gix_tempfile`, use `cargo add gix_tempfile` to add it to your `Cargo.toml`

error[E0433]: failed to resolve: use of unresolved module or unlinked crate `gix_tempfile`
   --> gix-pack/src/bundle/write/mod.rs:285:38
    |
285 | ...   let mut index_file = gix_tempfile::new(directory, ContainingDirectory::Exists, AutoRemove::Tempfile)?;
    |                            ^^^^^^^^^^^^ use of unresolved module or unlinked crate `gix_tempfile`
    |
    = help: if you wanted to use a crate named `gix_tempfile`, use `cargo add gix_tempfile` to add it to your `Cargo.toml`

error[E0433]: failed to resolve: use of unresolved module or unlinked crate `gix_tempfile`
  --> gix-pack/src/bundle/write/types.rs:65:76
   |
65 | pub(crate) type SharedTempFile = Arc<parking_lot::Mutex<std::io::BufWriter<gix_tempfile::Handle<Writable>>>>;
   |                                                                            ^^^^^^^^^^^^ use of unresolved module or unlinked crate `gix_tempfile`
   |
   = help: if you wanted to use a crate named `gix_tempfile`, use `cargo add gix_tempfile` to add it to your `Cargo.toml`

Some errors have detailed explanations: E0432, E0433.
For more information about an error, try `rustc --explain E0432`.
error: could not compile `gix-pack` (lib) due to 10 previous errors
```

While adding `--features wasm` causes it to succeed (also on the `main` branch):

```text
ek in 🌐 catenary in gitoxide on  main is 📦 v0.45.0 via 🦀 v1.89.0
❯ cargo build -p gix-pack --target wasm32-unknown-unknown --features wasm
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.13s
```

(I've shown shorter build output than usual, from builds where everything was built before already. But I have also verified the erroring and successful builds with a clean target directory.)

It may be that the added testing in adc3a41429ee0084e097b79b32ad5162fa18c55f actually obscures what's important, which is whether or not--and also how--`gix-pack` works on WASM, especially on `wasm32-unknown-unknown`.

Byron:
Thanks for taking another look.

Most definitely I am having a slow day, and that's the reason that I don't follow.

`gix-diff` is genuinely needed for the `generate` feature. What doesn't make sense to me is how `gix-tempile` is only used on non-wasm32 targets.
This means when building for `wasm32` it doesn't actually pull in the `gix-tempfile` dependency despite needed by the `streaming-input` feature. Yes, it compiles.
The only sane way this works is that it doesn't build the respective feature:

https://github.com/GitoxideLabs/gitoxide/blob/beba7204a50a84b30e3eb81413d968920599e226/gix-pack/src/bundle/mod.rs#L6-L7

So it's a bit of a split-brain problem that in code the `wasm` feature is used to detect the wasm target, but in the Cargo manifest the target is used.

With this patch it works in all the cases:

```patch
diff --git a/gix-pack/src/bundle/mod.rs b/gix-pack/src/bundle/mod.rs
index d8ef1107d7..d2e691820d 100644
--- a/gix-pack/src/bundle/mod.rs
+++ b/gix-pack/src/bundle/mod.rs
@@ -3,7 +3,7 @@ pub mod init;
 
 mod find;
 ///
-#[cfg(all(not(feature = "wasm"), feature = "streaming-input"))]
+#[cfg(all(not(target_arch = "wasm32"), feature = "streaming-input"))]
 pub mod write;
 
 ///
```

```
gitoxide ( main) [$]
❯ cargo build -p gix-pack --target wasm32-unknown-unknown
   Compiling gix-pack v0.60.0 (/Users/byron/dev/github.com/GitoxideLabs/gitoxide/gix-pack)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.72s

gitoxide ( main) +1 -1 [$!]
❯ cargo build -p gix-pack --target wasm32-unknown-unknown --features wasm
   Compiling gix-pack v0.60.0 (/Users/byron/dev/github.com/GitoxideLabs/gitoxide/gix-pack)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.33s
```

This would also mean that the `wasm` feature in `gix-pack` can be completely removed.

EliahKagan:
> https://github.com/GitoxideLabs/gitoxide/blob/beba7204a50a84b30e3eb81413d968920599e226/gix-pack/src/bundle/mod.rs#L6-L7
> 
> So it's a bit of a split-brain problem that in code the `wasm` feature is used to detect the wasm target, but in the Cargo manifest the target is used.

Thanks--I was not looking/thinking about the condition there!

> With this patch it works in all the cases:
> 
> *[...]*
> 
> This would also mean that the `wasm` feature in `gix-pack` can be completely removed.

I've been focusing on `wasm32-unknown-unknown` because that's where the build on CI breaks when bumping `getrandom` from 0.2 to 0.3. But I am wondering whether the approach shown in https://github.com/GitoxideLabs/gitoxide/pull/2093#issuecomment-3171844374 will do the right thing for other WASM targets, and also to what extent that should be a factor in the design here.

Currently the only other we attempt to build on CI is `wasm32-wasip1`. But other WASM targets have more of `std` available, and most of them--I think including `wasm32-wasip1`--can support `getrandom` 0.3.\* without requiring custom rustflags. This is my understanding from the mentions of various `wasm32` targets in the [Supported targets](https://docs.rs/getrandom/latest/getrandom/#supported-targets) and [Opt-in backends](https://docs.rs/getrandom/latest/getrandom/#opt-in-backends) tables in the `getrandom` documentation. More broadly, at least one `wasm32` target, [`wasm32-wasip2`](https://doc.rust-lang.org/rustc/platform-support/wasm32-wasip2.html#wasm32-wasip2), is [documented](https://doc.rust-lang.org/rustc/platform-support/wasm32-wasip2.html#requirements) to have full `std` support.

I think those others will also all match a check for `target_arch = "wasm32"`. Should there be a way to let the `streaming-input` feature (attempt to) build on them?

Byron:
To me it's all a bit confusing as WASM is this illusive target that exists in various flavors, and it's unclear which one is going to make it in the long run. On top of that, I don't know how downstream is using any of this, if at all.

So if there is a way to remove all additional `wasm` related complexities then I'd go for that. Additionally, I'd be perfectly fine to make `gix-pack` work by not providing a certain feature in the WASM target, and would wait for demand to be signalled by an issue for instance.

With the patch above, would be not be possible to remove all `wasm` features, along with the `getrandom` dependency in `gix-diff`?

EliahKagan:
> With the patch above, would be not be possible to remove all `wasm` features, along with the `getrandom` dependency in `gix-diff`?

I think I will need to expand testing of WASM to cover WASM targets where it is more fully functional, either expanding installation testing or, better, running parts of the test suite. That way:

- If the reason it doesn't seem to me that we can do any of these simple things is due to my own misunderstanding of something, then I'll discover it through such experimentation.
- If something more substantive is really required, then I'll have made progress toward it and also clarified it.

Moreover, even if this were very simple, I would still not likely get to it for a while, and as noted in https://github.com/GitoxideLabs/gitoxide/pull/2145#pullrequestreview-3171237101, the error described in #2092 when Dependabot encounters it is making Dependabot logs harder to understand, and possibly (though I'm not sure) making Dependabot sometimes take longer such that it fails. I may therefore hold back `getrandom` to SemVer-compatible versions as a temporary measure.

Since this PR adds testing and does little else, but the testing it adds actually obscures the problem because it keeps CI from getting to the `gix-path` tests where the conceptual difficulties arise, I may just retitle this PR and force-push something that modifies `dependabot.yml` accordingly. The discussion here will, of course, remain for whenever I (or anyone else interested) can pick it up.

Byron:
Alright then!

Let me restate that to me it seems the current way of doing things is overly complex, and that I had a patch somewhere that would remove `getrandom` from `gix-diff` entirely to push the burden of dealing with this downstream, which is breaking but also fine. Adding `getrandom` to `gix-diff` was a hack to help downstream to configure for `wasm`, but it's nothing that should remain.

EliahKagan:
> I had a patch somewhere that would remove `getrandom` from `gix-diff` entirely to push the burden of dealing with this downstream, which is breaking but also fine.

If such a patch is ready (or is found, or someone proposes an equivalent) and you think it's a good idea then I definitely have no objection to it being merged at any time.

I'll also try to look at this again tomorrow (or otherwise soon) to see if I'm overthinking it.

EliahKagan:
*Update:* When working on #2200, I've noticed that running `cargo-update` on a Dependabot version update PR (the test PR https://github.com/EliahKagan/gitoxide/pull/111) is [upgrading more things than Dependabot does](https://gist.github.com/EliahKagan/b9e39084a1c5b25b48245aa3ccb1ffb6). Since Dependabot should've been able to do those upgrades, that's strange. I'm going to hold back `getrandom` in case the inability to upgrade it is blocking bumping some transitive dependencies due to limitations of Dependabot. I do hope to make a better improvement here eventuallty. See also https://github.com/GitoxideLabs/gitoxide/pull/2200#issuecomment-3361407300 for context.

EliahKagan:
For now, this is superseded by #2201, which includes holding `getrandom` back. I don't think this speeds Dependabot up, or at least not much, but it at least makes its output easier to understand.

Unlike holding back `expectrl`, which may or may not be temporary (https://github.com/GitoxideLabs/gitoxide/pull/2200#issuecomment-3359015016), holding back `getrandom` is intended to be temporary. However, the actual changes made in this PR so far aren't what we'll want to CI to test the crates that use `getrandom`, since they focus on a crate that doesn't really use it rather than one that does.

Whoever resumes work on this (for #2092), which might be me, can do so by making a new PR or (especially if it's me) by reopening this PR, but for now I'm closing it. I am of course *not* closing issue #2092.

Issue #2089, opened 2025-07-31
Allow usage of rusqlite unbundled feature


svenstaro:
### Summary 💡

We use both `sqlx` and `gitoxide` in the same project. Or rather, we're trying to use this but it fails with:
```
error: failed to select a version for `libsqlite3-sys`.
    ... required by package `rusqlite v0.36.0`
    ... which satisfies dependency `rusqlite = "^0.36.0"` of package `gitoxide-core v0.48.0`
    ... which satisfies dependency `gitoxide-core = "^0.48.0"` of package `gitoxide v0.45.0`
...
versions that meet the requirements `^0.34.0` are: 0.34.0

package `libsqlite3-sys` links to the native library `sqlite3`, but it conflicts with a previous package which links to `sqlite3` as well:
package `libsqlite3-sys v0.30.1`
    ... which satisfies dependency `libsqlite3-sys = "^0.30.1"` of package `sqlx-sqlite v0.8.6`
    ... which satisfies dependency `sqlx-sqlite = "=0.8.6"` of package `sqlx v0.8.6`
...
Only one package in the dependency graph may specify the same links value. This helps ensure that only one copy of a native library is linked in the final binary. Try to adjust your dependencies so that only one package uses the `links = "sqlite3"` value. For more information, see https://doc.rust-lang.org/cargo/reference/resolver.html#links.

failed to select a version for `libsqlite3-sys` which could resolve this conflict
```
The problem is that this is a statically linked C dependency and both `sqlx` and `gitoxide` try to pull in the bundled version of sqlite by default.

However, this has a somewhat easy fix in that `rusqlite` supports the `unbundled` more where it will try to find a system libsqlite3 and will merely provide the bindings to that library. [sqlx already supports using this](https://github.com/launchbadge/sqlx/blob/bab1b022bd56a64f9a08b46b36b97c5cff19d77e/sqlx-sqlite/Cargo.toml#L29) but [gitoxide-core hardcodes it to bundled](https://github.com/GitoxideLabs/gitoxide/blob/8892cb18ccf4606df97181e9bc43461d1c71a28a/gitoxide-core/Cargo.toml#L82). I think it would be great if the gitoxide feature flags allowed for using the system-installed version via the `rusqlite` `unbundled` feature.

### Motivation 🔦

Allow for using multiple libraries that depend on sqlite in a single Rust project.

Skgland:
[...]

PR #2096, opened 2025-08-03
Be more explicit about how gitoxide should be used

svenstaro:
This is as per #2089 where suggested to mention the library more prominently.


Byron (approved):

I did take the liberty to significantly modify the text, which remains in spirit at least and hopefully still solves the original problem.

svenstaro:
> Thanks a lot for your help with this!
> 
> I did take the liberty to significantly modify the text, which remains in spirit at least and hopefully still solves the original problem.

That's alright so long as the core message is there :)

PR #2102, opened 2025-08-04
Extract UnifiedDiffSink

cruessler:
This is a draft PR. I’ll add more context as soon as I get to it.

### DIFF
--- a/gix-diff/src/blob/mod.rs
+++ b/gix-diff/src/blob/mod.rs
@@ -12,7 +12,7 @@ pub mod pipeline;
 pub mod platform;
 
 pub mod unified_diff;
-pub use unified_diff::_impl::UnifiedDiff;
+pub use unified_diff::_impl::{UnifiedDiff, UnifiedDiffSink};
 
 /// Information about the diff performed to detect similarity.
 #[derive(Debug, Default, Clone, Copy, PartialEq, PartialOrd)]
--- a/gix-diff/src/blob/unified_diff.rs
+++ b/gix-diff/src/blob/unified_diff.rs
@@ -26,6 +26,35 @@ impl ContextSize {
     }
 }
 
+/// Represents the type of a line in a unified diff.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum DiffLineType {
+    /// A line that exists in both old and new versions (context line).
+    Context,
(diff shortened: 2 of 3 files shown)


cruessler:
I’m going to mark this PR as ready for review in order to indicate that I’m now done with the initial version and would welcome a first round of feedback!

There’s a few genuine `TODO:` comments in places where I had questions myself. Feel free to mark this PR as draft in case you don’t think it’s ready yet!

cruessler:
I’ve now updated the PR.

- The API is a bit more verbose as the responsibility for handling newlines was moved to the consumer. This had become necessary as `HunkHeader::to_string()` doesn’t include any newline at all. For consistency, newlines for the actual lines in a diff have to be handled by the consumer as well.
- I did not manage to move `buffer2` to `UnifiedDiff` as I ran into lifetime and ownership issues, but I might have missed something. Please check carefully! :-)

Byron:
Apologies for the late response!

In any case, I think this is the way to go. Pushing newline handling further down seems like perfectly reasonable, the new 'abstract' version doesn't need to know about that, after all it has a header, it has each line in a buffer, that's all that it needs.

Regarding the lifetime issues with the buffer, I see why would that be the case. There is this hidden lifetime in `T` that one would want there. In theory, it should be possible, practically I'd also have trouble expressing this, even though I also think that it *should* be possible, maybe with more generics, maybe with an explicit lifetime that is added as constraint and can be used with the buffer, like `buffer: Vec<(Type, &'interned [u8])>`.

I set this PR back to draft, but invite you to change it back to what it is now to clearly indicate when it's ready for review.

cruessler:
The lifetime issue fortunately was much easier to solve than expected. I removed the intermediate `buffer2` and changed `self.buffer` to what you suggested, and that was already sufficient. The intermediate `buffer2` was only needed by the previous version because the previous version had to add newlines and that required a separate buffer.

Once CI passes, this will be ready for review!

Byron (approved):
Thanks a lot!

In my refactor, I mostly added another layer to help porting existing implementations that rely on ready-made hunks and headers, and removed the newline handling entirely from the base type. The refactor into a directory makes the diff terrible to read though, sorry for that.

Lastly, `DiffFileType` is now `DiffFileKind`, which might affect you.

PR #2105, opened 2025-08-04
fix(gix): extend lifetime of iterators

### LINKED ISSUE #2103: `gix` library returning too short of a lifetime
### Current behavior 😯

Trying to filter tags in a separate function fails currently as the lifetime is too short:

```rust
fn latest_tag(repo: &gix::Repository) -> gix::Reference {
    let refs = repo.references().unwrap();
    refs.tags() // .tags() borrows from `refs` and thus this fails to compile
        .unwrap()
        .filter_map(|tag| tag.ok())
        .map(|tag| (tag, semver::Version::parse(tag.name().shorten().to_str().unwrap())))
        .max_by_key(|(_tag, version)| version.clone())
        .map(|(tag, _version)| tag)
}
```

### Expected behavior 🤔

`.tags()` should instead borrow from `repo` and successfully compile.

I believe the issue is here: https://github.com/GitoxideLabs/gitoxide/blob/202bc6da79854d1fb6bb32b9c6bb2a6f882c77f5/gix/src/reference/iter.rs#L58-L60

`tags` should return `Result<Iter<'repo>, init::Error>` instead as otherwise it borrows from `&self`.

If this is desirable, I would be willing to contribute the fix (for tags and also the surrounding functions that look to have the same issue).

### Git behavior

Not applicable

### Steps to reproduce 🕹

Try to compile the following code:

```toml
# Cargo.toml
[package]
name = "gix-describe"
version = "0.1.0"
edition = "2024"

[dependencies]
gix = { version = "0.73.0", default-features = false, features = ["max-performance", "basic", "status"] }
semver = "1.0.26"
```

```rust
// src/main.rs
fn latest_tag(repo: &gix::Repository) -> gix::Reference {
    let refs = repo.references().unwrap();
    refs.tags() // .tags() borrows from `refs` and thus this fails to compile
        .unwrap()
        .filter_map(|tag| tag.ok())
        .map(|tag| (tag, semver::Version::parse(tag.name().shorten().to_str().unwrap())))
        .max_by_key(|(_tag, version)| version.clone())
        .map(|(tag, _version)| tag)
}

fn main() {}
```

jalil-salame:
Previously the iterators would return references with lifetimes that were shorter than the actual lifetimes of the `gix::Reference` themselves. This was dues to a footgun with `'_` (eliding the lifetime).

When a function returns an elided lifetime, this lifetime usually is the lifetime of the `&self` parameter, but sometimes it is the lifetime of the type itself (e.g. `Iter<'_>`).

I made the lifetimes explicit to ensure we were using the correct ones.

Closes #2103

### DIFF
--- a/gix-ref/src/store/file/overlay_iter.rs
+++ b/gix-ref/src/store/file/overlay_iter.rs
@@ -189,11 +189,11 @@ impl Iterator for LooseThenPacked<'_, '_> {
     }
 }
 
-impl Platform<'_> {
+impl<'repo> Platform<'repo> {
     /// Return an iterator over all references, loose or packed, sorted by their name.
     ///
     /// Errors are returned similarly to what would happen when loose and packed refs were iterated by themselves.
-    pub fn all(&self) -> std::io::Result<LooseThenPacked<'_, '_>> {
+    pub fn all<'p>(&'p self) -> std::io::Result<LooseThenPacked<'p, 'repo>> {
         self.store.iter_packed(self.packed.as_ref().map(|b| &***b))
     }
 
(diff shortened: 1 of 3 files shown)


Byron (approved):
Thanks a lot for helping with this!

The lifetime-part of my refactor you may find interesting - I hope that doesn't break it (again).
If it does, let's add another test and fix it for good.

PR #2106, opened 2025-08-07
Add `gix::open_with_environment_overrides`

cruessler:
Based on this suggestion: https://github.com/gitui-org/gitui/pull/2664#discussion_r2258725560.

### DIFF
--- a/gitoxide-core/src/repository/tag.rs
+++ b/gitoxide-core/src/repository/tag.rs
@@ -1,5 +1,7 @@
 use gix::bstr::{BStr, BString, ByteSlice};
 
+use crate::OutputFormat;
+
 #[derive(Eq, PartialEq, PartialOrd, Ord)]
 enum VersionPart {
     String(BString),
@@ -41,7 +43,11 @@ impl Version {
     }
 }
 
-pub fn list(repo: gix::Repository, out: &mut dyn std::io::Write) -> anyhow::Result<()> {
+pub fn list(repo: gix::Repository, out: &mut dyn std::io::Write, format: OutputFormat) -> anyhow::Result<()> {
+    if format != OutputFormat::Human {
+        anyhow::bail!("JSON output isn't supported");
+    }
+
     let platform = repo.references()?;
 
     let mut tags: Vec<_> = platform
--- a/gix/src/discover.rs
+++ b/gix/src/discover.rs
@@ -3,7 +3,7 @@ use std::path::Path;
 
 pub use gix_discover::*;
 
-use crate::{bstr::BString, Repository, ThreadSafeRepository};
+use crate::{bstr::BString, ThreadSafeRepository};
 
 /// The error returned by [`crate::discover()`].
 #[derive(Debug, thiserror::Error)]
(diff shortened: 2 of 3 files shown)


Byron:
Thanks for kicking this off.

I turned the PR back into draft mode as it seems like it was generated by an agent which can't run `cargo` locally.

And thinking about it, I think I completely failed to see that `open_with_environment_overrides()` already exists, so it should just be used in `gitui`. There probably is no need for this to be top-level, but I don't object to it either.

Can be closed, can be fixed, it's up to you.

cruessler:
> Thanks for kicking this off.
> 
> I turned the PR back into draft mode as it seems like it was generated by an agent which can't run `cargo` locally.

Actually, the agent was me, just copying `gix::discover` without running any checks, not taking into account that parameters and return type needed to be adapted. :-)

> And thinking about it, I think I completely failed to see that `open_with_environment_overrides()` already exists, so it should just be used in `gitui`. There probably is no need for this to be top-level, but I don't object to it either.

I find the addition of `gix::discover_with_environment_overrides` as well as `gix::open_with_environment_overrides` useful as I remember my confusion when there was only `gix::discover` in the docs, but no obvious way of doing the same with environment overrides. Or am I missing something? If so, I could extend the docs so that it becomes clearer which function to use for which use case.

Byron:
> Actually, the agent was me, just copying `gix::discover` without running any checks, not taking into account that parameters and return type needed to be adapted. :-)

I admit I am kind of paranoid these days 😅.

> I find the addition of `gix::discover_with_environment_overrides` as well as `gix::open_with_environment_overrides` useful as I remember my confusion when there was only `gix::discover` in the docs,

That's all I need to think these top-level functions really should be there. And it would certainly be nice if they could at least have the gist of why one would want to use them in their own docs before referring to the respective docs in `ThreadSafeRepository`.

EliahKagan:
It looks like the remaining CI failures are not due to code that is new in this PR.

1. [The lint failure](https://github.com/GitoxideLabs/gitoxide/actions/runs/16823918370/job/47656067995?pr=2106#step:5:973) is due to an improved dead code diagnostic in Rust 1.89; the error is fixed in #2108.

2. The [`test-fast`](https://github.com/GitoxideLabs/gitoxide/actions/runs/16823918370/job/47656067966?pr=2106) failure is due to a problem on the runner itself:

    > The hosted runner lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.

If the feature branch here is rebased onto the current tip of the main branch, then I expect (1) to pass because the fix is integrated into main, and (2) to pass because it looks like a failure with low probability that doesn't require any change to fix.

(I have refrained from making those changes myself and merging this PR, in case there are reasons unrelated to those CI failures that this PR has not yet been marked back from draft.)

Byron:
Thanks a lot for your help @EliahKagan , I was a bit slow yesterday.

cruessler:
> And it would certainly be nice if they could at least have the gist of why one would want to use them in their own docs before referring to the respective docs in `ThreadSafeRepository`.

I’ll extend the docs, some time this week probably!

cruessler:
I extended the comments and fixed a broken link in another doc comment. I hope my understanding of the context is correct!

PR #2110, opened 2025-08-11
fix(gix-date): add parse_raw()

### LINKED ISSUE #2095: gix::date::parse_header() is too lenient
### Current behavior 😯

...and because `parse_header()` is too lenient, `gix::date::parse()` is also too lenient.

```rust
#[test]
fn parse_header_is_too_lenient() {
    let now = SystemTime::now();
    for not_a_header_str in ["2005-04-07T22:13:09", "2005-04-07 22:13:09"] {
        assert_eq!(
            gix_date::parse_header(not_a_header_str),
            Some(Time {
                seconds: 2005,
                offset: 0
            })
        );

        assert_eq!(
            gix_date::parse(not_a_header_str, Some(now)).unwrap(),
            Time {
                seconds: 2005,
                offset: 0
            }
        );
    }
}
```

### Expected behavior 🤔

```rust
#[test]
fn parse_header_is_too_lenient() {
    let now = SystemTime::now();
    for not_a_header_str in ["2005-04-07T22:13:09", "2005-04-07 22:13:09"] {
        assert!(gix_date::parse_header(not_a_header_str).is_none());
        assert!(gix_date::parse(not_a_header_str, Some(now)).is_err());
    }
}
```

These come into play in StGit in the context of parsing command line arguments that set/modify either the author or commit date. We try to accommodate these iso8601-like strings that are missing the timezone specifier with the policy that if the timezone specifier is missing, we assume the user meant the local timezone.

Updating gix-date from 0.9.4 to 0.10.3 triggers a regression in this StGit code:

https://github.com/stacked-git/stgit/blob/f88c7f0e375e069ea35f7db1294d1408d69e7a1d/src/ext/time.rs#L20

To work around this in StGit, I'd have to bypass using `gix::date::parse()` and instead write my own similar version with the stricter semantics I'm indicating in this issue. Would be nice to be able to get this behavior from gix.

I also note [this comment from gix_date::parse()](https://github.com/GitoxideLabs/gitoxide/blob/8f51845ab1e27c38f1e202c93f003870d282cf0d/gix-date/src/parse.rs#L148) that seems to indicate that this problem may be on your radar somewhere (@Byron).

As always, thank you for your tremendous effort on gitoxide!

### Git behavior

_No response_

### Steps to reproduce 🕹

_No response_

jpgrayson:
When parsing arbitrary date inputs from the user with gix_date::parse(), the non-strict gix_date::parse_header() would accept a variety of inputs that do not match any of the known/accepted formats, but would nonetheless be accepted as being "close enough" to a raw date as found in git commit headers.

The new parse_raw() function takes a stricter approach to parsing raw dates as found in commit headers, only accepting inputs that unambiguously match a valid raw date.

The gix_date::parse() function is updated to use the new parse_raw() function instead of the less-strict parse_header(). This has the effect of making gix_date::parse() avoid mis-parsing various inputs that would otherwise have been accepted by gix_date::parse_header(). For example "28 Jan 2005".

-----

This resolves #2095 more robustly and avoids some regressions in the downstream StGit project.


Byron:

In my refactor I made the existing tests clearer, and added some 'positive' tests to show other aspects of the built-in leniency. The biggest observable change is that the `parse_raw()` function is now private.

Edit: there is also a new patch release of `gix-date` for consumption.

jpgrayson:
Thank you again, @Byron. With this fix and the gix-date 0.10.5 release, StGit is now up to date with the latest gix 0.73.

PR #2111, opened 2025-08-12
Persist object refs on disk with a trailing newline

handlerug:
Match git's behavior. This improves compatibility with alternative Git client programs such as Sublime Merge.


handlerug:
The "cargo-deny-advisories" failure doesn't seem to be related to my changes.

EliahKagan:
> The "cargo-deny-advisories" failure doesn't seem to be related to my changes.

Yes, it is unrelated. That failure is due to [RUSTSEC-2025-0047](https://rustsec.org/advisories/RUSTSEC-2025-0047.html). It has since been fixed in #2112 by updating the affected dependency.

Byron:
A patch-release of `gix-ref` has just been created.

PR #2120, opened 2025-08-17
perf(gix-utils): add normalization fast path

davidkna:
`unicode_normalization` provide fast-paths functions (`is_nfc_quick`/`is_nfd_quick`) for quick-checks to determine if (de-)composition is necessary. These functions may can return an ambiguous results in certain cases (`Maybe`), to avoid this `is_nfc`/`is_nfd` exists which do slow checks if these functions return `Maybe` (matching the current code in `gix-utils`).

For strings that do not require normalisation the `is_nfc`/`is_nfd` are about 75% faster, but 20% slower for strings that do require normalisation. As far as I can tell the slowdown can be avoided by avoiding the slow check e.g `is_nfc_quick(s.chars()) == IsNormalized::Yes`, but this may come at the expense of unnecessary allocations.


Byron (approved):
Wow, that's great, thanks so much!

PR #2130, opened 2025-08-22
Report August 2025

Byron:
### Tasks

* [x] report
* [x] timesheet
* [x] post

### DIFF
--- a/etc/reports/25-08.md
+++ b/etc/reports/25-08.md
@@ -1,0 +1,68 @@
+This month feels like one of these summer months when not much has been happening, *again*, 
+at least I didn't get to contribute more than basic maintenance.
+
+## GitButler - Core Engine Rewrite
+
+Just to be able to report something, let me share some of my GitButler work again. The last time this came up I changed the way GB sees the world by making it based on a Graph data structure.
+
+And now that this works and informs what the user sees in the app.
+What's new is that it is also used to mutate the repository, for instance by creating references.
+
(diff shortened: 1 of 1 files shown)


Byron:
https://github.com/GitoxideLabs/gitoxide/discussions/2131

Byron:
@EliahKagan As a heads-up, it seems CI is broken now because Git changed and probably packs references more often or more aggressively. There seems to be at least one script that assumes the presence of a loose reference that is now packed.
I will merge this PR as it's clearly unrelated, and hope to get to fixing this soon. If you get to it before me, I'd also appreciate it.

EliahKagan:
The new CI failures all strongly appear to be due to the change in https://github.com/git/git/commit/a5a727c448e56b212546e047031166ced82c7d1e. The `ubuntu-latest` runner image was recently upgraded to Git 2.51.0, which includes this change, which disallows something that would often be unintentional but that we are deliberately doing in a fixture script used in multiple tests.

I'm working on this now, and I expect to be able to fix it by using `git config` instead of `git remote`. If for some reason that doesn't work, then directly editing the `config` file should work.

If for some reason *that* doesn't work, I believe the previous version is still available from the git-core PPA so we may be able to downgrade to it on CI as a temporary measure. If for some reason we can't do that, we can downgrade even further as a temporary measure by using `git` from Ubuntu's own repositories (the runner has Ubuntu 24.04, so its non-PPA `git` is new enough for our test suite).

EliahKagan:
The new CI failures are fixed since #2132.

PR #2136, opened 2025-08-26
docs(README): replace external asciinema embed with local SVG

kingsword09:
Replace the external asciinema embed with a locally animated SVG, allowing users to view the demo directly in the README without needing to visit asciinema.


Byron:
Thanks a lot!

Somehow, I *currently* prefer the README static instead of auto-playing, even though I totally admit that seeing something automatically on a landing page helps to sell/elevate it, provided the demo is nicely made.

I am clearly torn :D.

My concern is that this recording is 3 years old, and it might be better to update it beforehand. Would you be able to share the code (maybe even in the `justfile`) that allows to generate these animated SVGs?

Thanks again.

Byron:
You know, let's just try it - the file costs 7kb when compresses, I think it's OK to experiment with that.
I'd still appreciate information on how this file can be (re)generated.

kingsword09:
@Byron  I used the [svg-term-cli](https://github.com/marionebl/svg-term-cli) tool to generate the SVG file.
The exact command was:
```bash
svg-term --cast=542159 --out etc/gix-asciicast.svg --window
```

PR #2138, opened 2025-08-26
Expose gix-pack caching features

tmm1:


Byron:
Thanks a lot!

Each of the added features would need to get their own documentation. While CI fails, I have set this PR back to draft.

Lastly, even if the PR is empty, the commit should lead with `feat: ` and contain a motivation in the body. You would have one, writing this will suffice.
Thanks

Byron:

PR #2143, opened 2025-08-29
various improvements

Byron:
### Tasks

* [x] fix CI

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -49,9 +49,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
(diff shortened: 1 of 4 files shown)


Byron:
The issue doesn't reproduce locally, but the problem is that an untrusted path ends up trusted.
Locally it doesn't reproduce either, and I wonder how this could relate to this commit 042efa0f0b45744aeb70edb48c156b44380132ae which is likely the culprit.

If `HOME` wasn't set, it would now try something else on other platforms, so maybe we should try that only on Windows?

PR #2147, opened 2025-08-31
fix: consider a Windows resource untrusted if security information could not be retrieved (#2128)

### LINKED ISSUE #2128: Trust-check doesn't always work on Windows
### Current behavior 😯

Depending on how that path to the valid repository is mounted, we may see this:

```
    Couldn't get security information for path 'Z:\home\user\repo\.git' with err Incorrect function. (os error 1)) path="Z:\\home\\user\\repo"
```

This leads to us not being able to open the repository.

Originally reported [on Discord](https://discord.com/channels/1060193121130000425/1407619340626300928/1407657416329265232).

### Expected behavior 🤔

_No response_

### Git behavior

_No response_

### Steps to reproduce 🕹

_No response_

Byron:
Fixes #2128 .

Related to https://github.com/gitbutlerapp/gitbutler/issues/9913

### DIFF
--- a/gix-sec/src/identity.rs
+++ b/gix-sec/src/identity.rs
@@ -123,9 +123,9 @@ mod impl_ {
                 );
 
                 if result != ERROR_SUCCESS {
-                    // We cannot determine ownership, so we default to reduced trust (false) rather
-                    // than failing completely.
                     if result == ERROR_INVALID_FUNCTION {
+                        // We cannot obtain security information, so we default to reduced trust
+                        // (false) rather than failing completely.
                         return Ok(false);
                     }
                     let inner = io::Error::from_raw_os_error(result as _);
--- a/gix-sec/src/identity.rs
+++ b/gix-sec/src/identity.rs
@@ -82,7 +82,7 @@ mod impl_ {
 
     pub fn is_path_owned_by_current_user(path: &Path) -> io::Result<bool> {
         use windows_sys::Win32::{
-            Foundation::{GetLastError, LocalFree, ERROR_INSUFFICIENT_BUFFER, ERROR_SUCCESS},
+            Foundation::{GetLastError, LocalFree, ERROR_INSUFFICIENT_BUFFER, ERROR_INVALID_FUNCTION, ERROR_SUCCESS},
             Security::{
                 Authorization::{GetNamedSecurityInfoW, SE_FILE_OBJECT},
                 CheckTokenMembership, EqualSid, GetTokenInformation, IsWellKnownSid, TokenOwner,
@@ -123,6 +123,11 @@ mod impl_ {
                 );
 
                 if result != ERROR_SUCCESS {
+                    // We cannot determine ownership, so we default to reduced trust (false) rather
+                    // than failing completely.
+                    if result == ERROR_INVALID_FUNCTION {
+                        return Ok(false);
+                    }
                     let inner = io::Error::from_raw_os_error(result as _);
                     error!(
                         inner,


gix-sec/src/identity.rs
```diff
@@ -123,6 +123,11 @@ mod impl_ {
                 );
 
                 if result != ERROR_SUCCESS {
+                    // We cannot determine ownership, so we default to reduced trust (false) rather
+                    // than failing completely.
+                    if result == ERROR_INVALID_FUNCTION {
+                        return Ok(false);
+                    }
```
EliahKagan: It seems to me that this describes the workaround more accurately and clearly, as we do still return errors for conditions other than the unavailability of (or inability to retrieve) security information.

```suggestion
                if result != ERROR_SUCCESS {
                    if result == ERROR_INVALID_FUNCTION {
                        // We cannot obtain security information, so we default to reduced trust
                        // (false) rather than failing completely.
                        return Ok(false);
                    }
```

EliahKagan (approved):
Per https://github.com/GitoxideLabs/gitoxide/issues/2128#issuecomment-3240295243, this looks fine to me. I've proposed a change to comments below that I think makes them clearer and more accurate to the condition under which this workaround applies.

It may be good to have a test that this workaround is not triggered on conditions that are decisively other errors, such as on a path where no file or directory exists at all. However, I don't think that needs to block this, and I can add it later.


(More extensive tests than that, such as those involving an actual 9P share, could also eventually be added but I'd personally hold off on doing those until next time I'm working on `gix-sec`, where I anticipate I'll be able to add them while augmenting the test suite in other ways.)

EliahKagan:
I think this doesn't require a re-amend, but I just noticed that a couple of the paths in my analysis (that you copied into the commit message) are not really readable in the commit message, because I had used explicit HTML for custom formatting internal to the code spans:

```markdown
Support for 9P shares was added to Windows to implement shares that access files in installed WSL distributions. These are shares named like <code>&bsol;&bsol;wsl$&bsol;<em>distro</em></code> or <code>&bsol;&bsol;wsl.localhost&bsol;<em>distro</em></code>, where *`distro`* is the distribution. Currently 9P is supported on Windows only for such WSL shares (though I don't know if being intended only for this WSL-related use is the reason security information isn't available through them).
```

Specifically:

```markdown
These are shares named like <code>&bsol;&bsol;wsl$&bsol;<em>distro</em></code> or <code>&bsol;&bsol;wsl.localhost&bsol;<em>distro</em></code>, where *`distro`* is
```

This is probably fine and still adequately informative, and it may even render properly in automatically generated changelog entries, or it can be fixed up there if not. Therefore, I suggest ignoring this unless you already need to amend the commit between now and when the PR merges. That is, noticing this has not changed my judgment that the PR looks good to me. :)

Byron:
I also just realised that this working hinges on the error-check to be correct, and admittedly I am not sure as I didn't validate this at all.

`gix free discover` would be a way to check this though, it lists the trust level and might also just fail if the trust-discovery fails.

EliahKagan:
> I also just realised that this working hinges on the error-check to be correct, and admittedly I am not sure as I didn't validate this at all.

It seems to be working. In the following, I've replaced some long output with `...` in a few places.

```text
C:\Users\ek\source\repos\gitoxide [main ≡]> cargo build
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 6.31s
C:\Users\ek\source\repos\gitoxide [main ≡]> net use
New connections will be remembered.


Status       Local     Remote                    Network

-------------------------------------------------------------------------------
OK           X:        \\localhost\C$            Microsoft Windows Network
OK           Y:        \\kip\ek                  Microsoft Windows Network
             Z:        \\wsl$\Ubuntu             Plan 9 Network Provider
The command completed successfully.

C:\Users\ek\source\repos\gitoxide [main ≡]> cd Z:\home\ek\repos-wsl\gitoxide
Z:\home\ek\repos-wsl\gitoxide [(main)]> git config -l --local
fatal: --local can only be used inside a git repository
Z:\home\ek\repos-wsl\gitoxide [(main)]> gix config
Error: Could not determine trust level for path '.\.git'.

Caused by:
    Couldn't get security information for path '.\.git' with err Incorrect function. (os error 1)
Z:\home\ek\repos-wsl\gitoxide [(main)]> C:\Users\ek\source\repos\gitoxide\target\debug\gix config
# From 'C:/Users/ek/scoop/apps/git/2.51.0/etc/gitconfig' (GitInstallation)
...

# From 'C:\Users\ek\.gitconfig' (User)
...

# From '.\.git\config' (Local, untrusted)
[core]
        repositoryformatversion = 0
        filemode = true
        bare = false
        logallrefupdates = true

[remote "origin"]
        url = git@github.com:EliahKagan/gitoxide.git
        fetch = +refs/heads/*:refs/remotes/origin/*

...
```

Byron:
Fantastic, thanks!

PR #2153, opened 2025-09-05
Add `Repository::blame_file()`

cruessler:
This PR adds `Repository::blame_file()`, reusing the existing feature flag `blame`. This is the minimal API I was able to come up with.

### Open questions

- The proposed API offers no way for the consumer to influence how `odb`, `cache` and `resource_cache` get created. Is this the correct choice?
- As opposed to other APIs, e. g. `Repository::references`, this API is not `Platform`-based, but directly calls the underlying function. Is this something that should be changed/needs to be changed? What are the trade-offs involved?
- There’s other APIs that provide a recorder-based API, such as `gix_diff::tree`, allowing consumers to process the API’s output one-by-one while also allowing them to cancel it after each piece of output. Is that something we might want for `gix_blame::file()` and, if so, would we have to account for that in this initial version of `Repository::blame_file()`?
- Is there anything that needs to be done in order for the tests in `gix/tests/gix/repository/blame.rs` to be run on CI? I’m asking because the code I added is behind a feature flag.


gix/src/repository/blame.rs
```diff
@@ -0,0 +1,43 @@
+#[cfg(feature = "blame")]
```
Byron: Since the whole module is behind a feature toggle, that toggle doesn't need to be used inside of it anymore.

gix/src/repository/blame.rs
```diff
@@ -0,0 +1,43 @@
...
+    #[error(transparent)]
+    Blame(#[from] gix_blame::Error),
+}
+
+impl Repository {
+    /// Produce a list of consecutive [`gix_blame::BlameEntry`] instances. Each `BlameEntry`
+    /// corresponds to a hunk of consecutive lines of the file at `suspect:<file_path>` that got
+    /// introduced by a specific commit.
+    ///
+    /// For details, see the documentation of [`gix_blame::file::function::file()`].
```
Byron: Shouldn't the linke be `gix_blame::file()`?

gix/src/repository/blame.rs
```diff
@@ -0,0 +1,43 @@
...
+    #[cfg(feature = "blame")]
+    pub fn blame_file(
+        &self,
+        file_path: &BStr,
+        suspect: ObjectId,
```
Byron: Often this is `impl Into<ObjectId>` to make it easier to pass `gix::Id` and the likes.

gix/tests/gix/repository/blame.rs
Byron, line 1: I don't know if I'd test anything here as the method itself is really just a forwarding. Probably it's good to avoid duplication unless we know what specific thing we should test.

In theory, we could start adding doc-tests, then redundancy doesn't matter as the purpose is to inform the user, but it's by no means an obligation as they are quite cumbersome to make look nice.

gix/src/repository/blame.rs
```diff
@@ -0,0 +1,43 @@
...
+#[derive(Debug, thiserror::Error)]
+#[allow(missing_docs)]
+pub enum Error {
+    #[error(transparent)]
+    CommitGraphIfEnabled(#[from] commit_graph_if_enabled::Error),
+    #[error(transparent)]
+    DiffResourceCache(#[from] diff_resource_cache::Error),
+    #[error(transparent)]
+    Blame(#[from] gix_blame::Error),
+}
```
Byron: In the absence of custom 'connected' blame related types, I think this whole module should be in `gix::repository::blame`, with the error type 'where the other errors types are' - there is pattern for this.

Alternatively `gix_blame::Outcome` would have to be turned into a duplicate structure with a reference to `repo: &Repository`, which doesn't seem worth it - it's the bane of the `gix` crate as it always has to repeat whole types just to add some convenience.

Byron (commented):
Thanks a lot for tackling this! Note that I put the PR back into draft while CI is failing and there are questions.

> This PR adds `Repository::blame_file()`, reusing the existing feature flag `blame`. This is the minimal API I was able to come up with.

I think this works, the feature-flag usage is great, and the API is exactly what it should be - minimal and to the point. If there are other uses that need more control, they can easily use `gix_blame` directly, or one could offer a `blame_opt()` of sort that has more parameters.

> 
> ### Open questions
> * The proposed API offers no way for the consumer to influence how `odb`, `cache` and `resource_cache` get created. Is this the correct choice?
A good start, let's keep it simple until there is demand.

> * As opposed to other APIs, e. g. `Repository::references`, this API is not `Platform`-based, but directly calls the underlying function. Is this something that should be changed/needs to be changed? What are the trade-offs involved?

It's not obvious to me how a platform would be helpful here, which probably means there should be none.

> * There’s other APIs that provide a recorder-based API, such as `gix_diff::tree`, allowing consumers to process the API’s output one-by-one while also allowing them to cancel it after each piece of output. Is that something we might want for `gix_blame::file()` and, if so, would we have to account for that in this initial version of `Repository::blame_file()`?

We talked about the final-fancy version of `gix-blame` which would be far more complex, but while we don't have it, we should keep it simple.

> * Is there anything that needs to be done in order for the tests in `gix/tests/gix/repository/blame.rs` to be run on CI? I’m asking because the code I added is behind a feature flag.

Let's skip the tests, they seem redundant.

gix/src/repository/blame.rs
cruessler: You’re right, that also fixes the only error on CI. :-)

gix/src/repository/blame.rs
cruessler: Thanks for the hint, changed.

gix/src/repository/blame.rs
cruessler: I’ve moved the enum, following the example of `diff_tree_to_tree::Error` and others. Was that correct?

gix/tests/gix/repository/blame.rs
cruessler: These tests indeed duplicate tests found elsewhere. I mainly added them to make sure that the whole `gix::Repository::blame_file` API could be used only relying on imports from `gix`. Feel free to remove them if you think they’re redundant!

cruessler:
From my side, this PR is now ready for review!

Byron:
That's perfect, thanks so much!

And yes, the test is definitely good to see how to access the return type, and I am grateful that you moved them instead of removing them as originally suggested.

PR #2154, opened 2025-09-06
fix `gix-blame` performance regresion

folkertdev:
fixes https://github.com/GitoxideLabs/gitoxide/issues/2148

The issue was introduced in https://github.com/GitoxideLabs/gitoxide/pull/1963, where it was incorrectly assumed that the `max-performance-safe` feature flag was only used to enable other feature flags. But this feature flag is itself used in various places, so failing to enable it leads to worse codegen. This issue was discussed/found in https://github.com/GitoxideLabs/gitoxide/issues/2148

### DIFF
--- a/gix/Cargo.toml
+++ b/gix/Cargo.toml
@@ -246,8 +246,8 @@ pack-cache-lru-static = ["gix-pack/pack-cache-lru-static"]
 ## Provide a hash-map based LRU cache whose eviction is based a memory cap calculated from object data.
 pack-cache-lru-dynamic = ["gix-pack/pack-cache-lru-dynamic"]
 
-## Enable all features required for performance. Currently, this is equivalent to `max-control`, as gix always uses zlib-rs.
-max-performance = ["max-control"]
+## Enable all features required for performance.
+max-performance = ["max-performance-safe"]
 
 ## Deprecated: gix always uses zlib-rs.
 zlib-ng = ["gix-features/zlib"]


cruessler:
I just ran the benchmark and can confirm that this fixes the performance regression on my linux/x86_64 machine! Thanks a lot!

```
# this PR
❯ hyperfine 'target/release/gix blame README.md'
Benchmark 1: target/release/gix blame README.md
  Time (mean ± σ):     164.2 ms ±   1.6 ms    [User: 135.1 ms, System: 28.2 ms]
  Range (min … max):   161.8 ms … 168.4 ms    17 runs

# the commit before the one that introduced the regression
❯ hyperfine 'bin/gix-f3684a4f6 blame README.md'
Benchmark 1: bin/gix-f3684a4f6 blame README.md
  Time (mean ± σ):     161.5 ms ±   1.8 ms    [User: 129.6 ms, System: 30.9 ms]
  Range (min … max):   158.5 ms … 164.9 ms    18 runs
```

Byron (approved):
Thanks SO much for figuring this out.
And I am so sorry I let that pass through the first time around.

In any case, I think digging into this also helped us gain something:

* zlib-rs, depending on the platform, can be slower when handling many small objects (even though `gix` doesn't notice this on MacOS at least). Now there is a test for that as a first step towards improving this.
* zlib-rs can be used directly, and the `flate2` dependency can be removed - I will see to bringing in your commit soon.

Byron:
And indeed, the natural order has been restored.

```
❯ hyperfine './gix-8d8dba2-bench-zlib-ng blame README.md' './gix-max-perf-safe blame README.md'
Benchmark 1: ./gix-8d8dba2-bench-zlib-ng blame README.md
  Time (mean ± σ):      47.1 ms ±   1.0 ms    [User: 36.9 ms, System: 7.6 ms]
  Range (min … max):    45.6 ms …  51.9 ms    61 runs

Benchmark 2: ./gix-max-perf-safe blame README.md
  Time (mean ± σ):      46.8 ms ±   3.0 ms    [User: 36.3 ms, System: 7.7 ms]
  Range (min … max):    45.0 ms …  69.2 ms    61 runs

Summary
  ./gix-max-perf-safe blame README.md ran
    1.01 ± 0.07 times faster than ./gix-8d8dba2-bench-zlib-ng blame README.md
```

PR #2155, opened 2025-09-06
in `gix-features`, use `libz-rs-sys` directly, skipping `flate2`

folkertdev:
feel free to steal from or force push to this.

Also, this was a good exercise in designing a safe API for zlib-rs (which, fingers crossed, may happen in Q4).

### DIFF
--- a/gix-features/src/zlib/mod.rs
+++ b/gix-features/src/zlib/mod.rs
@@ -1,4 +1,111 @@
-pub use flate2::{Decompress, Status};
+use std::fmt::Display;
+
+pub struct Decompress(Box<libz_rs_sys::z_stream>);
+
+unsafe impl Sync for Decompress {}
+unsafe impl Send for Decompress {}
+
+impl Decompress {
+    pub fn total_in(&self) -> u64 {
+        self.0.total_in
+    }
+
+    pub fn total_out(&self) -> u64 {
+        self.0.total_out
+    }
+
+    pub fn new(_zlib_header: bool) -> Self {
+        let mut this = Box::new(libz_rs_sys::z_stream::default());
+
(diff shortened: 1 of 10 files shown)


Byron:
This is fantastic, thanks so much!

I am particularly looking forward to getting to try out the safe API for `zlib-rs` :) - who knows, maybe things get even faster then.

gix-features/src/zlib/mod.rs
```diff
@@ -1,4 +1,132 @@
-pub use flate2::{Decompress, Status};
+use std::ffi::c_int;
+
+/// A type to hold all state needed for decompressing a ZLIB encoded stream.
+pub struct Decompress(libz_rs_sys::z_stream);
```
folkertdev, line 4: btw not having to `Box` (or otherwise `Pin`) here is something only zlib-rs can do because our state is not self-referential. I currently see no reason why that would not work in the future, but technically it's undocumented/unspecified behavior.

Byron: Oh wow, I wasn't aware!

For now I'd leave it like this, but would also hope that you would be able to declare something as `!Pin` so it's clear this can't be moved?

I don't think that this ultimately made a difference in terms of performance, so I'd rather box it again than to risk silent memory corruption in a potential future.

folkertdev: it won't be silent (e.g. zlib/zlib-ng explicitly error on this). Also looking at it some more, I really see no reason why we'd ever make the structure self-referential. For zlib they seem to do it to save passing a the `z_stream` down (so, it saves on a register?!). Not something we really have to worry about with modern compilers I don't think.

PR #2156, opened 2025-09-07
Reduce noise in `gix-blame` tests

cruessler:
This is a PR that reduces noise in `gix-blame` tests by implementing `From<(Range<u32>, ObjectId)> for UnblamedHunk`. This implementation allows us to replace `UnblamedHunk { range_in_blamed_file: 3..5, suspects: [(suspect, 3..5)].into(), source_file_name: None, }` by `(3..5, suspect).into()` and `new_unblamed_hunk(0..5, suspect, Offset::Added(0))` by `(0..5, suspect).into()`. I think this change significantly increases the signal-to-noise ratio.

In addition to this, we could also implement `From<(Range<u32>, (ObjectId, Range<u32>))> for UnblamedHunk` or `From<(Range<u32>, ObjectId, Offset)> for UnblamedHunk` in order to convert places where the range associated with a suspect is not identical to `range_in_blamed_file`, but the benefit would be smaller. What do you think?

### DIFF
--- a/gix-blame/src/file/tests.rs
+++ b/gix-blame/src/file/tests.rs
@@ -18,6 +18,18 @@ fn new_unblamed_hunk(range_in_blamed_file: Range<u32>, suspect: ObjectId, offset
     }
 }
 
+impl From<(Range<u32>, ObjectId)> for UnblamedHunk {
+    fn from(value: (Range<u32>, ObjectId)) -> Self {
+        assert!(value.0.end > value.0.start, "{:?}", value.0);
+
+        UnblamedHunk {
+            range_in_blamed_file: value.0.clone(),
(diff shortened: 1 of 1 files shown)


Byron:
This is awesome!

Sure, it's a bit less clear what these values mean at first glance, but tests can be expert mode where a thing or two have to be learned in order to write or understand them.

> In addition to this, we could also implement `From<(Range<u32>, (ObjectId, Range<u32>))> for UnblamedHunk` or `From<(Range<u32>, ObjectId, Offset)> for UnblamedHunk` in order to convert places where the range associated with a suspect is not identical to `range_in_blamed_file`, but the benefit would be smaller. What do you think?

Go for it :).

PR #2157, opened 2025-09-09
Reduce noise in `gix-blame` tests

cruessler:
This is a follow-up to #2156. It is a net reduction of about 270 lines in one test file. At the same time, it hopefully makes tests easier to understand as you can now better focus on the most important bits, i. e. the line numbers in the source file and the blamed file as well as the suspect’s `ObjectId`. I think this is a win.

Also, `impl From<(Range<u32>, ObjectId)> for UnblamedHunk` now reuses `impl From<(Range<u32>, ObjectId, Range<u32>)> for UnblamedHunk`, so the `assert!`s only have to be written once.

### DIFF
--- a/gix-blame/src/file/tests.rs
+++ b/gix-blame/src/file/tests.rs
@@ -2,29 +2,34 @@ use std::ops::Range;
 
 use gix_hash::ObjectId;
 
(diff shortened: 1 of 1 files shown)


Byron (approved):
Neat, thanks!

PR #2163, opened 2025-09-12
feat: replace peel_to_x_in_place by peel_to_x

cruessler:
This is a follow-up to/replacement for #2146. It renames methods on `ReferenceExt`, deprecating the previously used variants: `peel_to_id_in_place`, `peel_to_id_in_place_packed`, `follow_to_object_in_place_packed`.

This doesn’t touch `gix::types::Head`, the type that caused the initial PR, yet, as I want to wait for feedback before making any further changes.

PR #2171, opened 2025-09-14
feat: replace `peel_to_x_in_place` by `peel_to_x` in `gix::types::Head`

cruessler:
This is a follow-up to #2163. It deprecates `try_peel_to_id_in_place`, `peel_to_object_in_place` and `peel_to_commit_in_place` on `gix::types::Head` in favour of their non-suffixed counterparts.


Byron:
This is great, thanks a lot!

The API definitely feels more streamlined now!

PR #2173, opened 2025-09-15
feat: ability to change the fetch url of a remote

metlos:
The remote has a couple of "builder" methods to change is fields, e.g. `push_url` for setting the push url.

A builder method for changing the fetch url of a remote was missing. This makes it impossible to fully replicate  the functionality of `git remote set-url` (e.g. look at the lengths I'm trying to workaround this (imperfectly) in a PR to replicate `git remote set-url --push`) in `jj`: https://github.com/metlos/jj/blob/8ec7cd1f68a5279e6aff2faea248a4978d46d17a/lib/src/git.rs#L2024)

### DIFF
--- a/gix/src/remote/build.rs
+++ b/gix/src/remote/build.rs
@@ -2,6 +2,31 @@ use crate::{bstr::BStr, remote, Remote};
 
 /// Builder methods
 impl Remote<'_> {
+    /// Set the `url` to be used when fetching data from a remote.
+    pub fn with_url<Url, E>(self, url: Url) -> Result<Self, remote::init::Error>
+    where
+        Url: TryInto<gix_url::Url, Error = E>,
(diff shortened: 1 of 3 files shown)


Byron (approved):

Having API symmetry is certainly a good trait to have. Probably I avoided having these to avoid confusion with the `Repository::remote_at*()` methods which do the same, but I added a note that ideally these are used instead.

To me these kinds of API are still unsolved, as in theory one would either want to have free access to the fields (something I prefer), or one has to provide both builders and setters for everything which is cumbersome.
It's certainly possible to find more instances where certain methods are missing even though one would expect them to be there.

metlos:
Yeah, there's no silver bullet for the APIs. Thanks for merging this! 🫶

PR #2180, opened 2025-09-21
Strengthen and adjust `copy-royal` usage guidance and caveats

EliahKagan:
Copy-royal is useful for taking repository trees that triggered a bug, and producing trees that diff the same, allowing regression tests to be prepared and focusing on aspects that are relevant to diffing.

However, the pattern represented by the output of copy-royal is effectively equivalent to a regular grammar of the possible inputs; more specifically, a concatenation of character classes, which in practice are each small. For example, copy-royal on `Hello, world!` produces `Iette, yeuta!`. If the input is assumed to consist of ASCII characters, the possible inputs that produce `Iette, yeuta!` are the strings that match the regular expression `[HR][eoy][blv][blv][eoy],\ [cmw][eoy][hr][blv][dnx]!`.

Because real-world inputs are not random selections of letters, the code or text of the input will often be possible to reconstruct from this pattern, such as by using an autoregressive LLM with constrained decoding to sample only the small subset of logits at each time step that are not contradicted by the the regex, while exploring multiple paths with techniques such as beam search. This approach is explored with a proof-of-concept in [this Google Colab notebook](https://colab.research.google.com/drive/1jT1OnESk0oTR_B3N09WTeU1c_UgYH_HD?usp=sharing).

Because copy-royal is thus effectively reversible in many cases, likely including all practical cases, we should avoid saying or implying that the input text cannot be reconstructed or discovered, and we should especially avoid saying or implying that copy-royal could or should be used to turn private or sensitive data into data that are okay to share publicly. The documentation of copy-royal has suggested irreversibility for a long time, though not initially meaning it in a robust sense, and gradually shifted to incorporate language that further suggested that it could be used for for security or privacy related purposes.

This pull request rewrites the documentation for copy-royal and directly related facilities to avoid implying benefits or use cases that it does not or may not have, and instead to describe its useful purpose more clearly.

Besides in [the notebook](https://colab.research.google.com/drive/1jT1OnESk0oTR_B3N09WTeU1c_UgYH_HD?usp=sharing) linked above, there are also some more details in the commit message.


Byron:
Thanks a lot, much better!

cruessler:
Thanks a lot!

Issue #2052, opened 2025-06-20
Worktree resolves to wrong path


ibhagwan:
### Current behavior 😯

When using a git bare repo with submodules (dotfiles) `gix status` results in a crash, it seels that the submodule is resolved to the wrong path (missing the username, should be `/home/bhagwan...`):

![Image](https://github.com/user-attachments/assets/64f38dd5-8861-4ceb-a582-80350d24f64b)

Text version of the error:
```
→ gix status
Error: Could not determine status for submodule at '.config/awesome/lain'

Caused by:
    Worktree root at '/home/.config/awesome/lain' is not a directory

Stack backtrace:
   0: <unknown>
   1: <unknown>
   2: <unknown>
   3: <unknown>
   4: <unknown>
   5: <unknown>
   6: <unknown>
   7: <unknown>
   8: <unknown>
   9: __libc_start_call_main
             at ./csu/../sysdeps/nptl/libc_start_call_main.h:58:16
  10: __libc_start_main_impl
             at ./csu/../csu/libc-start.c:360:3
  11: <unknown>
```

### Expected behavior 🤔

Expected to return the output of the status command.

### Git behavior

`git status` succeeds with no issue:

![Image](https://github.com/user-attachments/assets/16ed6ff6-dc51-42da-86f4-e5b1bf9ef44d)

Output of submodule:

![Image](https://github.com/user-attachments/assets/37170dff-3802-4817-bd6a-99d9b69ba3d1)

Output of worktree list:

![Image](https://github.com/user-attachments/assets/68be3133-3527-4785-bad1-6b194c8a07a6)

### Steps to reproduce 🕹

1. setup git repository with submodules  in `~/dots` 
2. cd ~/dots
3. git config core.worktree $HOME
4. ln -fs ~/dots/.git ~/.git
5. cd ..
6. `gix status` (in $HOME)

If you wish to try it out, the process above is automated by [my dotfiles install script](https://github.com/ibhagwan/dots/blob/master/dots/install.sh).

Byron:
Thanks a lot for reporting!

I think the issue seems to be that the worktree isn't correctly computed when `core.worktree` is used.

* git: /home/bhagwan/dots
* gix: /home/bhagwan

When trying it myself, with an empty repository, I am getting even more interesting results:

```
~ ( main) [?]
❯ gix worktree list
/Users/byron [main]

~ ( main) [?]
❯ git worktree list
/Users/byron/dev/testing  0000000 [main]
[...]

PR #2182, opened 2025-09-21
fix: submodule worktree incorrect with symlinked `.git` dir (#2052)

arumugamramaswamy:
fixes https://github.com/GitoxideLabs/gitoxide/issues/2052

My fix introduced a regression. Looks like we never return relative paths for `repo.workdir()`. One way I see to fix this is by making the path relative to `cwd`, but not sure if this is desired. Any thoughts on how we can work around this?

```
failures:

---- repository::open::submodules::by_their_worktree_checkout_and_git_modules_dir stdout ----
[gix/src/open/repository.rs:308:17] &git_dir = "tests/fixtures/generated-do-not-edit/make_submodules/1482069896-unix/with-submodules/.git/modules/m1"
[gix/src/open/repository.rs:308:17] &wt_path = "tests/fixtures/generated-do-not-edit/make_submodules/1482069896-unix/with-submodules/.git/modules/m1/../../../m1"

thread 'repository::open::submodules::by_their_worktree_checkout_and_git_modules_dir' panicked at gix/tests/gix/repository/open.rs:338:17:
  left: "/home/aru/oss/gitoxide/gix/tests/fixtures/generated-do-not-edit/make_submodules/1482069896-unix/with-submodules/m1"
 right: "tests/fixtures/generated-do-not-edit/make_submodules/1482069896-unix/with-submodules/m1"

---- repository::worktree::with_core_worktree_config::relative stdout ----
[gix/src/open/repository.rs:308:17] &git_dir = "tests/fixtures/generated-do-not-edit/make_core_worktree_repo/4038243702-unix/absolute-worktree/.git"
[gix/src/open/repository.rs:308:17] &wt_path = "/home/aru/oss/gitoxide/gix/tests/fixtures/generated-do-not-edit/make_core_worktree_repo/4038243702-unix/worktree"
[gix/src/open/repository.rs:308:17] &git_dir = "tests/fixtures/generated-do-not-edit/make_core_worktree_repo/4038243702-unix/relative-worktree/.git"
[gix/src/open/repository.rs:308:17] &wt_path = "tests/fixtures/generated-do-not-edit/make_core_worktree_repo/4038243702-unix/relative-worktree/.git/../../worktree"

thread 'repository::worktree::with_core_worktree_config::relative' panicked at gix/tests/gix/repository/worktree.rs:51:17:
assertion `left == right` failed: relative-worktree|true: work_dir is set to core.worktree config value, relative paths are appended to `git_dir() and made absolute`
  left: "/home/aru/oss/gitoxide/gix/tests/fixtures/generated-do-not-edit/make_core_worktree_repo/4038243702-unix/worktree"
 right: "tests/fixtures/generated-do-not-edit/make_core_worktree_repo/4038243702-unix/worktree"


failures:
    repository::open::submodules::by_their_worktree_checkout_and_git_modules_dir
    repository::worktree::with_core_worktree_config::relative
```

### DIFF
--- a/gix/src/open/repository.rs
+++ b/gix/src/open/repository.rs
@@ -301,7 +301,18 @@ impl ThreadSafeRepository {
                     | gix_config::Source::EnvOverride => wt_path,
                     _ => git_dir.join(wt_path).into(),
                 };
-                worktree_dir = gix_path::normalize(wt_path, current_dir).map(Cow::into_owned);
+
+                // the reason we use realpath instead of gix_path::normalize here is because there
+                // could be any intermediate symlinks (for example due to a symlinked .git
+                // directory)
+                worktree_dir = gix_path::realpath(&wt_path).ok();
+                // restore the relative path if possible after resolving the absolute path
+                if wt_path.is_relative() {
+                    if let Some(rel_path) = worktree_dir.as_deref().and_then(|p| p.strip_prefix(current_dir).ok()) {
+                        worktree_dir = Some(rel_path.to_path_buf());
+                    }
+                }
+
                 #[allow(unused_variables)]
                 if let Some(worktree_path) = worktree_dir.as_deref().filter(|wtd| !wtd.is_dir()) {
                     gix_trace::warn!("The configured worktree path '{}' is not a directory or doesn't exist - `core.worktree` may be misleading", worktree_path.display());
--- a/gix/tests/gix/status.rs
+++ b/gix/tests/gix/status.rs
@@ -311,6 +311,27 @@ mod index_worktree {
         }
 
         #[test]
+        fn submodule_in_symlinked_dir() -> crate::Result {
+            use crate::util::named_subrepo_opts;
+            let repo = named_subrepo_opts(
+                "make_submodule_with_symlinked_git_dir.sh",
+                "symlinked-git-dir",
+                gix::open::Options::isolated(),
+            )?;
+            let status = repo
+                .status(gix::progress::Discard)?
+                .index_worktree_options_mut(|opts| {
+                    opts.sorting =
+                        Some(gix::status::plumbing::index_as_worktree_with_renames::Sorting::ByPathCaseSensitive);
+                })
+                .into_index_worktree_iter(None)?;
+            for change in status {
+                change?;
+            }
+            Ok(())
+        }
+
+        #[test]
         fn submodule_modification() -> crate::Result {
             let repo = submodule_repo("modified-untracked-and-submodule-head-changed-and-modified")?;
             let mut status = repo
--- a/gix/tests/fixtures/generated-archives/.gitignore
+++ b/gix/tests/fixtures/generated-archives/.gitignore
@@ -9,3 +9,4 @@
 /make_diff_repos.tar
 /make_submodule_with_worktree.tar
 /repo_with_untracked_files.tar
+/make_submodule_with_symlinked_git_dir.tar
--- a/gix/tests/fixtures/make_submodule_with_symlinked_git_dir.sh
+++ b/gix/tests/fixtures/make_submodule_with_symlinked_git_dir.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+set -eu -o pipefail
+
+git init -q module1
+(cd module1
+  touch this
+  mkdir subdir
+  touch subdir/that
+  git add .
+  git commit -q -m c1
+  echo hello >> this
+  git commit -q -am c2
+  touch untracked
+)
+
+mkdir symlinked-git-dir
+(cd symlinked-git-dir
+  git init -q r1
+  (cd r1
+    git commit -q --allow-empty -m "init"
+  )
+
+  git config -f r1/.git/config core.worktree "$(pwd)"
+  ln -s r1/.git .git
+
+  git -c protocol.file.allow=always submodule add ../module1 m1
+  git commit -m "add module 1"
+)


Byron:
Without having looked at it in detail, it's true that `gitoxide` attempts to retain the path given by the user, which also means that paths *can* be relative to the CWD. Or in other words, the paths passed by the user are retained.
I hope that helps.

arumugamramaswamy:
Would it be a breaking change if we use absolute paths for the worktree_dir instead?

Byron:
Absolutely, at least when done unconditionally. But maybe this even *should* be done when it makes sense?  Would the case here qualify using real paths?
Unfortunately, this is one of the cases where asking Git doesn't help as I think it will always use real-paths instead *or* it's so different in how it deals with paths that it's not really comparable.

arumugamramaswamy:
Looks like I got the tests to pass! However, I'm not sure how I should create the `.tar` archive.
(The old archive had a path prefixed by `/home/aru` which was weird)

gix/tests/fixtures/make_submodules.sh
```diff
@@ -151,4 +151,18 @@ git clone with-submodules not-a-submodule
...
+git init unborn
+
+mkdir symlinked-git-dir
+(cd symlinked-git-dir
+  git init -q r1
+  (cd r1
+    git commit -q --allow-empty -m "init"
+  )
+
+  git config -f r1/.git/config core.worktree `pwd`
```
EliahKagan: When not enclosed in double quotes, `` `pwd` ``, like `$(pwd)`, performs three [expansions](https://www.gnu.org/software/bash/manual/bash.html#Shell-Expansions), only the first which is intended here:

1. Run `pwd` and capture its output (dropping trailing newlines), expanding to that.
2. Split the result of the previous expansion on `IFS` whitespace characters, such as spaces, producing separate fields to be passed as separate command-line arguments.
3. In each of those separate fields, check if it contains globbing characters such as `*` and `?`, and if so then perform globbing (filename/pathname expansion), replacing the field with fields for each of the paths found in the expansion.

Since only (1) is intended, (2) and (3) should be suppressed by [`"`-quoting](https://www.gnu.org/software/bash/manual/bash.html#Double-Quotes) the command substitution as ``"`pwd`"`` or `"$(pwd)"` (the latter of which is usually preferred these days). The double quotes permit the command substitution itself to occur, but suppress word splitting and globbing.

(In addition to the second and third expansions not being part of the intended semantics here, these expansions will also cause problems for some developers who run the test suite with `GIX_TEST_IGNORE_ARCHIVES`, since directories with "weird" characters in their names or the names of directories leading up to them do occur in practice.)

I've included these changes in the first of the two commits I'm adding here (the second one has the regenerated archive described above).

EliahKagan:
[This test failure](https://github.com/GitoxideLabs/gitoxide/actions/runs/17956053167/job/51067510907?pr=2182#step:7:437) occurs on CI on macOS:

```text
        FAIL [   0.638s] gix::gix status::index_worktree::iter::submodule_in_symlinked_dir
  stdout ───

    running 1 test
    test status::index_worktree::iter::submodule_in_symlinked_dir ... FAILED

    failures:

    failures:
        status::index_worktree::iter::submodule_in_symlinked_dir

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 335 filtered out; finished in 0.61s
    
  stderr ───
    Error: IndexAsWorktreeWithRenames(DirWalk(WorktreeRootIsFile { root: "/System/Volumes/Data/home/ek/repos/gitoxide/gix/tests/fixtures/generated-do-not-edit/make_submodules/2131503439-unix/symlinked-git-dir" }))
```

But I don't think the problem is macOS-specific. I suspect it will happen whenever we don't set `GIX_TEST_IGNORE_ARCHIVES`.

This arises from the problem you noticed a hint of in https://github.com/GitoxideLabs/gitoxide/pull/2182#issuecomment-3324790414, regarding the inclusion of a hard-coded path to your home directory when you tried generating the archives. In addition, this path is somehow being appended--maybe as though it were a relative path?--to `/System/Volumes/Data`.

Feel free to rewind (i.e. drop) e1d9ae3c47917607634941a12ba4eff3d1b0ef6d, of course, if it turns out to be useful to do so. (By itself removing that won't fix whatever is causing this, though.)

*If the behavior of hard-coding an absolute path that is specific to the current machine in the fixture* is really intended, then this could be solved by adding the fixture archive to the appropriate `.gitignore` file so that `gix-testtools` will not use or expect a generated archive and will instead regenerate the fixture on every machine where the test suite is run. Since this can itself hide some kinds of problems, and since `make_submodules` already existed with code that does not require this measure, it might be best to split the new code out into its own fixture script, if feasible.

Of course, all of that is still only applicable if the behavior of the fixture script (and associated tests) is as intended, such that the generated fixture really should hard-code an absolute path that is only correct on the system where it is run.

cc @Byron

gix/src/open/repository.rs
```diff
@@ -301,7 +301,20 @@ impl ThreadSafeRepository {
...
                 };
-                worktree_dir = gix_path::normalize(wt_path, current_dir).map(Cow::into_owned);
+
+                // the reason we use realpath instead of gix_path::normalize here is because there
+                // could be any intermediate symlinks (for example due to a symlinked .git
+                // directory)
+                worktree_dir = gix_path::realpath(&wt_path).ok();
+                // restore the relative path if possible after resolving the absolute path
+                if wt_path.is_relative() {
+                    if let Ok(cwd) = std::env::current_dir() {
```
Byron: The `current_dir` variable is already available and should be used here.

gix/tests/fixtures/make_submodules.sh
```diff
@@ -151,4 +151,18 @@ git clone with-submodules not-a-submodule
...
+  (cd r1
+    git commit -q --allow-empty -m "init"
+  )
+
+  git config -f r1/.git/config core.worktree "$(pwd)"
+  ln -s r1/.git .git
+
+  git -c protocol.file.allow=always submodule add ../module1 m1
+  git commit -m "add module 1"
+)
```
Byron: This would have to go into its own script, and the script is placed in `.gitignore`, please see https://github.com/GitoxideLabs/gitoxide/pull/2182#issuecomment-3325222349 for details.

Byron (commented):
Thanks a lot for tackling this! The code itself looks mostly good to me, so I think the real issue here comes from the way the tests work.

I hope my comments in that regard make sense, but if not, I could also take it form here if you like. Just let me know.
Thanks again!

arumugamramaswamy:
Thanks @Byron, @EliahKagan! The PR is ready for another of review:)

Edit: sorry, let me get CI to pass 😅

gix/src/open/repository.rs
```diff
@@ -301,7 +301,18 @@ impl ThreadSafeRepository {
                     | gix_config::Source::EnvOverride => wt_path,
                     _ => git_dir.join(wt_path).into(),
                 };
-                worktree_dir = gix_path::normalize(wt_path, current_dir).map(Cow::into_owned);
+
+                // the reason we use realpath instead of gix_path::normalize here is because there
+                // could be any intermediate symlinks (for example due to a symlinked .git
+                // directory)
+                worktree_dir = gix_path::realpath(&wt_path).ok();
```
Byron, line 308: Using realpath should be conditional, based on if the path exists or not.
Because what's under test here really is that `normalize` can mess up paths that have symlinks in them, yielding an invalid path.

So I think there should even be a warning in `gix_path::normalize()` to tell people to check if the path is still valid.
In a follow-up, one should check all call-sites as they will have the same issue.

gix/tests/gix/status.rs
```diff
@@ -310,6 +310,27 @@ mod index_worktree {
             );
         }
 
+        #[test]
+        fn submodule_in_symlinked_dir() -> crate::Result {
```
Byron, line 314: This test should be removed in favor of specific tests which make clear what's under test.
This test 'only' tests if the status succeeds, otherwise it fails with something like.

```
Error: IndexAsWorktreeWithRenames(TrackedFileModifications(SubmoduleStatus { rela_path: "m1", source: SubmoduleStatus(NextStatusItem(IndexWorktree(IndexAsWorktreeWithRenames(DirWalk(WorktreeRootIsFile { root: "tests/fixtures/generated-do-not-edit/make_submodule_with_symlinked_git_dir/3844911589-unix/m1" }))))) }))
```

gix/src/open/repository.rs
```diff
@@ -301,7 +301,18 @@ impl ThreadSafeRepository {
...
+                // restore the relative path if possible after resolving the absolute path
+                if wt_path.is_relative() {
+                    if let Some(rel_path) = worktree_dir.as_deref().and_then(|p| p.strip_prefix(current_dir).ok()) {
+                        worktree_dir = Some(rel_path.to_path_buf());
+                    }
+                }
```
Byron, line 314: If we have this, there must be a test for this. Maybe it's OK to leave it out for now, since this symlink handling is already quite a special case.

Byron (changes requested):
Thanks a lot for taking another look!

I now spent some time with this and realise that some changes are needed.
The most important one is a test which at least makes clear what's under test, and something as complex as a status isn't that.

Thanks again for bearing with me here - if you are out of time I could also take it from here.

gix/src/open/repository.rs
arumugamramaswamy: Hey, sorry for the hiatus, life got in the way:)

For this, I don't think existence of the path is a sufficient condition to use the output of normalize.

Imagine that the directory exists, but it isn't the correct folder (the correct folder is behind a symlink) - this would lead to incorrect behavior.

Byron: To me, handling paths and symlinks is difficult, and really needs the right tests to motivate them. I can't argue that I might be out of my depth here, but in the end, we have to find a way to make it maintainable.

Right now, I guess I don't really understand it and admittedly, have lost most context that I once had.

Byron:

PR #2184, opened 2025-09-21
Fix discover::upwards when working dir is the repo

moroten:
The upwards search for the repository directory takes a directory as input and then walks through the parents. It turns out that it was broken when the repository was the same as the working directory.

The code checked when the directory components had been stripped to "", in case the directory was replaced with `cwd.parent()`, so the loop missed to check `cwd` itself. If the input directory was "./something", then "." was checked which then succeeded.


EliahKagan:
The failing test, `changed_and_untracked_and_renamed`, is not necessarily failing due to any change in this PR. It fails occasionally due to #1832, which seems to have been happening more often recently (though it's possible that is due to some other bug causing intermittent failure rather than that one; I have not investigated). The failure here seems to resemble other recent failures that have gone away when the CI job is rerun.

In view of this, I've gone ahead and rerun the failing CI job here to see whether if it fails again or not.

Byron:
Thanks a lot for contributing, and for the improved and added tests!

This portion of the code is hard to understand and admittedly, I don't know why this is working better now. However, I trust that the existing code-coverage is good enough and validated the new tests. it seems sound.

As I couldn't push into this PR, I am merging the commit with #2189 instead, while closing this one.

PS: Infrastructure seems really flaky right now as Rustc installations fail on Windows.

moroten:
The test fails without this patch. I discovered it when calling `gix::discover::upwards("src")` in the repository root failed, but `gix::discover::upwards("./src")`. The parent of `"src"` is `""` but the parent of `"./././src"` is `"."`. Just checking for empty string creates the different behaviour.

Byron:
I have no doubt it's not working, yet I felt I should disclose that I don't understand why it's working. It's merely me not knowing my own code anymore :D. The sibling PR is merged, so this change is included now.

PR #2187, opened 2025-09-22
Don't persist GitHub authentication token in `.git/config` on CI

EliahKagan:
When `actions/checkout` is used to check out the repository on CI, it persists credentials related to the GitHub token in the local scope configuration at `.git/config`, unless `persist-credentials` is explicitly set to `false`. This facilitates subsequent remote operations on the repository that could otherwise fail, but we have no such operations in any of our workflows.

As an added layer of protection to keep these credentials from leaking into logs (or otherwise being displayed or subject to exfiltration) in case there is ever unintended coupling between the operation of the test suite (or any step subsequent to checkout that is used to prepare or run tests or other checks) and the cloned `gitoxide` repository itself, this:

- Adds `persist-credentials: false` in a `with` mapping on every step that uses `actions/checkout`.
- Adds a new CI job that checks that every `actions/checkout` step in any job in any workflow sets `persist-credentials` to `false`.

In addition to usual testing on CI, the `release.yml` workflow is among the workflows changed here, and it has also been tested: https://github.com/EliahKagan/gitoxide/actions/runs/17899238656

See also:

- https://github.com/actions/checkout/blob/main/README.md
  (covers what happens with/without `persist-credentials: false`)
- https://github.com/actions/checkout/issues/485

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v5
+        with:
+          persist-credentials: false
       - uses: extractions/setup-just@v3
       - name: Read the MSRV
         run: |
@@ -60,6 +62,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v5
+        with:
+          persist-credentials: false
       - uses: extractions/setup-just@v3
       - name: Ensure we start out clean
         run: git diff --exit-code
@@ -75,6 +79,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v5
+        with:
+          persist-credentials: false
       - name: Prerequisites
         run: |
           prerequisites=(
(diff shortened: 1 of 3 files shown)


EliahKagan:
I just realized that the way I did the partial squash was different from the way I described it, so the commit that is described as refining the new check is actually the one that adds the new check (with all the refinements included). That is, I had planned to have three commits and I ended up having two, with the first commit claiming to contain some changes that only come in as of the second commit.

That might be better anyway, but it's not better that the commit messages are both slightly wrong; sorry about that, and I hope it doesn't cause any confusion. I think the actual changes are fine, and this issue with the messages doesn't hurt https://github.com/gitbutlerapp/gitbutler/pull/10399 either.

Byron:
I noticed it when looking at this PR but also didn't think it was worth fixing 😅.

PR #2194, opened 2025-09-26
Switch to CodeQL "advanced" setup so it can run on all PRs

### LINKED ISSUE #2193: CodeQL doesn't run on PRs from forks
### Current behavior 😯

We've been using the default setup of CodeQL. Since a few weeks ago, this has been working for scanning Rust code, with no errors. For much longer, it's been scanning our GitHub Actions workflows. However, while this setup scans the main branch and also PRs that originate from within this upstream repository, it doesn't scan PRs originating from forks.

I will fix this shortly. The main reason I'm opening this rather than just putting this information in a PR is so that it's clear why I'm turning off default scanning, which I have to do first before setting up "advanced" scanning (but it will not be off for long).

### Expected behavior 🤔

We would ideally run CodeQL on pull requests originating from forks, so that if such a pull request inadvertently introduces a security bug that CodeQL can catch, then it might catch it earlier. I think this is of especial value for PRs originating from forks that modify our GitHub Actions workflows--such as future PRs originating from my fork--because this can prevent the introduction of security vulnerabilities.

I've learned that CodeQL is actually much more effective than I had realized at identifying most of the common kinds of vulnerabilities that arise in GitHub Actions workflows, other than vulnerabilities due to the incorrect use of the `pull_request_target` event (which currently we do not use at all and do not have on any branch).

This is not to say it's perfect. For example, it did not identify that we could benefit from the changes in #2187, even when run with extended queries enabled. In addition, there is at least one other excellent static analyzer for GitHub Actions, [zizmor](https://github.com/zizmorcore/zizmor), which I think we should use, too. (To be clear, the purpose of this issue is not a feature request for zizmor; I mention it here only to avoid overstating the benefit of CodeQL for GHA, which I think is a considerable benefit even with no overstatement.)

I believe CodeQL is also valuable for Rust--of which we of course have much more of than YAML files defining workflows! Although it will try to run on Dependabot pull requests even in the default setup, I don't think this actually is able to work, whereas I think having a workflow (i.e. the "advanced setup") does work properly on Dependabot PRs. So that's another benefit--mainly for Rust scanning, since Dependabot PRs that upgrade Rust dependencies sometimes need to have commits added to them to adapt to SemVer-breaking changes, and it's useful to have static analysis for those changes; but also potentially for Actions, which may also be upgraded by Dependabot.

### Git behavior

Unsure. I believe the Git project does numerous CI checks and I believe they use some forms of static analysis, but I have not looked into this.

### Steps to reproduce 🕹

Observe that PRs from forks don't have CodeQL checks. Or, see [the documentation](https://docs.github.com/en/code-security/code-scanning/enabling-code-scanning/configuring-default-setup-for-code-scanning#about-default-setup).

EliahKagan:
Fixes #2193

See the individual commits for details, and #2193 for rationale.

This intentionally does not make or recommend that the CodeQL checks be made blocking for PR auto-merge, just as we have deliberately not blocked PRs on the dynamic CodeQL checks from the default setup. (We may of course sometimes wait to see what the outcome is before merging a PR about which there is concern, or which is substantial in ways relevant to something we hope CodeQL to check.)

**Testing:** I've tested this in https://github.com/EliahKagan/gitoxide/pull/107, and temporarily on the main branch of my fork (since rewound). Of course, that doesn't test it with a PR originating from outside of my fork. I'll revisit this configuration later to verify that it is working properly with such PRs.

**Outscoped:** I think we should enable more CodeQL queries such as extended queries, and possibly also make it so forks whose operators want to run CodeQL within the fork can use this configuration rather than overriding it while still changing what queries are run, such as by adjusting per-repo [variable](https://docs.github.com/en/actions/reference/workflows-and-actions/variables). But this PR deliberately does not include any of that. To be clear, the CodeQL configuration introduced here *does* scan with all the queries that were in use previously in this upstream repository (just not all the ones I had been testing out in my fork).

~~This is a draft while I wait for the CodeQL checks in this PR itself to complete.~~ Done.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -9,15 +9,20 @@
 # the `language` matrix defined below to confirm you have the correct set of
 # supported CodeQL languages.
 #
-name: "CodeQL Advanced"
+name: CodeQL
 
 on:
   push:
-    branches: [ "main" ]
+    branches:
+      - main
+      - 'run-ci/**'
+      - '**/run-ci/**'
   pull_request:
-    branches: [ "main" ]
+    branches:
+      - main
   schedule:
     - cron: '32 3 * * 6'
+  workflow_dispatch:
 
 jobs:
   analyze:
@@ -33,11 +38,11 @@ jobs:
       security-events: write
 
       # required to fetch internal or private CodeQL packs
-      packages: read
+      # packages: read
 
       # only required for workflows in private repositories
-      actions: read
-      contents: read
+      # actions: read
+      # contents: read
 
     strategy:
       fail-fast: false
@@ -57,7 +62,9 @@ jobs:
         # your codebase is analyzed, see https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/codeql-code-scanning-for-compiled-languages
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@v5
+      with:
+        persist-credentials: false
 
     # Add any setup steps before running the `github/codeql-action/init` action.
     # This includes steps like installing compilers or runtimes (`actions/setup-node`


EliahKagan (commented):
This seems to be working. Here's the checks it produced on this PR itself:

- [Actions code scan](https://github.com/GitoxideLabs/gitoxide/actions/runs/18047002598/job/51360097825?pr=2194)
- [Rust code scan](https://github.com/GitoxideLabs/gitoxide/actions/runs/18047002598/job/51360097842?pr=2194)
- [Code scanning results](https://github.com/GitoxideLabs/gitoxide/pull/2194/checks?check_run_id=51360148579)

PR #2197, opened 2025-09-28
[WIP] Add script to generate tests cases for `gix-diff`

cruessler:
This PR is about 75–80 % done (hopefully). It still needs comments and there’s also a couple of `TODO` comments that I want to go through. I already wanted to open it at this early stage, kind of as a sneak preview of the things I’m currently working on. :-) I’ll update the PR as well as this description once I get to it. If not this week, then the one after that.

### DIFF
--- a/gix-diff/tests/diff/blob/slider.rs
+++ b/gix-diff/tests/diff/blob/slider.rs
@@ -1,0 +1,148 @@
+use std::path::PathBuf;
+
+use gix_diff::blob::{
+    intern::TokenSource,
+    unified_diff::{ConsumeBinaryHunk, ContextSize},
+    Algorithm, UnifiedDiff,
+};
+use gix_object::FindExt;
+use gix_ref::{bstr::BString, file::ReferenceExt};
+
+struct Baseline {
+    hunks: (),
+}
+
+mod baseline {}
+
+#[test]
+fn sliders() -> gix_testtools::Result {
+    let worktree_path = fixture_path()?;
+    let git_dir = worktree_path.join(".git");
+    let odb = gix_odb::at(git_dir.join("objects"))?;
+    let refs = gix_ref::file::Store::at(git_dir.clone(), gix_ref::store::init::Options::default());
+    // In `../tree.rs`, there is `head_of` which would obviate the need for using `gix_ref`.
+    let mut head = refs.find("HEAD")?;
+    let head_id = head.peel_to_id(&refs, &odb)?;
+
+    let commits = gix_traverse::commit::Simple::new(Some(head_id), &odb)
+        .map(Result::unwrap)
+        .map(|commit| commit.id)
+        .collect::<Vec<_>>();
+
+    let index = gix_index::File::at(git_dir.join("index"), gix_hash::Kind::Sha1, false, Default::default())?;
+    let stack = gix_worktree::Stack::from_state_and_ignore_case(
+        worktree_path.clone(),
+        false,
+        gix_worktree::stack::State::AttributesAndIgnoreStack {
+            attributes: Default::default(),
+            ignore: Default::default(),
+        },
+        &index,
+        index.path_backing(),
+    );
+    let capabilities = gix_fs::Capabilities::probe(&git_dir);
+    let mut resource_cache = gix_diff::blob::Platform::new(
+        Default::default(),
+        gix_diff::blob::Pipeline::new(
+            gix_diff::blob::pipeline::WorktreeRoots {
+                old_root: None,
+                new_root: None,
+            },
+            gix_filter::Pipeline::new(Default::default(), Default::default()),
+            vec![],
+            gix_diff::blob::pipeline::Options {
+                large_file_threshold_bytes: 0,
+                fs: capabilities,
+            },
+        ),
+        gix_diff::blob::pipeline::Mode::ToGit,
+        stack,
+    );
+
+    assert!(commits.len() > 0);
+    assert!(commits.len().is_multiple_of(2));
+
+    let mut buffer = Vec::new();
+    let mut buffer2 = Vec::new();
+
+    let mut iter = commits.chunks(2);
+
+    while let Some(&[new, old]) = iter.next() {
+        // TODO: We can extract the duplicate code.
+        let old_commit = odb.find_commit(&old, &mut buffer)?.to_owned();
+        let new_commit = odb.find_commit(&new, &mut buffer)?.to_owned();
+        let file_path = BString::from(old_commit.message.trim_ascii_end());
+
+        eprintln!("diffing {old} and {new}, file path {file_path}");
+
+        let old_tree = old_commit.tree;
+        let old_blob_id = odb
+            .find_tree(&old_tree, &mut buffer2)?
+            .entries
+            .iter()
+            .find(|entry| file_path.eq(entry.filename))
+            .unwrap()
+            .oid
+            .to_owned();
+
+        let new_tree = new_commit.tree;
+        let new_blob_id = odb
+            .find_tree(&new_tree, &mut buffer2)?
+            .entries
+            .iter()
+            .find(|entry| entry.filename == file_path)
+            .unwrap()
+            .oid
+            .to_owned();
+
+        resource_cache.set_resource(
+            old_blob_id,
+            gix_object::tree::EntryKind::Blob,
+            file_path.as_ref(),
+            gix_diff::blob::ResourceKind::OldOrSource,
+            &odb,
+        )?;
+        resource_cache.set_resource(
+            new_blob_id,
+            gix_object::tree::EntryKind::Blob,
+            file_path.as_ref(),
+            gix_diff::blob::ResourceKind::NewOrDestination,
+            &odb,
+        )?;
+
+        let outcome = resource_cache.prepare_diff()?;
+
+        //let old_string: gix_ref::bstr::BString = outcome.old.data.as_slice().unwrap().into();
+        //let new_string: gix_ref::bstr::BString = outcome.new.data.as_slice().unwrap().into();
+        //eprintln!("{old_string}");
+        //eprintln!("{new_string}");
+
+        let interner = gix_diff::blob::intern::InternedInput::new(
+            tokens_for_diffing(outcome.old.data.as_slice().unwrap_or_default()),
+            tokens_for_diffing(outcome.new.data.as_slice().unwrap_or_default()),
+        );
+
+        let actual = gix_diff::blob::diff(
+            Algorithm::Myers,
+            &interner,
+            UnifiedDiff::new(
+                &interner,
+                // TODO: use `ConsumeHunk` instead.
+                ConsumeBinaryHunk::new(String::new(), "\n"),
+                ContextSize::symmetrical(3),
+            ),
+        )?;
+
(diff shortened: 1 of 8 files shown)


tests/it/src/commands/create_diff_cases.rs
```diff
@@ -0,0 +1,129 @@
...
+
+            blocks.push(format!(
+                r#"cp "$ROOT/{asset_dir}/{old_blob_id}.commit" ./{old_blob_id}
+git add {old_blob_id}
+git commit -m {old_blob_id}
+cp "$ROOT/{asset_dir}/{new_blob_id}.commit" ./{old_blob_id}
+git add {old_blob_id}
+git commit -m "{old_blob_id} -> {new_blob_id}"
+
+git diff HEAD^ HEAD > .git/$(git rev-parse HEAD^)-$(git rev-parse HEAD).baseline
```
Byron: I am wondering if we need to create a Git repository here as all we are interested in is pairs of blobs to diff.
It seems like `gix-diff` isn't particularly well equipped to deal with this as it can't/shouldn't use `gix`, so it feels like an unnecessary abstraction.

Something that also came to my mind is that in the places where we want a whole repository, encoding it as a "assets + script" seems like an over engineered solution compared to merely storing a Git repository using something like `mv .git assets/git.tracked && git add assets/git.tracked`.

gix-diff/tests/diff/blob/slider.rs
```diff
@@ -0,0 +1,298 @@
...
+            assert!(line.starts_with(b"+++ "));
+        }
+    }
+
+    impl Iterator for Baseline<'_> {
+        type Item = DiffHunk;
+
+        fn next(&mut self) -> Option<Self::Item> {
+            static HEADER_REGEX: Lazy<Regex> = Lazy::new(|| {
+                Regex::new(r"@@ -([[:digit:]]+),([[:digit:]]+) \+([[:digit:]]+),([[:digit:]]+) @@")
```
Byron: Let's not use the `regex` dependency to parse strings of a diff header, it's trivial enough to write or generate one that is good enough for testing and deals with number conversions as part of that.

tests/it/src/commands/create_diff_cases.rs
Byron: Regarding my previous comment: I see how the Git repo is 'double-duty' - it holds the assets, but it also creates the baseline using Git itself which is much easier with a repository.

And as we already have a precedent to 'encode' Git repositories like this, I suppose we can keep doing it just because. However, I'd love to know if you also have a hunch that it's over-engineered compared to storing Git repositories as a whole.

cruessler: That’s a very helpful suggestion as it made me go read `git diff`’s docs again. I discovered that there is `git diff [<options>] --no-index [--] <path> <path> [<pathspec>...]` which looks like we’ll be able to use it to produce diffs without first creating a history. https://git-scm.com/docs/git-diff#Documentation/git-diff.txt-gitdiffoptions--no-index--pathpathpathspec

To address that larger question: yes, it felt like some effort was necessary to create the script and then access the repository using only lower-level crates. It looks like all the information we need could also be encoded as `<blob id before>-<blob id after>.baseline`. I’ll simplify the code accordingly!

cruessler:
> There is a major question I have that isn't mentioned inline: Is it easy to read/understand the output of the pretty assertion that compares both diffs using the internal hunk format? How does that fare in comparison to diffing unified diffs?

This is two different diffs, the first one uses the internal hunk format, the second one uses unified diffs. I think the second one is significantly easier to understand, so I changed the test to use that format in the last commit.

### Using the internal hunk format

diffing-non-textual-diffs

### Using unified diffs

diffing-textual-diffs

tests/it/src/commands/create_diff_cases.rs
cruessler: I was indeed able to delete a lot of code by using `git diff --no-index` to create the baselines.

gix-diff/tests/diff/blob/slider.rs
cruessler: I removed the dependency on `regex` and replaced it with a trivial parser.

tests/it/src/commands/create_diff_cases.rs
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        let mut buf = Vec::new();
+
+        let script_name = "make_diff_for_sliders_repo.sh";
+
+        let mut blocks: Vec<String> = vec![format!(
+            r#"#!/usr/bin/env bash
+# TODO:
+# `git diff --no-index` returns 1 when there's differences, but 1 is treated as an error by the
```
Byron: I'd keep the `set -eu…` and do `git diff … || true`  or something like that, assuming that it will keep working.

tests/it/src/commands/create_diff_cases.rs
```diff
@@ -0,0 +1,129 @@
...
+                let old_blob = repo.objects.find_blob(&old_blob_id, &mut buf)?.data;
+                std::fs::write(dst_old_blob, old_blob)?;
+
+                let new_blob = repo.objects.find_blob(&new_blob_id, &mut buf)?.data;
+                std::fs::write(dst_new_blob, new_blob)?;
+            }
+
+            blocks.push(format!(
+                r#"git diff --no-index "$ROOT/{asset_dir}/{old_blob_id}.commit" "$ROOT/{asset_dir}/{new_blob_id}.commit" > .git/{old_blob_id}-{new_blob_id}.baseline
+cp "$ROOT/{asset_dir}/{old_blob_id}.commit" assets/
```
Byron: Are these `.commit`s?

Byron (commented):
Thanks, this looks so much simpler now!

The diffs are also much easier to understand, yet I am wondering: would an actual unified diff be better to diff against another unified diff? That would retain diff lines, and I think the common term for it is 'interdiff'.

tests/it/src/commands/create_diff_cases.rs
cruessler: Thanks for the suggestion, changed!

tests/it/src/commands/create_diff_cases.rs
cruessler: Thanks for the suggestion, changed to `.blob`.

cruessler:
> The diffs are also much easier to understand, yet I am wondering: would an actual unified diff be better to diff against another unified diff? That would retain diff lines, and I think the common term for it is 'interdiff'.

That’s possible and it’s easier than I thought it would be. This is the result:

diffing-unified-diffs

Byron:
Thanks for trying it out, and… it does seem to be easier on the eyes. If you think it's better, I guess that will be the way differences are presented.

On another note, I did take a look at how `gitoxide` is using imara-diff and realised that what we'd really want is a mix of imara 0.1 and 0.2!
0.1 is great for line counting (and the fastest possible implementation as it avoids allocation), while 0.2 is better for post-processing, and in theory we could use the slider implementation from there and test it here.

This is just to throw it out there - right now we have some tech-debt due to the 0.1/0.2 chasm, and this feels like a good opportunity to fix it.

cruessler:
Quick update: this is almost ready for a round of reviews! There’s one thing I want to add before I declare it ready, and that is setting the diff algorithm for creating the baseline as well as the comparison diffs, but that should hopefully not take much more than a couple of days.

cruessler:
@Byron I think/hope that this PR is now ready for a first round of reviews! (If you don’t feel it is, feel free to set it back to draft. :smile:) My main question is: is this useful in its current state? Is there enough context for people to understand what the new command is supposed to do?

There’s still a few `todo!()`s in the code. I would be fine with leaving them as as, turning them into `unimplemented!()` or implementing some kind of error handling if you think this would be useful. I wasn’t sure what level of error handling is desired in this developer tool, so that’s still an open question.

gix-diff/tests/fixtures/make_diff_for_sliders_repo.sh
Byron, line 1: I love to keep this info, but it feels it would be more accessible in `gix-diff/README.md` (instead of a seemingly fake shell script).
Or am I missing something, maybe the placeholder is needed for something?

Byron: And in that file, could you provide a snipped, behind `<details><summary>…` that I can use as a diff slider example?
That way I don't have to repeat all these steps and… run `python`… (yes, I am snobby ;)), just to try it out.

Byron (changes requested):
Thanks for the ping :)!

I also agree it's ready for review, and I even think it's *basically* ready for merge.
However, before that, I want to try the tests myself, comments regarding that are below.

Once I tried it, I think I will merge it right away.

Thanks again.

gix-diff/tests/fixtures/make_diff_for_sliders_repo.sh
cruessler: As far as I can tell, `make_diff_for_sliders_repo.sh` needs to be present for this line not to fail: https://github.com/cruessler/gitoxide/blob/dd5127f54ce4e055e6a87719b4962482cd82090d/gix-diff/tests/diff/blob/slider.rs#L257. I just deleted the script locally which made the associated test fail with a file not found message.

What do you mean by “diff slider example”? If you want the test to run on one single diff in order to compare the git baseline against `gix-diff`, you would need two `.blob` files in `assets` as well as the corresponding code in the script to generate the `.baseline` files (one for each of the 2 diff algorithms currently being tested). While that, in an of itself, would be easy to add, the resulting test would most likely fail as `gix-diff` doesn’t yet match the git baseline in most cases as far as I can tell. Which is why I didn’t even try in the first place. 😄

cruessler: Apart from that, I was also wondering where the best place for the info that currently lives in the script is. Feel free to move it to a more appropriate place, the `README.md` seems like a good choice to me!

Byron: This clearly shows that I don't fully understand the setup yet. What I read made me think the slider file is all that's needed and the test only depends on that, but that's not the case as it doesn't include the actual blobs. It just refers to them by ID.

So I will be back with this tomorrow probably and spend the time that I probably tried to safe today 😅.

cruessler: That kind of feedback is already very valuable! 😄 I’m happy to provide more context, hoping to make this a bit more self-contained.

Byron: I actually think you already did everything you could and I will use my ignorance and naiveté towards that matter to make it work for other readers with a similar level of ignorance and naiveté :D. So I think I am predestined to do that 😁.

Byron (approved):
Thanks a lot! This is a nice 'baseline' for future improvements to testing the diverging sliders.

The failure case seems pretty human-readable as well.

Image

Something I can imagine as future steps is to make use of a more recent version of imara-diff for diffing blobs, which would allow the usage of the new slider heuristics.

Alternatively, we could implement our own based in imara-diff v0.1, whatever is easier.

I still like v0.1, but also see that for many typical usecases, one will have to collect hunks anyway.

PR #2201, opened 2025-10-02
Tell Dependabot to keep `expectrl` and `getrandom` back for now

EliahKagan:
This allows them to have patch version updates but not major and minor version updates. Becuase they are not yet at 1.0.0, this has the effect of allowing SemVer-compatible but not SemVer-breaking updates to them via Dependabot version update PRs.

This may be temporary and, in the case of `getrandom`, is intended to be temporary. For details, see comments in `dependabot.yml`, and:

- https://github.com/GitoxideLabs/gitoxide/pull/2200#issuecomment-3361407300
- https://github.com/GitoxideLabs/gitoxide/pull/2093#issuecomment-3361449228

This also expands the comment for holding back `imara-diff` (all referenced PRs in the file now have full URLs, not just numbers).

The effect of keeping back `expectrl` was tested and verified to be the only new hold needed to allow version updates to work, in: https://github.com/EliahKagan/gitoxide/pull/111

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,7 +10,19 @@ updates:
     allow:
       - dependency-type: all
     ignore:
-      # Keep imara-diff at 0.1.* for now (see comments in #2068).
+      # Keep `expectrl` at 0.7.* for now. See https://github.com/GitoxideLabs/gitoxide/pull/2200.
+      - dependency-name: expectrl
+        update-types:
+          - 'version-update:semver-major'
+          - 'version-update:semver-minor'
+      # Some `getrandom` dependencies are at 0.2.*. Keep them there for now. (Some are at 0.3.*.
+      # This allows them to remain there but keeps them from going to a future 0.4.*.) This is
+      # hopefully temporary. See https://github.com/GitoxideLabs/gitoxide/pull/2093.
+      - dependency-name: getrandom
+        update-types:
+          - 'version-update:semver-major'
+          - 'version-update:semver-minor'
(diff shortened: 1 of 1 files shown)


EliahKagan (commented):
It's a bit annoying that the automatic CI job GitHub provides to check that `dependabot.yml` is valid does't usually run on PRs. After this merges, that check will be done, revealing whether or not I've made a mistake here.

(Sometimes I merge these in my own fork first to test their effects, but in this case that wasn't necessary to test whether version updates would be able to proceed with no other changes we can't easily adapt to. Also, it's time-consuming to merge these into my own fork, test them, and force-push back, partly because of the time to do it and the way it interferes with other work in my fork when done, but more so because Dependabot version update scans take a very long time very time they're run, both here and in my fork. The less imporant of the changes here--keeping back `getrandm`--will hopefully help somewhat with that, but I don't know how much.)

Byron:
> (Sometimes I merge these in my own fork first to test their effects, but in this case that wasn't necessary to test whether version updates would be able to proceed with no other changes we can't easily adapt to. Also, it's time-consuming to merge these into my own fork, test them, and force-push back, partly because of the time to do it and the way it interferes with other work in my fork when done, but more so because Dependabot version update scans take a very long time very time they're run, both here and in my fork. The less imporant of the changes here--keeping back `getrandom`--will hopefully help somewhat with that, but I don't know how much.)

No worries at all. I am totally OK with 'risking' minor breakage in the main repository if this saves a lot of hassle, while being easy to fix once it's clear if it's breaking or not.

PR #2204, opened 2025-10-06
Improve blame ranges

cruessler:
This PR is a continuation of #1976. I took the original PR by @holodorum and applied the suggestions I had given in my reviews of his PR. I also added a few tests and adapted some parts to `gitoxide`’s conventions.

These are the most important changes:
- a rename from `range` to `ranges` that was done so the variable name now corresponds to the type name `BlameRanges`.
- the conversion of `BlameRanges` into an enum, to more idiomatically cover the special case `WholeFile`.
- making sure `BlameRanges` keeps its ranges sorted as that is important for the blame algorithm.

There’s one decision that requires particular attention. When you add a range to a `WholeFile`, it is turned into a `PartialFile` containing the given range. This mirrors the difference of `git blame` vs. `git blame -L a,b`. The initial version of the PR returned an error instead, but that makes the API less ergonomic, so I decided to make the change.

### DIFF
--- a/gix-blame/src/error.rs
+++ b/gix-blame/src/error.rs
@@ -30,7 +30,7 @@ pub enum Error {
     #[error(transparent)]
     DiffTreeWithRewrites(#[from] gix_diff::tree_with_rewrites::Error),
     #[error("Invalid line range was given, line range is expected to be a 1-based inclusive range in the format '<start>,<end>'")]
-    InvalidLineRange,
+    InvalidOneBasedLineRange,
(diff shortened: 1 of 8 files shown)


Byron:
Did you say this is ready for merging and hanging on me?
Maybe I missed this completely.

cruessler:
Yes! If I remember correctly, nothing is missing from my side (though it has been a couple of weeks since I last touched this PR 😅).

PR #2207, opened 2025-10-07
fix: assure that blob merges recognize conflicts in cojunction with empty bases.

Byron:
Previously, there were special cases that would cause it to auto-merge conflicts even though it shouldn't.

This is now fixed just by removing the special case entirely, which seems unexpected in any of the test-suite.


gix-merge/src/blob/builtin_driver/text/function.rs
```diff
@@ -115,65 +115,60 @@ pub fn merge<'a>(
                 let last_hunk = last_hunk(front_hunks, our_hunks, their_hunks, back_hunks);
                 write_ancestor(input, ancestor_integrated_until, first_hunk.before.start as usize, out);
                 write_hunks(front_hunks, input, &current_tokens, out);
-                if their_hunks.is_empty() {
-                    write_hunks(our_hunks, input, &current_tokens, out);
-                } else if our_hunks.is_empty() {
-                    write_hunks(their_hunks, input, &current_tokens, out);
-                } else {
```
Byron, line 122: @Caleb-T-Owens Removing this special case resolved the issue you discovered. Turns out it was entirely unnecessary, and besides the test-case that validates it specifically, there was only one more failure due to that in existing tests, and as it was already a deviation, it was overlooked.

Caleb-T-Owens: Neat, thank you for looking into this!

Byron:
Forcing merge as Windows has weird errors that seem related to … OS flakiness.

EliahKagan:
Regarding the Windows failures:

```text
       ABORT [   0.018s] gix-ref-tests::refs file::log::iter::backward::with_buffer_big_enough_for_largest_line::realistic_logs_can_be_read_completely
           - with code 0xc0000142: A dynamic link library (DLL) initialization routine failed. (os error 1114)
    (test aborted)
       ABORT [   0.019s] gix-ref-tests::refs file::log::iter::backward::with_buffer_big_enough_for_largest_line::single_line
           - with code 0xc0000142: A dynamic link library (DLL) initialization routine failed. (os error 1114)
       ...
```

I haven't seen that one before! I don't have time to look into it now but I'll try to do so later. I'm also reruning the test to see if it happens again. Maybe there is a problem on the runner?

PR #2208, opened 2025-10-07
[WIP] Add commit signing

cruessler:
This is a rough, incomplete prototype, mainly intended to allow me to ask a couple of questions on how to implement various things.

### DIFF
--- a/gitoxide-core/src/repository/commit.rs
+++ b/gitoxide-core/src/repository/commit.rs
@@ -1,6 +1,14 @@
-use std::{io::Write, process::Stdio};
+use std::{
+    borrow::Cow,
+    io::{Read, Write},
+    process::Stdio,
+};
 
 use anyhow::{anyhow, bail, Context, Result};
+use gix::{
+    bstr::{BStr, ByteSlice},
+    objs::commit::SIGNATURE_FIELD_NAME,
+};
 
 /// Note that this is a quick implementation of commit signature verification that ignores a lot of what
 /// git does and can do, while focussing on the gist of it.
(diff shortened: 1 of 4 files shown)


gitoxide-core/src/repository/commit.rs
```diff
@@ -39,6 +47,60 @@ pub fn verify(repo: gix::Repository, rev_spec: Option<&str>) -> Result<()> {
...
+
+    let extra_header: Cow<'_, BStr> = Cow::Borrowed(signed_data.as_bstr());
+
+    // TODO:
+    // What do we want to do when there is already a signature?
+    commit_ref
+        .extra_headers
+        .push((BStr::new(SIGNATURE_FIELD_NAME), extra_header));
+
+    eprintln!("{commit_ref:?}");
```
cruessler: @Byron At this point, I would like to write the signed commit back to the filesystem. I searched the codebase for hints on how this is usually done, but couldn’t really find anything. Would you be able to point me to an example?

Byron: That would be `repo.write_object(&commit_ref)?`.

It's great you use the `commit_ref` for this by the way as it means any commit, even *somewhat* malformed ones, can be signed with this.

cruessler: Thanks, that’s what I was looking for! I assume that, once the signed object has been written, `rev_spec` needs to be made to point to the new commit, and the reflog needs to be updated. Is that what I’d use `repo.edit_reference()` for?

Byron: Right - you'd want to now point `HEAD` to the new commit, which won't change the worktree at all.
Editing the head can really use a specific API, but while it's not there, you'd indeed need to use `edit_references([head_edit])` to edit only `HEAD` with `deref: true`. No value should stay the default just to be explicit.

cruessler:
> There was some amount of refactor if you want to take a look.

Thanks for the refactoring, helpful as always! 🎉

PR #2219, opened 2025-10-15
Replace `split_at_pos` by `split_at_checked`

cruessler:
Now that #2217 has been merged, this is a follow-up to replace `split_at_pos` by `split_at_checked` which was stabilized in 1.80. This does not change behaviour, it just replaces a custom function by the equivalent one from `std`.


Byron (approved):
That is a great catch, thanks o much!

Issue #2152, opened 2025-09-05
Non-additive transport features are annoying to deal with


djc:
### Current behavior 😯

Cargo features are intended to be additive and non-additive features are making it substantially harder to deal with the gix crates. It would be very nice if the curl and reqwest backends could be rewired so that the caller must pick a backend and the crates can still be compiled if both are enabled.

(Noticed this again while working on https://github.com/rustsec/rustsec/pull/1408.)

### Expected behavior 🤔

Compile with `--all-features` without this:

```rust
error[E0428]: the name `Impl` is defined multiple times
   --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/gix-transport-0.48.0/src/client/blocking_io/http/mod.rs:220:1
    |
217 | pub type Impl = curl::Curl;
    | --------------------------- previous definition of the type `Impl` here
...
220 | pub type Impl = reqwest::Remote;
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `Impl` redefined here
    |
    = note: `Impl` must be defined only once in the type namespace of this module

error: Cannot set both 'http-client-reqwest' and 'http-client-curl' features as they are mutually exclusive
  --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/gix-transport-0.48.0/src/client/blocking_io/http/mod.rs:26:1
   |
26 | compile_error!("Cannot set both 'http-client-reqwest' and 'http-client-curl' features as they are mutually exclusive");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0428`.
```

Byron:
Thanks for reporting, and putting it onto the map.

When implementing it like this, mutually exclusive features that should be additive, I didn't have a solution as the goals are competing. How can a rewiring be done given such constraints?

If no feature is selected, that's fine I think as there doesn't have to be support for transports at all (and the local transport will always work if I remember correctly). So what's left is to decide which feature takes precedence if multiple are selected, but that looks like a footgun that is rather be avoided: ambiguity, while silently making a choice for the user.

Right now, the constraint is violated to not have to make such a choice.

[...]

PR #2220, opened 2025-10-15
Deduplicate packetline crates

djc:
Trying to make a little progress on #2152 (and this copied crate stuff seems extremely ugly anyway).

Why does `cargo check` pass while `cargo check -p gix` break? Something to do with feature unification? This means I'm not getting feedback from VS Code on errors, which makes it much harder to get stuff done.

### DIFF
--- a/gix-packetline/src/async_io/encode.rs
+++ b/gix-packetline/src/async_io/encode.rs
@@ -209,14 +209,14 @@ pub async fn flush_to_write(mut out: impl AsyncWrite + Unpin) -> io::Result<usiz
     Ok(4)
 }
 
-/// Write `data` of `kind` to `out` using side-band encoding.
+/// Write `data` of `kind` to `out` using sideband encoding.
 pub async fn band_to_write(kind: Channel, data: &[u8], out: impl AsyncWrite + Unpin) -> io::Result<usize> {
     prefixed_data_to_write(&[kind as u8], data, out).await
 }
 
-/// Serialize this instance to `out`, returning the amount of bytes written.
+/// Serialize `band` to `out`, returning the amount of bytes written.
 ///
-/// The data written to `out` can be decoded with [`Borrowed::decode_band()]`.
+/// The data written to `out` can be decoded with [`crate::PacketLineRef::decode_band()`].
 pub async fn write_band(band: &BandRef<'_>, out: impl AsyncWrite + Unpin) -> io::Result<usize> {
     match band {
         BandRef::Data(d) => band_to_write(Channel::Data, d, out),
@@ -226,19 +226,19 @@ pub async fn write_band(band: &BandRef<'_>, out: impl AsyncWrite + Unpin) -> io:
     .await
 }
 
-/// Serialize this instance to `out`, appending a newline if there is none, returning the amount of bytes written.
+/// Serialize `band` to `out`, appending a newline if there is none, returning the amount of bytes written.
 pub async fn write_text(text: &TextRef<'_>, out: impl AsyncWrite + Unpin) -> io::Result<usize> {
     text_to_write(text.0, out).await
 }
 
-/// Serialize this line as error to `out`.
+/// Serialize `error` to `out`.
 ///
-/// This includes a marker to allow decoding it outside of a side-band channel, returning the amount of bytes written.
+/// This includes a marker to allow decoding it outside a sideband channel, returning the amount of bytes written.
 pub async fn write_error(error: &ErrorRef<'_>, out: impl AsyncWrite + Unpin) -> io::Result<usize> {
     error_to_write(error.0, out).await
 }
 
-/// Serialize this instance to `out` in git `packetline` format, returning the amount of bytes written to `out`.
+/// Serialize `line` to `out` in git `packetline` format, returning the amount of bytes written to `out`.
 pub async fn write_packet_line(line: &PacketLineRef<'_>, out: impl AsyncWrite + Unpin) -> io::Result<usize> {
     match line {
         PacketLineRef::Data(d) => data_to_write(d, out).await,
--- a/gix-packetline/src/lib.rs
+++ b/gix-packetline/src/lib.rs
@@ -36,9 +36,10 @@ pub mod blocking_io {
     pub use write::Writer;
 }
 
+/// Various utilities for `io::Read` trait implementation.
 ///
+/// Only useful in conjunction with the `async-io` and `blocking-io` cargo features.
 pub mod read;
-pub use read::{ProgressAction, StreamingPeekableIterState};
 
 const U16_HEX_BYTES: usize = 4;
 const MAX_DATA_LEN: usize = 65516;
@@ -48,7 +49,7 @@ const DELIMITER_LINE: &[u8] = b"0001";
 const RESPONSE_END_LINE: &[u8] = b"0002";
 const ERR_PREFIX: &[u8] = b"ERR ";
 
-/// One of three side-band types allowing to multiplex information over a single connection.
+/// One of three sideband types allowing to multiplex information over a single connection.
 #[derive(PartialEq, Eq, Debug, Hash, Ord, PartialOrd, Clone, Copy)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub enum Channel {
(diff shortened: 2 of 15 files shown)


Byron:
Thanks for giving it a shot, it's much appreciated!

Feature-unification is the reason for requiring a differently named copy. It's very ugly, but there was no other way.
It all goes down to wanting to be able to 'switch' async on or off with minimal code duplication and shared tests. This is how one atrocity led to another.
Like mentioned before, I'd live better without `async` entirely (in this codebase).

djc:
> Feature-unification is the reason for requiring a differently named copy. It's very ugly, but there was no other way.
> It all goes down to wanting to be able to 'switch' async on or off with minimal code duplication and shared tests. This is how one atrocity led to another.

Do you have reason to think the approach in this PR would not work? It seems like an existence proof contradicting that "there was no other way" (and I don't think it introduces substantial code duplication, although the tests might be a bit harder).

Byron:
Apologies, I didn't even look at the code yet and wanted to wait until you think it's ready for review.
My statement was about how this came to be, it's not the wish to torture users, but to solve a specific problem that led down this path.
Sometimes other solutions aren't seen in time leading to suboptimal results, and I genuinely hope that this PR can show the way.

djc:
> Apologies, I didn't even look at the code yet and wanted to wait until you think it's ready for review.

It obviously doesn't pass CI yet, but I'd like some early feedback before I try to tackle the remaining issues. I think the idea is conceptually simple: the gix-packetline crate offers both async and sync APIs, separately (both are toggled by a Cargo feature flag), and consumers explicitly use the API for the mode they want to use. I don't think there's substantially more library-level code duplication. For now, this leaves mutually exclusive features in higher-level crates alone, at some point the I/O mode will bubble up to the top-level API and then we have to decide how to best present that.

gix-filter/Cargo.toml
```diff
@@ -22,7 +22,7 @@ gix-command = { version = "^0.6.2", path = "../gix-command" }
 gix-quote = { version = "^0.6.0", path = "../gix-quote" }
 gix-utils = { version = "^0.3.0", path = "../gix-utils" }
 gix-path = { version = "^0.10.20", path = "../gix-path" }
-gix-packetline-blocking = { version = "^0.19.1", path = "../gix-packetline-blocking" }
+gix-packetline = { version = "^0.19.1", path = "../gix-packetline", features = ["blocking-io"] }
```
Byron: And this is what can't be happening. Setting this to blocking sets it to blocking for everyone in the dependency tree, even though there maybe others who want it async.

This PR solves this by making `gix-packetline` features additive, which *might* be OK to adjust here assuming that the code that uses it already is clearly split into sync and async.

If that's the case, I think it makes total sense to continue this effort.

Byron (commented):
I think I understand now what this PR is going for and have a feeling that it *could* work. Maybe `gix-packetline` never needed have offer one unified codepath that can be 'toggled' async due to its primitive nature.

Now that I think about it, maybe the tests for `gix-packetline` then will be more duplicated, but that should be alright as they are stable.

And yes, I agree that from here more opportunities for similar updates may show up.

gix-filter/Cargo.toml
lorenzleutgeb: In principle, if there are two features `"gix-packetline/blocking-io"` and `"gix-packetline/async-io"` (which seems to be the case IIUC), and these are *additive*, this could be fine.

However, it'd be good to understand why this can't be:
```toml
[features]
blocking-io = ["gix-packetline/blocking-io"]
[dependencies]
gix-packetline = { version = "...", path = "...", features = [] } # for emphasis
```
That is, why can't `gix-filter` also have a feature toggle on its I/O, and propagate down to `gix-packetline`?
If this is impossible, then this must imply that `gix-filter` *somehow intrinsically requires blocking I/O*. And then this would be the next best frontier to push, i.e. to work on getting `gix-filter` to support both blocking and async. Or you decide that this is not worth doing and just declare `gix-filter` to require blocking I/O, full stop. You could future-proof by requiring the feature flag `"blocking-io"` for the crate to compile at all (like `gix-hash` now requires `"sha1"`), just to make room for a potential `"async-io"` down the line.

So long as this is not mutually exclusive with enabling async I/O, this might just be a minor annoyance...

djc: On `main`, gix-filter unconditionally depends on gix-packetline-blocking, which I take to mean that gix-filter indeed depends on the blocking I/O implementation in gix-packetline. Solving that problem is not my goal for now -- allowing the features not to be mutually exclusive is.

djc:
This feels like it's pretty close. Somehow some builds in the full tests end up enabling both `async-client` and `blocking-client` on gix-transport which are supposed to be mutually exclusive on this branch as before -- I'm not sure what's triggering that? It doesn't feel like any of the changes I made should be capable of that, they're all supposed to be upstream of gix-transport.

```rust
djc-2021 dedup-packetline gitoxide $ cargo check --features lean-async
    Checking gix-packetline v0.19.1 (/Users/djc/src/gitoxide/gix-packetline)
    Checking gix-filter v0.20.0 (/Users/djc/src/gitoxide/gix-filter)
    Checking gix-transport v0.48.0 (/Users/djc/src/gitoxide/gix-transport)
error: Cannot set both 'blocking-client' and 'async-client' features as they are mutually exclusive
  --> gix-transport/src/lib.rs:92:1
   |
92 | compile_error!("Cannot set both 'blocking-client' and 'async-client' features as they are mutually exclusive");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

djc:
> This feels like it's pretty close. Somehow some builds in the full tests end up enabling both `async-client` and `blocking-client` on gix-transport which are supposed to be mutually exclusive on this branch as before -- I'm not sure what's triggering that? It doesn't feel like any of the changes I made should be capable of that, they're all supposed to be upstream of gix-transport.
> 
> ```rust
> djc-2021 dedup-packetline gitoxide $ cargo check --features lean-async
>     Checking gix-packetline v0.19.1 (/Users/djc/src/gitoxide/gix-packetline)
>     Checking gix-filter v0.20.0 (/Users/djc/src/gitoxide/gix-filter)
>     Checking gix-transport v0.48.0 (/Users/djc/src/gitoxide/gix-transport)
> error: Cannot set both 'blocking-client' and 'async-client' features as they are mutually exclusive
>   --> gix-transport/src/lib.rs:92:1
>    |
> 92 | compile_error!("Cannot set both 'blocking-client' and 'async-client' features as they are mutually exclusive");
>    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ```

Ahh, I think I just didn't understand the justfile `!` prefix syntax. I think this next run might pass CI. Honestly not sure why it's useful to have CI check that certain configurations don't compile.

djc:
I've tacked on a second commit (which can be reviewed independently) to clean up the gix-packetline API in a way that seems quite a bit nicer to me.

gix-filter/Cargo.toml
Byron: I can confirm that there is no use in `gix-filter` supporting async as it's implemented as sync API, which is also where the need for `gix-packetline-blocking` came from. With `gix-packetline`'s IO features turned additive, it will be able to depend on `gix-packetline` normally.
That should be OK.

gix-packetline/src/async_io/encode.rs
```diff
@@ -212,3 +213,37 @@ pub async fn flush_to_write(mut out: impl AsyncWrite + Unpin) -> io::Result<usiz
 pub async fn band_to_write(kind: Channel, data: &[u8], out: impl AsyncWrite + Unpin) -> io::Result<usize> {
     prefixed_data_to_write(&[kind as u8], data, out).await
 }
+
+/// Serialize this instance to `out`, returning the amount of bytes written.
+///
+/// The data written to `out` can be decoded with [`Borrowed::decode_band()]`.
+pub async fn write_band(band: &BandRef<'_>, out: impl AsyncWrite + Unpin) -> io::Result<usize> {
```
Byron, line 220: Wouldn't this generally be more common to write as `&self` on `BandRef`?
(The same comment applies to the other functions)

gix-packetline/src/read.rs
Byron, line 1: I think this should be `read/mod.rs`.

gix-packetline/src/lib.rs
Byron, line 1: This can be `write.rs`.

gix-packetline/src/lib.rs
```diff
@@ -9,6 +8,38 @@
...
+    pub use read::StreamingPeekableIter;
+    mod sidebands;
+    pub use sidebands::WithSidebands;
+    mod write;
+    pub use write::Writer;
+}
+
+///
+#[cfg(feature = "blocking-io")]
+pub mod blocking_io {
```
Byron, line 28: That's cool, I don't think I have ever seen this kind of style before. I will take some time to wrap my head around it :).

gix-packetline/src/async_io/encode.rs
djc: It can't trivially be because there are two variants, an async and blocking one, which can't share the same name. IMO it's cleaner to keep them separate in their respective I/O mode namespace rather than attached as methods with different names.

djc:
> and didn't yet have time for my refactoring round.

What do you mean by this?

gix-packetline/src/async_io/encode.rs
Byron: Right, it's now additive. Thank you for the clarification.

Byron:
This means I have only done a partial armchair review without actually opening the code in an editor. When I do, I usually just make the changes I need and merge. That saves a lot of time as I don't have to go back and forth.

gix-packetline/src/read.rs
djc: Why? There are no nested modules, right? Putting it in a directory without a need for further nesting seems pretty unidiomatic.

gix-packetline/src/lib.rs
djc: It's not clear what this is referring to.

gix-packetline/src/read.rs
Byron: That's true. To me it looked like there is `read` on the top-level, or does it not?

Screenshot 2025-10-22 at 08 39 04

AaronDewes: Image

The view is a bit confusing, it shows a read directory because it was there previously and has now been deleted.

djc:
Rebased onto current main.

gix-packetline/src/lib.rs
Byron: Sorry, this was me not knowing how to interpret the sidebar on the GH website.

djc:
> Sorry for the wait, and thanks again for tackling this!
> 
> I hope this won't discourage you from continuing on these tracks, and right the wrong in the process 🙏.

See

- #2231 

Also happy to take more feedback and incorporate it, so I can learn any style preferences you have.

PR #2230, opened 2025-10-23
Remove `doc_auto_cfg` feature

sylfn:
It is part of `doc_cfg` feature since https://github.com/rust-lang/rust/pull/138907

Should fix docs.rs build. I have tested locally with `cargo docs-rs -p gix` and documentation builds correctly


Byron (approved):
Thanks a lot for the fix, it's much appreciated!

It's bad timing that `docs.rs` is down today, so I can't see how the documentation looks like now. But maybe there is nothing do to the failure, and that's it.

With that said, I will be using this PR to push new versions of the affected crates as well.

*Please note that for my own convenience, I edited your commit message (while breaking its signature) to put the commit-message into the changelog automatically.*

sylfn:
That's a bit weird, but I feel like all links in changelog got corrupted and now point to my fork instead of upstream even in historical entries

Byron:
Oh no :D - next time they are regenerated, they will be fixed, I think.
And even if not, it's a nice historical artefact of tooling gone slightly wrong 😁.

EliahKagan:
> but I feel like all links in changelog got corrupted and now point to my fork

The good news is that they are full OIDs, so they are not likely to ever point to a different/wrong commit, though in the event that you were ever to delete your fork, they would break.

(Renaming your fork or even moving it into an organization or otherwise transferring it would probably not break them, though, because GitHub makes URLs with an old repository name redirect to corresponding URLs with the new repository name, such as how `Byron/gitoxide` URLs redirect to `GitoxideLabs/gitoxide` URLs.)

> next time they are regenerated, they will be fixed, I think.

@Byron I should be able to fix these easily to be `GitoxideLabs/gitoxide` instead of `yuki0iq/gitoxide` URLs. Do you prefer I not do so?

My argument for doing so is that extra changes across tags may make it harder for people to look at what changed when comparing two tags.

Of course, if it is *known* that these changelogs will all be regenerated before the next tag, then no such confusion would arise, whether or not they are fixed before then.

Byron:
Thank you @EliahKagan, but I think it's Ok to keep it until someone complains, and assuming it will be auto-fixed next time. I think it regenerates old blocks anyway so in a month it's all gone.

PR #2231, opened 2025-10-24
fix!: move TransportWithoutIO::request() method to I/O-specific traits

djc:
The `RequestWriter` type is specific to the I/O mode, so this should not live in the I/O-generic trait.


Byron (approved):

To me it seems that the now removed conditional compilation in `gix-transport/src/client/traits.rs` were a clear indication that something is wrong that needs fixing. I also don't recall why I would have thought this was necessary in the first place, so let's just roll with it.

PR #2232, opened 2025-10-24
fix!: prefer curl over reqwest connections if both are enabled

djc:
Instead of colliding `Impl` type aliases for `http-client-curl` and `http-client-reqwest`, prefer curl (the one that is enabled by default) if both are available.

### DIFF
--- a/gix-transport/src/client/blocking_io/connect.rs
+++ b/gix-transport/src/client/blocking_io/connect.rs
@@ -4,7 +4,7 @@ pub(crate) mod function {
     #[cfg(feature = "http-client-curl")]
     use crate::client::blocking_io::http::curl::Curl;
     #[cfg(feature = "http-client-reqwest")]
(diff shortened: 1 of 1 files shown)


gix-transport/src/client/blocking_io/http/mod.rs
```diff
@@ -215,13 +212,6 @@ impl Default for Options {
     }
 }
 
-/// The actual http client implementation, using curl
-#[cfg(feature = "http-client-curl")]
-pub type Impl = curl::Curl;
-/// The actual http client implementation, using reqwest
-#[cfg(feature = "http-client-reqwest")]
-pub type Impl = reqwest::Remote;
-
```
Byron, line 224: Initially I wanted to bring `Impl` back as it's a nice way to get the implementation of what's actually implemented, but then I realised that this is quite an implementation detail and it's probably unused outside of this crate anyway.

Byron (approved):

PR #2235, opened 2025-10-26
fix: use `std::env::home_dir()` and remove the `home` crate from dependencies. (#2234)

### LINKED ISSUE #2234: Remove use of the home crate
The `home` crate is no longer intended for general use, but instead only for use by cargo and rustup. It is recommended to switch to [`std::env::home_dir`](https://doc.rust-lang.org/nightly/std/env/fn.home_dir.html) instead.

Note that there are a few things to consider here:

- `home_dir`'s behavior was changed on Windows to use USERPROFILE (similar to how `home` worked) in 1.85 via https://github.com/rust-lang/rust/pull/132515.
- `home_dir`'s deprecation was removed in 1.87 via https://github.com/rust-lang/rust/pull/137327.
- `home_dir` uses a fallback on an empty HOME as of 1.90 via https://github.com/rust-lang/rust/pull/141840.

Byron:
Let's hope this works with WASM.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2126,7 +2126,6 @@ dependencies = [
  "gix-testtools",
  "gix-trace",
  "gix-validate",
- "home",
  "serial_test",
  "thiserror 2.0.17",
  "windows 0.62.1",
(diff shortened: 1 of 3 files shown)


Byron:
Also CC @EliahKagan as I am pretty sure this will change behaviour in some circumstances. To my mind this shouldn't be exploitable though, it's *just* a correctness issue. I am thinking that this will affect Windows users if the application was built with an older Rust compiler.

PR #2236, opened 2025-10-26
transport: make I/O mode features additive

djc:


gix-transport/src/client/blocking_io/connect.rs
```diff
@@ -10,10 +10,10 @@ pub(crate) mod function {
     /// A general purpose connector connecting to a repository identified by the given `url`.
     ///
     /// This includes connections to
-    /// [local repositories](crate::client::blocking_io::file::connect()),
```
Byron, line 13: Please do not turn `[name](link)` into `[name][link]` - even though rustdoc is fine with it, my IDE is not and it wants standard markdown for me to be able to jump there.
I do hope that you are not in an IDE or editor where it's the other way around 😅.

gix/src/clone/mod.rs
```diff
@@ -1,12 +1,17 @@
 #![allow(clippy::result_large_err)]
 use crate::{bstr::BString, remote};
 
+#[cfg(feature = "async-network-client")]
+use gix_transport::client::async_io::Transport;
+#[cfg(feature = "blocking-network-client")]
+use gix_transport::client::blocking_io::Transport;
```
Byron, line 7: I am particularly interested to see how this is ever (of course not in this PR) becoming additive without code duplication. But maybe it doesn't have to be, as making the plumbing crates additive is already a win.
In any case, I am staying tuned 🤩.

Byron (commented):
Thanks so much for contributing this!

I am pleasantly surprised how well this works, and it clearly was a mistake to not start additively in the first place. After all, on the consuming side, one can just import the same type name (either async or sync), and existing code will work as before.
So even if this for some reason stops translating up the dependency tree, there is a net-win.

This was just a cursory review, but I think once I can have my rustdoc links back, I can finish it and get this merged.

gix/src/clone/mod.rs
djc: Yes, I think there will be a bit more duplication at the higher levels, hopefully that will still be acceptable.

gix-transport/src/client/blocking_io/connect.rs
djc: Huh, yeah, not sure why that happened. Reverted this change.

gix/src/clone/mod.rs
Byron: I also realise now that as long as callers can use traits, these will work perfectly due to `maybe_async`. So maybe, just maybe, this will work just fine.

Byron (approved):
Fantastic work! Even after looking carefully at everything, I wasn't able to make any of my usual refactoring 😁, so merging as is.

PR #2240, opened 2025-10-30
fix!: respect `diff.algorithm` in `Repository::blame_file`

cruessler:
Previously, consumers of `Repository::blame_file()` had to call `Repository::diff_algorithm()` themselves if they wanted `gix` to take `diff.algorithm` into account. Now, they can pass `diff_algorithm: None` and `Repository::diff_algorithm()` will be called by `Repository::blame_file()`.


Byron (approved):

PR #2241, opened 2025-11-01
Make gix-transport compile with all features

djc:
Some miscellaneous stuff unrelated to transport implementations, and a bunch of stuff I missed because some async code was previously configured out if blocking code was enabled. Test that all features work in CI.

### DIFF
--- a/gix-transport/Cargo.toml
+++ b/gix-transport/Cargo.toml
@@ -18,6 +18,8 @@ doctest = false
 default = []
 
 ## If set, blocking implementations of the typical git transports become available in `crate::client::blocking_io`
+##
+## If used in conjunction with an async implementation, this one takes precedence.
 blocking-client = ["gix-packetline/blocking-io"]
 ## Implies `blocking-client`, and adds support for the http and https transports.
 http-client = [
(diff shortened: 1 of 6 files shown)


Byron (approved):
Thanks so much, and sorry for the delay!

The most notable change that I did was the renaming of `Outcome::from_lines…()` to `Handshake`, which seemed like a better fit.
Now I don't know if this is a mistake because it introduced ambiguity elsewhere, but if you encounter it, I am sure you'd be most qualified to find better names.

gix-transport/Cargo.toml
```diff
@@ -18,6 +18,8 @@ doctest = false
 default = []
 
 ## If set, blocking implementations of the typical git transports become available in `crate::client::blocking_io`
+##
+## If used in conjunction with an async implementation, this one takes precedence.
```
Byron, line 22: This might also be something to double-check @djc, I hope it's correct.

djc:
> The most notable change that I did was the renaming of `Outcome::from_lines…()` to `Handshake`, which seemed like a better fit.
> Now I don't know if this is a mistake because it introduced ambiguity elsewhere, but if you encounter it, I am sure you'd be most qualified to find better names.

The `Outcome` name was pre-existing -- I didn't change it. I agree that it's not a great name, and `Handshake` is definitely an improvement. I think I saw some other types named `Outcome` as well...

gix-transport/Cargo.toml
djc: I think the async and blocking implementations are fully parallel -- I don't remember seeing any code where either one is prioritized, and I don't intend to introduce any such code.

Byron: Thanks! I fixed that in the other PR which is about to merge.

Byron:
> I think I saw some other types named `Outcome` as well...

Yes, there are everywhere, and typically they really are nothing more than the outcome of some function. In the handshake case though, it made sense to name it, and in the other PR I also changed it, I think for the better. Please feel entitled to review these common names as well.

PR #2242, opened 2025-11-01
Isolate logic surrounding ls-refs command

djc:
Rejiggers the code surrouding the `ls-refs` part of gix-protocol to split pre- and post-transport logic from the core networking logic, with the goal of avoiding code duplication for blocking/async I/O down the line.

### DIFF
--- a/gitoxide-core/src/pack/receive.rs
+++ b/gitoxide-core/src/pack/receive.rs
@@ -4,6 +4,7 @@ use std::{
     sync::{atomic::AtomicBool, Arc},
 };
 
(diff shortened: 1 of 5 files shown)


Byron (approved):
Thanks a lot!

This is wonderful and thanks to my additional refactor I think I am also restoring my feeling for this part of the codebase, a part that I have neglected quite a bit, and the happier I am that you are doing this work!

Byron:
And I forgot to mention: please use the conventional commit prefixes sparingly, particularly with `refactor` which can be a catch all. Refactors aren't usually relevant in changelogs unless they are breaking. For that reason, I took the liberty of removing all `refactor: ` prefixes from your commits as well.

PR #2249, opened 2025-11-07
Move worktree_branches call out of loop in ref update

iczero:
`worktree_branches` opens every worktree as a `Repository` to resolve its `HEAD`. This is currently getting called in the loop over ref mappings, but I do not see why it needs to be. Moving it outside the loop results in a pretty significant performance increase (~10x in my case) for fetch when worktree(s) are present, dependent on how many refs exist in the repository and the value configured for `object_store_slots`. In my case, `object_store_slots` was set relatively high, resulting in slow drop for `gix_odb::Store` and hence `gix::Repository`.


Byron:
Thanks so much, a great catch!

For a moment I was surprised that this worktree mapping is mutable, but it seems that it only sorts the worktree directories which indeed is idempotent in this context.

Issue #2177, opened 2025-09-19
The `CommitRef` from `invalid_email_or_committer` test cannot be materialized or round-tripped


ilyagr:
### Current behavior 😯

At the latest commit a78bf84, I modified the test as follows:

```diff
diff --git a/gix-object/tests/object/commit/from_bytes.rs b/gix-object/tests/object/commit/from_bytes.rs
index b89213394..c67a9383d 100644
--- a/gix-object/tests/object/commit/from_bytes.rs
+++ b/gix-object/tests/object/commit/from_bytes.rs
@@ -37,9 +37,15 @@ fn invalid_email_of_committer() {
         email: b"gh <Gregor Hartmann<gh@openoffice.org".as_bstr(),
         time: "1282910542 +0200",
     };
+    let fixture = fixture_name("commit", "invalid-actor.txt");
+    let commit_ref = CommitRef::from_bytes(&fixture).expect("ignore strangely formed actor format");
+    let write_result = commit_ref.write_to(&mut Vec::new());
+    dbg!(&write_result);
+    assert!(write_result.is_ok(), "write result is not OK");
+    // The assert above fails
+
     assert_eq!(
-        CommitRef::from_bytes(&fixture_name("commit", "invalid-actor.txt"))
-            .expect("ignore strangely formed actor format"),
+        commit_ref,
         CommitRef {
             tree: b"220738fd4199e95a2b244465168366a73ebdf271".as_bstr(),
             parents: [b"209fbe2d632761b30b7b17422914e11b93692833".as_bstr()].into(),
```

The resulting test fails:

```

---- commit::from_bytes::invalid_email_of_committer stdout ----
[gix-object/tests/object/commit/from_bytes.rs:43:5] &write_result = Err(
    Custom {
        kind: Other,
        error: IllegalCharacter,
    },
)

thread 'commit::from_bytes::invalid_email_of_committer' (43934500) panicked at gix-object/tests/object/commit/from_bytes.rs:44:5:
write result is not OK
```

### Expected behavior 🤔

AFAIU, `CommitRef` should always be possible to materialize. If commits like the one in `invalid-actor.txt` exist in the wild, I'd want to round-trip them.

Byron:
Thanks a lot for reporting!

Indeed, `CommitRef` should have such properties.
Let me try to get a fix going.

Byron:
[...]

PR #2253, opened 2025-11-09
fix!: store raw commit/tag actor headers and parse lazily (see #2177)

Pingasmaster:
This PR makes signature handling truly lossless for "creative" emails and other info. We now stash the raw name <email> slice on IdentityRef/SignatureRef and fall back to it when rewriting, so even commits with embedded angle brackets round-trip cleanly (might want to expand to other malformed characters before merging? idk). Parsing and serialization honor that flag but still keep strict validation for normal input. I also added regression coverage for these scenarios. Might not be the most elegant solution, as now every ctor/helper that builds signatures or identities explicitly sets raw: None. This is only me taking a stab at it for fun. It is not prod ready but just an idea.

Tries to help with #2177.


Byron (commented):
Thanks a lot for making this happen.

I took a first look and don't like it at all, but also wouldn't know how to achieve round-tripping differently.

If you wouldn't mind, breaking changes should be in a separate commit and prefixed with `fix!: ` probably, and all adjustments to other crates go into a separate commit (not marked as breaking (assuming they aren't breaking).

Byron:
> Might not be the most elegant solution, as now every ctor/helper that builds signatures or identities explicitly sets raw: None.

First of all, the `raw` field is only in the `*Ref` versions of the type, and there already was a precedent for altering these to support round-tripping. From that point of view, I think it's acceptable, but… this never had to go so far and implement a by-pass.

I was wondering… what if this `raw` information would be stored on the `CommitRef` instead? Then I'd even go as far as to turn the existing fields [with parsed `*Ref` types](https://github.com/GitoxideLabs/gitoxide/blob/6f469a6fea59c88e6c69a5f94b0bc8a5977cb75b/gix-object/src/lib.rs#L94-L101) into `&BStr`, turning it into the `raw` field effectively.

Then one can use `commit.author` to access the raw field, and `commit.author()` to get a fallible parsed version of it just like before.

So this PR is definitely good at showing that alternative solutions like the one mentioned here a worth exploring.
Is this something you'd be interested in?

Pingasmaster:
I'm not sure I can do exactly what you have in mind but I gave it a try. It's much cleaner code-wise at least. Divided it into 2 commits like you asked too, but I'm still not sure this is 100% ready. It's 2AM right now for me so I'll sleep and see tomorrow if I have anything else which might make this cleaner. Thanks for the feedback!

gix-object/src/commit/mod.rs
```diff
@@ -95,14 +99,14 @@ impl<'a> CommitRef<'a> {
     ///
     /// This is different from the `author` field which may contain whitespace.
     pub fn author(&self) -> gix_actor::SignatureRef<'a> {
```
Byron: These must be fallible, panics aren't allowed. It's OK to use the error type returned by `SignatureRef::from_bytes()`.

gix-object/src/commit/write.rs
```diff
@@ -4,6 +4,32 @@ use bstr::ByteSlice;
...
+fn signature_len(raw: &bstr::BStr) -> usize {
+    if signature_requires_raw(raw) {
+        raw.len()
+    } else {
+        parse_signature(raw).size()
+    }
+}
+
+fn write_signature(mut out: &mut dyn io::Write, field: &[u8], raw: &bstr::BStr) -> io::Result<()> {
+    if signature_requires_raw(raw) {
```
Byron: I wonder why this differentiation is still required. In theory, the committer and author are now verbatim, which should always be what's written back.

gix-object/src/object/convert.rs
```diff
@@ -7,15 +7,20 @@ impl From<TagRef<'_>> for Tag {
             name,
             target_kind,
             message,
-            tagger: signature,
+            tagger,
             pgp_signature,
         } = other;
+        let tagger = tagger.map(|raw| {
+            gix_actor::SignatureRef::from_bytes::<()>(raw.as_ref())
+                .expect("signatures were validated during parsing")
```
Byron: Every time the signature is parsed it must be fallible.

gix-object/src/tag/write.rs
```diff
@@ -1,10 +1,27 @@
...
+use bstr::{BStr, ByteSlice};
 use gix_date::parse::TimeBuf;
 
 use crate::{encode, encode::NL, Kind, Tag, TagRef};
 
+fn parse_signature(raw: &BStr) -> gix_actor::SignatureRef<'_> {
+    gix_actor::SignatureRef::from_bytes::<()>(raw.as_ref()).expect("signatures were validated during parsing")
+}
+
+fn signature_requires_raw(raw: &BStr) -> bool {
```
Byron: Again, I think differentiating between these shoudln't be necessary.

Pingasmaster:
I've adjusted the commit and tag APIs so they return fallible results. `CommitRef::author/committer/time` report `Result<SignatureRef, decode::Error>` and after successful decoding trim the parsed value. `TagRef::tagger` also follows the same pattern. I made the owned conversions (`CommitRef::into_owned/to_owned` and `TagRef::into_owned`) use TryFrom so everything propagates cleanly, but don't hesitate to tell me if you have a better idea.

I’ve also removed the raw/canonical split. You were right, just streamring the stored header bytes and sizing them via `raw.len()` for commits and tag writers is much better. 

Every place that re-parses actors during conversion or utility code should be fallible now. Tell me if you see anything else.

Byron:
My apologies for letting this PR wait for so long. I do hope to get to it this weekend.

Byron:
Thanks for the patience - I will do a final pass now and make sure this PR gets merged - no more action is needed from your side.

PR #2254, opened 2025-11-10
fix: allow credential fill to run outside repo

### LINKED ISSUE #2198: `credential fill` subcommand fails outside of git repository
### Current behavior 😯

`gix credential fill` fails if the current directory isn't part of a git repository with the message `Error: Could not find a git repository in '.' or in any of its parents`
From a bit of testing it seems to be originating from this line: https://github.com/GitoxideLabs/gitoxide/blob/0a576b968e61e9a675ae630bae3f6c0fd9422ac3/src/plumbing/main.rs#L661

### Expected behavior 🤔

`gix credential fill` should run, no matter if in a git repository or not

### Git behavior

`git credential fill` runs, no matter if in a git repository or not

### Steps to reproduce 🕹

1. Go into a directory that isn't part of any git repository
2. Run both gix and git with the subcommand
```bash
$ git credential fill << EOF
protocol=https
host=github.com
EOF
```

ralphmodales:
fixes #2198

### DIFF
--- a/gitoxide-core/src/repository/credential.rs
+++ b/gitoxide-core/src/repository/credential.rs
@@ -33,10 +33,10 @@ pub fn function(repo: Option<gix::Repository>, action: gix::credentials::program
                     gix::config::credential_helpers(
                         gix::url::parse(url.as_ref())?,
                         &config,
-                        false,
-                        |_| true,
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot, this will work!

Issue #2265, opened 2025-11-23
gix cat: Encountered unsupported command code: 0


ElvishJerricco:
### Current behavior 😯

I encountered this after updating `jj`, but I bisected it to `jj`'s update from `gix` `0.73.0` to `0.74.1`, and bisected further to https://github.com/GitoxideLabs/gitoxide/pull/2155

In a large repo I have, reading a certain object always fails.

```
$ gix cat 70b81af0814b0620db4af251b3f0e9530457104b
Error: An error occurred while obtaining an object from the packed object store

Caused by:
    Encountered unsupported command code: 0

Stack backtrace:
   0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
   1: <core::result::Result<T,F> as core::ops::try_trait::FromResidual<core::result::Result<core::convert::Infallible,E>>>::from_residual
   2: gitoxide_core::repository::cat::display_object
   3: gitoxide_core::repository::cat::function::cat
   4: gitoxide::plumbing::main::main::{{closure}}
   5: gitoxide::shared::pretty::prepare_and_run
   6: gitoxide::plumbing::main::main
   7: gix::main
   8: core::ops::function::FnOnce::call_once
   9: std::sys::backtrace::__rust_begin_short_backtrace
  10: std::rt::lang_start::{{closure}}
  11: std::rt::lang_start_internal
  12: std::rt::lang_start
  13: main
  14: __libc_start_call_main
  15: __libc_start_main_alias_1
  16: _start
```

### Expected behavior 🤔

The object should be readable, as it is with gix `0.73.0`.

### Git behavior

`git cat-file -p 70b81af0814b0620db4af251b3f0e9530457104b` in the same repo outputs the contents of a file in the repo.

### Steps to reproduce 🕹

Unfortunately, I can only narrow this down to a 5.1GiB pack file, and I don't know how this problem was created. It seems like the pack file is valid, given that both `git` and earlier versions of `gix` read it fine. Copying the `.pack` file to a new repo and regenerating the `.idx` and `.rev` with `git-index-pack` preserves the problem, so the problem is in the `.pack`. Is there a way I can narrow down the pack file to just the problematic object?

Byron:
Thanks for reporting and the initial investigation, and sorry for the hassle.

I understand this issue isn't present before #2155 was merged, and that it's unrelated to the pack index which was also regenerated. Thus, let me CC @folkertdev. 

[...]

Issue #2245, opened 2025-11-02
Dependabot fails with "No Cargo.toml!"


EliahKagan:
### Current behavior 😯

When a Dependabot version update runs for Rust dependencies (i.e. the "cargo" ecosystem), it fails with:

```text
Dependabot encountered '1' error(s) during execution, please check the logs for more details.
+-----------------------------------------------+
|                    Errors                     |
+---------------+-------------------------------+
| Type          | Details                       |
+---------------+-------------------------------+
| unknown_error | {                             |
|               |   "message": "No Cargo.toml!" |
|               | }                             |
+---------------+-------------------------------+
Failure running container d89a028c2d61c4b2e4b792c3b0eece859214657bd263b037ccbb4baa1f54422a: Error: Command failed with exit code 1: /bin/sh -c $DEPENDABOT_HOME/dependabot-updater/bin/run update_files
Cleaned up container d89a028c2d61c4b2e4b792c3b0eece859214657bd263b037ccbb4baa1f54422a
  proxy | 2025/11/01 08:01:47 0/11 calls cached (0%)
2025/11/01 08:01:47 Posting metrics to remote API endpoint
Error: Dependabot encountered an error performing the update
```

There may be more relevant details in the logs, but if so, I haven't found them. [Here's the most recent run](https://github.com/GitoxideLabs/gitoxide/actions/runs/18993727840/job/54250504894) in this upstream repository. That links to the CI run for it that includes the full log. One can get there also from [this page](https://github.com/GitoxideLabs/gitoxide/network/updates/9454592/jobs) by clicking "View logs" and clicking "View logs" again.

I've also been having this problem in my fork, where it seems to have started at some point between October 15th and October 17th, based on the test PR https://github.com/EliahKagan/gitoxide/pull/117 that was able to be created but not to be updated. It looks like #2208 was the only thing that came in during that time, which doesn't look like it could cause or have anything to do with such a problem.

I've had various Dependabot failures in my fork for a while, and I didn't assume this would happen here; but this is the first that happens upstream as well, and seems to happen every time.

Unfotunately I don't remember the messages from those previous self-resolving patterns of failure that had resolved by retrying, so I can't compare them to this one, and they were long enough ago that their logs are probably not preserved.

This only affects version updates for Rust dependencies. Version updates for GitHub Actions dependencies still work.

### Expected behavior 🤔

Dependabot version updates should work, as they have worked before.

[...]

PR #2266, opened 2025-11-23
various improvements

Byron:
A catch-all PR with various changes.

### Tasks

* [ ] TBD

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,15 +3,6 @@
 version = 3
 
 [[package]]
-name = "addr2line"
-version = "0.25.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
-dependencies = [
- "gimli",
-]
-
-[[package]]
 name = "adler2"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


Byron:
Actually there are no more plans for now.

EliahKagan:
Was b130277420d8ecc9a02f7c4e0aa759f4210e0fca meant to upgrade dependencies in any `Cargo.toml` files? The change it makes is only to `Cargo.lock`. I can't tell from the wording if it means that it upgrades them in `Cargo.lock` in accordance with the version ranegs in `Cargo.toml` files, which does appear to be what was done, or if it instead was meant to upgrade versions in `Cargo.toml` files themselves.

(The detailed message suggests that modifying only `Cargo.lock` may be intentional. But I would still expect these changes to be observed by CLI users if they install with `--locked`. So I'm not sure if this is really referring to all possible effects or instead to particular effects such as those related to #2265.)

Normally we upgrade versions listed in `Cargo.toml` files monthly via Dependabot, but that hasn't happened recently, due to #2245.

Byron:
My apologies, I could have been more specific about the motivation. 

First of all, I didn't see `--locked` as the feature that would observe the `Cargo.lock` changes, even though it certainly would if people install with `--git`.

Besides that, it was really only meant to assure that local builds come out 'fresher' as I thought it might affect #2044. I didn't follow this further, the `cargo update` was just a blind shot I did without much thinking.

EliahKagan:
Thanks, no problem! Also, sorry about my confusion: while CLI users who install from this repository, either directly or with `--git`, wil use the lock file when passing `--locked`, you're right that it's not otherwise used. I don't know why `cargo` seems to accept `--locked` in scenarios where it has no effect, but you're right that it doesn't do anything unless something like `--git` (or `--path`) is passed.

Anyway, it looks like the #2245 may be fixed soon, since https://github.com/dependabot/dependabot-core/pull/13359 has been merged to fix the underlying bug https://github.com/dependabot/dependabot-core/issues/13345. But I don't know it the fix has been deployed yet.

PR #2274, opened 2025-11-26
protocol: split async and blocking APIs for fetching refmaps

djc:
This is just one part of the problem in gix-protocol, but I'm happy that I was able to resolve it with only about +20 lines despite some duplication.

Happy to take feedback if you can articulate it (to save you on refactoring time).

### DIFF
--- a/gix/src/remote/connection/ref_map.rs
+++ b/gix/src/remote/connection/ref_map.rs
@@ -157,7 +157,7 @@ where
             extra_refspecs,
         };
 
-        let fetch_refmap = handshake.fetch_or_extract_refmap(
+        let fetch_refmap = handshake.prepare_lsrefs_or_extract_refmap(
             self.remote.repo.config.user_agent_tuple(),
             prefix_from_spec_as_filter_on_remote,
             context,
--- a/gitoxide-core/src/pack/receive.rs
+++ b/gitoxide-core/src/pack/receive.rs
@@ -88,7 +88,7 @@ where
         extra_refspecs: vec![],
     };
 
-    let fetch_refmap = handshake.fetch_or_extract_refmap(user_agent.clone(), true, context)?;
+    let fetch_refmap = handshake.prepare_lsrefs_or_extract_refmap(user_agent.clone(), true, context)?;
 
     #[cfg(feature = "async-client")]
     let refmap = fetch_refmap
--- a/gix-protocol/src/fetch/negotiate.rs
+++ b/gix-protocol/src/fetch/negotiate.rs
@@ -111,7 +111,7 @@ pub struct Round {
 /// * `graph`
 ///     - The commit-graph for use by the `negotiator` - we populate it with tips to initialize the graph traversal.
 /// * `ref_map`
-///     - The references known on the remote, as previously obtained with [`crate::Handshake::fetch_or_extract_refmap()`].
+///     - The references known on the remote, as previously obtained with [`crate::Handshake::prepare_lsrefs_or_extract_refmap()`].
 /// * `shallow`
 ///     - How to deal with shallow repositories. It does affect how negotiations are performed.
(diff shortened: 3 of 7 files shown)


gix-protocol/tests/protocol/fetch/_impl.rs
```diff
@@ -91,17 +103,27 @@ mod fetch_fn {
...
-                    &mut progress,
-                    trace,
-                    ("agent", Some(Cow::Owned(agent.clone()))),
-                )
-                .await?
-            }
+            None => match delegate.action() {
+                Ok(RefsAction::Skip) => Vec::new(),
+                Ok(RefsAction::Continue) => {
+                    #[cfg(feature = "async-client")]
```
Byron, line 109: It would probably be nice if the test would fail at compile time with a clear message if both features are enabled. Now it would probably fail more indirectly due to invalid code.

gix/src/remote/connection/ref_map.rs
```diff
@@ -156,16 +156,21 @@ where
...
+        )?;
+
+        #[cfg(feature = "async-network-client")]
+        let ref_map = fetch_refmap
+            .fetch_async(progress, &mut self.transport.inner, self.trace)
             .await?;
+
+        #[cfg(feature = "blocking-network-client")]
+        let ref_map = fetch_refmap.fetch_blocking(progress, &mut self.transport.inner, self.trace)?;
+
```
Byron, line 173: This is interesting - ~~now it will compile with both features toggled on, but also yield a warning about the first `ref_map` never being read from.~~

Probably that wouldn't compile with both features toggled on, and even though it's not the time to have a test for that just yet (unless it should be expected to fail), I think it's a good time to figure out what should happen in this case.

One thing is for sure: `    #[gix_protocol::maybe_async::maybe_async]` decides if the function is async, and that decides if `await` is possible at all. Only if it is possible, it should be used, and the blocking version then isn't necessary anymore.
The question is if it is easy enough to 'sync' with what the parent macro does.

It's totally conceivable as well to bring `maybe-async` like functionality into `gix-macros` and make it work as we need it.

Maybe you have plans as well that you could share?

Byron (commented):
Thanks a lot for keeping these PR coming, it's much appreciated!

Regarding the CI failure, it can be reproduced with `cargo test -p gix-protocol --features blocking-client`.

There are a few more general questions that arose from a first relatively brief look.

gix/src/remote/connection/ref_map.rs
djc: Given that the gix crate's `async-network-client` and `blocking-network-client` enable the relevant upstream features in gix-protocol, the gix crate efffectively cannot be compiled with both features enabled anyway. So for now, this is not an issue, and I'd like to defer looking into it in more detail -- I suspect it will be something like the refactoring in this PR: isolate the network calls in a relatively small function and duplicate that part into explicit `async` and `blocking` API.

gix-protocol/tests/protocol/fetch/_impl.rs
djc: There's still a `compile_error!()` in the gix-protocol crate root that requires the client features to be mutually exclusive, which will trigger before anything here. So it seems okay for now?

djc:
Tests should be passing now.

Byron:
Thanks a lot, I will take a look soon and see to merging it.
Meanwhile, I am unleashing an AI reviewer to see what it can do, please feel free to ignore trash or to ignore it entirely.

djc:
AI code review comments were pretty good, addressed all of them.

gix-protocol/tests/protocol/fetch/_impl.rs
Byron: Oh, right, that's totally enough. Sorry for the false flag.

PR #2276, opened 2025-12-01
Report number of non-matching diffs in slider test

cruessler:
This PR improves the test reporting for the diff slider test.

In addition to the first non-matching diff, the test now also reports the fixture it came from and the number of total as well as non-matching diffs. This makes the slider test more suitable as a benchmark for comparing different slider heuristics.

This is a follow-up to #2197.

Comparison of new test report (above) vs. old test report (below):

Screenshot 2025-12-01 at 10 15 00

### DIFF
--- a/gix-diff/tests/diff/blob/slider.rs
+++ b/gix-diff/tests/diff/blob/slider.rs
@@ -95,11 +95,13 @@ fn baseline() -> gix_testtools::Result {
         .filter(|(_, _, actual_matches_baseline, _)| *actual_matches_baseline)
         .count();
 
-    assert!(
-        matching_diffs == total_diffs,
-        "assertion failed: total diffs {} == matching diffs {}\n\n{}",
-        total_diffs,
+    assert_eq!(
         matching_diffs,
+        total_diffs,
+        "matching diffs {} == total diffs {} [{:.2} %]\n\n{}",
+        matching_diffs,
+        total_diffs,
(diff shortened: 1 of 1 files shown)


gix-diff/tests/diff/blob/slider.rs
```diff
@@ -80,13 +81,39 @@ fn baseline() -> gix_testtools::Result {
...
     }
 
+    let total_diffs = diffs.len();
+    let matching_diffs = diffs
+        .iter()
+        .filter(|(_, _, actual_matches_baseline, _)| *actual_matches_baseline)
+        .count();
+
+    assert!(
+        matching_diffs == total_diffs,
```
Byron: Would this not be more readable with `assert_eq!`?

gix-diff/tests/diff/blob/slider.rs
```diff
@@ -80,13 +81,39 @@ fn baseline() -> gix_testtools::Result {
...
+        "assertion failed: total diffs {} == matching diffs {}\n\n{}",
```
Byron: I think the assertion failure will come across by merit of `assert_eq`, so the text can skip `assertion failed:`.

cruessler:
Thanks for the suggestions, they make a lot of sense! I also added the ratio of matching diffs as a percentage, to make the comparison between two sets of diffs even easier in case they contain a different number of diffs.

Byron:
Wonderful 🙏!

kritsada1998mt-create:
AI

PR #2281, opened 2025-12-02
gix-mailmap: Derive `std::hash::Hash` for `Snapshot`

caio:
Hello!

I'd like to cache output that relies on mailmaps.

Using the object id or a hash of the raw bytes is an alternative but the format allows for comments and whatnot. At present, the best alternative I think is newtyping and implementing Hash through `Snapshot::iter()`. Having hash on `Snapshot` lets me avoid that.


Byron:
Thanks a lot! Makes sense.

PR #2282, opened 2025-12-03
[do not merge] use zlib-rs directly

folkertdev:
just checking that this works out. It's not clean, but it compiles.


Byron:
Actually this is implemented now :).

PR #2284, opened 2025-12-05
fix: Probe executable bits by checking if we can change them

isuffix:
After resolving https://github.com/jj-vcs/jj/issues/3949 via https://github.com/jj-vcs/jj/pull/7484, I realized that `jj git init` differs from `git init` and still causes issues with executable bits.

The problem is that `jj git init` sets up the `.git/config` file with `core.filemode = true`, despite file modes not being modifiable. Following the code, the reason is that `jj` uses the [default file capabilities](https://github.com/jj-vcs/jj/blob/bcfd5c977c26ccffd55fe8c69943a1af04b1c461/lib/src/git_backend.rs#L241) from gitoxide when initializing the repo, and gitoxide sets the capabilities [by probing the filesystem](https://github.com/GitoxideLabs/gitoxide/blob/4b00596e7c025b18b07ad9ea1f78a5884244cbfe/gix/src/create.rs#L209). However, the probe for executable bits only tests whether the executable bit can be set when a file is _created_, but not whether it can be _modified_.

This PR changes the filemode probe to also verify that the executable bit can be modified. I tried to do so by changing as little surrounding code as possible.

The equivalent test by git itself for `git init` is here:
https://github.com/git/git/blob/f0ef5b6d9bcc258e4cbef93839d1b7465d5212b9/setup.c#L2367-L2379

### DIFF
--- a/gix-fs/src/capabilities.rs
+++ b/gix-fs/src/capabilities.rs
@@ -73,14 +73,15 @@ impl Capabilities {
             .and_then(|file| {
                 let old_mode = file.metadata()?.mode();
                 let is_executable = old_mode & 0o100 == 0o100;
-                Ok(is_executable && {
-                    let new_mode = old_mode ^ 0o100;
-                    match file.set_permissions(PermissionsExt::from_mode(new_mode)) {
-                        Ok(()) => new_mode == file.metadata()?.mode(),
+                let exe_bit_flip_works_in_filesystem = {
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot, great work!

I particularly like the reference to the relevant code in Git, and of course the motivation coming from JJ along with references.

It also usually makes me a little sad that GitHub de-emphasises commits to the point where PRs have all the interesting background information, but the commit has not. Maybe next time :).

PR #2285, opened 2025-12-07
Account for "\ No newline at end of file" in baseline diff parser

cruessler:
This is a small PR that fixes an oversight in the baseline diff parser in the diff slider tests. The parser now also parses `\ No newline at end of file` lines. This is something I found while using the script to run more tests. Off the top of my head, I don’t know whether there’s more cases the parser doesn’t handle yet.

### DIFF
--- a/gix-diff/tests/diff/blob/slider.rs
+++ b/gix-diff/tests/diff/blob/slider.rs
@@ -214,14 +214,7 @@ mod baseline {
                         hunk_lines.push(b'\n');
                     }
                     b'\\' => {
-                        if line == "\\ No newline at end of file".as_bytes() {
-                            // Do nothing.
-                        } else {
-                            unreachable!(
-                                "BUG: expecting unified diff format, found line: `{}`",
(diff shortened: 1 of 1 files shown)


gix-diff/tests/diff/blob/slider.rs
```diff
@@ -213,7 +213,20 @@ mod baseline {
...
-                    _ => unreachable!("BUG: expecting unified diff format"),
+                    b'\\' => {
+                        if line == "\\ No newline at end of file".as_bytes() {
+                            // Do nothing.
+                        } else {
+                            unreachable!(
+                                "BUG: expecting unified diff format, found line: `{}`",
+                                line.to_str_lossy()
+                            )
+                        }
```
Byron: I think this should just be an `assert!` with message.

cruessler: Makes a lot of sense, thanks for the suggestion!

PR #2289, opened 2025-12-09
Mark `gix_hash::Kind` as `non_exhaustive`

cruessler:
This is a follow-up to #2218 (which is a rewritten version of #2213). It marks `gix_hash::Kind` as `non_exhaustive`. This is a breaking change now, but will prevent further breaking changes whenever we add a new hashing algorithm in the future.

### DIFF
--- a/gix-worktree-stream/src/protocol.rs
+++ b/gix-worktree-stream/src/protocol.rs
@@ -111,6 +111,7 @@ fn byte_to_mode(b: u8) -> gix_object::tree::EntryMode {
 fn hash_to_byte(h: gix_hash::Kind) -> u8 {
     match h {
         gix_hash::Kind::Sha1 => 0,
+        _ => unreachable!("BUG: not implemented for hash kind {h}"),
     }
 }
 


gix-worktree-stream/src/protocol.rs
```diff
@@ -111,6 +111,7 @@ fn byte_to_mode(b: u8) -> gix_object::tree::EntryMode {
 fn hash_to_byte(h: gix_hash::Kind) -> u8 {
     match h {
         gix_hash::Kind::Sha1 => 0,
+        _ => unreachable!("BUG: not implemented for hash kind {h}"),
```
Byron, line 114: It's too bad that it's not possible to `allow` the lint, instead of adding a fallback. That way, Gitoxide could opt-in to 'breaking' loudly once a new variant is added.
This makes me feel this non_exhaustive feature is really half-baked… why I can't we opt-out 😅?

cruessler: It certainly tackles the problem from a specific angle. :-) I find it very useful, but as you say, there’s downsides in a monorepo.

PR #2290, opened 2025-12-09
fix tar feature in gix-archive

gruebel:
Currently it is not possible to only choose the `tar` feature in `gix-archive`, also `Format::TarGz` was not correctly instantiated and throw an additional error.


Byron (changes requested):
Thanks a lot for the fix, it's much appreciated.

Unfortunately, I can't merge this without a test that triggers it, ideally in a separate commit and push, so CI can turn green with the fix.

Let's fix the reason this bug was present - the missing test.
Thanks again 🙏.

gruebel:
@Byron 
I checked and actually there is already a test for it, but the problem was the test run for `gix-archive` were not correctly setup and I fixed in the `justfile`. I also had to enable the `deflate-flate2-zlib-rs` feature in `zip` for the `gzip` feature, otherwise it complains not having a backend. let me know, if it works for you.

PR #2292, opened 2025-12-10
Add `ObjectId::Sha256` and `Kind::Sha256`

cruessler:
This is a draft that adds `ObjectId::Sha256` and `Kind::Sha256` to `gix-hash`. I’m opening this to get early feedback, in particular from a CI run.

Some remarks:

- It feels wrong to put `empty_blob_sha256` and `empty_tree_sha256` into `oid` as the comment says `oid` is related to `Sha1`, but I’m unsure how to best resolve this tension.
- There’s inconsistencies with respect to spelling: `SHA256`, `SHA-256`, `Sha256`.
- `ObjectId::null_sha1` hardcodes 20. Should this be `SIZE_OF_SHA1_DIGEST`? `oid::null_sha1` uses `SIZE_OF_SHA1_DIGEST`.
- `Kind::len_in_hex` and `Kind::len_in_bytes` also hardcode values.
- Hardcoded empty trees and empty blobs appear in more than place (for both SHA-1 and SHA-256). I feel like we should have a single source of truth here.
- Some comments in `ObjectId` would probably benefit from being rephrased. See e. g. `ObjectId::new_sha1` or `ObjectId::from_20_bytes`.

Sources for the hardcoded values related to SHA-256:

https://github.com/git/git/blob/e85ae279b0d58edc2f4c3fd5ac391b51e1223985/hash.h
https://github.com/git/git/blob/e85ae279b0d58edc2f4c3fd5ac391b51e1223985/hash.c

### DIFF
--- a/gix-hash/Cargo.toml
+++ b/gix-hash/Cargo.toml
@@ -17,9 +17,9 @@ test = false
 
 [features]
 default = ["sha1"]
-## Support for SHA1 digests.
+## Support for SHA1 hashes and digests.
 sha1 = []
-## Support for SHA256 digests.
+## Support for SHA256 hashes and digests.
 sha256 = ["dep:sha2"]
 ## Data structures implement `serde::Serialize` and `serde::Deserialize`.
 serde = ["dep:serde", "faster-hex/serde"]
--- a/gix-hash/src/hasher.rs
+++ b/gix-hash/src/hasher.rs
@@ -53,10 +53,9 @@ pub(super) mod _impl {
         /// Finalize the hash and produce an object id.
         ///
         /// Returns [`Error`] if a collision attack is detected.
-        ///
-        /// TODO: Since SHA256 has an infallible `finalize`, it might be worth investigating
-        /// turning the return type into `Result<crate::ObjectId, Infallible>` when this crate is
-        /// compiled with SHA256 support only.
+        // TODO: Since SHA-256 has an infallible `finalize`, it might be worth investigating
+        //       turning the return type into `Result<crate::ObjectId, Infallible>` when this crate is
+        //       compiled with SHA-256 support only.
         #[inline]
         pub fn try_finalize(self) -> Result<crate::ObjectId, Error> {
             match self {
@@ -65,8 +64,9 @@ pub(super) mod _impl {
                     CollisionResult::Mitigated(_) => {
                         // SAFETY: `CollisionResult::Mitigated` is only
                         // returned when `safe_hash()` is on. `Hasher`’s field
-                        // is private, and we only construct it in the
-                        // `Default` instance, which turns `safe_hash()` off.
+                        // is private, and we only construct the SHA-1 variant
+                        // via `Hasher::new_sha1()` (and thus through `hasher()`),
+                        // which configures the builder with `safe_hash(false)`.
                         //
                         // As of Rust 1.84.1, the compiler can’t figure out
                         // this function cannot panic without this.
--- a/gix-hash/src/kind.rs
+++ b/gix-hash/src/kind.rs
@@ -23,9 +23,9 @@ impl FromStr for Kind {
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         Ok(match s {
-            "sha1" | "SHA1" => Kind::Sha1,
+            "sha1" | "SHA1" | "SHA-1" => Kind::Sha1,
             #[cfg(feature = "sha256")]
-            "sha256" | "SHA256" => Kind::Sha256,
+            "sha256" | "SHA256" | "SHA-256" => Kind::Sha256,
             other => return Err(other.into()),
         })
     }
@@ -34,9 +34,9 @@ impl FromStr for Kind {
 impl std::fmt::Display for Kind {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            Kind::Sha1 => f.write_str("SHA1"),
+            Kind::Sha1 => f.write_str("sha1"),
             #[cfg(feature = "sha256")]
-            Kind::Sha256 => f.write_str("SHA256"),
+            Kind::Sha256 => f.write_str("sha256"),
         }
     }
 }
@@ -101,8 +101,8 @@ impl Kind {
     }
 
     /// Returns the kind of hash that would fit the given `hex_len`, or `None` if there is no fitting hash.
-    /// Note that `0` as `hex_len` up to 40 always yields `SHA1` while anything in the range 41..64
-    /// always yields `SHA256` if it is enabled.
+    /// Note that `0` as `hex_len` up to 40 always yields SHA-1 while anything in the range 41..64
+    /// always yields SHA-256 if it is enabled.
     #[inline]
     pub const fn from_hex_len(hex_len: usize) -> Option<Self> {
         Some(match hex_len {
--- a/gix-hash/src/object_id.rs
+++ b/gix-hash/src/object_id.rs
@@ -194,6 +194,8 @@ impl ObjectId {
     pub fn from_bytes_or_panic(bytes: &[u8]) -> Self {
         match bytes.len() {
             SIZE_OF_SHA1_DIGEST => Self::Sha1(bytes.try_into().expect("prior length validation")),
+            #[cfg(feature = "sha256")]
+            SIZE_OF_SHA256_DIGEST => Self::Sha256(bytes.try_into().expect("prior length validation")),
             other => panic!("BUG: unsupported hash len: {other}"),
         }
     }
--- a/gix-hash/src/oid.rs
+++ b/gix-hash/src/oid.rs
@@ -99,8 +99,8 @@ impl oid {
         }
     }
 
-    /// Create an `oid` from the input `value` slice without performing any safety check.
-    /// Use only once sure that `value` is a hash of valid length.
+    /// Create an `oid` from the input `value` slice without performing any length check.
+    /// Use only once you are sure that `value` is a hash of valid length, or panics will occur on most uses.
     pub fn from_bytes_unchecked(value: &[u8]) -> &Self {
         Self::from_bytes(value)
     }
@@ -154,12 +154,15 @@ impl oid {
 
     /// Write ourselves to the `out` in hexadecimal notation, returning the hex-string ready for display.
     ///
-    /// **Panics** if the buffer isn't big enough to hold twice as many bytes as the current binary size.
+    /// # Panics
+    ///
+    /// If the buffer isn't big enough to hold twice as many bytes as the current binary size.
     #[inline]
     #[must_use]
     pub fn hex_to_buf<'a>(&self, buf: &'a mut [u8]) -> &'a mut str {
         let num_hex_bytes = self.bytes.len() * 2;
-        faster_hex::hex_encode(&self.bytes, &mut buf[..num_hex_bytes]).expect("to count correctly")
+        faster_hex::hex_encode(&self.bytes, &mut buf[..num_hex_bytes])
+            .expect("buffer size must be at least twice the hash digest size in bytes")
     }
 
     /// Write ourselves to `out` in hexadecimal notation.
@@ -202,7 +205,7 @@ impl oid {
     }
 }
 
-/// Methods for creating special SHA1 and SHA256 `oid`s
+/// Methods for creating special-case `oid`s (null, empty blob, empty tree)
 impl oid {
     /// Returns a SHA1 digest with all bytes being initialized to zero.
     #[inline]
--- a/gix-hash/src/verify.rs
+++ b/gix-hash/src/verify.rs
@@ -10,9 +10,9 @@ pub struct Error {
 }
 
 impl oid {
-    /// Verify that `self` matches the `expected` object id.
+    /// Verify that `self` matches the `expected` object ID.
     ///
-    /// Returns an [`Error`] containing both object ids if they differ.
+    /// Returns an [`Error`] containing both object IDs if they differ.
     #[inline]
     pub fn verify(&self, expected: &oid) -> Result<(), Error> {
         if self == expected {
--- a/gix-hash/tests/hash/prefix.rs
+++ b/gix-hash/tests/hash/prefix.rs
@@ -43,15 +43,23 @@ mod cmp_oid {
     #[test]
     #[cfg(all(feature = "sha1", feature = "sha256"))]
     fn it_detects_inequality_sha1_and_sha256() {
-        let prefix_sha1 = gix_hash::Prefix::new(&hex_to_id("bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"), 7).unwrap();
+        let len = 7;
+        let prefix_sha1 = gix_hash::Prefix::new(&hex_to_id("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"), len).unwrap();
         let prefix_sha256 = gix_hash::Prefix::new(
             &hex_to_id("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"),
-            7,
+            len,
         )
         .unwrap();
-        assert_eq!(prefix_sha256.cmp(&prefix_sha1), Ordering::Greater);
-        assert_eq!(prefix_sha1.to_string(), "bbbbbbb");
-        assert_eq!(prefix_sha256.to_string(), "aaaaaaa");
+        assert_eq!(
+            prefix_sha256.cmp(&prefix_sha1),
+            Ordering::Greater,
+            "prefixes of larger hashes are always larger"
+        );
+        assert_eq!(
+            prefix_sha1.to_string(),
+            prefix_sha256.to_string(),
+            "even though they look the same"
+        );
     }
 
     #[test]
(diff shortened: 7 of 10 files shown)


Byron:
You are on fire :)! Let me share some early thoughts, without having looked at the code at all.

> * It feels wrong to put `empty_blob_sha256` and `empty_tree_sha256` into `oid` as the comment says `oid` is related to `Sha1`, but I’m unsure how to best resolve this tension.

And indeed, `&oid` is supposed to represent any hash. if `empty_blob` somehow is hardcoded to always be SHA1, it's a bug and it needs to take a `gix-hash::Kind`. The `empty_something` could also be implemented on `gix-hash::Kind` while at it.
> * There’s inconsistencies with respect to spelling: `SHA256`, `SHA-256`, `Sha256`.

Should we fix them in favor of what Git seems to do: SHA256?

Screenshot 2025-12-11 at 04 59 53

I am definitely open to your suggestions.

> * `ObjectId::null_sha1` hardcodes 20. Should this be `SIZE_OF_SHA1_DIGEST`? `oid::null_sha1` uses `SIZE_OF_SHA1_DIGEST`.

Yes, that sounds like an oversight.
> * `Kind::len_in_hex` and `Kind::len_in_bytes` also hardcode values.

Yes, let's use constants where we have them. And since there is already a precedent with `SIZE_OF_SHA1_DIGEST` we might also have the respective SHA256 versions and use them.

> * Hardcoded empty trees and empty blobs appear in more than place (for both SHA-1 and SHA-256). I feel like we should have a single source of truth here.

I am open to suggestions as to where the empty-* methods should be, but would also not lightheartedly introduce breaking changes. If the new way feels better, we should have it. By now, I always think in terms of `gix-hash::Kind::null()/empty_*()` by the way, which aligns well with `repo.object_hash().…()`

> * Some comments in `ObjectId` would probably benefit from being rephrased. See e. g. `ObjectId::new_sha1` or `ObjectId::from_20_bytes`.

Yes, it's a good time to unify this and make it less 'hardcoded'.

Thanks again 🙏

gix-hash/tests/hash/hasher.rs
```diff
@@ -14,10 +14,21 @@ fn size_of_hasher() {
...
     );
 }
+
+#[test]
+#[cfg(all(feature = "sha256", feature = "sha1"))]
+fn size_of_try_finalize_return_type_sha1_and_sha256() {
+    assert_eq!(
+        std::mem::size_of::<Result<ObjectId, gix_hash::hasher::Error>>(),
+        34,
+        "The size of the return value is just 1 byte larger than just returning the object hash itself"
```
Byron: The message seems to be incorrect now, as it now has a 2 bytes discriminator?

gix-hash/tests/hash/prefix.rs
```diff
@@ -18,7 +18,30 @@ mod cmp_oid {
     }
 
     #[test]
-    fn it_detects_equality() {
+    #[cfg(feature = "sha256")]
+    fn it_detects_inequality_sha256() {
```
Byron, line 22: It would be interesting to see inequality between hashes of different kinds. Of course we know this will always be true, but… it's to be sure? Maybe make both have the same prefix, so an incorrect implementation *could* get it wrong.

cruessler: Makes total sense! Are you suggesting a test such as the following one?

```
#[test]
#[cfg(all(feature = "sha1", feature = "sha256"))]
fn it_detects_inequality_sha1_and_sha256() {
    let prefix_sha1 = gix_hash::Prefix::new(&hex_to_id("bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"), 7).unwrap();
    let prefix_sha256 = gix_hash::Prefix::new(
        &hex_to_id("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"),
        7,
    )
    .unwrap();
    assert_eq!(prefix_sha256.cmp(&prefix_sha1), Ordering::Greater);
    assert_eq!(prefix_sha1.to_string(), "bbbbbbb");
    assert_eq!(prefix_sha256.to_string(), "aaaaaaa");
}
```

`Prefix` has `#[derive(Ord, PartialOrd)]`, so this is what the default implementation gives us.

gix-hash/tests/hash/hasher.rs
cruessler: You’re absolutely right, fixed in the latest commit!

gix-hash/tests/hash/prefix.rs
Byron: Yes, that would be it.
We'd probably also want to ensure we have a test that turns both on at the same time then.

cruessler: I’m not entirely sure I understand what you mean by “a test that turns both on at the same time”. Can you elaborate? :-)

Byron: Sorry, I meant a test that runs if both feature toggles are enabled so both hashes are available.

cruessler:
I’ve now finally been able to address your feedback and a couple of notes that I had taken. Most of my recent changes are related to polishing the API: normalizing the spelling to `SHA1` and `SHA256`, rephrasing some comments, moving 2 methods into a different `impl oid` block so they better match the associated comment, and adding a test for the inequality of a `SHA1` and a `SHA256` prefix.

## Observation

- The tests in `prefix.rs` use `.unwrap()` a couple of times, in case that’s something we want to also change in this PR.

gix-hash/src/hasher.rs
```diff
@@ -67,6 +67,8 @@ pub(super) mod _impl {
     pub fn hasher(kind: crate::Kind) -> Hasher {
         match kind {
             crate::Kind::Sha1 => Hasher::default(),
+            #[cfg(feature = "sha256")]
+            crate::Kind::Sha256 => Hasher::default(),
```
Byron: This doesn't look right, I'd put the PR back to draft until a Sha256 hashes is available.
I also don't think it can be `unimplemented!()`, but… maybe it would be an intermediate option if there is no other way *now* and there is a plan to fix it.

gix-hash/src/kind.rs
```diff
@@ -36,13 +45,27 @@ impl Kind {
...
+        }
+        #[cfg(feature = "sha1")]
+        {
+            Self::Sha1
+        }
     }
 
     /// Returns the longest hash we support.
     #[inline]
     pub const fn longest() -> Self {
```
Byron, line 60: Now that there is this logic, it feels like tests for `longest/shortest` would be good. They'd have to repeat the feature flags so I am not a huge fan, but it really feels like there should be something in place to prevent this from going wrong, ever.
I'd be OK if you said 'better not' as this seems to be correct and when Git introduces a new hash in 10y+, we'd be looking at a completely different environment anyway.

gix-hash/src/hasher.rs
cruessler: Thanks for catching that! I should have noticed, but didn’t that `Hasher` is particular to SHA1. I was too focused on other things. :-)

Byron: Ok, no problem. Putting the PR back to draft so you can mark it as ready when ready.

gix-hash/src/kind.rs
cruessler: Done.

gix-hash/src/hasher.rs
cruessler: `Sha256` now uses the correct hasher. I hope I didn’t miss anything and marked the PR as ready again.

gix-hash/Cargo.toml
```diff
@@ -17,8 +17,10 @@ test = false
 
 [features]
 default = ["sha1"]
-## Support for SHA-1 digests.
+## Support for SHA1 digests.
 sha1 = []
+## Support for SHA256 digests.
+sha256 = []
```
Byron: I was about to give this PR my final review and merge it, but noticed that it *appears* that `sha2` should be optional, and probably with it, `sha1-checked` is as well.
With that, we'd need some new test entries in the `just` file to assure all combinations of these are compiling and can be tested, for good measure, after all this is very much at the code.

cruessler: You’re right, I forgot to make the `sha2` dependency optional. I’ve changed that.

In order to facilitate review, I have not yet begun the work to make `sha1` optional as well. In its current state, this PR tries to not make any changes to existing behaviour. That being said, feel free to make `sha1` optional as well!

gix-hash/src/kind.rs
```diff
@@ -61,23 +84,31 @@ impl Kind {
...
-    /// Note that `0` as `hex_len` up to 40 always yields `Sha1`.
+    /// Note that `0` as `hex_len` up to 40 always yields `SHA1` while anything in the range 41..64
+    /// always yields `SHA256` if it is enabled.
     #[inline]
     pub const fn from_hex_len(hex_len: usize) -> Option<Self> {
         Some(match hex_len {
-            0..=40 => Kind::Sha1,
+            0..=SIZE_OF_SHA1_HEX_DIGEST => Kind::Sha1,
+            #[cfg(feature = "sha256")]
+            0..=SIZE_OF_SHA256_HEX_DIGEST => Kind::Sha256,
```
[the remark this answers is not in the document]
Byron: Not how it works.
It matches Sha1 while it can, then it matches Sha256.

gix-hash/src/lib.rs
```diff
@@ -48,15 +48,37 @@ pub struct Prefix {
...
+#[cfg(feature = "sha256")]
+const SIZE_OF_SHA256_DIGEST: usize = 32;
+/// The size of a SHA256 hash digest in hex.
+#[cfg(feature = "sha256")]
+const SIZE_OF_SHA256_HEX_DIGEST: usize = 2 * SIZE_OF_SHA256_DIGEST;
+
+const EMPTY_BLOB_SHA1: &[u8; SIZE_OF_SHA1_DIGEST] =
+    b"\xe6\x9d\xe2\x9b\xb2\xd1\xd6\x43\x4b\x8b\x29\xae\x77\x5a\xd8\xc2\xe4\x8c\x53\x91";
+const EMPTY_TREE_SHA1: &[u8; SIZE_OF_SHA1_DIGEST] =
+    b"\x4b\x82\x5d\xc6\x42\xcb\x6e\xb9\xa0\x60\xe5\x4b\xf8\xd6\x92\x88\xfb\xee\x49\x04";
```
Byron, line 64: Finally there is all the constants that were scattered before in one place.

gix-hash/src/object_id.rs
```diff
@@ -159,41 +193,68 @@ impl ObjectId {
     /// Use `Self::try_from(bytes)` for a fallible version.
     pub fn from_bytes_or_panic(bytes: &[u8]) -> Self {
         match bytes.len() {
-            20 => Self::Sha1(bytes.try_into().expect("prior length validation")),
+            SIZE_OF_SHA1_DIGEST => Self::Sha1(bytes.try_into().expect("prior length validation")),
```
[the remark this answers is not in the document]
Byron: Great catch!

gix-hash/src/oid.rs
```diff
@@ -140,12 +152,32 @@ impl oid {
...
+    #[inline]
+    #[must_use]
+    pub fn hex_to_buf<'a>(&self, buf: &'a mut [u8]) -> &'a mut str {
+        let num_hex_bytes = self.bytes.len() * 2;
+        faster_hex::hex_encode(&self.bytes, &mut buf[..num_hex_bytes]).expect("to count correctly")
+    }
+
+    /// Write ourselves to `out` in hexadecimal notation.
+    #[inline]
+    pub fn write_hex_to(&self, out: &mut dyn std::io::Write) -> std::io::Result<()> {
```
Byron, line 170: Yeah, why not.
These methods feel oddly familiar, even though they are new.

Ah, now I see they just moved.

gix-hash/src/verify.rs
```diff
@@ -10,9 +10,9 @@ pub struct Error {
 }
 
 impl oid {
-    /// Verify that `self` matches the `expected` object ID.
+    /// Verify that `self` matches the `expected` object id.
```
Byron: I didn't fully internalise why `id` is better than `ID`. Maybe because it's `ObjectId`, it's best to write `object id`?

gix-hash/tests/hash/hasher.rs
```diff
@@ -14,10 +14,34 @@ fn size_of_hasher() {
...
         "The size of the return value is just 1 byte larger than just returning the object hash itself"
     );
 }
+
+#[test]
+#[cfg(all(feature = "sha256", feature = "sha1"))]
+fn size_of_try_finalize_return_type_sha1_and_sha256() {
+    assert_eq!(
+        std::mem::size_of::<Result<ObjectId, gix_hash::hasher::Error>>(),
+        34,
```
Byron, line 44: This is a little strange. Maybe it's padding that makes the tag one byte larger than I think it should be.

gix-hash/tests/hash/prefix.rs
```diff
@@ -28,6 +65,19 @@ mod cmp_oid {
...
     }
+
+    #[test]
+    #[cfg(feature = "sha256")]
+    fn it_detects_equality_sha256() {
+        let id = hex_to_id("a920bbb055e1efb9080592a409d3975738b6efb338b6efb338b6efb338b6efb3");
+        let prefix = gix_hash::Prefix::new(&id, 6).unwrap();
+        assert_eq!(prefix.cmp_oid(&id), Ordering::Equal);
+        assert_eq!(
+            prefix.cmp_oid(&hex_to_id("a920bbffffffffffffffffffffffffffffffffff")),
```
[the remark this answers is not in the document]
Byron: Good point

Byron (commented):
Thanks so much!

It took me a while to throughly review this, but I *think* it should be Ok for now ("he says with a trembling voice").

> * It feels wrong to put `empty_blob_sha256` and `empty_tree_sha256` into `oid` as the comment says `oid` is related to `Sha1`, but I’m unsure how to best resolve this tension.

Documenting `oid` as related to hashes, or specific hashes, works for me, but maybe I am missing something.

> * There’s inconsistencies with respect to spelling: `SHA256`, `SHA-256`, `Sha256`.
Yes, that's quite unfortunate. The official way of referring to these seems to be `SHA-1` and `SHA-256` respectively, and I'd love this to be done in the docs at least. This PR unifies this towards SHA1 and SHA256, which can also be argued for as it's the capitalised version of `core.objectFormat`. As long as it is consistent, it's fine, with said preference.

> * Hardcoded empty trees and empty blobs appear in more than place (for both SHA-1 and SHA-256). I feel like we should have a single source of truth here.
Sometimes one has one type or the other, but one as Deref implemented, which should automatically allow to invoke everything on `oid` from `ObjectId`. But it's implemented on ObjectId for practical reasons. Please feel free to submit a PR though if you have something specific in mind.

> * Some comments in `ObjectId` would probably benefit from being rephrased. See e. g. `ObjectId::new_sha1` or `ObjectId::from_20_bytes`.

Maybe the function names could also be better. Please feel free to submit a PR here as well.

gix-hash/Cargo.toml
```diff
@@ -17,9 +17,9 @@ test = false
 
 [features]
 default = ["sha1"]
-## Support for SHA1 digests.
+## Support for SHA-1 hashes and digests.
```
cruessler: The change to `SHA1` was intentional and based on this comment: https://github.com/GitoxideLabs/gitoxide/pull/2292#issuecomment-3639991397 Though I’m fine with any of the variants. :-)

gix-hash/src/verify.rs
cruessler: In this instance, this could have been the context. The comment is in a module called `oid` which would probably be `oId` if it were `ID`. But then, it doesn’t matter terribly much, I guess. Also, I fell `ID` stands out quite a bit and people generally seem to prefer `id` these days (though my impression could of course also be wrong).

gix-hash/Cargo.toml
cruessler: I added my comment before noticing your later comment: https://github.com/GitoxideLabs/gitoxide/pull/2292#pullrequestreview-3620383434.

Byron: I will revert this for now, as this change happened before I saw that SHA1/SHA256 are now used consistently. Which I really don't want to change 😅.
One fine day I will ask Copilot to change it in comments, but not today.

gix-hash/tests/hash/prefix.rs
```diff
@@ -107,13 +184,23 @@ mod try_from {
...
     }
 
+    #[test]
+    #[cfg(all(feature = "sha256", feature = "sha1"))]
+    fn id_too_long() {
+        let input = "abcdefabcdefabcdefabcdefabcdefabcdefabcd123123123123123123123123123123";
+        let expected = Error::TooLong { hex_len: 70 };
+        let actual = Prefix::try_from(input).unwrap_err();
+        assert_eq!(actual, expected);
+    }
```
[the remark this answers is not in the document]
Byron: Not really, but it turns out the annotation isn't needed.

gix-hash/src/verify.rs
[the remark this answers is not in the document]
Byron: @cruessler An alternative would be to call it object identifier. Or use the type name in backticks. But this is also confusing, and the reasoning behind making it lower case is unclear to me.

gix-hash/src/oid.rs
```diff
@@ -162,46 +198,51 @@ impl oid {
...
         match self.kind() {
             Kind::Sha1 => &self.bytes == oid::empty_tree_sha1().as_bytes(),
+            #[cfg(feature = "sha256")]
+            Kind::Sha256 => &self.bytes == oid::empty_tree_sha256().as_bytes(),
         }
     }
 }
 
-/// Sha1 specific methods
+/// Methods for creating special SHA1 and SHA256 `oid`s
```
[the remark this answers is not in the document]
Byron: Good!

PR #2301, opened 2025-12-22
update to zlib-rs 0.5.5

folkertdev:
fixes https://github.com/GitoxideLabs/gitoxide/issues/2265

zlib-rs now correctly handles an input/output buffer size larger than `c_uint`. Internally `c_uint::MAX` is now used as the length of the buffer. The caller is responsible for handing the (unlikely) case of zlib actually using that full buffer and not having found the logical end of the input.


Byron:

PR #2307, opened 2025-12-28
Add feature flag `blame-experimental` to `gix`

cruessler:
The proposed feature flag makes it easier to run tests comparing the output of `gix blame` using `imara-diff` 0.1 vs. 0.2.


Byron:
Thanks a lot!

This should make it easy to use blame with mostly fixed sliders anywhere, and also make testing via `gix` (CLI) much easier as one can just create a local build with it.

In the spirit of supporting experiments, I made the feature more prominent in the generated feature docs by putting it under a new headline.

PR #2324, opened 2025-12-31
Fix various Windows test bugs

EliahKagan:
This fixes a few bugs in the test suite, which either primarily or solely affect Windows environments. These bugs do not currently appear to cause any tests to wrongly fail or wrongly pass on CI, but in some common local Windows setups they prevent a few tests from passing that ought to pass. See the commit messages for complete details.

### DIFF
--- a/gix-filter/tests/filter/driver.rs
+++ b/gix-filter/tests/filter/driver.rs
@@ -493,7 +493,7 @@ pub(crate) mod apply {
                     .map(BStr::new)
                     .map(gix_quote::single)
             });
-            CAT.as_deref().map(BStr::new).unwrap_or_else(|| b"cat.exe".into())
+            CAT.as_deref().map_or_else(|| b"cat.exe".into(), BStr::new)
         } else {
             b"cat".into()
         }


gix-command/tests/command.rs
```diff
@@ -549,7 +622,7 @@ mod spawn {
 
         #[test]
         fn sh_shell_specific_script_code_with_multiple_extra_args() -> crate::Result {
-            let out = gix_command::prepare("printf")
+            let out = gix_command::prepare(":;printf")
```
Byron, line 625: A nice way to force a shell to be used!

EliahKagan: Thanks--in the context of this test, I agree. We already had:

https://github.com/GitoxideLabs/gitoxide/blob/30e1989fdc05159f13b14dbfa86711bcd4810c1a/gix-command/tests/command.rs#L601-L609

So this is really just the same technique, for which I believe it is you and not I who should be credited. :smile:

We've not been using this technique more broadly to force commands to run in a shell, and I would be reluctant to start doing so. Outside of this test setting, it's not very self-documenting--in contrast, the tests make the point clear. (I would guess that this may also be why you had not used it more broadly.)

Also inapplicable here but a concern for broader use is that it's not obvious to me that this will always work: there are situations where one could optimize a shell script with multiple commands into a sequence of calls that do not use a shell. If Git ever does that, then this might be one of the simpler and easier cases (the `:` builtin is used, but the `:` builtin can usually be optimized away because it's guaranteed to have no side effects and complete with a status of 0), which it might cover. If Git does make such a change, then presumably we would follow suit, either by switching to that strategy entirely, or by adding an option to use it for compatibilty purposes. Then some commands that contained `;` to force running in a shell on Windows would stop running in a shell.

In some cases, that wouldn't be readily detected. Here, it would be: the tests would fail. (It would also be a good thing, as it would remind us of the incompatibility.) So the possibility that the optimization might change in the future is no problem for these tests either.

I thought about this while working on other parts of this PR, since we do have some scenarios where something passes through to `gix-command` in a way that doesn't let its caller signify that a shell absolutely must be used--such as in the regression tests that use the `cat` filter. That's the main reason I implemented the "private" test helper `cat_invocation()`--which fortunately we should be able to greatly simplify or eliminate once more functionality of `gix-path` is polished and exposed for use--instead of just writing `;cat` as the filter name.

PR #2328, opened 2026-01-02
Bump reqwest to 0.13.1

NobodyXu:


Byron:
Thanks! 

It looks like `Cargo.lock` wasn't updated. But besides that, there should be no reason not to merge this.
I just wonder if this might be a breaking change, thinking it probably isn't as `reqwest` isn't exposed.

NobodyXu:
Thanks updated `Cargo.lock`, but it fails in pure-rust dependencies checking, the regex is so complex that I didn't understand it😅

https://github.com/GitoxideLabs/gitoxide/actions/runs/20654302166/job/59304263816?pr=2328#step:10:1

NobodyXu:
Ok I got it, basically that regex forbidden all -sys except for libz-rs-sys and linux-raw-sys

it failed because of new aws-lc-sys that is pulled in, which replaces ring.

I think this is fine as ring also has a bunch of pre-generated assemblies/c that requires a compiler, cc @Byron should I update CI to whitelist aws-rc-sys?

NobodyXu:
Pushed a new change to whitelist aws-rc-sys for pure-rust-build

> I just wonder if this might be a breaking change, thinking it probably isn't as reqwest isn't exposed.

I agree, it's probably fine to have a compatible release, I'd appreciate that

Byron:
It seems to build, which really does mean that `aws-lc-sys` doesn't invoke the `cc` compiler. Great!
Should be ready to merge.

NobodyXu:
Thanks!

what is the estimated release date for next release?

Byron:
I just published `gix-transport` as patch release. A great way to not have to wait till the 22nd :D, and also to see if I was wrong about my "it's not breaking" assessment ;).

NobodyXu:
Thank you!

PR #2332, opened 2026-01-02
Make sha1 optional in `gix-hash`

cruessler:
This PR makes the `sha1` flag optional in `gix-hash`. It modifies `justfile` to run `cargo check` as well as `cargo nextest` with 3 configurations: with `sha1` only, with `sha256` only, with both `sha1` and `sha256`.

It’s mostly done, but still opened as a draft, so I can wait for CI to be green before I’m going to mark it as ready.

I haven’t checked all of `gix-hash`’s comments yet, some might need minor changes in case they still reference a `sha1`-only world.

### DIFF
--- a/justfile
+++ b/justfile
@@ -159,7 +159,7 @@ unit-tests:
     cargo nextest run -p gix-worktree-tests --features gix-features-parallel --no-fail-fast
     cargo nextest run -p gix-hash --no-fail-fast
     cargo nextest run -p gix-hash --features sha256 --no-fail-fast
-    cargo nextest run -p gix-hash --no-default-features --features sha256 --no-fail-fast
+    cargo nextest run -p gix-hash --no-default-features --features sha256 --no-fail-fast # TODO: make this actually work by removing 'sha1' from default features.
     cargo nextest run -p gix-object --no-fail-fast
     cargo nextest run -p gix-object --features verbose-object-parsing-errors --no-fail-fast
     cargo nextest run -p gix-tempfile --features signals --no-fail-fast


cruessler:
I just went through all instances of `feature = "sha1"` to check whether there was any context that we would need to update in comments or code, but I only found one `TODO:` that I was able to remove as the tests that it mentioned should be added were already there.

cruessler:
I’m not entirely sure I get what you mean. :-)

This is the error messages I get when trying to compile `gix-hash`:

```
branch-8 on  make-sha-1-optional is 📦 v0.49.0 via 🦀 v1.92.0
❯ cargo check --package gix-hash --no-default-features
    Checking gix-hash v0.21.1 (/home/christoph/worktrees/gitoxide/branch-8/gix-hash)
error: Please set either the `sha1` or the `sha256` feature flag
  --> gix-hash/src/lib.rs:13:1
   |
13 | compile_error!("Please set either the `sha1` or the `sha256` feature flag");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

[…]

Some errors have detailed explanations: E0004, E0308, E0665.
For more information about an error, try `rustc --explain E0004`.
warning: `gix-hash` (lib) generated 39 warnings
error: could not compile `gix-hash` (lib) due to 15 previous errors; 39 warnings emitted
```

Is there something missing?

Apart from that, I’m very much interested in moving hash selection to `gix` (and possibly other crates that need to be involved). :smile:

Byron:
I am talking about `cargo test -p gix-hash --no-default-features`, which doesn't fail.

And indeed, to fix it one would have to remove the default `sha1` feature, and leave the selection to `gix`.
But… I realise this probably doesn't work as so many crates need access to a selected hash.
So `gix-object` for instance also needs to expose the hash selection, and those who need it, and… it's a nightmare.
Maybe one can let the tests for each consumer of `gix-hash` set its hash, which would make sense as eventually we have to add hash-specific tests and that needs a way to be controlled.

Then, one will probably get around exposing the hash toggle in every consumer of `gix-hash`.

Issue #2338, opened 2026-01-04
New `aws-lc-sys` dependency breaks `max-pure` on powerpc64le


EliahKagan:
### Current behavior 😯

#2328 upgraded `reqwest`. As noted in https://github.com/GitoxideLabs/gitoxide/pull/2328#issuecomment-3704783897, this replaced `ring` with `aws-lc-sys`. This was allowlisted as a recognized `*-sys` dependency in 28b675e453de39bea0dbf66eb47f90c73671437b, which fixed CI and allowed #2328 to be integrated. Subsequently, a patch release of `gix-transport` was made. But no release of the `gitoxide` crate has yet been made since.

**This change breaks the `release.yml` workflow.** Specifically, the workflow depends on being able to build `gix` and `ein` binaries for a variety of architectures. Failure to do so will not prevent `gitoxide` from being published to crates.io, and with the way we currently do things it won't prevent a GitHub release from being made, but it does block attaching any assets to that GitHub release.

It looks like the only broken architecture is powerpc64le. Notably, `max-pure` is broken. [**Here's the failing CI job, where the failure is reported.**](https://github.com/EliahKagan/gitoxide/actions/runs/20690681225/job/59398335972#step:7:814) This is in my fork, where I test releasing.

(I actually do not test it regularly. I found this because #2337 modifies `release.yml`, so I tested it on my fork with a test tag on a commit in that PR, and I found that it broke in a way that didn't make sense for any of the changes there, and that the build failed when retried. So I made a test tag at the tip of the main branch in my fork and tested with that, and found it produced the same failure. The above-linked CI run is from that tag at the current tip of main.)

### Expected behavior 🤔

The release workflow should be able to build `max-pure` for all the targets it's been able to build it on for a long time and that we regularly release binaries for.

I'm also not clear on the justification for allowing `max-pure` to link to `aws-lc-sys` in the first place. It's a binding to the [aws-lc](https://github.com/aws/aws-lc) library. This looks like a C library, with some code in assembly language and some in C++ (though perhaps the latter is just bindings). The error messages (shown below) are failures to link to missing symbols in an object file named `bcm.c.o`; this name suggests it was built from `bcm.c`. My understanding in general of the reason only recognized `-sys` crates are allowed for `max-pure` is that, otherwise, we get native code that doesn't satisfy goals for `max-pure`.

[...]

PR #2340, opened 2026-01-04
Replace `dtolnay/rust-toolchain` with manual `rustup`

EliahKagan:
This builds on #2337 by replacing all uses of `dtolnay/rust-toolchain` with `rustup` commands, as described in the "Plan" section there. See the commit message here (dc980b24ed5ede328394d5aa976f513c22af404e) for considerable details including background, rationale, and why an alternative action was not as good of a choice.

I first attempted to do this change (within my fork) via a Copilot PR. An error occurred, but the generated prompt in https://github.com/EliahKagan/gitoxide/pull/141 (if one expands the details) is potentially a useful alternative summary. The diff here is the same as what Copilot would've done.

Analogous to [that release test](https://github.com/EliahKagan/gitoxide/actions/runs/20691845377) of changes to `release.yml` in #2337, a release test achieved by cherry-picking the commit used there onto this branch was done in [this release test](https://github.com/EliahKagan/gitoxide/actions/runs/20691893860). (That was before rebasing this, but the rebase was just to bring this ahead of the merge commit--it didn't change the contents of `release.yml` or anything else.) Based on that, this seems not to break `release.yml` (so it should work again once some production-suitable fix/workaround for #2338 is devised).

However, this PR remains a draft until I check that the expected security scanning alerts, for both CodeQL and Zizmor, have appeared in the Security tab due to #2337. (Most of them should then go away as a result of this PR.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -187,7 +187,10 @@ jobs:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           persist-credentials: false
-      - uses: dtolnay/rust-toolchain@stable
+      - name: Install Rust
+        run: |
+          rustup update stable
+          rustup default stable
       - uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2.8.2
         with:
           save-if: ${{ github.ref == 'refs/heads/main' }}
@@ -211,7 +214,10 @@ jobs:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           persist-credentials: false
-      - uses: dtolnay/rust-toolchain@stable
+      - name: Install Rust
+        run: |
+          rustup update stable
+          rustup default stable
       - uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2.8.2
         with:
           save-if: ${{ github.ref == 'refs/heads/main' }}
@@ -239,7 +245,11 @@ jobs:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           persist-credentials: false
-      - uses: dtolnay/rust-toolchain@stable
+      - name: Install Rust
+        shell: bash
+        run: |
+          rustup update stable
+          rustup default stable
       - uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2.8.2
         with:
           save-if: ${{ github.ref == 'refs/heads/main' }}
@@ -297,7 +307,11 @@ jobs:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           persist-credentials: false
-      - uses: dtolnay/rust-toolchain@stable
+      - name: Install Rust
+        shell: bash
+        run: |
+          rustup update stable
+          rustup default stable
       - uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2.8.2
         with:
           save-if: ${{ github.ref == 'refs/heads/main' }}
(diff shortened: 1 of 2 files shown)


EliahKagan (commented):
The expected code scanning alerts, from both CodeQL and Zizmor, were created and are visible in the Security tab, as intended in #2337. That's the main thing I wanted to check before merging this (which is expected to close most of the alerts, since it fixes the condition it describes).

The failure here that I mentioned in https://github.com/GitoxideLabs/gitoxide/pull/2339#pullrequestreview-3624964205 looks like it's an occasional network error and not due to any changes here.

Therefore, I think this is ready to merge once its tests pass.

EliahKagan:
> Adds explicit `shell: bash` specifications where needed for cross-platform compatibility

These are actually just to be able to run a script step of multiple commands that stops and reports failure on the first failing command. Maybe this reason for using step-level `shell: bash` isn't obvious and should be made clearer in some way.

Byron:
> > Adds explicit `shell: bash` specifications where needed for cross-platform compatibility
> 
> These are actually just to be able to run a script step of multiple commands that stops and reports failure on the first failing command. Maybe this reason for using step-level `shell: bash` isn't obvious and should be made clearer in some way.

If Copilot doesn't know that even, then I think it's very non-obvious. Even reading this I can't imagine why it has this behaviour except that it's something they explicitly put in to make using scripts easier, while having to keep the default behaviour backwards-compatible to not fail fast.

EliahKagan:
> If Copilot doesn't know that even, then I think it's very non-obvious.

I can put comments on every step-level `shell: bash`, much as we do for `shell: bash` when it is set at the workflow or job level. I think that should help both humans and LLMs.

I think we should also somehow make clearer (at least to humans reading the workflow file) why we're using different shells in the first place in some jobs, even when setting `shell: bash` at the job level would simplify commands, and even though we're using job-level `shell: bash` in some other jobs. There are good reasons to run `cargo nextest` from different shells depending on the operating system--`bash` on a Windows GHA runner is Git Bash, whose environment subtly differs, and we intend the test suite to work properly even when run from outside a Git Bash environment. But I don't think this is currently clear enough.

> Even reading this I can't imagine why it has this behaviour except that it's something they explicitly put in to make using scripts easier, while having to keep the default behaviour backwards-compatible to not fail fast.

I realized I forgot to reply to something earlier about this. I've done so now; see https://github.com/GitoxideLabs/gitoxide/pull/1363#issuecomment-3711904953. In short, PowerShell doesn't quite have this feature, and also it makes different judgments from POSIX-compatble shells about what constitutes an error.

But it *almost* has it, such that the default behavior on CI for PowerShell should maybe be different from what it is currently. The situation is:

1. `set -e` itself is subject to [huge mismatches](https://mywiki.wooledge.org/BashFAQ/105) between its actual effect and its intuitive effects, and many shells intentionally avoid replicating that design. This includes not only shells like PowerShell that are meant to be very different from Unix-style shells, but also shells like `fish` that are meant to be intuitive for users of other Unix-style shells.
2. PowerShell does have an `$ErrorActionPreference` variable. Setting `$ErrorActionPreference = 'stop'` fails fast on errors. When an external command fails, however, that's not considered an error in PowerShell. This *is* set in PowerShell script steps in GHA.
3. PowerShell doesn't even automatically return the exit status of the last command to have run. However, this can be done reliably by appending a line to the script. GHA runners append `if ((Test-Path -LiteralPath variable:\LASTEXITCODE)) { exit $LASTEXITCODE }` when making the PowerShell script file for a script step.
4. *Some versions* of PowerShell support `$PSNativeCommandErrorActionPreference`. If this is set to `$true` then something approximating `set -e` is achieved. I'm not sure how good the approximation is. I think this is still conisdered an experimental feautre, but I'm not sure. This would be usable in `pwsh` but not in `powershell`. But GitHub-hosted Windows runners have both. `pwsh` is the default on Windows when present.
5. The GitHub Actions documentation [documents](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#exit-codes-and-error-action-preference) points (2) and (3). But it characterizes the effect of the former as, "Fail-fast behavior when possible." It seems to me that this is misleading, since `$PSNativeCommandErrorActionPreference` is not set even when doing so is possible. (But it could be that I'm overestimating the uplift that confers in practice for achieving fail-fast behavior across separate commands in a PowerShell script.)

The different `bash` and `pwsh`/`powershell` behavior on CI led to the bug in #1556 that was fixed in #1559, which was not caused by any change in #1363 but would perhaps have been found in review had I not forgotten to reply to https://github.com/GitoxideLabs/gitoxide/pull/1363#issuecomment-2100695412 there. (I was reminded recently of the bug fixed in #1559, because the more severe bug I introduced in #2337 and fixed in #2342 resembles it, though this recent bug did not relate to fail-fast behavior in scripts.)

Byron:
Thanks a lot for sharing, I learned a lot!

Issue #2308, opened 2025-12-28
Tracking: match `git diff` slider algorithm more closely in `gix-diff`


cruessler:
This is a tracking issue related to what’s known as the “diff slider problem”. When calculating the diff between two versions of a file, a hunk can be ambiguous with respect to where it starts and ends. Consider the following example:

```diff
 fn test_a_different_thing() {
 }
 
+#[test]
+fn test_something() {
+}
+
 #[test]
 fn test_something_else() {
 } 
```

The following diff is, technically speaking, equally valid, even though human readers most likely would prefer the above diff.

```diff
 fn test_a_different_thing() {
 }
 
 #[test]
+fn test_something() {
+}
+
+#[test]
 fn test_something_else() {
 } 
```

`gix-diff` relies on `imara-diff` 0.1 under the hood for it’s diffing algorithm. There’s a significant number of cases, though, where `imara-diff` 0.1 and `git diff` yield different results as they employ different heuristics with respect to moving ambiguous diff hunks up or down.

There is an experimental feature flag in `gix-diff` (`blob-experimental`) that uses `imara-diff` 0.2. `imara-diff` 0.2 comes closer to what `git diff` returns, though it still doesn’t match its output in all cases.

We assume that most people would expect `gix-diff` to yield the same results as `git diff`. Based on this assumption, we want to track progress towards closing the gap between `gix-diff` and `git diff` with respect to slider placement.

## Tasks

* [ ]  Extract a set of baseline tests for use in `gix-diff/tests`
* [ ] Document how closely `imara-diff` matches `git diff`
* [ ] Port `git diff` postprocessing heuristic for use with `imara-diff` 0.2

## Sources

https://github.com/mhagger/diff-slider-tools

## Related

* [x] #2307
* [x] #2288
* [x] #2287
* [x] #2286
* [x] #2285
* [x] #2276
* [x] #2269
* [x] #2197

- Merged, but later reverted: #2011

Byron:
Thanks for getting this started!

Could we have a task-list/checkbox list in the issue description to show what's there and what's still planned to do? This is more of a tool for both of us to align, and I would then add what I think to your initial take on it.

Thanks again!

cruessler:
I’ve updated the issue, adding the PRs that are related to `imara-diff` 0.2. Feel free to make further changes if something’s missing or not accurate!

Byron:
[...]

PR #2343, opened 2026-01-04
[not intended for merging] experiment: use `git diff` in `blob_changes`

cruessler:
This is an experiment that’s supposed to help me validate the `gix-blame` algorithm. `gix-blame` currently rests on two foundations: 1. is the algorithm that traverses a file’s history and keeps track of blamed and unblamed hunks. 2. is the algorithm used for diffing subsequent states of a file throughout its history. Basically, (1.) orchestrates `gix-blame`, using (2.) to do most of the work of following individual lines throughout a file’s history. `gix-blame` currently doesn’t match `git blame` in all cases, but we set out to change that in #2308 as we expect most people to want `gix-blame` to match `git blame`’s output.

Since the algorithm used for (2.) differs from the one used by `git`, it has been somewhat challenging to validate (1.) on a large scale as it is quite hard to tell whether a divergence between `gix-blame` and `git blame` is due to differences in (1.) or (2.). This PR tries to approach this issue from a different angle, substituting `gix-diff` in (2.) by just shelling out to `git diff`. That way, it hopefully becomes much easier to compare `gix-blame` using `gix-diff` vs. `gix-blame` using `git diff` vs. `git blame` on the scale of entire repositories.

I’m opening this PR mostly to get CI feedback and to make this work visible. We probably can close it once I’m done with it.

### DIFF
--- a/gitoxide-core/src/repository/blame.rs
+++ b/gitoxide-core/src/repository/blame.rs
@@ -39,6 +39,7 @@ pub fn blame_file(
     let cache: Option<gix::commitgraph::Graph> = repo.commit_graph_if_enabled()?;
     let mut resource_cache = repo.diff_resource_cache_for_tree_diff()?;
     let outcome = gix::blame::file(
+        repo.workdir().expect("TODO").to_path_buf(),
         &repo.objects,
         suspect,
         cache,
--- a/gix-blame/src/file/function.rs
+++ b/gix-blame/src/file/function.rs
@@ -981,6 +981,7 @@ fn blob_changes(
 #[allow(clippy::too_many_arguments)]
 #[cfg(feature = "blob-experimental")]
 fn blob_changes(
+    _worktree_path: PathBuf,
     odb: impl gix_object::Find + gix_object::FindHeader,
     resource_cache: &mut gix_diff::blob::Platform,
     oid: ObjectId,
--- a/gix-blame/src/file/function.rs
+++ b/gix-blame/src/file/function.rs
@@ -1,4 +1,4 @@
-use std::num::NonZeroU32;
+use std::{num::NonZeroU32, path::PathBuf};
 
 use gix_diff::{blob::intern::TokenSource, tree::Visit};
 use gix_hash::ObjectId;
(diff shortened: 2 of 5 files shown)


Byron:
This is genius, loving it!

If you want, you can also put the `git diff` shelling behind an experimental Cargo feature toggle, or even a Rustc feature flag.
That way you can turn it on with ` RUSTFLAGS="--cfg my_custom_flag"` (maybe like this, I asked LLM).

That way it might be easier to test in the real world, and we could merge this.

cruessler:
I was hoping that merging behind a flag might be possible and I really like the idea of using `RUSTFLAGS`! There’s one issue, though, as I had to change the API of `gix_blame::file::function::file` to include `worktree_path` as that’s needed by `git diff`. Is there a way to get `worktree_path` from any of the other parameters maybe?

Byron:
Since the `gix` based diff definitely has that information, I am surprised it's not readily available. But somewhere, a level up, it must definitely be there, so I think this can be cfg_if'ed on that level, with adjustments to the callsite. It's not super pretty to adjust the call site (one with, and one without the worktree path), but it gets the job done and is probably needed to make this testable in the real world.

cruessler:
From what I can tell, the issue is that we pass `repo.objects` as `odb: impl gix_object::Find + gix_object::FindHeader`. While `repo.objects` has `.store().path()` which we could use to infer the workdir (at that level, we can also get it through `repo.workdir()`), `odb` does not provide this piece of information. So I agree that the easiest workaround probably is to feature flag the `worktree_path` as an additional argument.

Byron:
Let's try that then. If it doesn't come together, I would also take a look and actually try it myself, so far I was just looking at this from my armchair.

Byron:
Is this PR still relevant?
If so it might be a good time to rebase given the recent changes to the diff implementation.

Byron:
A lot has changed since this PR was first created.
Is it still relevant?

cruessler:
> A lot has changed since this PR was first created.
> 
> Is it still relevant?

Sorry for not responding earlier! This can be closed, I’ll move this branch/PR to my own fork. (I still like the idea and want to investigate whether this will uncover any issues in the blame implementation, but currently lack the time to properly focus on it.)

Byron:
> Sorry for not responding earlier! 

How dare you sleep at night ;)!

Of course, no worries, and sorry for not reading what it's for. I just saw it rots and thought then it's probably time to do something.

Closing it here seems like a good choice now, and I hope to have it back once there is more time.

PR #2346, opened 2026-01-06
gix-pack release

Byron:
This is just a release of `gix-pack`, going through a PR due to branch protections.


gix-transport/Cargo.toml
```diff
@@ -79,8 +79,8 @@ path = "tests/async-transport.rs"
 required-features = ["async-client"]
 
 [dependencies]
-gix-command = { version = "^0.6.4", path = "../gix-command" }
-gix-features = { version = "^0.45.1", path = "../gix-features" }
+gix-command = { version = "^0.6.5", path = "../gix-command" }
+gix-features = { version = "^0.45.2", path = "../gix-features" }
```
EliahKagan, line 83: The `[dependencies]` section is the only change in `gix-transport/Cargo.toml`. I don't expect this to require a release of `gix-transport`, and the version number above in this `Cargo.toml` file is unchanged. But I figured I'd ask just in case, in view of the PR title.

Byron: The PR title definitely is wrong, my apologies. I corrected it now, just to be less confusing.

Only `gix-pack` along with a couple of crates were supposed to be released via `cargo smart-release --allow-empty-release-message --allow-fully-generated-changelogs  --auto-publish-of-stable-crates gix-pack`. However, while it ran I panicked for a moment as it seemed to release a whole lot of other crates.

In the end, it created GitHub releases only for the crates I thought it would create releases for, but the number of actual releases seems higher.

Screenshot 2026-01-06 at 17 50 14

Comparing the number of tags created with the number of releases should bring clarity though, and despite being puzzled, I have trouble to believe that it could do such a thing.

EliahKagan: Thanks for explaining the situation! Should there be an issue for this behavior of `cargo-smart-release`? Or did this happen due to factors outside it?

Another thing I noticed, and wasn't sure if was intentional, is that [the changelog entry](https://github.com/GitoxideLabs/gitoxide/releases/tag/gix-pack-v0.64.1) was written to the GitHub release but not to the [committed file](https://github.com/GitoxideLabs/gitoxide/blob/main/gix-pack/CHANGELOG.md). That seems bad to be, though I haven't looked into whether it's happened before or if it might somehow be intentional.

EliahKagan: It looks like changelogs not being commited is a broder problem:

- https://github.com/GitoxideLabs/gitoxide/discussions/2362

I don't know if this is due to some regression in `cargo-smart-release` or something else.

Byron: > Thanks for explaining the situation! Should there be an issue for this behavior of cargo-smart-release? Or did this happen due to factors outside it?

I don't think it was outside influence, as it did just run through without issues.

> Another thing I noticed, and wasn't sure if was intentional, is that [the changelog entry](https://github.com/GitoxideLabs/gitoxide/releases/tag/gix-pack-v0.64.1) was written to the GitHub release but not to the [committed file](https://github.com/GitoxideLabs/gitoxide/blob/main/gix-pack/CHANGELOG.md). That seems bad to be, though I haven't looked into whether it's happened before or if it might somehow be intentional.

This is a new problem for me, and it's expected to write changelogs to files and then add them to a commit.
Fortunately, that issue will fix itself the next time the changelogs are generated, after all, they always regenerate from the entire history.

`cargo changelog --write gix` already does the trick, but I noticed that the tags are out of order. Try it yourself and you know what I mean. Maybe it's us misusing `jiff`? It says that v0.74.1 is 1 day older than the tag before, 0.77.0, which is months in the future.

> I don't know if this is due to some regression in cargo-smart-release or something else.

It feels like a regression, but it's probably also related to some input data that it doesn't manage well. After all, CSR didn't really change, despite dependency updates maybe.

PR #2354, opened 2026-01-08
fix: respect .git/info/exclude in worktrees

canac:
When running `gix status` inside of a worktree, it wasn't respecting `.git/info/exclude` entries in the common git dir. This PR fixes that.

**AI Disclosure** 🤖: I used Claude Code to assist with writing this PR. I manually refined the test case and verified the fix.

### DIFF
--- a/gix/tests/fixtures/generated-archives/.gitignore
+++ b/gix/tests/fixtures/generated-archives/.gitignore
@@ -9,3 +9,4 @@
 /make_diff_repos.tar
 /make_submodule_with_worktree.tar
 /repo_with_untracked_files.tar
+/make_worktree_repo_with_info_exclude.tar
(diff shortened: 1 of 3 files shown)


Byron:
Thanks a lot for the fix, and for the disclaimer!

I will take care of merging it very soon.

canac:
Absolutely! Thanks for your hard work on this project! I use it indirectly through Starship.

Byron:
And I just double-checked the [Git configuration](https://git-scm.com/docs/gitignore), which persistently mentions `$GITDIR` without ever talking about worktrees.

This makes me think it's probably out of date, and is what I used to implement this in the first place.
Most importantly, there doesn't seem to be a worktree-specific excludes file yet.

PR #2365, opened 2026-01-10
Upgrade `gix-testtools` dependencies to have a version with `gix-date` v0.12.0 or larger

Byron:
Everything below that has a security advisory.

This also removes the crate from the workspace to allow a release without auto-bumping
dependencies.


Byron:
I wanted to, but realized no change was needed too late. So I used this commit for release tag.
Plain cargo publish was used as well, and I have no clue if this breaks future csr releases,but will find out.

EliahKagan:
Thanks for clarifying!

> Plain cargo publish was used as well, and I have no clue if this breaks future csr releases,but will find out.

I don't know why that would break future releases--but there's also a huge amount I don't know about `cargo-smart-release`, most things actually, so me not understanding why there would be a problem shouldn't be taken to mean there won't be one. :smile:

Byron:
I also don't know why it would break, I just vaguely remember that there where issues in the past and I don't know the why's and what's, just that… everything seems possible with CSR 😅.
I found that doing a physical release dance with a ceremonial sacrifice of compiler code printed on paper greatly increases the chances, so I will just keep doing that ;).

PR #2368, opened 2026-01-11
Set default-features to false for curl

bbqsrc:
curl pulls in openssl by default, and even specifying rustls will cause this.

This enables a pure rustls build.


Byron (approved):
Thanks a lot, great catch.

The CI failure seems to be legit, but I guess that's good as one will now have to select which backend to use.

PR #2370, opened 2026-01-12
chore: upgrade to latest `flate2` to remove `zlib-rs-sys` from dependencies

Byron:


EliahKagan:
Per https://github.com/GitoxideLabs/gitoxide/pull/2372#issuecomment-3739837677, I had Copilot post-review this to check my hypothesis that it would not be able to identify that the changes here allowed the `-sys` crate to be removed from the `-sys` allowlist for "pure" builds, since done in c97ef4e62be7723a5eb078db07cd96a6f9c7e981 (#2372).

This was a Copilot experiment, not any concern about the changes here, which were good (even though the change in c97ef4e62be7723a5eb078db07cd96a6f9c7e981 could've also been included). As expected, Copilot was not able to identify this--I don't think code far away in the project is automatically included in its context, so it would have to use its search tool to find relevant TODOs.

Possibly it could be made to do this by editing the copilot instructions. Searching for potentially relevant TODOs everywhere in the project, even in other files not referred to from where one makes a change, could be tedious and flow-interrupting for developers to do as a regular practice. Yet I suspect it wouldn't degrade Copilot performance to do it.

Byron:
Thanks so much for catching this!

When I had the opportunity today to remove `libz-rs-sys` from the dependency tree, it felt very important and I wondered if there was anything else, but came up empty :D.

But this was it, `max-pure` just became a little bit purer.

> Possibly it could be made to do this by editing the copilot instructions. Searching for potentially relevant TODOs everywhere in the project, even in other files not referred to from where one makes a change, could be tedious and flow-interrupting for developers to do as a regular practice. Yet I suspect it wouldn't degrade Copilot performance to do it.

Evaluating the effectiveness of these instructions is pretty hard as it's quite subjective. But maybe that's all we need - in any case, please do feel free to experiment.

PR #2374, opened 2026-01-13
Turn `Exn::into_box()` to `Exn::into_inner()`.

Byron:


Byron:
@EliahKagan It really seems that `Rebase` is now very sticky here, and I don't know why that is. This is just FYI.

PR #2375, opened 2026-01-13
feat: filters and partial cloning: initial support

### LINKED ISSUE #1046: Minimal implementation of partial clones & promisor objects
### Summary 💡

Tracking a minimal implementation of partial clones and their corresponding promisor objects/remotes, as discussed in #1041.

**What is a partial clone?**

[Partial clone](https://www.git-scm.com/docs/partial-clone/2.42.0) is a recent(ish) feature of `git` that allows the client to fetch a subset of objects from a remote repository based on criteria (i.e. a "filter") of its choosing. The missing objects are referred to as "promisor objects", and are expected to be able to be provided by "promisor remotes" on-demand after the clone, as needed.

The most common use-case of partial-clone are where the client requests a clone with either no historical blobs (e.g. `--filter=blob:none`), or only historical blobs under some size threshold (e.g. `--filter=blob:512k`). Tree objects can also be filtered by a partial clone, however that use-case is far less common.

**Lessons learned from `git`**

Because partial clone was retrofitted into `git`, there are several performance gaps that have not yet been resolved. Operations like `fetch` and `checkout` behave exactly as one would expect - the missing objects are fetched in a single transaction with the remote. Other operations, such as `blame` and `rebase`, do *not* do this, and instead end up lazily fetching missing objects one at a time (each with a separate transaction to the remote), which significantly slows things down.

To implement partial clones efficiently, operations that traverse history and require inspecting the contents of blobs and trees need to:

1. Determine the set of object IDs needed by the operation (typically by walking a commit graph)
2. Fetch any missing objects in a single transaction to the remote
3. Continue with their "business logic"

That said, this feature does *not* aim to implement the optimized approach to partial clones across the board. However we would like to see APIs designed to facilitate the optimized approach, and possibly one implementation of the optimized approach to be used as a reference and proof that things can be made to work as expected.

**Tasks**

* [ ] minimal implementation to handle existing (created by `git`) partial clones
   * [ ] create a text fixture to generate a repository that will result in missing blobs in a blobless clone
   * [ ] implement a repository verification test that includes verifying all historical blobs
       * I'm not seeing anything like this today in `gitoxide`, but I may be missing something...
   * [ ] create a unit test that fails a repository verification when using a blobless clone
   * [ ] implement (in `gix-odb`?) handling of missing objects when promisor remotes are available
      * This will be the naive/inefficient approach, which will cause the verification unit test above to perform a remote transaction per missing object
* [ ] minimal implementation to perform a fresh partial clone
   * **TBD** Need to dig in more and fill this in...

### Motivation 🔦

_No response_

staehle:
Resolves #1046 

Hi @Byron! I'm Jake Staehle and I work at Roku here with @cesfahani. We've had an internal version of this for our internal tools for quite some time, but I recently updated our `gitoxide` version to `0.49` and had to refactor this code myself for all the changes since `0.40` -- I would very much rather not do that again, so it's upstreaming time :) 

Introduce `gix::remote::fetch::ObjectFilter` (currently blob filters only) and plumb it through clone/fetch all the way into the fetch protocol, returning a clear error if the remote doesn’t advertise `filter` capability.

Also persist the partial-clone configuration on clone (`remote.<name>.partialclonefilter`, `remote.<name>.promisor`, `extensions.partialclone`) so the repository is marked as a promisor/partial clone in the same way as Git.

On the CLI/plumbing side, add `--filter <spec>` to `gix clone`, and allow fetch arguments to be either refspecs or raw object IDs (treated as additional wants).

Includes tests for filter parsing and for persisting partial-clone settings during clone, as well as a full round-trip test creating a blobless bare clone using `gix`, creating (using real `git`) a worktree from that, and then using `gix` verify that `--filter=blob:none` is working!

I did create a new repo https://github.com/staehle/gitoxide-testing for that last test to work properly, with a small set of known commits, tags, and blob objects. If you think this is useful, I'd be glad to transfer the repo to you / the `GitoxideLabs` group!

Let me know what you think! Thanks!

Credit:
 - Original work by Cameron Esfahani <cesfahani@roku.com>
 - Rewritten for modern Gitoxide by Jake Staehle <jstaehle@roku.com>


gix/src/remote/connection/fetch/receive_pack.rs
```diff
@@ -98,6 +98,12 @@ where
             });
         }
 
+        let filter_argument = self
+            .filter
+            .as_ref()
+            .map(crate::remote::fetch::ObjectFilter::to_argument_string)
+            .filter(|_| self.additional_wants.is_empty());
```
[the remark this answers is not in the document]
staehle: Hard disagree. This is intentional -- `additional_wants` SHOULD ignore blob filters, since if you specifically want to checkout a commit, you NEED those blobs.

gix/src/remote/connection/fetch/receive_pack.rs
```diff
@@ -274,18 +283,71 @@ impl gix_protocol::fetch::Negotiate for Negotiate<'_, '_, '_> {
...
     fn add_wants(&mut self, arguments: &mut Arguments, remote_ref_target_known: &[bool]) -> bool {
-        negotiate::add_wants(
-            self.objects,
-            arguments,
-            self.ref_map,
-            remote_ref_target_known,
-            self.shallow,
-            negotiate::make_refmapping_ignore_predicate(self.tags, self.ref_map),
-        )
+        if self.additional_wants.is_empty() {
```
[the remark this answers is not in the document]
staehle: It looks confusing so I get why copilot doesn't understand, but the condition is not inverted, and the behavior matches the surrounding design.

When `additional_wants` is empty, `add_wants()` delegates to `negotiate::add_wants` and then iterates `self.additional_wants` (which is empty), so that loop is a harmless no-op. This is the "normal fetch" path.

When `additional_wants` is non-empty, the code intentionally avoids `negotiate::add_wants` and instead builds the wants list itself (including `want_ref()` when supported) and then appends the additional object-id wants. This aligns with the other deliberate behavior you already saw:

- `mark_complete_and_common_ref()` forces `MustNegotiate` when additional wants exist.
- `one_round()` returns an "empty round" and intentionally skips have exchange/negotiation rounds since we want to ask for these objects directly.

So, flipping the condition to `!self.additional_wants.is_empty()` would make the non-empty case call `negotiate::add_wants` and then fall through into the manual wants logic, which would duplicate/alter behavior and be wrong.

Byron:
Thanks a lot for upstreaming this feature, it's much appreciated!

I hope to get to giving it the attention it deserves this weekend.

gix/src/clone/fetch/util.rs
```diff
@@ -16,13 +16,49 @@ enum WriteMode {
...
+
+    if let Some(filter_spec) = filter {
+        let subsection = remote_name.to_str().map_err(|err| Error::RemoteNameNotUtf8 {
+            remote_name: remote_name.clone(),
+            source: err,
+        })?;
+        let mut remote_section = config.section_mut_or_create_new("remote", Some(subsection.into()))?;
+
+        while remote_section.remove("partialclonefilter").is_some() {}
+        while remote_section.remove("promisor").is_some() {}
```
Byron, line 39: Note to self: Create `gix-config` remove all.

gix/src/clone/fetch/util.rs
```diff
@@ -16,13 +16,49 @@ enum WriteMode {
...
+
+        let partial_clone_filter = ValueName::try_from("partialclonefilter").map_err(|err| Error::ConfigValueName {
```
Byron: Create new config-tree keys for `partialCloneFilter` and `promisor`.

gix/src/remote/connection/fetch/receive_pack.rs
```diff
@@ -274,18 +283,71 @@ impl gix_protocol::fetch::Negotiate for Negotiate<'_, '_, '_> {
...
             self.shallow,
             negotiate::make_refmapping_ignore_predicate(self.tags, self.ref_map),
-        )
+        )?;
+
+        if self.additional_wants.is_empty() {
+            return Ok(action);
+        }
+
+        Ok(match action {
```
Byron, line 310: Note to self: try to understand this.

gix/src/remote/connection/fetch/receive_pack.rs
```diff
@@ -294,12 +356,24 @@ impl gix_protocol::fetch::Negotiate for Negotiate<'_, '_, '_> {
...
+                self.negotiator.deref_mut(),
+                &mut *self.graph,
+                state,
+                arguments,
+                previous_response,
+            );
+        }
+
+        Ok((
+            negotiate::Round {
```
Byron, line 388: Note to self: Oh try to understand this.

gix/src/remote/fetch/object_filter.rs
```diff
@@ -0,0 +1,111 @@
...
+    Unsupported(String),
+    /// The provided blob size limit could not be parsed as an integer.
+    #[error("invalid blob size limit '{value}'")]
+    InvalidBlobLimit {
+        /// The string that failed to parse.
+        value: String,
+        /// The underlying parse error.
+        source: ParseIntError,
+    },
+}
```
Byron, line 49: Note to self: Turn this to `gix-error::ValidationError` right away.

Byron (commented):
I'd hope to finish the review and changes today. It should be alright, but I hope I will get the negotiation logic.

Due to the lack of a server implementation this is always tricky to test, but maybe that's what the journey tests are there for.

I also wonder what it would take to port them into the `gix` (crate) test-suite.


### Tasks

* [ ] review no-network commit
* [ ] address all comments

staehle:
Apologies, GitHub just did something stupid I didn't expect it to do

(I reset this branch back to the `2c60401` commit you were working on)

Byron:
@codex review

Byron:
Apologies for the long wait, I won't rest now until this is merged. Feel free to ignore the additional auto-reviews unless of course you think something is worth addressing. I thought they might help me, primarily.

laris:
Hope this PR can be merged into upstream, my current dev project require this feature.
Thanks

Byron:
While I plan to get back to this PR and continue on [my comments](https://github.com/GitoxideLabs/gitoxide/pull/2375#pullrequestreview-3736107912), a lot has changed since I last gave it a shot. My tooling changed, my workflow changed, and… agents.

For now I just use this capability to blindly fix merge conflicts, but I may not continue reviewing this feature before Git3 compatibility landed.

staehle:
> While I plan to get back to this PR and continue on [my comments](https://github.com/GitoxideLabs/gitoxide/pull/2375#pullrequestreview-3736107912), a lot has changed since I last gave it a shot. My tooling changed, my workflow changed, and… agents.
> 
> For now I just use this capability to blindly fix merge conflicts, but I may not continue reviewing this feature before Git3 compatibility landed.

Thanks @Byron! Let me know if you need assistance!

PR #2376, opened 2026-01-13
bugfix: gix-worktree-state: Reduced-size files on checkout should be truncated to correct length

staehle:
gix-worktree-state: Upon checkout of a file, truncate it to the length indicated by the blob

Hi @Byron! This is a second PR for a different bugfix we have in our copy of `gitoxide`. Originally authored here at Roku by @cesfahani . I don't have tests for this one, its just a bugfix we have internally that you should probably know about.

Move the file stat update of the index entry to after the truncate

Fixes issue with checkout on top of existing (larger) file having garbage data appended to the end of file unintentionally.

### DIFF
--- a/gix-worktree-state/tests/state/checkout.rs
+++ b/gix-worktree-state/tests/state/checkout.rs
@@ -710,3 +710,38 @@ fn setup_filter_pipeline(opts: &mut gix_filter::pipeline::Options) {
         required: true,
     }];
 }
+
+#[test]
+fn checkout_truncates_existing_longer_files() -> crate::Result {
+    let mut opts = opts_from_probe();
+    opts.overwrite_existing = false;
+    opts.destination_is_initially_empty = false;
+
+    // Use existing fixture and modify one file to be longer
+    let (_source_tree, destination, _index, _outcome) = checkout_index_in_tmp_dir_opts(
+        opts.clone(),
+        "make_mixed_without_submodules_and_symlinks",
+        None,
+        |_| true,
+        |dest| {
+            // Create a longer version of the "executable" file before checkout
+            let file_path = dest.join("executable");
+            std::fs::create_dir_all(dest)?;
+            std::fs::write(
+                &file_path,
+                b"This is much longer content that should be truncated to match git's version",
+            )?;
+            Ok(())
+        },
+    )?;
+
+    let file_path = destination.path().join("executable");
+    let final_content = std::fs::read(&file_path)?;
+    assert_eq!(
+        final_content[..].as_bstr(),
+        "content",
+        "File content should match git's version"
+    );
+
+    Ok(())
+}


gix-worktree-state/src/checkout/entry.rs
```diff
@@ -286,6 +287,9 @@ pub(crate) fn finalize_entry(
     if set_executable_after_creation {
         set_executable(&file)?;
     }
+    if num_bytes > u32::MAX as usize || entry.stat.size != num_bytes as u32 {
+        file.set_len(num_bytes as u64)?;
+    }
```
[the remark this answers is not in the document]
staehle: Not wrong, but I'm not sure the best way to do this.

gix-worktree-state/src/checkout/entry.rs
```diff
@@ -286,6 +287,9 @@ pub(crate) fn finalize_entry(
     if set_executable_after_creation {
         set_executable(&file)?;
     }
+    if num_bytes > u32::MAX as usize || entry.stat.size != num_bytes as u32 {
```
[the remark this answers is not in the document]
staehle: Okay, decent suggestion

gix-worktree-state/src/checkout/entry.rs
```diff
@@ -286,6 +287,9 @@ pub(crate) fn finalize_entry(
     if set_executable_after_creation {
         set_executable(&file)?;
     }
+    if num_bytes <= u32::MAX as usize {
+        file.set_len(num_bytes as u64)?;
+    }
```
Skgland: Why not

```suggestion
    if let Ok(num_bytes) = num_bytes.try_into() {
        file.set_len(num_bytes)?;
    }
```

so that `num_bytes` in the range (u32::MAX+1)..=min(usize::MAX, u64::MAX) isn't ignored?

Skgland: If the reason is that the rust api takes u64 even though say on some OSes the syscall only takes u32 and as such the call can't be made and an InvalidInput error is returned instead[^1]. Maybe it would still be better to try and just to ignore the error rather than not trying at all.

[^1]: https://doc.rust-lang.org/std/fs/struct.File.html#errors-2

staehle: Thanks! I agree, will edit.

Byron:
Thanks a lot, it's much appreciated.

I just wanted to get this merged to say hi and thanks for the upstreaming :) - I hope to get to the bigger PR in the coming days.
I will see if Copilot can create a test for that.

Byron:
It turns out that whatever this was fixing for you, it was fixed here already in 2024 by opening files with `.truncate(true)`. 

But with a test, and the condition that the truncation only happens with needed, I think this can be left in as additional safeguard. I could imagine that truncation takes time and we could consider to disable that one day, and with this fix it would still work, with a test that proves it does.
The test was generated by Copilot (CLI) by the way while I was doing something else, with some adjustments on top. So that seems to work pretty well.

staehle:
Sweet thanks! I had not checked if this was still an issue when rebasing our commits to the new version, but at least there's a unit test now! Hopefully just as an additional safeguard this code will end up being a no-op anyways :)

PR #2377, opened 2026-01-13
Add GIX_TEST_FIXTURE_HASH to gix-testtools

cruessler:
This PR adds the `sha256` feature to `gix-commitgraph`. I opened it as a draft, so we have a space where we can discuss how to best approach testing this new feature.

This PR was triggered by a comment in #2359.

### DIFF
--- a/justfile
+++ b/justfile
@@ -155,8 +155,8 @@ unit-tests:
     cargo nextest run -p gix-archive --no-default-features --features tar_gz --no-fail-fast
     cargo nextest run -p gix-archive --no-default-features --features zip --no-fail-fast
     cargo nextest run -p gix-status-tests --features gix-features-parallel --no-fail-fast
-    env GIX_TEST_FIXTURE_HASH=sha1 cargo nextest run -p gix-worktree-state-tests --features gix-features-parallel --no-fail-fast
-    env GIX_TEST_FIXTURE_HASH=sha1 cargo nextest run -p gix-worktree-tests --features gix-features-parallel --no-fail-fast
+    cargo nextest run -p gix-worktree-state-tests --features gix-features-parallel --no-fail-fast
+    cargo nextest run -p gix-worktree-tests --features gix-features-parallel --no-fail-fast
     cargo nextest run -p gix-error --no-fail-fast --test auto-chain-error --features auto-chain-error
     cargo nextest run -p gix-error --no-fail-fast
     cargo nextest run -p gix-hash --no-fail-fast
@@ -175,7 +175,7 @@ unit-tests:
     env GIX_TEST_FIXTURE_HASH=sha256 cargo nextest run -p gix-pack --all-features --no-fail-fast
     cargo nextest run -p gix-pack-tests --features all-features --no-fail-fast
     cargo nextest run -p gix-pack-tests --features gix-features-parallel --no-fail-fast
-    env GIX_TEST_FIXTURE_HASH=sha1 cargo nextest run -p gix-index-tests --features gix-features-parallel --no-fail-fast
+    cargo nextest run -p gix-index-tests --features gix-features-parallel --no-fail-fast
     cargo nextest run -p gix-packetline --features blocking-io,maybe-async/is_sync --test blocking-packetline --no-fail-fast
     cargo nextest run -p gix-packetline --features async-io --test async-packetline --no-fail-fast
     cargo nextest run -p gix-transport --features http-client-curl,maybe-async/is_sync --no-fail-fast
--- a/tests/tools/src/lib.rs
+++ b/tests/tools/src/lib.rs
@@ -1,5 +1,14 @@
 //! Utilities for testing `gitoxide` crates, many of which might be useful for testing programs that use `git` in general.
 //!
+//! ## Environment Variables
+//!
+//! ### `GIX_TEST_FIXTURE_HASH`
+//!
+//! Set this variable to control which hash function is used when creating or loading test fixtures.
+//! Valid values are the names of hash functions supported by `gix_hash::Kind` (e.g., `sha1`, `sha256`).
+//! If not set, the default hash function via `gix_hash::Kind::default()` is used.
+//!
+
 //! ## Feature Flags
 #![cfg_attr(
     all(doc, feature = "document-features"),
--- a/gix-commitgraph/tests/commitgraph.rs
+++ b/gix-commitgraph/tests/commitgraph.rs
@@ -81,6 +81,14 @@ pub fn graph_and_expected_named(
     let expected = inspect_refs(&repo_dir, refs);
     let cg =
         Graph::from_info_dir(&repo_dir.join(".git").join("objects").join("info")).expect("graph present and valid");
+    let object_hash = cg.object_hash();
+    let any_ref = expected.values().next().expect("at least one ref");
+    assert_eq!(
+        object_hash,
+        any_ref.id().kind(),
+        "graph hash kind should match fixture object IDs"
+    );
+
     (cg, expected)
 }
 


Byron:
Thanks a lot, I am looking forward to sorting this out!

There is also #2378, hopefully done in the next couple of days (it might conflict).

Byron:
Thanks a lot for starting this! This crate is special as it's the first one that parses a file-format that changes depending on the hash kind.

It seems like one commit currently in the `gix-object` PR would be useful here as well, in case you want to bring it up in its own PR.

cruessler:
Assuming you’re referring to the one adding `Kind::all` (https://github.com/GitoxideLabs/gitoxide/pull/2359/changes/6d950f036366ac3f53743c808be5cb149e4c9460), I’ll extract that addition into its own PR!

Apart from that, my plan is to go through the tests and, for every tests that currently uses `sha1`, add a corresponding one that uses `sha256`. Is that the way to go? Are there other things that would need to be done in this PR?

Byron:
> Assuming you’re referring to the one adding Kind::all (https://github.com/GitoxideLabs/gitoxide/commit/6d950f036366ac3f53743c808be5cb149e4c9460), I’ll extract that addition into its own PR!

Yes, that's the one, thanks 🙏.

> Apart from that, my plan is to go through the tests and, for every tests that currently uses sha1, add a corresponding one that uses sha256. Is that the way to go? Are there other things that would need to be done in this PR?

Sorry for the confusion, that's the way. The only thing tests have to do is to actively set all hashes using feature toggles, and they should just set sha1/sha256 while they are at it. Once a crate has been converted, the tests should automatically test both hashes, and make sure they are compiled in.

cruessler:
I added fixtures for SHA256 hashes to `gix-testtools` and changed `gix-commitgraph` tests to run for all available hash kinds (those depend on the feature flags passed to `gix-commitgraph`). This seems to have worked on my machine, but I’m not sure I got the logic in `gix-testtools` right (any pointers would be much appreciated!). I also added `gix-commitgraph` to `unit-test` in `justfile`, and now I want to see what feedback CI has to give. :-)

tests/tools/src/lib.rs
```diff
@@ -509,8 +558,13 @@ fn scripted_fixture_read_only_with_args_inner(
                 }
                 ArgsInHash::No => "".into(),
             };
+            let potential_hash_suffix = if hash_kind == gix_hash::Kind::Sha1 {
+                "".into()
+            } else {
+                format!("_{}", hash_kind)
```
EliahKagan: I recommend this change, as [suggested](https://github.com/GitoxideLabs/gitoxide/actions/runs/21113354969/job/60715240145?pr=2377#step:5:517) by clippy, which I've [verified will fix](https://github.com/EliahKagan/gitoxide/actions/runs/21115921310/job/60721426119) the `lint` failure:

```suggestion
                format!("_{hash_kind}")
```

What happened was that one of the fixes clippy suggested previously, which was applied in 227d8df7a9f977c80b24c7fd2c6efb3c6444f07b, made it possible to do this further simplification, the absence of which clippy now separately flags.

cruessler: Thanks, that was helpful!

cruessler:
I think I’ve now gotten to a point where I need some feedback in order to be able to continue. (CI is still running, so I also might need to address any issue that comes up there.)

Specifically, I’ve got the following questions:

- The PR runs tests for both hashes through `cargo nextest run -p gix-commitgraph --features sha256 --no-fail-fast`. Is that correct?
- Last time CI ran it did not complain about any of the generated archives in `gix-commitgraph`, but when I delete both `generated-archives` as well as `generated-do-not-edit` in order to regenerate the archives, `git status` shows that most files have changed their size by 512 bytes and that one file has been both deleted and modified. (I’m running `cargo test --features sha256` on Ubuntu 25.10.) What’s the best way to debug this?
- Is there anything else I would need to change or include before I can mark this PR as ready?

```
# output of `gix status` inside `gix-commitgraph/tests/fixtures`
modified:   generated-archives/generation_number_overflow.tar
modified:   generated-archives/generation_number_overflow_sha256.tar
modified:   generated-archives/octopus_merges.tar
modified:   generated-archives/single_commit.tar
modified:   generated-archives/single_commit_huge_dates.tar
modified:   generated-archives/single_parent.tar
deleted:    generated-archives/single_parent_huge_dates.tar
modified:   generated-archives/two_parents.tar
```

tests/tools/src/lib.rs
```diff
@@ -519,14 +571,12 @@ fn scripted_fixture_read_only_with_args_inner(
...
-                    script_identity,
-                    family_name()
-                )),
-                root,
-            );
+            let mut path = Path::new("generated-do-not-edit").join(script_basename);
+            if hash_kind != gix_hash::Kind::Sha1 {
+                path = path.join(hash_kind.to_string());
+            }
+            path = path.join(format!("{}-{}", script_identity, family_name()));
```
Byron: While it would be more cluttered, I seem to prefer the idea of not nesting `hash_kind`, but add it to the format string here.

In future, there will be even more of these as another parameter, the one for the refs backend, is added.

Alternatively… what do you think about looking at both parameters and using them for nesting?

`path.join(hash_kind).join(refs_kind)`

The refs_kind is nothing that exists now, but it would be added at some point.

Yes, that seems like the way to go.

So let's make a mild breaking change and always add the hash_kind to the path via `join`.

tests/tools/src/lib.rs
```diff
@@ -644,6 +695,7 @@ fn configure_command<'a, I: IntoIterator<Item = S>, S: AsRef<OsStr>>(
         .env_remove("GIT_WORK_TREE")
         .env_remove("GIT_COMMON_DIR")
         .env_remove("GIT_ASKPASS")
+        .env_remove("GIT_DEFAULT_HASH")
```
Byron: I think the remove here isn't needed if we are always overriding. If there is something special, then let's keep this and document it though.

gix-commitgraph/tests/access/mod.rs
```diff
@@ -2,112 +2,129 @@ use crate::{check_common, graph_and_expected, graph_and_expected_named};
...
-    let (cg, refs) = graph_and_expected_named("single_commit_huge_dates.sh", "v2", &["HEAD"]);
-    let info = &refs["HEAD"];
-    let actual = cg.commit_by_id(info.id).expect("present");
-    assert_eq!(
-        actual.committer_timestamp(),
-        1,
-        "overflow happened, can't represent huge dates"
-    );
-    assert_eq!(
+    for hash_kind in gix_hash::Kind::all() {
```
Byron: Could `all()` rather return an iterator that hands out owned versions of `hash_kind`? I think that should work with `impl Iterator<Item = Kind>`.

gix-commitgraph/Cargo.toml
```diff
@@ -16,6 +16,8 @@ rust-version = "1.82"
 doctest = false
 
 [features]
+## Support for SHA256 hashes and digests.
+sha256 = ["gix-hash/sha256"]
```
Byron: I think we'd want to use `dev-dependencies` to always add `sha256` to the features when testing. That way, there is only one way to run tests: all of them.

Byron:
As I am looking at Git for how they tackle this I wonder this: Should we, rather than changing every test and looping over it, prefer to build with all hashes, but run tests for each hash selectively, maybe by setting an environment variable that controls which hash is under test?

Performance-wise it should be very similar, even though development is a bit less convenient as one will have to run twice with different flags or environment variables set. There, having the tests run everything automatically, always, and what we have now, seems better.

Also, how do we handle expectations on hashes, particularly those with insta? It basically requires to chose one hash and find/replace the actual output to match that. I.e. `insta::assert_debug_snapshot(hash.norm(actual), @r"expected with sha1")`. And that remapping table, where would that come from? Is it passed in per function, or is it global? It can be global, and maybe that's how Git does it? I never checked, but it would be interesting to know.

So I think it's fine to keep this PR as is as it's happy with this simple and convenient solution, and we can probably layer moer on top of that when a need arises.

cruessler:
I’m glad you mention this other option because I’m still very hesitant to force every test to be rewritten to make it loop over hash kinds as well. What I could try instead is adding an environment variable, e. g. `GIX_TEST_HASH`, that defaults to `sha1` if not set, but can be changed to `sha256`. Any place that wants code to work regardless of hash kind wouldn’t have to bother, but could rely on CI as well as fixture script creation to take care of running with all hash kinds. I’ll try that approach in this PR!

cruessler:
I’ve added a `dev-depencency` on `gix-hash` that activates the feature `sha256`, but I suspect that’s the reason why a couple of tests related to the size of objects now fail. At first sight, it looks like the feature is now enabled for every `cargo nextest` run. Apart from that, I think using an environment variable to set the hash per `cargo nextest` run works quite well.

Byron:
> I’m glad you mention this other option because I’m still very hesitant to force every test to be rewritten to make it loop over hash kinds as well. What I could try instead is adding an environment variable, e. g. `GIX_TEST_HASH`, that defaults to `sha1` if not set, but can be changed to `sha256`. Any place that wants code to work regardless of hash kind wouldn’t have to bother, but could rely on CI as well as fixture script creation to take care of running with all hash kinds. I’ll try that approach in this PR!

That's fair, and probably that's best whenever there is no actual assertions against a hash.
When there is such assertions, one will have to touch these tests to have some sort of mapping, i.e. like with translations the test mentions only one kind of hash, probably SHA1, but it auto-translates to SHA256 when that's run.
Doing this ergonomically *probably* requires globals so one doesn't have to pass the hash kind in everywhere.

The manual looping is always something one *can* implement if that's a good choice in specific cases.

With such a system, we'd need just one more job to run the SHA256 version of the test suite, which just sets an environment variable.

> I’ve added a `dev-depencency` on `gix-hash` that activates the feature `sha256`, but I suspect that’s the reason why a couple of tests related to the size of objects now fail. At first sight, it looks like the feature is now enabled for every `cargo nextest` run. Apart from that, I think using an environment variable to set the hash per `cargo nextest` run works quite well.

Oh, right, the `nextest` builds everything at once so has only one dependency tree for resolving feature flags.

If `[dev-dependencies]` would instead pull in `gix-hash` to enable one or more hashes (probably always all of them), then there would be no need for duplicating the hash related cargo features in the crate that uses `gix-hash`, and instead one would have to tell plumbing users to depend on `gix-hash` just to select hash features. This is what `gix-features` is there for, and maybe the hash selection could be added there as well.

**In any case, it seems we'd have to find a way to get CI green**. Something that I'd not want to lose is to see what the best possible size is, i.e. if just one hash is supported and that is the smallest hash. But then one would have to test the individual crate just with one hash kind which may also be difficult in future. So… maybe just change the size expectations to the new value, while also using `[dev-dependencies]` to enforce building with dual-hash support.

Byron:
Oh, and what happens if the crate is now tested individually?
My feeling is that it needs its own feature toggles so that it knows what to test for… . If that's the case, I think we should add the respective statements to the `justfile`, every crate must be testable in isolation and work just the same.
Or… their tests can be adjusted to enforce all hashes as well, then there should be no difference. Probably my preference for now to keep it as simple as possible.

cruessler:
We could add the following line to `gix-pack/Cargo.toml` which would make the failing test in `gix-pack` pass:

```
gix-hash = { version = "^0.22.0", path = "../gix-hash", features = ["sha256"] }
```

This would duplicate the dependency, though.

Byron:
> We could add the following line to `gix-pack/Cargo.toml` which would make the failing test in `gix-pack` pass:

Yes, such a line would be needed via `[dev-dependencies]`, for all crates that had to be adjusted to compile with nextest.
On top of that, to play it safe, I think the `justfile` entry to test them individually would still be needed.

Lastly, it looks like there will never be a need to add `sha1/sha256` feature toggle forwarding to crates that depend on `gix-hash`, so plumbing stays clean and expects its direct consumers to know they have to configure `gix-hash` on application level.

`gix` (the crate) would offer such a forwarding though with reasonable defaults. That seems quite minimal then.

justfile
```diff
@@ -162,14 +163,17 @@ unit-tests:
...
+    env GIX_TEST_HASH=sha256 cargo nextest run -p gix-commitgraph --no-fail-fast
     cargo nextest run -p gix-object --no-fail-fast
     cargo nextest run -p gix-object --features verbose-object-parsing-errors --no-fail-fast
     cargo nextest run -p gix-tempfile --features signals --no-fail-fast
     cargo nextest run -p gix-features --all-features --no-fail-fast
     cargo nextest run -p gix-ref-tests --all-features --no-fail-fast
     cargo nextest run -p gix-odb --all-features --no-fail-fast
     cargo nextest run -p gix-odb-tests --features gix-features-parallel --no-fail-fast
-    cargo nextest run -p gix-pack --all-features --no-fail-fast
+    env GIX_TEST_HASH=sha1 nextest run -p gix-pack --all-features --no-fail-fast
```
EliahKagan:
```suggestion
    env GIX_TEST_HASH=sha1 cargo nextest run -p gix-pack --all-features --no-fail-fast
```

cruessler: Thanks a lot!

gix-commitgraph/Cargo.toml
cruessler: That means removing the feature here, is that correct?

cruessler:
CI is now green! Quick summary of this PR’s changes:

- It adds `GIX_TEST_HASH` which affects the default hash used in fixture scripts via `GIT_DEFAULT_HASH`. (We might want to consider renaming to `GIX_TEST_FIXTURE_HASH` or something that makes it clearer this is related to fixtures.)
- It adapts size-of-various-objects-related expectations in tests so that the sizes always take both hash kinds into account.
- For that to work consistently, it adds `gix-hash = { …, features = ["sha256"] }` to `dev-dependencies` in several crates.

gix-commitgraph/Cargo.toml
Byron: Yes, let's start with that.

This means we say that plumbing will never set hash features, and it expects to be able to see the hash to be used at runtime via `gix-hash::Kind`. Most implementations do that already, and I think this should work for *most* cases.

gix-diff/src/tree/visit.rs
```diff
@@ -151,9 +151,10 @@ mod tests {
     #[test]
     fn size_of_change() {
         let actual = std::mem::size_of::<Change>();
+        let ceiling = 72;
         assert!(
-            actual <= 48,
-            "{actual} <= 48: this type shouldn't grow without us knowing"
```
Byron: Could we record the old number? Maybe by doing:

```Rust
let sha1 = 48;
let sha256_extra = 24
actual <= sha1 + sha256_extra
```

This will just help document how much memory can be saved by not compiling for SHA256.

gix-pack/tests/Cargo.toml
```diff
@@ -6,7 +6,7 @@ version = "0.0.0"
 repository = "https://github.com/GitoxideLabs/gitoxide"
 authors = ["Sebastian Thiel <sebastian.thiel@icloud.com>"]
 license = "MIT OR Apache-2.0"
-description = "Please use `gix-<thiscrate>` instead ('git' -> 'gix')"
+description = "Tests for the gix-pack crate"
```
Byron, line 9: Great catch, thanks!

tests/tools/src/lib.rs
Byron: Did you see this comment?

justfile
```diff
@@ -106,6 +106,7 @@ check:
     cargo check -p gix-features --features zlib
     cargo check -p gix-features --features cache-efficiency-debug
     cargo check -p gix-commitgraph --all-features
+    cargo check -p gix-commitgraph --features sha256
```
Byron: Never mind, I think this is related to the comment at the top. What follows is what I wrote before.

----
Now I wonder: does `gix-commitgraph` still need its own feature? Or could it be told by `gix_hash::Kind` what it should expect?
My hope is that this can all be done with runtime values, and not forwarding `sha1/sha256` features should make that clearer as well.

gix-commitgraph/Cargo.toml
Byron, line 1: What surprises me is that there is no way to check what hash format the graph is using. Could we have an implementation of `object_hash` (which exists on `File`) for the `Graph`, and assert on the value.
But then we'd need a way to get the currently set hash, which had to come from `gix-testtools`.
I think this would be very useful to have and some assertion should definitely be made.

tests/tools/src/lib.rs
cruessler: My bad, sorry! I had seen the comment, but wasn’t 100 % sure what you had in mind at first, and then I forgot to address it. :smile: I think making the hash kind a part of the path is a good idea, though.

gix-commitgraph/Cargo.toml
cruessler: Makes sense, removed.

gix-commitgraph/Cargo.toml
cruessler: Are you thinking about something like the following:

```rust
/// The kind of hash used in this `Graph`.
///
/// Note that it is always conforming to the hash used in the owning repository.
///
/// # Panics
///
/// If the graph does not contain any `File`.
pub fn object_hash(&self) -> gix_hash::Kind {
    self.files
        .first()
        .map(super::File::object_hash)
        .expect("graph to have at least one file")
}
```

gix-diff/src/tree/visit.rs
cruessler: I like the idea and have changed all occurrences to explicitly contain that piece of information.

cruessler:
@Byron I’ve merged `main` and CI is green, so you can have another look if you want! Feel free to convert back to draft if you think it’s not ready yet!

gix-commitgraph/Cargo.toml
Byron: Are you interested in a follow-up? There is the `nonempty` crate which I think would do wonders here (and in other places).

But besides that, yes, this is what I imagined.

Byron:
Thanks a lot! Yes, I think it's ready. I gave it another quick look and had a comment about `object_hash()`, ideally we can have it and use it in a test or two.
Once this is addressed, and it's non-conflicting, I will bring this PR to merge.

justfile
```diff
@@ -155,24 +155,27 @@ unit-tests:
     cargo nextest run -p gix-archive --no-default-features --features tar_gz --no-fail-fast
     cargo nextest run -p gix-archive --no-default-features --features zip --no-fail-fast
     cargo nextest run -p gix-status-tests --features gix-features-parallel --no-fail-fast
-    cargo nextest run -p gix-worktree-state-tests --features gix-features-parallel --no-fail-fast
-    cargo nextest run -p gix-worktree-tests --features gix-features-parallel --no-fail-fast
+    env GIX_TEST_HASH=sha1 cargo nextest run -p gix-worktree-state-tests --features gix-features-parallel --no-fail-fast
+    env GIX_TEST_HASH=sha1 cargo nextest run -p gix-worktree-tests --features gix-features-parallel --no-fail-fast
```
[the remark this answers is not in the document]
Byron: I agree.

tests/tools/src/lib.rs
```diff
@@ -470,14 +470,26 @@ fn scripted_fixture_read_only_with_args_inner(
         gix_tempfile::signal::handler::Mode::DeleteTempfilesOnTerminationAndRestoreDefaultBehaviour,
     );
 
+    let gix_test_hash = env::var_os("GIX_TEST_FIXTURE_HASH")
+        .and_then(|os_string| os_string.into_string().ok())
+        .unwrap_or_default();
+    let hash_kind = gix_hash::Kind::from_str(&gix_test_hash).unwrap_or_default();
+
```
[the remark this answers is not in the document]
Byron: Yes, we should fail loudly.

gix-commitgraph/Cargo.toml
```diff
@@ -33,6 +33,7 @@ document-features = { version = "0.2.0", optional = true }
 [dev-dependencies]
 gix-testtools = { path = "../tests/tools" }
 gix-date = { path = "../gix-date" }
+gix-hash = { path = "../gix-hash", features = ["sha256"] }
```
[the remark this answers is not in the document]
Byron: Yeah, right. But I think it's fine. We also said that consumers of plumbing are expected to pull in `gix-hash` to configure the hash - that inconvenience is acceptable.

gix-commitgraph/Cargo.toml
cruessler: Absolutely! I’ll have a look at `nonempty`. I assume we want to wrap `self.files`? Or are you thinking about a more local use just in `object_hash`?

cruessler: I’ve now added `Graph::object_hash`.

tests/tools/src/lib.rs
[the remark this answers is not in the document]
cruessler: I like that! Changed. (You might want to tune the specific behaviour.)

justfile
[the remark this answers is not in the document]
cruessler: My oversight, fixed.

justfile
```diff
@@ -155,24 +155,27 @@ unit-tests:
...
     cargo nextest run -p gix-ref-tests --all-features --no-fail-fast
     cargo nextest run -p gix-odb --all-features --no-fail-fast
     cargo nextest run -p gix-odb-tests --features gix-features-parallel --no-fail-fast
-    cargo nextest run -p gix-pack --all-features --no-fail-fast
+    env GIX_TEST_FIXTURE_HASH=sha1 cargo nextest run -p gix-pack --all-features --no-fail-fast
+    env GIX_TEST_FIXTURE_HASH=sha256 cargo nextest run -p gix-pack --all-features --no-fail-fast
     cargo nextest run -p gix-pack-tests --features all-features --no-fail-fast
     cargo nextest run -p gix-pack-tests --features gix-features-parallel --no-fail-fast
-    cargo nextest run -p gix-index-tests --features gix-features-parallel --no-fail-fast
+    env GIX_TEST_HASH=sha1 cargo nextest run -p gix-index-tests --features gix-features-parallel --no-fail-fast
```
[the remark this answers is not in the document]
cruessler: My oversight, fixed.

cruessler:
> Thanks a lot! Yes, I think it's ready. I gave it another quick look and had a comment about `object_hash()`, ideally we can have it and use it in a test or two. Once this is addressed, and it's non-conflicting, I will bring this PR to merge.

Awesome! I’ve tried to addressed the outstanding comments. I’ve also merged `main`. You probably want to double-check the result as I had to merge my changes with the ones coming from #2412. In particular, you might want to change the order of arguments or `hash_kind` as an argument to some functions.

Byron:
Uff, I think I should add to the contribution guidelines to never merge main, but always rebase. While the diff is still what it's supposed to be, I feel 'unsafe' when it's unclear which commits are actually added here.

Would you fix that for me please?

cruessler:
I’ve now rebased onto `upstream/main`. I’ll address any CI failures as they come in (which is why I’ve marked the PR as draft in the meantime). Unfortunately I had to already squash all commits into one as `git` only picked up the resolutions for `tests/tools/src/lib.rs` that were recorded during my first merge that way (I did want to manually resolve the same conflicts again). I hope this works for you and nothing got lost! This process also removed quite a few commits that had become redundant in the meantime, so it might even have improved the PR.

Byron (commented):
Thanks a lot, it's so much better!

If you don't mind, in the very sense of the words, could you split the commits up so only `gix-commitgraph` gets the `feat: …` ? Everything could probably be `Adapt to changes…` then.

Otherwise, and I'd understand, please feel free to bring this PR out of draft mode and I will edit some commits.

Thanks again (I really want to merge this one, has been going on for too long 😅, as becomes obvious by the website sluggishness).

cruessler:
I’ve now split the PRs focusing on the addition of `GIX_TEST_FIXTURE_HASH` as most of the other changes were in response to that change. Feel free, though, to split along other lines if you think there’s a better way!

justfile
```diff
@@ -162,14 +162,17 @@ unit-tests:
     cargo nextest run -p gix-hash --no-fail-fast
     cargo nextest run -p gix-hash --features sha256 --no-fail-fast
     cargo nextest run -p gix-hash --no-default-features --features sha256 --no-fail-fast # TODO: make this actually work by removing 'sha1' from default features.
+    env GIX_TEST_FIXTURE_HASH=sha1 cargo nextest run -p gix-commitgraph --no-fail-fast
+    env GIX_TEST_FIXTURE_HASH=sha256 cargo nextest run -p gix-commitgraph --no-fail-fast
```
Byron, line 166: @cruessler I have removed the other usages of `gix-commitgraph` as they just seemed to set the hash to SHA1 which doesn't seem necessary or effective anymore.

cruessler:
> Thanks a lot! Overall this took quite a while to review and I have changed `gix-testtools` quite significantly to make it non-breaking.

Thanks a lot for the review as well as for the patience! Out of curiosity: was there anything I could have done that would have made reviewing or working with this PR in general easier?

> It would probably be advisable to build up a guide that helps AI to formalise this conversion (similar to what's happening in #2423).

I’ll have a look! (Or are you already on it?) Are there any examples I could follow?

Byron:
> Thanks a lot for the review as well as for the patience! Out of curiosity: was there anything I could have done that would have made reviewing or working with this PR in general easier?

Thanks for asking. Unfortunately I have no answer except for a learning for myself: If it's something 'fundamental', I should just dabble in it myself and get the basics going, and of course, also get an understanding myself first. This PR had a lot of rounds, and I think there are diminishing returns if it's not merged fast enough.

> I’ll have a look! (Or are you already on it?) Are there any examples I could follow?

I am not on it, but have an example in `gix-error/src/lib.rs`. That detailed guide was AI written, based on actual mistakes from the real world. I helped it write its own guide to hopefully do better in future. Maybe pick a next crate from the [tasks section](https://github.com/GitoxideLabs/gitoxide/issues/281) and see how the AI thing goes (if you want to use that at all). It certainly is a double-edged sword, as it's so great at making it appear like it's all sorted, even though only half of the work is done, and none of it properly 😅 (and I am saying that from my `gix-error` conversion experience which takes a ton of time to review).

PR #2383, opened 2026-01-18
`anyhow` integration for `gix-error`

Byron:
Make it possible to follow `gix-error` types with `source()` chains, and
provide a conversion to anyhow (beyind a feature toggle).

### Tasks

* [x] anyhow conversion from `Exn`
* [x] source-chain for `Exn`
    - [x] make it possible to keep the location.
* [x] source-chain for `Error`
* [x] A way to turn on auto-source-chain for `Error` so `gix` errors will be compatible automatically

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -429,7 +429,7 @@ jobs:
       - name: Test (nextest)
         env:
           GIX_TEST_IGNORE_ARCHIVES: '1'
-        run: cargo nextest run --workspace --no-fail-fast
+        run: cargo nextest run --workspace --no-fail-fast --exclude gix-error
 
   test-32bit-windows-size-doc:
     runs-on: windows-latest
--- a/gix-error/Cargo.toml
+++ b/gix-error/Cargo.toml
@@ -18,7 +18,7 @@ rust-version = "1.82"
 ## [`into_box()`](crate::Exn::into_box()) or [`into_inner()`](crate::Exn::into_inner()).
 anyhow = ["dep:anyhow"]
 ## The [`Error`](crate::Error) type is always flattening the [`Exn`](crate::Exn) error tree
-## into a chain of errors, while keeping there locations and runtime type-information.
+## into a chain of errors, while keeping their locations and runtime type-information.
 auto-chain-error = []
 ## The opposite of `auto-chain-error` and implicitly enabled by default. Use it to override `auto-chain-error`.
 tree-error = []
--- a/gix-error/src/lib.rs
+++ b/gix-error/src/lib.rs
@@ -79,6 +79,9 @@ pub use exn::{ErrorExt, Exn, Frame, OptionExt, ResultExt, Something, Untyped};
 ///
 /// If it's enabled, this type is merely a wrapper around [`ChainedError`]. This happens automatically
 /// so applications that require this don't have to go through an extra conversion.
+///
+/// When both the `tree-error` and `auto-chain-error` features are enabled, the `tree-error`
+/// behavior takes precedence and this type uses the tree-based representation.
 pub struct Error {
     #[cfg(any(feature = "tree-error", not(feature = "auto-chain-error")))]
     inner: error::Inner,
--- a/gix-error/tests/error/exn.rs
+++ b/gix-error/tests/error/exn.rs
@@ -630,7 +630,7 @@ fn into_chain() {
     ]
     "#);
 
-    // But these can alos be turned off
+    // But these can also be turned off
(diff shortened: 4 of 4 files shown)


EliahKagan:
The `test-32bit` [test failures](https://github.com/GitoxideLabs/gitoxide/actions/runs/21118673176/job/60728060192?pr=2383#step:10:1311) (on arm32v7) were:

```text
        FAIL [   0.003s] (1196/2983) gix-error::auto-chain-error from_exn_error
  stdout ───

    running 1 test
    test from_exn_error ... FAILED

    failures:

    failures:
        from_exn_error

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2 filtered out; finished in 0.00s
    
  stderr ───

    thread 'from_exn_error' (19568) panicked at gix-error/tests/auto_chain_error.rs:7:5:
      left: "Message(\"one\")"
     right: "one"

        FAIL [   0.003s] (1197/2983) gix-error::auto-chain-error from_exn_error_tree
  stdout ───

    running 1 test
    test from_exn_error_tree ... FAILED

    failures:

    failures:
        from_exn_error_tree

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2 filtered out; finished in 0.00s
    
  stderr ───

    thread 'from_exn_error_tree' (19570) panicked at gix-error/tests/auto_chain_error.rs:20:5:
      left: "Message(\"topmost\")\n|\n└─ Message(\"E6\")\n    |\n    └─ Message(\"E5\")\n    |   |\n    |   └─ Message(\"E3\")\n    |   |   |\n    |   |   └─ Message(\"E1\")\n    |   |\n    |   └─ Message(\"E10\")\n    |   |   |\n    |   |   └─ Message(\"E9\")\n    |   |\n    |   └─ Message(\"E12\")\n    |       |\n    |       └─ Message(\"E11\")\n    |\n    └─ Message(\"E4\")\n    |   |\n    |   └─ Message(\"E2\")\n    |\n    └─ Message(\"E8\")\n        |\n        └─ Message(\"E7\")"
     right: "topmost"
```

Given that no tests fail in the `test-fast` jobs that have completed so far, I don't know why this would happen. The error seems somehow specific to 32-bit targets, but I don't know why that would be.

Byron:
I don't think it is specific to 32bits, I hope it will resolve itself. In any case, I will see tomorrow 😅.

EliahKagan:
Sounds good! It seemed weird enough that I figured I'd mention it. (Also, sorry about not saying what commit, fd8c038b7f2998ccfe71f38015ff28dc1a5453b9, those CI results were from--a new commit was pushed around the same time as I posted that comment.)

But actually it wasn't as weird as I thought:

- In `ci.yml`, it looks like some relevant tests were excluded here in the `test-fast` job. That might be sufficient to explain the combination of passing and failing jobs. *If not...*

- The failures seem like the usually happened when `GIX_TEST_IGNORE_ARCHIVES=1` is set: the `text-fixtures-windows` job also failed, in [the same way](https://github.com/GitoxideLabs/gitoxide/actions/runs/21118673176/job/60728060191#step:6:1374), it just hadn't finished yet when I commented. This is set in `test-32bit`, too, but not in `test-fast`. This is to say that, *other* than the "full" `test` job, they always happened in the jobs that rerun all fixture scripts. I mention this because, if that turns out to be relevant, then it suggest there may be less than the intended amout of feature coverage--for example, some missing `nextest` commands--in the `justfile` recipes used by the `test` job.

In any case, more pass since the subsequent commit a435a29c7655ab0a820ce2f1e696d9eecb2320b0.

*Edit:* Those two tests [still fail](https://github.com/GitoxideLabs/gitoxide/actions/runs/21118904164/job/60728604288?pr=2383#step:6:1373) in `test-fixtures-windows` though.

Byron:
Thanks @EliahKagan. I have to confess that I did something quite messy: there is `gix-error` with a positive and a negative feature toggle. The positive one turns on a feature which disturbs `gix-error`'s own tests which expect it to be off. `cargo nextest` doesn't handle feature resolution correctly by building everything at once seemingly, so the normal nextest run now needs `gix-error` barred from running, whenever it is called.

Maybe this knowledge should be put into the `justfile`.

The origin of this issue is `cargo test -p gix` which must disable default features in `gix`, but instead now has to use that negative feature toggle to do it.

Maybe a better way to fix this is to change the `gix-error` tests to expect either this or that depending on how they are run, and, I think that would be a good follow-up. Keeping note of that.

aliali05340-tech:
hi

PR #2384, opened 2026-01-18
refactor!: catch up Exn designs with the upstream

tisonkun:
This follows https://github.com/GitoxideLabs/gitoxide/pull/2352#discussion_r2702101315.

I push several changes in separate commits. If you'd like to handle them in multiple PRs, please let me know.

### DIFF
--- a/gix-error/src/error.rs
+++ b/gix-error/src/error.rs
@@ -61,11 +61,7 @@ impl std::error::Error for Error {
     /// Return the first source of an [Exn] error, or the source of a boxed error.
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
         match &self.inner {
-            Inner::ExnAsError(frame) | Inner::Exn(frame) => {
-                frame.children().first().map(|f| {
-                    f.error() as &(dyn std::error::Error + 'static)
-                })
-            }
+            Inner::ExnAsError(frame) | Inner::Exn(frame) => frame.children().first().map(|f| f.error() as _),
         }
     }
 }


gix-error/src/exn/impls.rs
```diff
@@ -94,6 +90,14 @@ impl<E: Error + Send + Sync + 'static> Exn<E> {
         new_exn
     }
 
+    /// Use the current exception the head of a chain, adding `err` as its children.
```
Byron:
```suggestion
    /// Use the current exception as the head of a chain, adding `err` to its children.
```

Byron (commented):
Thanks a lot, that's much appreciated!

All looks good, but it looks like `deref` is required for convenient access to the error, but I am open to alternative solutions that don't add bloat/inconvenience to the site of creation.

tisonkun:
> it looks like deref is required for convenient access to the error

This is correct. I revert the change to keep the current solution. I'll consider whether to add this `impl Deref` to the `exn` crate.

gix-error/src/exn/impls.rs
```diff
@@ -94,6 +91,14 @@ impl<E: Error + Send + Sync + 'static> Exn<E> {
...
+    #[track_caller]
+    pub fn chain<T: Error + Send + Sync + 'static>(mut self, err: impl Into<Exn<T>>) -> Exn<E> {
+        let err = err.into();
+        self.frame.children.push(*err.frame);
+        self
+    }
+
     /// Use the current exception the head of a chain, adding `errors` to its children.
     #[track_caller]
     pub fn chain_iter<T, I>(mut self, errors: I) -> Exn<E>
```
tisonkun: Perhaps change this to `chain_all` as well.

P.S. I'll start a PR on `exn` to add these two `chain` methods. In the initial use case of `exn` in ScopeDB, we can always know all the children when constructing the chain/tree, so we doubt whether to make `children` mutable after the chain has been established. But reading `gix`'s use cases, it seems `chain`s that adding children afterwards can be helpful.

Byron: > Perhaps change this to chain_all as well.

That sounds great, please feel free to do this here, otherwise I will do it in a follow-up.

gix-error/src/exn/impls.rs
tisonkun, line 1: Now I know gix-error has to wrap once more the inner error into `Untyped` because `downcast` must have the type param matched. In this case, it seems to be inevitable.

Byron: In theory, the `into_inner()` and `into_box()` can be removed as there is always `gix_error::Error` to hold these types.

Let's leave it as is for now, and I will consider removing them to avoid that double-box.

gix-error/src/exn/impls.rs
tisonkun: Added.

tisonkun:
All tests passed. Welcome to drop another review.

Byron:
Oh, may I ask to squash this PR into one commit?
That commit would have to be prefixed with `refactor!: ` as it breaks the API. That doesn't matter yet as the crate isn't released, but we may as well try to stay consistent.

tisonkun:
> Oh, may I ask to squash this PR into one commit?

Sure. This is what I typically expected - Since we rely on GitHub, we use PR to track the dev history and squash commits for an easy-reading main linear commit history.

> commit would have to be prefixed with `refactor!: `

Agree. I'd change the PR title now.

Byron:
Sorry, I meant the commits. I can always squash via GH or rewrite history, but I'd put words into your mouth and destroy your signatures.

If you don't mind, I can do that of course.

tisonkun:
> Sorry, I meant the commits. I can always squash via GH or rewrite history, but I'd put words into your mouth and destroy your signatures.
> 
> If you don't mind, I can do that of course.

Never mind. Please do that.

PR #2385, opened 2026-01-18
refactor!: ErrorExt::raise_iter to raise_all + remove Frame::downcast

tisonkun:
This should be included in #2384, but I failed to find it at that time :\

### DIFF
--- a/gix-error/src/error.rs
+++ b/gix-error/src/error.rs
@@ -6,13 +6,13 @@ impl Error {
     /// Return the error that is most likely the root cause, based on heuristics.
     /// Note that if there is nothing but this error, i.e. no source or children, this error is returned.
     pub fn probable_cause(&self) -> &(dyn std::error::Error + 'static) {
-        let root = self.inner.as_frame();
+        let root = self.inner.frame();
         root.probable_cause().unwrap_or(root).error()
     }
 
     /// Return an iterator over all errors in the tree in breadth-first order, starting with this one.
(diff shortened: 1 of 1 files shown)


gix-error/src/exn/impls.rs
```diff
@@ -307,11 +307,6 @@ impl Frame {
         &*self.error
     }
 
-    /// Try to downcast this error into the exact type T, or return `None`
-    pub fn downcast<T: Error + 'static>(&self) -> Option<&T> {
-        self.error.downcast_ref()
-    }
```
tisonkun, line 313: The doc comment should be "Try to downcast this error **as a ref to** the exact type T, or return `None`"

But since it can be written as a fluent call `f.error().downcast_ref::<T>()`, I tend not to add this method to keep the API less.

Byron:
Thanks you, this looks great!

Happy to shed this extra method as well.

Byron:
@tisonkun Please note that I think it's best for me to see the conversion through to the end, and by that time I'd expect `gix-error::exn` to be stable enough to look at migrating back to `exn` again.

tisonkun:
> @tisonkun Please note that I think it's best for me to see the conversion through to the end, and by that time I'd expect `gix-error::exn` to be stable enough to look at migrating back to `exn` again.

Ack. Try to fill the gap in https://github.com/GitoxideLabs/gitoxide/issues/2351#issuecomment-3765986744.

PR #2391, opened 2026-01-20
fix: correct typo in error message for gix-fs directory creation

lunbun:
Corrects typo `Intermediae` -> `Intermediate`


Byron:
Good catch, thanks!

PR #2395, opened 2026-01-22
improvements

Byron:
Don't fail when untracked files can't be read.

### Tasks

* [x] reproduce issue with test
* [x] fix it
* [x] adapt to Rust 1.93

### DIFF
--- a/gix-command/tests/command.rs
+++ b/gix-command/tests/command.rs
@@ -362,7 +362,7 @@ mod prepare {
         );
         assert_eq!(
             format!("{cmd:?}"),
-            quoted(&[*SH, "-c", r#"\'ls\' \"$@\""#, "--", "--foo=a b"]),
+            quoted(&[*SH, "-c", r#"'ls' \"$@\""#, "--", "--foo=a b"]),
             "looks strange thanks to debug printing, but is the right amount of quotes actually"
         );
     }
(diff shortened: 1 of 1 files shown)


gix-status/tests/fixtures/status_many.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+ei/usr/bin/env bash
```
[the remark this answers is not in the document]
Byron: Silly mistake, somehow this was 'vimmed' into it, and I even copied it to the new fixture.

gix-command/tests/command.rs
```diff
@@ -380,9 +380,9 @@ mod prepare {
             quoted(&[
                 *SH,
                 "-c",
-                r#"\'C:\\Users\\O\'\\\'\'Shaughnessy\\with space.exe\' \"$@\""#,
+                r#"'C:\\Users\\O'\\''Shaughnessy\\with space.exe' \"$@\""#,
                 "--",
-                r"--foo=\'a b\'"
+                r"--foo='a b'"
```
EliahKagan, line 385: I'm glad the `Debug` trait for `OsString` no longer prints `'` characters as `\'`--it makes things a lot easier to read! Thanks for updating this to adapt to that!

I wonder if we should condition this on the Rust version for a while so the tests can run and pass on previous Rust versions. I might look into doing that--whether it makes sense might depend, in part, on how much else there is in the test suite that relies on very recent Rust.

(It looks like the adapted-to change in the standard library was done in https://github.com/rust-lang/rust/pull/148798 for https://github.com/rust-lang/rust/issues/114583. If I divide this into cases, I can mention that somewhere so it's clear why.)

Byron: Right now the test-suite is susceptible to CI image changes to Rust as well as Git (to a lesser extend), and I think it's fine for now: the failure rate due to that is quite low, with Clippy the number one offender here, but somehow I am not bothered by it.

I am just saying that in case becoming more resilient against it would mean slower CI, `rust-toolchain.toml` or more complexity for some tests.

PR #2399, opened 2026-01-25
Update zlib-rs to 0.6.0

musicinmybrain:
Release notes: https://github.com/trifectatechfoundation/zlib-rs/releases/tag/v0.6.0

Source diff: https://github.com/trifectatechfoundation/zlib-rs/compare/v0.5.5...v0.6.0

Particularly, upstream writes about `zlib-rs`, “The internals are no longer publicly exposed. If you were using some unstable symbol before, and it is no longer exported, but you do believe it would be reasonable for us to provide it, let us know.” I therefore needed to make a couple of source-code changes to replace constants from `zlib_rs::c_api`. I tried to keep the behavior and semantics the same.

----

Currently, the [`window_bits` is `MAX_WBITS` by default](https://docs.rs/zlib-rs/latest/src/zlib_rs/deflate.rs.html#235) for all built-in compression configurations, so we could reasonably change

```rust
        // Maximum value from the documentation for zlib_rs::Deflate::new.
        let max_wbits: u8 = 15;
        let inner = zlib_rs::Deflate::new(zlib_rs::DeflateConfig::best_speed().level, true, max_wbits);
```

to

```rust
        let config = zlib_rs::DeflateConfig::best_speed();
        let inner = zlib_rs::Deflate::new(config.level, true, config.window_bits as u8);
```

which would currently be equivalent in practice, and is probably better semantically.

----

For inflation, there is probably value in explicitly asking for the documented largest possible window size, even if that happens to be the same as the default in practice.

----

My interest in this is because I want to update `rust-zlib-rs` / `rust-libz-rs-sys` in Fedora without introducing compat packages.

----

Tested with `cargo test -F zlib` in `git-features/`.

### DIFF
--- a/gix-features/src/zlib/mod.rs
+++ b/gix-features/src/zlib/mod.rs
@@ -22,9 +22,9 @@ impl Decompress {
 
     /// Create a new instance. Note that it allocates in various ways and thus should be re-used.
     pub fn new() -> Self {
-        // Maximum value from the documentation for zlib_rs::Inflate::new.
-        let max_wbits: u8 = 15;
-        let inner = zlib_rs::Inflate::new(true, max_wbits);
+        let config = zlib_rs::InflateConfig::default();
+        let header = true;
+        let inner = zlib_rs::Inflate::new(header, config.window_bits as u8);
         Self(inner)
     }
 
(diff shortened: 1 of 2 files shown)


Byron:
Thanks a lot for the update and for the suggestion to use the configuration structs instead - I added this to the PR and will merge it now.

Issue #2388, opened 2026-01-19
Option to use `ring` as `rustls` provider


fenhl:
### Summary 💡

Currently, the `blocking-http-transport-reqwest-rust-tls-trust-dns` feature uses [`aws-lc-rs`](https://docs.rs/aws-lc-rs) as the cryptography provider for [`rustls`](https://docs.rs/rustls). I would like to request the ability (either as the default or as an alternative crate feature) to use [`ring`](https://docs.rs/ring) instead.

### Motivation 🔦

`aws-lc-rs` has nontrivial C dependencies and as a result it has a history of builds breaking in various configurations, e.g. it currently [fails to compile for Android](https://github.com/aws/aws-lc-rs/issues/1006). While `ring` is not technically pure Rust as it extensively uses assembly code, it does not have these build problems in my experience.

stormshield-gt:
Maybe a feature that doesn't set any provider would be more future proof ?
That way, any provider can be chosen. For instance, https://github.com/ctz/graviola/ which is also really easy and fast to compile

Some other libraries have gone that way, for instance `bollards` and `testcontainers`. The next release of `rustls` (https://github.com/rustls/rustls/issues/2400) has also removed the built-in provider.

Byron:
Thanks for sharing.

While I think that the most common features of `reqwest` should be forwarded via the `gix` crate, it should also always be possible to not select any feature, and instead depend on `reqwest` yourself and set it up as needed.

Is there anything that you'd like to do that doesn't work?

EliahKagan:
Related: #2338

fenhl:
> Is there anything that you'd like to do that doesn't work?

It wasn't obvious from the documentation, but the `blocking-http-transport-reqwest` feature doesn't configure any TLS backend, and therefore doesn't pull in `aws-lc-rs`. So using that feature and configuring `reqwest` and `rustls` separately works. I could send a small docs PR to clarify this if you'd like.

Byron:
Yes, please do!

Does that mean you brought `ring` back that way? Somehow I thought it disappeared from `reqwest` completely, but if not I'd also love to have the pure Rust build back and use `ring` there instead.

And that docs PR would definitely be welcome, go for it :).

fenhl:
> Does that mean you brought `ring` back that way? Somehow I thought it disappeared from `reqwest` completely, but if not I'd also love to have the pure Rust build back and use `ring` there instead.

[...]

PR #2404, opened 2026-01-29
Clarify that `blocking-http-transport-reqwest` doesn't enable `reqwest`'s default features

fenhl:
As suggested in #2388.


Byron:
Thanks a lot!

PR #2405, opened 2026-01-31
Add benchmark for counting additions/removals in diff

cruessler:
This PR adds a benchmark for comparing `imara-diff` 0.1 vs. 0.2 with respect to the performance of counting additions and removals in diffs. I hope this is the benchmark you had in mind in this comment: https://github.com/GitoxideLabs/gitoxide/pull/2279#issuecomment-3797796643. :-)

The benchmark itself sets up a simple `TokenSource` that yields unique lines that each contain a single number. The `TokenSource` can be configured to skip lines at a regular interval in order to trivially simulate files with additions and removals. On my machine, `imara-diff` 0.2 seems to be slightly faster than 0.1.

```
     Running benches/line_count.rs (/home/christoph/worktrees/gitoxide/branch-3/target/release/deps/line_count-c71fca099d57846a)
imara-diff 0.1          time:   [5.1431 ms 5.1484 ms 5.1537 ms]
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

imara-diff 0.2          time:   [4.9442 ms 4.9502 ms 4.9567 ms]
Found 6 outliers among 100 measurements (6.00%)
  6 (6.00%) high mild
```

### DIFF
--- a/gix-diff/benches/line_count.rs
+++ b/gix-diff/benches/line_count.rs
@@ -1,5 +1,4 @@
 use criterion::{criterion_group, criterion_main, Criterion};
-use imara_diff::intern::{InternedInput, TokenSource};
 
 struct BenchmarkTokenSource {
     number_of_lines: u32,
@@ -51,7 +50,7 @@ impl Iterator for BenchmarkTokenizer {
     }
 }
 
-impl TokenSource for BenchmarkTokenSource {
+impl imara_diff::intern::TokenSource for BenchmarkTokenSource {
     type Token = String;
 
     type Tokenizer = BenchmarkTokenizer;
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot! I think this is quite a cool way to test them, with an artificial token source :).

To me (on ARM64) the picture is quite different and v2 seems ~10% slower.

```
> cargo bench -p gix-diff --bench line-count
imara-diff 0.1          time:   [3.3331 ms 3.3506 ms 3.3688 ms]
                        change: [−4.9312% −4.2166% −3.4898%] (p = 0.00 < 0.05)
                        Performance has improved.

imara-diff 0.2          time:   [3.6174 ms 3.6317 ms 3.6470 ms]
                        change: [+1.0939% +1.5115% +2.0251%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

This is all I'd need to know that I want to keep v1 around for line counting at least, as it really seems faster there. For everything else, i.e. actual diffs, I think v2 should be used so we have one implementation to tune and eventually match Git perfectly :).

cruessler:
> This is all I'd need to know that I want to keep v1 around for line counting at least, as it really seems faster there. For everything else, i.e. actual diffs, I think v2 should be used so we have one implementation to tune and eventually match Git perfectly :).

Sounds good! I’ll put the migration on my todo list, then! :-)

PR #2411, opened 2026-02-03
chore: update `bytes` to a fixed version

Byron:
Fix CI


EliahKagan (commented):
Sorry, I just merged #2410, which I think made this redundant.

I think they both came in--that one, and then this--but that's actually fine.

PR #2426, opened 2026-02-08
Remove unsafe UTF-8 conversions and add named relative date parsing

apollocatlin:
- Replace unsafe `from_utf8_unchecked` with safe `from_utf8().expect()` in TimeBuf::as_str() and Time::fmt() , i was not sure from comment if its intended to remain unsafe, or why, but I did not encounter failing tests with making it safe or perf issues.
- Add support for parsing "now", "today", and "yesterday" as relative dates (case-insensitive)
- Refactor Time::size() to use a lookup table instead of a long if-else chain
- Remove unused bstr::ByteSlice import

I have been working on a number of gixy changes in a personal project, want to help fill some gaps if I can. Happy to take any feedback. I have some whole crates that are missing here.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,7 @@
 ### Please disclose the use AI…
 
-…if it's *full agent mode* or *multi*-line completions in the PR comment.
+…if it's editing files for you please let us know in the PR comment _or preferably_ using
+`Co-authored-by: <agent-identity>` message trailers.
 
 For everything else, please have a look at the respective section in the [README] file.
 


gix-date/src/parse/mod.rs
```diff
@@ -14,13 +14,9 @@ impl TimeBuf {
...
     /// signature fields in Git commits, also known as anything parseable as [raw format](function::parse_header()).
     pub fn as_str(&self) -> &str {
-        // SAFETY: We know that serialized times are pure ASCII, a subset of UTF-8.
-        //         `buf` and `len` are written only by time-serialization code.
+        // Time serializes as ASCII, which is a subset of UTF-8.
         let time_bytes = self.buf.as_slice();
-        #[allow(unsafe_code)]
-        unsafe {
-            std::str::from_utf8_unchecked(time_bytes)
-        }
```
Byron, line 23: The idea here was to skip UTF8 checks as we know these aren't needed.
Getting rid of `unsafe` is generally a good thing and I would be surprised if there was any code that would have benefited from being 30% or so faster.
So let's have this fix, and call the prior version a premature optimisation.

Byron (approved):
Thanks a lot!

I gave it a good refactor in case you want to take a look. While doing that I had a suspicion, due to duplication, that maybe this PR has AI involvement. If so, please declare it next time or add a `Co-authored-by` trailer.

apollocatlin:
hey sorry i use AI for comments and was moving stuff into your format. Im also booting back into rust.

PR #2436, opened 2026-02-15
Fix ls-refs ref-prefix'ing

FintanH:
See discussion at https://github.com/GitoxideLabs/gitoxide/issues/2429

I didn't quite have the correct environment set up, so I'm not sure if all the tests are passing, but let's see if CI catches anything I messed up :)

### DIFF
--- a/gitoxide-core/src/pack/receive.rs
+++ b/gitoxide-core/src/pack/receive.rs
@@ -98,7 +98,7 @@ where
     #[cfg(feature = "blocking-client")]
     let refmap = fetch_refmap.fetch_blocking(&mut progress, &mut transport.inner, trace_packetlines)?;
 
-    if refmap.mappings.is_empty() && !refmap.remote_refs.is_empty() {
+    if refmap.explicit_refspecs_have_no_mapping() {
         return Err(Error::NoMapping {
             refspecs: refmap.refspecs.clone(),
             num_remote_refs: refmap.remote_refs.len(),
--- a/gix-protocol/src/fetch/types.rs
+++ b/gix-protocol/src/fetch/types.rs
@@ -39,6 +39,8 @@ pub struct Context<'a, T> {
 
 #[cfg(feature = "fetch")]
 mod with_fetch {
+    use bstr::ByteSlice;
+
     use crate::fetch::{self, negotiate, refmap};
 
     /// For use in [`fetch`](crate::fetch()).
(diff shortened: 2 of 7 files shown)


gix-refspec/tests/refspec/spec.rs
```diff
@@ -28,14 +28,14 @@ mod prefix {
...
-            "refs/remote/"
+            "refs/remote/main"
         );
     }
 
     #[test]
     fn full_names_have_a_prefix() {
-        assert_eq!(parse("refs/heads/main").to_ref().prefix().unwrap(), "refs/heads/");
-        assert_eq!(parse("refs/foo/bar").to_ref().prefix().unwrap(), "refs/foo/");
+        assert_eq!(parse("refs/heads/main").to_ref().prefix().unwrap(), "refs/heads/main");
```
Byron, line 37: The test now tests for the wrong thing, and I think it tries to test that a ref acts as prefix.
Probably that's wrong to start with and might have been added as convenience.
Is there some notes that we can add here to show why this is desired behaviour now?

PS: It seems this would affect prefix generation negatively, at least if one assumes prefixes really are prefixes. It seems more versatile to support exact matches though, i.e. the new behaviour here.

gix-refspec/src/spec.rs
```diff
@@ -138,10 +138,17 @@ impl<'a> RefSpecRef<'a> {
...
             return source.into();
         }
-        let suffix = source.strip_prefix(b"refs/")?;
-        let slash_pos = suffix.find_byte(b'/')?;
-        let prefix = source[..="refs/".len() + slash_pos].as_bstr();
-        (!prefix.contains(&b'*')).then_some(prefix)
+
+        let sans_refs_prefix = source.strip_prefix(b"refs/")?;
+        if let Some(star_pos) = sans_refs_prefix.find_byte(b'*') {
+            // Disallow `*` glob star components after `refs/`
```
Byron: Seeing this here makes me think that this should be documented, to know where this is coming from.
Do you have context to share?

gix-refspec/src/spec.rs
```diff
@@ -138,10 +138,17 @@ impl<'a> RefSpecRef<'a> {
...
+            if star_pos == 0 {
+                return None;
+            }
```
Byron: It seems the `return None` might not match what Git does. Maybe there is a good reason to do this? Maybe the analysis is wrong?


<details><summary>Codex Analysis (unvalidated)</summary>

> Look at https://github.com/GitoxideLabs/gitoxide/pull/2436 and gather context on how Git is handling prefixes. Use the Git codebase at /Users/byron/dev/github.com/git/git if it helps.

----

From PR [GitoxideLabs/gitoxide#2436](https://github.com/GitoxideLabs/gitoxide/pull/2436), upstream Git context for `ref-prefix` is:

- Prefix generation happens in [`refspec_ref_prefixes()`](https://github.com/git/git/blob/master/refspec.c#L257-L295).
- For fetch, Git uses `src` (except exact-OID refspecs); for push, it prefers `dst` and falls back to `src` ([same function](https://github.com/git/git/blob/master/refspec.c#L268-L282)).
- Pattern refspecs are truncated at `*` before sending `ref-prefix` ([`strchr(prefix, '*')`](https://github.com/git/git/blob/master/refspec.c#L287-L291)).
- Non-pattern refspecs are expanded through [`expand_ref_prefix()`](https://github.com/git/git/blob/master/refspec.c#L293), which uses these rules in [`refs.c`](https://github.com/git/git/blob/master/refs.c#L595-L603) and implementation at [`refs.c#L631-L638`](https://github.com/git/git/blob/master/refs.c#L631-L638).

Server-side behavior:

- `ls-refs` treats each `ref-prefix` as a plain string prefix check (`starts_with`) in [`ls-refs.c`](https://github.com/git/git/blob/master/ls-refs.c#L54-L67).
- Protocol docs say prefix filtering is an optimization and clients should still filter results locally: [`gitprotocol-v2.adoc`](https://github.com/git/git/blob/master/Documentation/gitprotocol-v2.adoc#L215-L221).
- Tests confirm exact full-ref prefixes are valid (for example, `ref-prefix refs/heads/main`) in [`t5701-git-serve.sh`](https://github.com/git/git/blob/master/t/t5701-git-serve.sh#L181-L200).

How this maps to #2436:

- The PR’s change (moving away from “two path components only”) aligns with Git’s model of “full ref until `*`, or full ref if no `*`”.
- This resolves namespace cases like `refs/namespaces/<id>/...` that were previously over-truncated.
- One notable edge case: Git will emit broad prefixes like `refs/` for patterns such as `refs/*/main`; if #2436 intentionally rejects that case, that is a behavior divergence worth documenting in the PR.

</details>

FintanH: Ya, so this is where I opted to try and reach some kind of hybrid approach. There were test examples, for example `refs/*/main`, that would expect `None` as the return value. So I was attempting to maintain behaviour on that front.

As the analysis points out, Git would drop back to a prefix that looks like `refs/` which amounts to give me everything. Afaict, `refs/` and `None` will result in the same behaviour for `ls-refs`, but perhaps I'm wrong or missing a case :)

gix-refspec/src/spec.rs
FintanH: This relates to https://github.com/GitoxideLabs/gitoxide/pull/2436/changes#r2826753347.

Does that answer it, or do you need more context?

gix-refspec/tests/refspec/spec.rs
FintanH: Ya, perhaps I can add more context here for the use case in Radicle.

We actually want to ask for exact references, e.g. `refs/rad/id`, `refs/namespaces/me/refs/rad/sigrefs`. This allows us to compute the references we want to ask for next.

If this is not expected or desired, then I think we can take a hit and ask for `refs/rad/`, and `refs/namespaces/me/refs/rad/`.

I forget what I found Git to do in these cases.

Byron:
Thanks a lot for the patience, it's finally done!

Meantime, I changed a lot which helps me do reviews on PRs that otherwise would take me a lot of time to research. One of the advantages is much higher-quality reviews, and a review-commit for each input commit if changes were needed.

Most notably, I had to change the message of your second commit to make it breaking via `!`, and I added a sentence about the actual breaking change (the parameter type change of a public method).

FintanH:
Thanks for getting around to it!

I'm not precious about my changes, especially when I'm not so familiar with the codebase, so I'm delighted that you messed around and changed things. Fwiw, my workflow is the same over on `heartwood` :blush:

PR #2441, opened 2026-02-21
Remove `sha1` from default features in `gix-hash`

cruessler:
- **feat!: remove `sha1` from default features in `gix-hash`**
  

- **Adapt to `sha1` not being default feature of `gix-hash`**
  

- **Make `gix` depend on `gix-hash` with `sha1` enabled**
  
This PR removes `sha1` from `gix-hash`’s default features. It adds a feature `sha1` to all depending crates that did not compile anymore after the removal. Now, `just check` passes and I want to get feedback from CI to see where else I need to adapt to these changes.

This PR also adds `sha1` to `features` in the `[package.metadata.docs.rs]` section of all crates that now have a feature flag `sha1`.

This PR is part of #281.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1569,6 +1569,7 @@ dependencies = [
  "bstr",
  "criterion",
  "document-features",
+ "gix-config",
  "gix-config-value",
  "gix-features",
  "gix-glob",
@@ -2571,6 +2572,7 @@ dependencies = [
  "gix-object",
  "gix-path",
  "gix-worktree",
+ "gix-worktree-state",
  "io-close",
  "thiserror 2.0.18",
 ]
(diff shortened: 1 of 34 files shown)


Byron:
Thanks a lot for keeping the sha256-ball rolling!

Actually this is really nice as it reveals crates with a direct relationship to `gix-hash`, which then can probably be a focus for testing as well.

I will go back into hiding until the PR comes out of draft.

cruessler:
Quick status update: I believe I’ve now made all the follow-up changes necessary to make CI green. The last test failure seems not related to my changes as far as I can tell. I plan on updating the description soon, giving a higher-level overview of all the changes and some observations I made along the way.

EliahKagan:
> Quick status update: I believe I’ve now made all the follow-up changes necessary to make CI green.

That seems likely--I just reran the one failed job and it succeeded.

Byron:
### Review Notes

- `gix` also must make the default hashes configurable, as users should be able to turn it off by not enabling default features.
- in order to build crate documentation successfully, all crates now have to forward `sha1` effectively to configure docs.rs to set it.

PR #2446, opened 2026-02-24
`gix-ref` traits

Byron:
This is in preparation for supporting reftables

### Tasks

* [ ] design trait as parts
* [ ] implement it

This is also interesting for implementing optional in-memory handling of references later via a proxy database.


Byron:

PR #2448, opened 2026-02-28
Run `gix-refspec` tests with both `sha1` and `sha256`

cruessler:
This PR adds `sha256` support to `gix-refspec` tests. It adds the `sha256` feature flag to `dev-dependencies` and runs the test suite for `gix-refspec` for both `sha1` and `sha256`. It adds separate branches for SHA-256 in a few places where SHA-1 hashes were hardcoded.

## Open questions

- `gix-refspec/tests/fixtures/generated-archives/.gitignore` includes `/match_baseline.tar`. Does it also need to include `/match_baseline_sha256.tar`? (I haven’t committed the latter yet.)
- `gix_refspec::parse::function::looks_like_object_hash` uses `gix_hash::Kind::shortest`. Does it need to be changed? (I suspect it doesn’t.)

This PR is related to #281.

### DIFF
--- a/gix/Cargo.toml
+++ b/gix/Cargo.toml
@@ -90,6 +90,7 @@ comfort = [
 
 ## Enable support for the SHA-1 hash throughout the `gix` stack.
 sha1 = [
+    "gix-hash/sha1",
     "gix-archive?/sha1",
     "gix-blame?/sha1",
     "gix-commitgraph/sha1",
@@ -98,7 +99,6 @@ sha1 = [
     "gix-dir?/sha1",
     "gix-discover/sha1",
     "gix-filter?/sha1",
-    "gix-hash/sha1",
     "gix-hashtable/sha1",
     "gix-index?/sha1",
     "gix-merge?/sha1",
@@ -121,8 +121,6 @@ sha1 = [
 ]
 
 ## Enable support for the SHA-256 hash throughout the `gix` stack.
-##
-## Note that, as of 2026-03-02, only `gix-hash` offers that flag yet.
 sha256 = [
     "gix-hash/sha256",
 ]


Byron:
Sorry for the late response, and thanks for keeping this topic going!

As you can see, I changed my review style and create changes on top of each commit (if needed).
While doing so I also noticed that `gix` doesn't actually compile just yet, and I hope that getting it to compile can be prioritized (via `gix-pack` for instance, and probably more).

> * `gix-refspec/tests/fixtures/generated-archives/.gitignore` includes `/match_baseline.tar`. Does it also need to include `/match_baseline_sha256.tar`? (I haven’t committed the latter yet.)
Yes, this came up naturally and it's ignored now. Codex also pointed it out.

> * `gix_refspec::parse::function::looks_like_object_hash` uses `gix_hash::Kind::shortest`. Does it need to be changed? (I suspect it doesn’t.)

I think that's fine as well.

Finally, I hope I will not get into the situation again that so many PRs are piling up - have to take the time now to review everything 😅.

cruessler:
> Sorry for the late response, and thanks for keeping this topic going!

No worries, I’ll happily continue whenever I get a new review. :-)

> As you can see, I changed my review style and create changes on top of each commit (if needed). While doing so I also noticed that `gix` doesn't actually compile just yet, and I hope that getting it to compile can be prioritized (via `gix-pack` for instance, and probably more).

`gix` not compiling, is that something that still needs to be addressed? (I feel I might be missing context.)

Byron:
> gix not compiling, is that something that still needs to be addressed? (I feel I might be missing context.)

I think I meant that `gix` doesn't compile with `sha256` enabled. My confusion stemmed from me being able to use Cargo features that didn't actually compile.

cruessler:
> I think I meant that `gix` doesn't compile with `sha256` enabled. My confusion stemmed from me being able to use Cargo features that didn't actually compile.

Understood. I’ll have a look!

cruessler:
@Byron Which command did you run exactly that did not work?

On my Ubuntu machine, all of the following work:

- `cargo build --package gix --features=sha1,sha256` works.
- `cargo build --package gix --features=sha256` works.
- Adding a feature flag `sha256` to the root `Cargo.toml` and running `cargo build --features sha1,sha256` also works.

Byron:
Thanks for double-checking - I retract my statement :D. Maybe I really was confused, and will take a look if it happens again.

PR #2455, opened 2026-03-05
Revert "fix(gix-index): handle loongarch64-musl stat struct field names"

huajingyun01:
This reverts commit 7e25c3426b382493007fa6982873f76baa9d020d.

The [issue](https://github.com/rust-lang/libc/issues/4939) was fixed in libc: https://github.com/rust-lang/libc/pull/4958

### DIFF
--- a/gix-index/src/fs.rs
+++ b/gix-index/src/fs.rs
@@ -54,30 +54,14 @@ impl Metadata {
     pub fn modified(&self) -> Option<SystemTime> {
         #[cfg(not(windows))]
         {
-            #[cfg(not(any(
-                target_os = "aix",
-                target_os = "hurd",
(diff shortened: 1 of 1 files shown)


Byron:
Sorry for the long wait, and thanks @weihanglo for your appearance. That helped me to prioritise.

EliahKagan:
Will this change appear as intended in the changelog, or should something else be done to achieve that? I don't think cargo-smart-release recognizes commit messages of the form `Revert "..."` where `...` is conventional.

Byron:
That's a good catch! You are right, this won't be picked up and isn't going into the changelog.
The easiest would be to use `cargo changelog` to write the unreleased section, and then add the desired text directly.

huajingyun01:
Thanks!

PR #2457, opened 2026-03-06
gix-blame: Incremental

vmg:
Hiiiiii @Byron! Thanks for all your work on the library!

I've been playing around with the new blame APIs that @cruessler developed. The existing `gix_blame::file` was not fitting the use case we needed at Cursor, so I took a stab at implementing an equivalent to `git blame --incremental`. The changes were quite minimal because I just left `gix_blame::file` as a thin wrapper over `gix_blame::incremental`.

I then tried benchmarking the `incremental` API against Git itself and the numbers were not good at all. After some review, I noticed that the `gix-commitgraph` crate just didn't support the changed-paths bloom filter cache from Git, so I took a stab at implementing those too.

The results are very good. These are for `tools/clang/spanify/Spanifier.cpp` in the Chromium repository, which is a _very very_ hairy file:

```
gix-blame::incremental/without-commit-graph
                        time:   [14.852 s 14.895 s 14.944 s]
                        change: [+0.2968% +0.7623% +1.2529%] (p = 0.00 < 0.05)
gix-blame::incremental/with-commit-graph
                        time:   [287.55 ms 290.30 ms 292.85 ms]
                        change: [−3.1181% −1.6720% −0.4502%] (p = 0.11 > 0.05)
                        No change in performance detected.
```

Since all these changes are quite related, I'm putting them up here in a single PR. Every commit is self contained and explains the changes on the commit message so if you'd like me to split this into smaller PRs just let me know.

Thanks!


gix-blame/src/types.rs
```diff
@@ -10,6 +10,19 @@ use std::{
 use crate::file::function::tokens_for_diffing;
 use crate::Error;
 
+/// Receives [`BlameEntry`] values incrementally as they are discovered by
+/// [`incremental()`](crate::incremental()).
+pub trait BlameSink {
+    /// Receive a single blame chunk in generation order.
+    fn push(&mut self, entry: BlameEntry);
```
cruessler: What do you think about having `push` return `std::ops::ControlFlow`? https://doc.rust-lang.org/stable/std/ops/enum.ControlFlow.html

gix-blame/tests/blame.rs
```diff
@@ -621,6 +675,165 @@ mod rename_tracking {
...
+                    }
+
+                    acc
+                } else {
+                    acc.push(entry);
+
+                    acc
+                }
+            })
+    }
```
cruessler: This version of `coalesce_blame_entries` looks identical to the one in `function.rs`. Given that the latter also has context in a comment, I think it might make sense to export `coalesce_blame_entries` so the implementation can be shared. What do you think?

cruessler (commented):
Thanks a lot for the effort, that’s much appreciated as making blame incremental was on our TODO list as well! I had a first look at the changes and left 2 comments (though this is far from an in-depth review).

vmg:
Thanks for the review @cruessler! I did the two changes; interrupting the incremental blame is a little bit involved and makes the code harder to follow so I've kept it in a separate commit. Let me know what else would you like to see changed.

vmg:
Anything else I could do to move this PR forward, @Byron? I think it'd a nice addition upstream and we would very much prefer _not_ to carry a Gitoxide fork. Thank you so much in advance!

Byron:
Hi, I'd also like to get this merged, and it seems there is more and more code coming up and never enough time to review.

These days that's all I am doing 😅.

@codex review

cruessler:
@vmg I’m really sorry I haven’t responded earlier, I had a few very busy weeks at work, so was lacking review capacity.

@Byron Things have gotten quieter this week, so I can prioritize reviewing the blame-related parts this weekend if you want an extra pair of eyes.

vmg:
No worries! I know you're all busy! I'm gonna go through all the AI comments first (I can see a couple legit ones) and then I'll hand it over so you guys can do human review. Thanks again.

Byron:
Thanks everyone! Yes, I will be happy to wait for your review @cruessler, and you can hand it over to me even with things that are still to be addressed, as long as *you* want to go forward with the changes. Then I will take it towards merging, and will also be looking forward to seeing more contributions and improvements from @vmg 🙏.

cruessler:
> No worries! I know you're all busy! I'm gonna go through all the AI comments first (I can see a couple legit ones) and then I'll hand it over so you guys can do human review. Thanks again.

I’ve gone through the changes related to incrementally emitting hunks at a high level, and they look very promising to me! Looking forward to you handing over after you’ve gone through AI’s comments!

One thing I was wondering: in `gitoxide` tests, we usually try to run our implementations against a `git` baseline if there is one. In this instance, do you think it would make sense to test the incremental part of the blame algorithm against `git`’s implementation? We know that `gix-blame` doesn’t match `git blame` in 100 % of cases due to small differences in their respective diff algorithms, but there’s still such a significant overlap that writing such tests could be useful.

vmg:
Just went through the AI review comments, they were all basically the same (good) point: since we're not supporting the deprecated v1 format, we should explicitly ensure that we don't load the bloom filter cache when it's v1.

Let me know what else needs changing!

Byron:
@vmg Maybe we can work towards merging the bloom-filter support first?

From there, `gix_blame::incremental` appears to be a re-implementation of the original non-incremental algorithm that lives side-by-side, and @cruessler wonders how that can be validated more with baseline tests. From all I know, we already have baseline tests and I'd think they can be adapted to run against the incremental implementation as well.
And I agree, I think having the same test-coverage as the existing non-incremental implementation would be a requirement for a merge (and I think it's in the interest of everyone).

cruessler:
Thanks a lot for all the changes, so far the code has been really easy to follow and understand which is much appreciated! I started reviewing the `gix log PATHSPEC`/`log_file` part and will go through this PR’s major topics one by one (where I think my review and testing might be helpful). In terms of getting it merged as efficiently as possible, I think it would benefit from being split into smaller PRs as it looks like it will take me a couple of hours to do all parts I want to review the justice they deserve. And I might need to spread the reviews a little to keep my eyes fresh. :-)

I just tested the output of running `gix log README.md` and `gix log Cargo.lock` in this branch, running a diff with the output of `git log --oneline --full-history README.md` and `Cargo.lock`, respectively. Apart from the fact that `git log` returned hashes that were one character longer than the ones returned by `gix` (I don’t think that needs to be changed), there were some minor differences in commit titles (I didn’t dig further) and some larger ones with respect to the order of commits in two blocks that had about 2 dozen commits each. At this point, I assume these differences are related to `gix::traverse::commit::topo::Builder` and not to the newly added code, though.

Based on these observations, I think `log_file` is already good to go (I know it depends on the commit graph-related changes)!

gix-blame/src/file/mod.rs
```diff
@@ -322,24 +322,36 @@ fn process_change(
...
+
+    for hunk in hunks_to_blame {
+        if !hunk.has_suspect(&suspect) {
+            new_hunks_to_blame.push(hunk);
+            continue;
+        }
 
-    for mut hunk in hunks_to_blame.into_iter().map(Some) {
-        let mut offset_in_destination = Offset::Added(0);
+        let suspect_range = hunk.get_range(&suspect).expect("has_suspect was true");
```
cruessler, line 345: Could this also be shortened to?

```
        let Some(suspect_range) = hunk.get_range(&suspect) else {
            new_hunks_to_blame.push(hunk);
            continue;
        };
```

gix-blame/src/file/mod.rs
```diff
@@ -322,24 +322,36 @@ fn process_change(
...
 fn process_changes(
     hunks_to_blame: Vec<UnblamedHunk>,
-    changes: Vec<Change>,
+    changes: &[Change],
     suspect: ObjectId,
     parent: ObjectId,
 ) -> Vec<UnblamedHunk> {
     let mut new_hunks_to_blame = Vec::new();
+    let mut offset_in_destination = Offset::Added(0);
+    let mut changes_pos = 0;
```
cruessler, line 336: Instead of using `changes_pos`, would the following also be an option?

```rust
// Move these to lines out of the `for` loop
let mut changes_iter = changes.iter().cloned();
let mut change = changes_iter.next();

// And then, instead of `changes_pos = 0`
changes_iter = changes.iter().cloned();
change = changes_iter.next();
```

I’m mainly asking because I find it easier to wrap my head around iterators as opposed to manual tracking of indices.

gix-blame/src/file/mod.rs
```diff
@@ -322,24 +322,36 @@ fn process_change(
 /// Consume `hunks_to_blame` and `changes` to pair up matches ranges (also overlapping) with each other.
 /// Once a match is found, it's pushed onto `out`.
 ///
-/// `process_changes` assumes that ranges coming from the same *Source File* can and do
-/// occasionally overlap. If it were a desirable property of the blame algorithm as a whole to
-/// never have two different lines from a *Blamed File* mapped to the same line in a *Source File*,
-/// this property would need to be enforced at a higher level than `process_changes`.
-/// Then the nested loops could potentially be flattened into one.
+/// Uses a cursor to avoid restarting the changes iteration from the beginning for each hunk.
```
cruessler, line 325: In `subsequent_hunks_overlapping_end_of_addition`, I think it would make sense to make the following change:

```rust
let changes = vec![
    Change::AddedOrReplaced(5..9, 4),
    Change::Unchanged(9..10),
    Change::AddedOrReplaced(10..14, 0),
];
```

The test that was written as a small regression test when the original nested loops were introduced currently only has one change, so it doesn’t trigger resetting `changes_pos` in a meaningful way.

cruessler (changes requested):
This is a review for the commit `optimize process_changes to avoid redundant work`. Just a few suggestions.

On a separate note: both @Byron and I already mentioned so-called baseline tests. In the context of this PR, this would mean creating `.baseline` files similar to how it is done, e. g., in this script: https://github.com/GitoxideLabs/gitoxide/blob/5b9460485488ff209bc8ce77c5db97c9e32d7944/gix-blame/tests/fixtures/make_blame_repo.sh#L262-L297. Then, there would be a parser similar to this one: https://github.com/GitoxideLabs/gitoxide/blob/5b9460485488ff209bc8ce77c5db97c9e32d7944/gix-blame/tests/blame.rs#L11-L146. And then, the test would parse the relevant `.baseline` file and compare the output of `gix_blame::incremental` to `git`’s output. Can you add such baseline tests or do you see any issues with that approach?

gix-blame/benches/incremental_blame.rs
```diff
@@ -0,0 +1,131 @@
...
+            .expect("incremental blame should succeed");
+        });
+    });
+    group.finish();
+}
+
+fn has_commit_graph_cache(git_dir: &Path) -> bool {
+    let info_dir = git_dir.join("objects/info");
+    info_dir.join("commit-graph").is_file() || info_dir.join("commit-graphs").is_dir()
+}
```
cruessler, line 101: Is this potentially available as an API on `repo`?

@Byron It it is not, should it be?

gix-blame/benches/incremental_blame.rs
```diff
@@ -0,0 +1,131 @@
...
+
+fn write_changed_path_commit_graph(worktree_path: &Path) -> std::io::Result<()> {
+    let config_status = Command::new("git")
```
cruessler, line 104: Is there a way to set this value through an API on `repo`?

gix-blame/benches/incremental_blame.rs
```diff
@@ -0,0 +1,131 @@
...
+        .args(["config", "commitGraph.changedPathsVersion", "2"])
+        .current_dir(worktree_path)
+        .status()?;
+    assert!(
+        config_status.success(),
+        "setting commitGraph.changedPathsVersion should succeed"
+    );
+
+    let write_status = Command::new("git")
```
cruessler, line 113: What do you think about adding a comment that we have to shell out to `git` because `gix-commitgraph` is currently read-only?

gix-blame/Cargo.toml
```diff
@@ -38,6 +43,8 @@ smallvec = "1.15.1"
 thiserror = "2.0.18"
 
 [dev-dependencies]
+criterion = "0.8.0"
+gix = { path = "../gix", default-features = false, features = ["basic", "sha1"] }
```
cruessler, line 47: @Byron Is it ok to depend on `gix` here? Since the feature `blame` is not included, this does not create a circular dependency, but as far as I can tell, currently only `gitoxide-core` and `internal-tools` depend on `gix`, so I wanted to flag this.

cruessler (changes requested):
This is a review for the commit `implement benchmarks for incremental blame`. Just a couple of minor suggestions/questions.

gix-blame/src/file/function.rs
```diff
@@ -100,19 +95,19 @@ pub fn file(
...
 
-    let mut out = Vec::new();
     let mut diff_state = gix_diff::tree::State::default();
     let mut previous_entry: Option<(ObjectId, ObjectId)> = None;
     let mut blame_path = if options.debug_track_path {
         Some(Vec::new())
     } else {
         None
     };
+    let mut stopped_early = false;
```
cruessler, line 110: What do you think about turning the `while` loop into an actual `loop` and have it return a value, like so:

```
let stopped_early = 'outer: loop {
    …
};
```

gix-blame/src/file/function.rs
```diff
@@ -879,8 +924,111 @@ fn collect_parents(
...
+
+#[cfg(test)]
+mod tests {
+    use gix_object::bstr::BString;
+    use gix_testtools::Result;
+
+    use super::{file, incremental};
+    use crate::{file::coalesce_blame_entries, BlameRanges, Options};
+
+    struct Fixture {
```
cruessler, line 968: By moving the `mod tests` block to `tests/blame.rs`, would it be possible to reuse the existing `struct Fixture`? https://github.com/GitoxideLabs/gitoxide/blob/5b9460485488ff209bc8ce77c5db97c9e32d7944/gix-blame/tests/blame.rs#L148-L152

gix-blame/tests/blame.rs
```diff
@@ -621,6 +675,159 @@ mod rename_tracking {
...
+            &mut resource_cache,
+            source_file_name.as_ref(),
+            &mut sink,
+            options(),
+        )?;
+
+        assert!(
+            sink.chunks_received > 0,
+            "incremental blame should stream at least one chunk"
+        );
```
cruessler, line 746: Do you think it makes sense to strengthen the assertion to `assert_eq!`?

gix-blame/tests/blame.rs
```diff
@@ -621,6 +675,159 @@ mod rename_tracking {
...
+        assert!(
+            !outcome.blob.is_empty(),
+            "incremental blame should still return partial metadata on early stop"
+        );
+        Ok(())
+    }
+}
+
+fn write_changed_path_commit_graph(worktree_path: &Path) -> gix_testtools::Result {
```
cruessler, line 805: This function now exists twice. Does it make sense to only keep a single instance?

cruessler (changes requested):
This is a review for the 2 commits related to adding the actual incremental blame. I really like your implementation and how it fits the existing code! I just have a couple of mostly smaller questions.

As far as I can tell, I’m now done with reviewing all the commits that I feel I have sufficient context on reviewing. My knowledge of the commit graph and bloom filters is very limited, so I wouldn’t really be able to add much here. I’m handing this back to you! @vmg

gix-blame/benches/incremental_blame.rs
Byron: I don't think so, and judging from how it is used, I think it's fair to just use https://docs.rs/gix/latest/gix/struct.Repository.html#method.commit_graph_if_enabled. While it's doing more than needed here, it's fine in a test.
And I also think it should be used rather than having this helper.

gix-blame/benches/incremental_blame.rs
Byron: It's totally possible, even though I have seen code like this generated a lot as the APIs are cumbersome enough to use at least compared to `git2`.

I'd even be OK with it but would prefer something like `gix_testtools::invoke_bash()` as a helper. I will [whip one up](https://github.com/GitoxideLabs/gitoxide/pull/2511).

gitoxide-core/src/repository/log.rs
Byron, line 56: We could make this `info.id().shorten()` to match Git behaviour. That needs the ODB of the repo to be configured differently though to get proper performance. This is just a note, and not a requirement for this PR.

gix-blame/Cargo.toml
Byron: It's better to avoid it as it slows down tests tremendously due to increased build times.
As the `repo` seems to be used only  to obtain the commit-graph, we should remove this dependency and open it via plumbing directly. Should be easy enough even without re-implementing all Git configuration, so that typical commit-graphs can be opened.

It's a benchmark after all.

Byron:
It's strange that this PR is shown as non-conflicting as when rebasing it, it does conflict immediately.
I have a clanker fix it up and maybe I can push the final patch here for comparison, here is the result: https://github.com/GitoxideLabs/gitoxide/pull/2625. The PR is for convenience only, inspection, and to see if CI passes now, not to supersede this one.

This PR has been open for a long time and I wonder how to better facilitate innovation like this.
Maybe something we can catch up on one day, and I think @cruessler can be our liaison.

PR #2489, opened 2026-03-25
fix(tree): peel to the final tree entry

### LINKED ISSUE #2456: `Tree::peel_to_entry{_by_path}` does not point to to "last seen tree" if leaf entry is a `Tree`
### Current behavior 😯

Calling `Tree::peel_to_entry{_by_path}` does not update the tree itself to "the last seen tree" if the returned entry itself is a tree. For example, looking up the entry `path/to/a/tree` causes the tree that is being peeled to point at `path/to/a/` instead of `path/to/a/tree`, but from the documentation that is not entirely clear.

I feel like the behaviour is intentional, but the documentation is ambiguous in my opinion.

Naturally I am more than willing to contribute a documentation change to clarify this, but I wanted to make sure it was intentional first.

### Expected behavior 🤔

Peeling takes all "seen" elements into account, including the returned `Entry`.

### Git behavior

N/A

### Steps to reproduce 🕹

1. Obtain a tree `tree` that contains a tree `sub_tree`.
2. Call `tree.peel_to_entry_by_path("sub_tree")`
3. `tree` is unchanged, despite `peel_to_entry_by_path` having "seen" the tree.

```rust
#[test]
fn repro() -> Result<(), Box<dyn std::error::Error>> {
    use gix::object::tree::EntryKind;

    let repo = gix::open(".")?;
    let mut middle_tree = repo.empty_tree().edit()?;
    middle_tree.upsert("sub_tree", EntryKind::Tree, repo.empty_tree().id)?;
    let middle_tree_id = middle_tree.write()?;

    let mut top_tree = repo.empty_tree().edit()?;
    top_tree.upsert("sub_tree", EntryKind::Tree, middle_tree_id)?;
    let top_tree_id = top_tree.write()?;

    let mut top_tree = repo.find_tree(top_tree_id)?;

    // Peeling into more than one level of trees changes
    // `peel_to_bottom`.
    let mut peel_to_bottom_tree = top_tree.clone();
    peel_to_bottom_tree.peel_to_entry_by_path("sub_tree/sub_tree")?;
    assert_ne!(peel_to_bottom_tree.id, top_tree_id);

    // But peeling into a single level does not change
    // `top_tree`, despite the fact that `peel_to_entry_by_path`
    // has "seen" `sub_tree`.
    top_tree.peel_to_entry_by_path("sub_tree")?;
    assert_ne!(
        top_tree_id, top_tree.id,
        "We saw(?) \"sub_tree\", but \"tree\" is unchanged"
    );

    Ok(())
}
```

Byron:
## Summary
- update `Tree::peel_to_entry{_by_path}` to move `self` to the final tree entry when the leaf entry is a tree
- clarify the behavior in the API docs
- add a regression test for peeling to a tree leaf

## Testing
- `cargo test -p gix --test gix object::tree::`

Refs #2456
Context: https://github.com/GitoxideLabs/gitoxide/issues/2456#issuecomment-4126248450

### DIFF
--- a/gix/src/object/tree/mod.rs
+++ b/gix/src/object/tree/mod.rs
@@ -88,15 +88,14 @@ impl<'repo> Tree<'repo> {
 
     /// Follow a sequence of `path` components starting from this instance, and look them up one by one until the last component
     /// is looked up and its tree entry is returned, while changing this instance to point to the last seen tree.
-    /// Note that if the lookup fails, it may be impossible to continue making lookups through this tree.
+    /// If the returned entry itself is a tree, this instance is updated to point to it as well.
     /// It's useful to have this function to be able to reuse the internal buffer of the tree.
     ///
     /// # Performance Notes
     ///
-    /// Searching tree entries is currently done in sequence, which allows to the search to be allocation free. It would be possible
-    /// to reuse a vector and use a binary search instead, which might be able to improve performance over all.
-    /// However, a benchmark should be created first to have some data and see which trade-off to choose here.
-    ///
+    /// Searching tree entries is currently done in sequence, which allows to the search to be allocation free.
+    /// This is beneficial for most 'common' repositiries, but for other cases an allocation might be preferable
+    /// to allow a bisect.
     pub fn peel_to_entry<I, P>(&mut self, path: I) -> Result<Option<Entry<'repo>>, find::existing::Error>
     where
         I: IntoIterator<Item = P>,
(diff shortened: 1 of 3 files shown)


gix/src/object/tree/mod.rs
```diff
@@ -127,7 +115,16 @@ impl<'repo> Tree<'repo> {
...
+                    let mapped = entry.map(|e| {
+                        let entry: gix_object::tree::Entry = e.into();
+                        Entry { inner: entry, repo }
+                    });
+                    if let Some(entry) = &mapped {
+                        if entry.mode().is_tree() {
+                            self.repo.find(entry.oid(), &mut self.data)?;
+                            self.id = entry.object_id();
+                        }
+                    }
```
datdenkikniet: To avoid the (effective) double `if let Some(_)`:

```suggestion
                    if let Some(entry) = entry {
                        let repo = self.repo;
                        let entry: gix_object::tree::Entry = entry.into();

                        if entry.mode.is_tree() {
                            self.repo.find(&entry.oid, &mut self.data)?;
                            self.id = entry.oid;
                        }

                        break Ok(Some(Entry { inner: entry, repo }));
                    } else {
                        break Ok(None);
                    }
```

Byron: Thanks for taking a look.
I don't think this can be done with the MSRV compiler level. Maybe I am wrong, so please feel free to follow it up with the improvements.

PR #2509, opened 2026-04-10
upgrade getrandom to 0.4

alexanderkjall:
from: https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/gix-diff/debian/patches/getrandom-0.4.patch?ref_type=heads


Byron:
Thank you!

PR #2511, opened 2026-04-12
various improvements

Byron:
### Tasks

* [x] review `invoke_bash()`

### DIFF
--- a/tests/tools/src/lib.rs
+++ b/tests/tools/src/lib.rs
@@ -260,6 +260,27 @@ pub fn run_git(working_dir: &Path, args: &[&str]) -> std::io::Result<std::proces
         .status()
 }
 
+/// Run `script` with [`bash_program()`] in `cwd`.
+///
+/// Standard input is disconnected while standard output and error stay attached to the inherited
+/// handles.
+///
+/// # Panics
+///
+/// This function expects the script to succeed and will panic otherwise.
+pub fn invoke_bash(cwd: impl AsRef<Path>, script: &str) {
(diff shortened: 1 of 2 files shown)


Byron:
@EliahKagan [This](https://github.com/GitoxideLabs/gitoxide/actions/runs/24317419214/job/70997454794?pr=2511#step:7:1) isn't the first time I have seen this failure mode.

```
Run cargo nextest run --workspace --no-fail-fast --exclude gix-error -- --skip fuzzed_timeout --skip performance --skip speed
error: no such command: `nextest`

help: a command with a similar name exists: `test`

help: view all installed commands with `cargo --list`
help: find a package to install `nextest` with `cargo search cargo-nextest`
```

Just wanted to flag this in case you have some intuition about it - it's like there is a race between putting something in the `PATH` and being able to invoke it.

EliahKagan:
> Just wanted to flag this in case you have some intuition about it

I vaguely recall reading that there was a bug in one of the actions we use in CI to install packages that intermittently causes it to fail silently to install a package on ARM64 Windows 11 systems--and that was fixed, but only recently. I don't know for sure whether this recollection is correct or not. I'll try to look into it.

EliahKagan:
I've upgraded `install-action` in https://github.com/GitoxideLabs/gitoxide/pull/2517, which should fix the `` no such command: `nextest` `` error. See https://github.com/GitoxideLabs/gitoxide/pull/2517#pullrequestreview-4108665488 for details.

PR #2512, opened 2026-04-13
Implement `TokenSource` for `BStr`

cruessler:
This PR implements `TokenSource` for `BStr`. This makes the following code work:

```rust
let before: &BStr = "...";
let after: &BStr = "...";

let input = v2::InternedInput::new(before, after);
let printer = v2::BasicLineDiffPrinter(&input.interner);
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1922,6 +1922,7 @@ dependencies = [
 name = "gix-imara-diff"
 version = "0.2.0"
 dependencies = [
+ "bstr",
  "cov-mark",
  "expect-test",
  "hashbrown 0.16.1",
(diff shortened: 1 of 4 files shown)


Byron:
Thanks for sharing!

Now that I saw it I realize that the BasicUnifiedDiffPrinter isn't anything `gix` needs unless it's for performance reasons. The version in `gix-diff` provides finer-grained access which is what's needed downstream.

And performance isn't really an issue as the slow part is always the diffing, so I am thinking in the direction of removing the unused parts from `gix-imara-diff` entirely, up to the point where maybe I will put all relevant code into one module with license and attribution and bring it into `gix-diff`, to not add yet another crate to the tree.

Thus I think this PR can be closed unless you think differently about the direction.

cruessler:
That’s absolutely fine by me! One less thing I have to keep in mind. :-)

PR #2514, opened 2026-04-13
fix(gix-object): detect trailers when they are the sole body content

mtsgrd:
## Problem

`BodyRef::from_bytes` searched for `\n\n` *within the body* to find the trailer block. But `MessageRef::from_bytes` already consumed the first `\n\n` when splitting the title from the body — so for a message like:

```
Fix the thing

Signed-off-by: Alice <alice@example.com>
```

the body bytes arriving at `BodyRef::from_bytes` are just `"Signed-off-by: Alice <alice@example.com>"` with no inner `\n\n`. The result was zero trailers, diverging from `git interpret-trailers --parse`:

```
$ printf 'Fix the thing\n\nSigned-off-by: Alice <alice@example.com>\n' | git interpret-trailers --parse
Signed-off-by: Alice <alice@example.com>
```

This is the common case in practice: a subject line followed directly by trailers with no other body text (Gerrit `Change-Id`, GitHub `Co-authored-by`, etc.).

## Fix

When no `\n\n` is found within the body *and* the first non-empty line parses as a valid `Token: value` trailer, treat the entire body as the trailer block (with an empty `body_without_trailer`). Non-trailer body text followed by a trailer on the same paragraph without a blank-line separator is correctly left unchanged — a blank line is still required to delimit the trailer block from preceding prose.

## Tests added

- `trailer_as_sole_body_content` — single trailer, no preceding body text  
- `multiple_trailers_as_sole_body_content` — multiple trailers, no preceding body text  
- `body_text_then_trailer_without_blank_line_is_not_a_trailer` — guard: body text + trailer in same paragraph (no `\n\n`) is **not** a trailer block  
- `trailer_as_sole_body_content_via_message_ref` — end-to-end through the public `MessageRef` API

### DIFF
--- a/gitoxide-core/src/hours/mod.rs
+++ b/gitoxide-core/src/hours/mod.rs
@@ -2,6 +2,7 @@ use std::{collections::BTreeSet, io, path::Path, time::Instant};
 
 use anyhow::bail;
 use gix::{
+    actor::{Identity, IdentityRef},
     bstr::{BStr, ByteSlice},
     prelude::*,
     progress, Count, NestedProgress, Progress,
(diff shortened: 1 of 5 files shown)


Byron:
Thanks so much to get this started!

I ended up having Codex implement a compliant version of this, along with support for continuations/multi-line (those are folded into a single line).

PR #2519, opened 2026-04-16
fix CI (cargo deny)

Byron:


Byron:
Bypassed CI as this really only affects the `deny` job, and that one passed.

Issue #1595, opened 2024-09-14
Streaming Blob Contents


willstott101:
### Summary 💡

I'd like to get a stream of bytes from the odb without having to load the entire blob into memory. I've had a peruse of the pack parsing code and the official git docs for the format, and it seems like there are no guarantees that the chunks are stored in order in the packfile, would you expect that to be a sticking point here, or in reality would they always be in order?

I'm intending to explore implementing this myself.

### Motivation 🔦

I'm starting to use gitoxide to serve files from git repos on a server. I was previously shelling out to `git cat-file` in nodejs, but rust is proving to be a nicer environment for the few other bits this service performs. I would like to keep memory usage to a minimum however, and we have some tens-of-megabyte blobs which are a bit awkward to have to load into memory before serving them.

Byron:
You are not the first to desire such a feature, and here is my thoughts on it:

* object streaming isn't reasonably possible for delta-packed objects
* object streaming is quite straightforward for
    - loose objects
    - non-deltified objects

`gitoxide` doesn't have an API that would allow streaming, but for two out of three cases it *could* have one. It's just a question how often an object can actually be streamed, but that could be something the server-side can influence.

I can imagine, for example, that the server finds big objects that it wants to make streamable, and then writes them out as loose-objects. Unfortunately, these would probably be removed next time a `gc` runs, but that would be under control of the server as well.

From an implementation point of view, I can imagine offering a way to stream an object, which then either streams it from an opened loose file, or 'fake'-streams it from an in-memory copy. The first version of the implementation probably wouldn't want to deal with streaming from a pack as most packed objects are deltified.

My recommendation is to start with a non-streaming version, and see how you do in the memory-usage department right away. I suspect real memory usage to be pretty low and very much defined by the size of objects that are held in memory during the streaming, assuming not much else is going on.
From there, with the effort described here, one should be able to lower the peak memory usage.

The reason I am saying this is that there might be other factors that drive memory usage, and it would be good to find out quickly before optimizing something that in reality isn't a problem.

willstott101:
I can't pretend I'm not partly pursuing this for academic reasons. I'm sure for 99% of files stored in git repos it would never result in a perceivable improvement, but at the same time there's a chicken and egg problem - tooling proficiency influences usage patterns which influences tooling engineering effort. Anyway that's all a bit philosophical...

[...]

PR #2526, opened 2026-04-20
gix,gix-odb,gix-pack: stream blob contents from the ODB

mjgil:
Add streaming blob lookup support for loose and packed objects, avoid eagerly materializing packed results into memory, and expose repository helpers with regression coverage for loose, packed, empty, and wrong-type cases.

Addresses #1595.

## Summary

This adds blob-only streaming reads from the object database so callers can obtain blob contents without first materializing the full object in memory.

At the repository level, this introduces blob streaming helpers that:
- stream loose blobs directly
- stream packed blobs through a tempfile-backed reader
- reject non-blob object IDs
- preserve the existing empty-blob fast path

## Implementation

To support this, the change adds a streaming object path through the ODB stack.

Loose objects are exposed as a reader over incremental decompression.

Packed objects are decoded into a temporary file and then returned as a readable stream. This avoids holding the fully resolved blob in RAM, including for delta-resolved packed blobs, while still fitting the existing pack decoding model.

The repository-facing API is intentionally blob-only for now, which keeps the surface area aligned with the use case described in this issue.

## Tests

Added coverage for:
- loose blob streaming
- packed blob streaming
- empty blob streaming
- rejection of non-blob IDs

Also validated with:
- `cargo test -p gix`
- `cargo test -p gix-odb`
- `cargo test -p gix-pack`

## Notes

Co-authored-by: Claude
Co-authored-by: Codex

### DIFF
--- a/gix-odb/src/cache.rs
+++ b/gix-odb/src/cache.rs
@@ -91,6 +91,22 @@ impl<S> Cache<S> {
     }
 }
 
+impl<S> Cache<crate::store::Handle<S>>
+where
+    S: Deref<Target = crate::Store> + Clone,
+{
+    /// Find an object and return its decoded bytes as a stream.
+    pub fn try_find_stream(
+        &self,
+        id: &gix_hash::oid,
+    ) -> Result<Option<(crate::find::Stream, Option<gix_pack::data::entry::Location>)>, gix_object::find::Error> {
+        match self.pack_cache.as_ref().map(RefCell::borrow_mut) {
+            Some(mut pack_cache) => self.inner.try_find_stream(id, pack_cache.deref_mut()),
+            None => self.inner.try_find_stream(id, &mut gix_pack::cache::Never),
+        }
+    }
+}
+
 impl<S> From<S> for Cache<S>
 where
     S: gix_pack::Find,
--- a/gix-odb/src/find.rs
+++ b/gix-odb/src/find.rs
@@ -1,3 +1,73 @@
+use std::io::{self, Cursor, Read};
+
+/// A streaming view over an object's decoded bytes.
+pub struct Stream {
+    kind: gix_object::Kind,
+    size: u64,
+    inner: StreamInner,
+}
+
+enum StreamInner {
+    InMemory(Cursor<Vec<u8>>),
+    File(std::fs::File),
+    Loose(crate::store_impls::loose::find::StreamReader),
+}
+
+impl Stream {
+    /// Return the kind of the object yielded by this stream.
+    pub fn kind(&self) -> gix_object::Kind {
+        self.kind
+    }
+
+    /// Return the decoded object size in bytes.
+    pub fn size(&self) -> u64 {
+        self.size
+    }
+
+    /// Return an empty blob stream.
+    pub fn empty_blob() -> Self {
+        Self::from_bytes(gix_object::Kind::Blob, Vec::new())
+    }
+
+    pub(crate) fn from_bytes(kind: gix_object::Kind, data: Vec<u8>) -> Self {
+        Self {
+            kind,
+            size: data.len() as u64,
+            inner: StreamInner::InMemory(Cursor::new(data)),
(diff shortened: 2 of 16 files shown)


Byron:
Thanks for taking a stab at this. If this would work, that would be a huge step forward for our Gitoxide.

For the PR to have a chance of being merged, I'd need proof of "it's working". For that, you could create a test with a custom 'measuring' allocator that shows how an optimally compressed and thus huge base object and deltified object are decompressed with less memory usage compared to the current eager decompression that stores the whole object in memory.

I may also add that it will take considerable time to get to it, but if you stay on top of it and provide the requested information, it should happen eventually.

Byron (changes requested):
Let's also add benchmark comparisons to see if this affects the performance of packed object decompression in the non-streaming case. The `gix` CLI has subcommands for that.

mjgil:
Addressed the review items in `84931e7c3`.

Changes:
- reject truncated loose-object streams instead of returning short reads
- restore delta output-size validation after the `Write`-based `delta::apply()` refactor
- add an allocator-backed proof test for packed delta blob streaming memory behavior

Validation:
- `cargo test -p gix-pack`
- `cargo test -p gix-odb`
- `cargo test -p gix`

Memory proof:
- added `cargo test -p gix --test blob_stream_memory`
- this creates a packed delta blob and compares peak allocations between `find_blob()` and `find_blob_stream()` on fresh repo instances
- the streaming path shows a lower peak allocation than the eager path

Non-streaming benchmark:
- command: `gix free pack verify -s <idx>`
- `main`: `111.2 ms ± 1.0 ms`
- this branch: `113.6 ms ± 2.1 ms`

So on this fixture the non-streaming packed-object path is slightly slower, about 2%, but not by a large margin.

Byron:
Thanks! Please note that I am a bit allergic against PRs that appear fully automated. This translates to a lot of time for me and possibly very little for the author.

What's the story related to `gix-filter`? Right now this also isn't streaming, and to checkout or checkin/track a file one will have to run them. That is, any worktree interaction both from or to Git objects may require a filter invocation.

willstott101:
In the issue here: https://github.com/GitoxideLabs/gitoxide/issues/1595 @Byron you mentioned that you'd prefer streaming to fail if it wasn't possible rather than the use something like `Cursor<Vec<u8>>` - but it seems like that's a code path in this PR.

I suppose the context is different. If the goal is for gitoxide to incrementally be able to improve it's memory usage by reducing how often entire blobs must be loaded into memory then I prefer this invisible `Vec` approach. But I suppose if the goal is for users of the library to request a file stream for server streaming like in my old issue then knowing if the stream was possible or not would be helpful.

I would be happy to help review and extend this PR, it's still been in the back of my mind since I made that issue.

My inclination would be to make this API surface available similar to it's current state, and then smaller PRs/improvements can be made to consume the streaming API surface within gix - for working directory stuff like filters support - as well as the underlying streaming mechanics - to stream more efficiently in more scenarios.

Does that sound reasonable? Will offer an actual review later.

Byron:
Hi @mjgil, are you still interested in this? The PR looks fully automated, and the recently changed contribution guidelines require disclosure, so I think it's fair to just reboot this in a new PR, agents tend to adhere to the guidelines and self-disclose (also in the commits).

Hi @willstott101, sorry for the late response.

> In the issue here: https://github.com/GitoxideLabs/gitoxide/issues/1595 [Sebastian Thiel (@Byron)](https://github.com/Byron) you mentioned that you'd prefer streaming to fail if it wasn't possible rather than the use something like Cursor<Vec<u8>> - but it seems like that's a code path in this PR.

That issue seems quite old and today I'd also think that streaming shouldn't fail if it can also not stream, if the object is deltified for instance.

> I would be happy to help review and extend this PR, it's still been in the back of my mind since I made that issue.

See my comment above, but in general it makes sense to re-generate such attempts, with most recent models, etc.

PR #2532, opened 2026-04-21
Use `GIX_TEST_FIXTURE_HASH` for `gix-diff`

cruessler:
This is a follow-up to #2497 which made tests in `gix-diff-tests` run with `GIX_TEST_FIXTURE_HASH=sha256`.

This PR adds:
- Remove `profile.release` section in `gix-imara-diff`
  This fixes the following warning:

  ```
  warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
  package:   .../gix-imara-diff/Cargo.toml
  workspace: .../Cargo.toml
  ```

  There’s also a commented out `[[bench]]` section in `gix-imara-diff/Cargo.toml` that could be removed as well.
- Run `gix-diff-tests` with `GIX_TEST_FIXTURE_HASH=sha1` as well (before, they were only run with `GIX_TEST_FIXTURE_HASH=sha256`)
- Use `GIX_TEST_FIXTURE_HASH` for `gix-diff` as well (as opposed to `gix-diff-tests` which is a separate crate)

This is part of #281

### DIFF
--- a/gix-diff/tests/diff/blob/platform.rs
+++ b/gix-diff/tests/diff/blob/platform.rs
@@ -48,7 +48,11 @@ fn resources_of_worktree_and_odb_and_check_link() -> crate::Result {
     assert_eq!(new.data.as_slice().expect("present").as_bstr(), a_content);
     assert_eq!(new.driver_index, Some(0));
     assert_eq!(new.mode, EntryKind::BlobExecutable);
-    assert_eq!(new.id, hex_to_id("4c469b6c8c4486fdc9ded9d597d8f6816a455707"));
+    let new_id = hex_to_id(
+        "4c469b6c8c4486fdc9ded9d597d8f6816a455707",
+        "061557fb6e67e1be5d41f8e83962699fce2a7d168ccfb2b5743d27ab06038da8",
+    );
+    assert_eq!(new.id, new_id);
     assert_eq!(new.rela_path, "a", "location is kept directly as provided");
 
     let out = platform.prepare_diff()?;
@@ -70,7 +74,14 @@ fn resources_of_worktree_and_odb_and_check_link() -> crate::Result {
             2,
             3
         )),
-        format!("{}test a <tmp-path> 0000000000000000000000000000000000000000 100644 <tmp-path> 4c469b6c8c4486fdc9ded9d597d8f6816a455707 100755", if !cfg!(windows) { "GIT_DIFF_PATH_COUNTER=3 GIT_DIFF_PATH_TOTAL=3 GIT_DIR=. " } else { Default::default() }),
+        format!(
+            "{}test a <tmp-path> 0000000000000000000000000000000000000000 100644 <tmp-path> {new_id} 100755",
+            if !cfg!(windows) {
+                "GIT_DIFF_PATH_COUNTER=3 GIT_DIFF_PATH_TOTAL=3 GIT_DIR=. "
+            } else {
+                Default::default()
+            }
+        ),
         "in this case, there is no rename-to field as last argument, it's based on the resource paths being different"
     );
 
@@ -95,7 +106,7 @@ fn resources_of_worktree_and_odb_and_check_link() -> crate::Result {
         a_content,
         "despite the same content"
     );
-    assert_eq!(new.id, hex_to_id("4c469b6c8c4486fdc9ded9d597d8f6816a455707"));
+    assert_eq!(new.id, new_id);
     assert_eq!(new.rela_path, "a");
 
     let out = platform.prepare_diff()?;
@@ -117,7 +128,14 @@ fn resources_of_worktree_and_odb_and_check_link() -> crate::Result {
             0,
             1
         )),
-        format!("{}test a <tmp-path> 0000000000000000000000000000000000000000 100644 <tmp-path> 4c469b6c8c4486fdc9ded9d597d8f6816a455707 120000", if !cfg!(windows) { r#"GIT_DIFF_PATH_COUNTER=1 GIT_DIFF_PATH_TOTAL=1 GIT_DIR=. "# } else { Default::default() }),
+        format!(
+            "{}test a <tmp-path> 0000000000000000000000000000000000000000 100644 <tmp-path> {new_id} 120000",
+            if !cfg!(windows) {
+                r#"GIT_DIFF_PATH_COUNTER=1 GIT_DIFF_PATH_TOTAL=1 GIT_DIR=. "#
+            } else {
+                Default::default()
+            }
+        ),
         "Also obvious that symlinks are definitely special, but it's what git does as well"
     );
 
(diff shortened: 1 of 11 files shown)


Byron:
Thanks a lot! I am looking forward to merge this!

Not having seen the warning really shows I have to be more mindful about my workflow. Apparently, it was only the agent who ran this, neatly hiding it from my view 🤦‍♂️.

cruessler:
I’ll mark the PR as ready for review now! My main focus was on making the tests pass with as few changes as possible, so I did not change the rewrite tracker in this iteration. If you feel like you want to take this as an opportunity to refactor, go ahead! 😀

## Decisions that probably need to be double-checked

- I’ve replaced `assert_hash_agnostic_eq!` by `pretty_assertions::assert_eq!` because the former was producing snapshots that depended on how hashes sort alphabetically, and that order is not necessarily the same for SHA-1 and SHA-256. Because of that, there’s now some branching based on the current hash, so slightly different sets of expectations run.
- I changed `hex_to_id` to require a hardcoded hash of each type. If I remember correctly, before that change, the actual hash did not matter in snapshots because it would be normalized anyway. That felt less safe than actually hardcoding the expected hashes, so I went with the latter. Plenty of opportunities for more refactoring, though. 😀
- Some of the assertions have, as their message in case of failure, strings that contain SHA-1 hashes. I did not touch any of them as changes here probably depend on how the two points above are addressed, with the exception of some that were already produced through `format!`.

gix-diff/tests/diff/index.rs
```diff
@@ -328,21 +328,24 @@ fn renames_by_similarity_with_limit() -> crate::Result {
...
+        ),
+        crate::hex_to_id(
+            "3bb459b831ea471b9cd1cbb7c6d54a74251a711b",
+            "19ebb6b2c2f3a64e6578013f680ec39330ce158af5977a1b17be0d551185fbab",
+        ),
+        crate::hex_to_id(
+            "0a805f8e02d72bd354c1f00607906de2e49e00d6",
+            "6271a9ad76b692e75a96d260cf02c4cd89e1d2071256447ff50e8a8f443299b4",
+        ),
+    ];
```
Byron, line 348: Loving this! 🙏

gix-diff/tests/diff/main.rs
```diff
@@ -121,16 +125,6 @@ fn assert_hash_agnostic_patch_eq(actual: &str, expected: &str) {
...
-macro_rules! assert_hash_agnostic_eq {
-    ($left:expr, $right:expr $(, $($arg:tt)+)?) => {{
-        pretty_assertions::assert_eq!(
-            crate::normalize_debug_snapshot(&($left)),
-            crate::normalize_debug_snapshot(&($right))
-            $(, $($arg)+)?
-        );
-    }};
-}
-
```
Byron, line 133: 🙏

gix-diff/tests/diff/rewrites/mod.rs
```diff
@@ -37,28 +39,28 @@ impl gix_diff::rewrites::tracker::Change for Change {
     }
 }
 
-const NULL_ID: gix_hash::ObjectId = gix_hash::Kind::Sha1.null();
+static NULL_ID: LazyLock<gix_hash::ObjectId> = LazyLock::new(|| crate::fixture_hash_kind().null());
```
Byron, line 42: I like `LazyLock`, something we can use more in tests in particular.

I'd actually see it in `fixture_hash_kind()` so we avoid accessing the environment each time.
It would probably be best in `gix-testtools`, so the underlying function does the Lazy update on first access to store the value globally.
But then again, I think environment access is also resonably fast, one lock, one hashmap, so probably nothing to think about.

gix-diff/tests/diff/rewrites/tracker.rs
```diff
@@ -151,7 +151,10 @@ fn copy_by_id() -> crate::Result {
         let out = util::assert_emit_with_objects(
             &mut track,
             |dst, src| {
-                let id = hex_to_id("2e65efe2a145dda7ee51d1741299f848e5bf752e");
+                let id = hex_to_id(
+                    "2e65efe2a145dda7ee51d1741299f848e5bf752e",
+                    "eb337bcee2061c5313c9a1392116b6c76039e9e30d71467ae359b36277e17dc7",
+                );
```
Byron, line 157: When seeing this followed by the next hunk that also maps `2e65efe2a145dda7ee51d1741299f848e5bf752e`, I remember that in another crate Codex added the mapping directly in `hex_to_id` with a simple hardcoded mapping. This was the most unintrusive solution there as there were only 10'ish different hashes in the whole test-suite anyway.

Here it's 24 though.

Byron (approved):
Thanks so much for the cleanup!

This is the first PR that I used my new and improved review & refactor workflow on. I think it's working, too :). (mostly, actually, but maybe there is a way to fix the last minor inconvenience)

gix-diff/tests/diff/main.rs
```diff
@@ -180,7 +180,11 @@ mod util {
     impl ObjectDb {
         /// Insert `data` and return its hash. That can be used to find it again.
         pub fn insert(&mut self, data: &str) -> Result<gix_hash::ObjectId, Error> {
-            let id = gix_object::compute_hash(gix_hash::Kind::Sha1, gix_object::Kind::Blob, data.as_bytes())?;
+            let id = gix_object::compute_hash(
+                gix_testtools::hash_kind_from_env().unwrap_or_default(),
+                gix_object::Kind::Blob,
```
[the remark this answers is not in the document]
Byron: Don't think so.

gix-diff/tests/diff/blob/platform.rs
```diff
@@ -70,7 +74,14 @@ fn resources_of_worktree_and_odb_and_check_link() -> crate::Result {
...
         )),
-        format!("{}test a <tmp-path> 0000000000000000000000000000000000000000 100644 <tmp-path> 4c469b6c8c4486fdc9ded9d597d8f6816a455707 100755", if !cfg!(windows) { "GIT_DIFF_PATH_COUNTER=3 GIT_DIFF_PATH_TOTAL=3 GIT_DIR=. " } else { Default::default() }),
+        format!(
+            "{}test a <tmp-path> 0000000000000000000000000000000000000000 100644 <tmp-path> {new_id} 100755",
+            if !cfg!(windows) {
+                "GIT_DIFF_PATH_COUNTER=3 GIT_DIFF_PATH_TOTAL=3 GIT_DIR=. "
+            } else {
+                ""
+            }
+        ),
```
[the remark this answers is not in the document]
Byron: That's Ok

gix-diff/tests/diff/rewrites/tracker.rs
cruessler: That’s a great suggestion, thanks! I’m going to try this with `gix-traverse`. I already discovered `git cat-file --batch-check --batch-all-objects` with makes it quite straightforward to create such a mapping.

PR #2536, opened 2026-04-22
Myers performance and more tests

Byron:
With a pathological testcase it turned out that Git is 1.7x faster than `gix`,
at least in this case.

This PR investigates ways to become faster, mostly by picking lower-hanging fruit.

This is a backburner task that isn't particularly important just yet.

### Tasks

* [x] let it run
* [ ] decide if it's worth taking the risk of keeping the improvements if they are not substantial.
* [ ] put `it` utilities into separate commit, and/or throw them away.
* [ ] refackiew

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3146,8 +3146,10 @@ name = "internal-tools"
 version = "0.0.0"
 dependencies = [
  "anyhow",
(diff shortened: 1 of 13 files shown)


Byron:
Slippery slope, let's not go here.

PR #2538, opened 2026-04-23
Git to Gix Ralph Wiggum Progress

Mbd06b:
Hello Sebastian, or do you prefer @Byron?   

I'd call this a fan letter, not really a PR.  I've got two branches, and this is my... get it done branch. 

This is really more of a ... hi, love what you're doing here and am seeing if I can brute force a gix branch with git feature parity using Claude Code,  and ralph-wiggum loops on each git command, with demonstrated testable parity following the testing framework your using, and scaffolding the problem

I don't know how far I'll get towards reaching git parity, but the goal is if I can get gix feature parity with git, I can start dogfooding this tool. If this branch succeeds at that, then then the idea will become to create narrowly-scoped PRs, from the get it done branch, to something that is much more timid and respectful of the contributing guidelines.

In the meantime, you COULD in theory, use something from this branch. As a courtesy, I just wanted to let you know what shenanigans I was up to, incase you wanted anything from this sooner rather than later.  (as to be efficient, in our use of time and tokens, in the pursuit of any mutual goals here)

Lastly, what is "[brit](https://github.com/ethosengine/brit)"?  Well, it's thought to be a fit-for-purpose tooling for the [Elohim Protocol](https://github.com/ethosengine/elohim) network, a moonshot vision to subsume the current extractive internet with digital public, human-scale infrastructure which forces a richer multi-dimensional view of information primitives: provenance, economic events, governance context, and content-addressed links that know where your code is running, and an unfinished gix is sort of an unfinished dependency in regards to what I really want to do, which is  redesigning a rust-based version control protocol tooling for a DAG-CBOR, libp2p, DHT vision for the shape of a future I'd like to see considered.  One where grandma can keep her family photos on a network of peers, on an uncapturable system of compute and relationships, that are as reliable and as easy to access as any Platform-based social media library is today,  but in such an architecture that by design respects her embodied life and inherent human dignity in relationship with others.   

So in the meantime. That's what I'm up to.  (note as of drafting this, push parity is theoretically delivered here according to the gix-steward agent I designed). 

Claude is doing all the dev, I'm the  driver ( I attest that I wrote this message explicitly w/o any AI edits. )
 
With love and respect, 

Matthew


Mbd06b:
Here's where we're at:  fetch is going right now. 

` ok gix push passed parity, what's next plumbing or another ralph loop?                                                                                                                             
                                                                                                                                                                                                     
● Another ralph — specifically fetch. Strategic logic:                                                                                                                                               
                                                                                                                                                                                                   
  Why fetch over plumbing now                                                                                                                                                                        
  - The fetch-pack substrate already exists (Task 5.1 mirrored that pattern). Unlike push, no substrate sprint needed first. Expect iter-0 → iter-~25 to close it straight.
  - fetch + push together make gix network-functional end-to-end. That's a bigger unlock than any single plumbing item.                                                                              
  - You'll learn what the parity loop looks like against a mostly-working command (vs. push, which was substrate-blocked). That calibrates expectations for log, diff, status, etc.
                                                                                                                                                                                                     
  Ranking of what to ralph next (in the 11-daily-driver scope)                                                                                                                                       
  1. fetch — substrate exists; complements push; likely closes quickly.                                                                                                                              
  2. log — impl exists; no parity scaffold yet; probably many flag gaps but each is a cheap row.                                                                                                     
  3. status — probably small surface, fast win.                                                                                                                                                      
  4. diff — impl exists (223 LOC); parity scope is real work but no substrate gap.                                                                                                                   
  5. clone — network command, shares plumbing with fetch; natural follow-up to fetch.                                                                                                                
  6. blame, branch, tag, cat — smaller but still useful.                                                                                                                                             
  7. commit — partially ralph-able but hook rows will hit the gix-hook stub. Save until hooks are done.                                                                                              
                                                                                                                                                                                                     
  When to switch from ralph to plumbing                                                                                                                                                              
  Turn to foundational work when:                                                                                                                                                                    
  - Ralph hits commit and needs hooks (gix-hook crate → unblocks commit + pre-push).                                                                                                                 
  - You want add/rm/mv (index mutation + tree-from-index → unblocks staging and also closes the last commit gap).                                                                                    
  - 7.1-followup pack-copy optimization becomes visible (large-repo push benchmarks).                                                                                                                
                                                                                                                                                                                                     
  Quick kickoff:                                                                                                                                                                                     
  CMD=fetch                                                                                                                                                                                          
  /ralph-loop "$(sed "s/\$CMD/$CMD/g" etc/parity/prompt.md)" \                                                                                                                                       
    --max-iterations 50 \                                     
    --completion-promise "PARITY-git-$CMD"                                                                                                                                                           
                                          
  Expected path: scaffold iter-1 creates tests/journey/parity/fetch.sh + Clap stub, then ~25 row closures. Foundational interruption unlikely for pure fetch.'

Byron:
Hi @Mbd06b, thanks so much for sharing!

This reminds me very much of https://schacon.github.io/grit/, but started from `gix` instead of a green field. 

`brit` sounds great too, and its "information primitives" reminded me a bit of [`git-meta`](https://git-meta.com), which is also very recent and mostly generated.

I agree with the general approach and am happy to host this PR here. Hence you can keep it running and see when it claims parity. From there we can evaluate how to divide and conquer.

Something I'd be interested in is a central prompt, if such a thing exists. Because getting the spirit of this PR and its prompt right probably goes a long way of aligning with what I would find acceptable in the first place. Knowing it isn't going to be used to criticize or reject, but merely to have a chance to amend it with bits that I would find important.

In any case, thanks for the tokens and happy looping!

Byron:
**I thought I'd intervene early so we have a chance to ever merging this and it doesn't go off in the wrong direction for too long.**

Just one thought about this. In theory, the changes to `gitoxide-core` that give you a `git` command-like surface is something we could merge even without great review: having more code that exercises the existing crates (without changes) is useful.

However, I really want these clearly marked and show up as generated, so you would have to change the author to the model and version that was generating it, and if you want, make you a co-author, just so we know the human behind the machine.

Also, it would be required if `gix` isn't the target for these changes, but instead that there is a new binary that clearly states what it is, that can be in a separate crate that has its own core. So `gitoxide-core` really is only for the `gix` CLI.

**Also, the following changes would be needed:**

* do not litter `Claude.md` everywhere.
* do not add `/docs`, use `etc/**` for that
* use a new binary, let's call it `git` in the `gix-git` crate, clearly stating it should be a stand-in for C-`git`. Open to suggestions here as well, but we can keep it this simple and not bike-shed names too much.
* let it first do all the commands that it can implement today, **without touching existing crates**
* author: make it claude + version, feel free to be the co-author

**This way, we have a Git-compatible CLI that can do what `gitoxide` can do today. This should be merged early as checkpoint.**

Then each newly added command capability that needs a change to plumbing goes into its own commit so it can be reviewed individually. This would be a new PR that effectively is this one, but with an organization that makes it easier to review it commit by commit. Oh, and each of these commits definitely has to pass CI. Let's be sure the quality doesn't slip. This also slows it down which is a good thing 😅.

The ping to @Mbd06b just to be sure this message is read.


PS: Maybe find me on the GitButler Discord server so we have a more realtime comms channel when needed. I am willing to take this on if I can count on your support, but this needs closer collaboration I think.

Byron:
And one more thing. I think we should put validation first and foremost. So plan around running the git test suite against the new binary that is produced here.

Ideally, it's able to only consider the test cases that pertain to particular arguments that it implements. That's probably the hardest part. But ideally, there is a way to see that number of passing tests go up even while not everything is fully implemented.

Also, I think it's fine to, even when lacking certain APIs, work around that and implement certain things directly in the core of the new binary, so it can serve as an example on what to, one day, port over into the plumbing. That way it might be possible to get to a fully working and fully compatible Git CLI substitute, as measured by Git's own test-suite, even without touching any plumping.

All this we should definitely talk through on day on Discord.

Byron:
@Mbd06b anybody home?

Byron:
Closing due to lack of activity, without replies neither here nor by email.

PR #2549, opened 2026-04-28
Avoid duplicate hash computation during object creation

### LINKED ISSUE #2516: Add API so that `Repository` write functions only compute hash once
### Summary 💡

`write_blob` looks like it computes the hash twice. Could a variant of `write_buf`/`write_stream` be added that avoids the second hash computation?

To detail, on `Repository`, I believe [`write_blob`](https://github.com/GitoxideLabs/gitoxide/blob/8f091d108cd75371be2ed9de6e81f785cda53d92/gix/src/repository/object.rs#L290-L301) will:

1. Calls `compute_hash` to produce a hash.
2. Call `exists` with the hash for a possible early exit.
3. Call [`write_buf`](https://github.com/GitoxideLabs/gitoxide/blob/8f091d108cd75371be2ed9de6e81f785cda53d92/gix-object/src/traits/mod.rs#L15-L17), which calls [`write_stream`](https://github.com/GitoxideLabs/gitoxide/blob/8f091d108cd75371be2ed9de6e81f785cda53d92/gix-odb/src/sink.rs#L23-L57), which again computes a hash.

The hashes computed by `compute_hash` (in `write_blob`) and `write_stream` should be identical, implying `write_stream` could avoid a hash computation and instead rely on the earlier value.

### Motivation 🔦

I was looking at this while examining the `write_blob` implementation as part of digging into performance for https://github.com/jj-vcs/jj/pull/9304, and I'd probably use it in the related code. I think it'd be an incremental performance improvement. If you're supportive of this idea, I'd be happy to contribute a PR for it.

Byron:
Fixes  #2516

### Tasks

* [x] refackiew
* [x] commits with decent changelog entries

### DIFF
--- a/gix-object/src/traits/mod.rs
+++ b/gix-object/src/traits/mod.rs
@@ -15,6 +15,16 @@ pub trait Write {
     fn write_buf(&self, object: crate::Kind, mut from: &[u8]) -> Result<gix_hash::ObjectId, crate::write::Error> {
         self.write_stream(object, from.len() as u64, &mut from)
     }
+    /// As [`write_buf`](Write::write_buf), but the object `id` has already been computed by the caller.
+    ///
+    /// Implementations may trust the given `id` and avoid computing it again. Callers must make sure `id` matches
+    /// the provided `object` and `from` bytes.
+    fn write_buf_with_known_id(
+        &self,
+        object: crate::Kind,
+        from: &[u8],
+        id: gix_hash::ObjectId,
+    ) -> Result<gix_hash::ObjectId, crate::write::Error>;
     /// As [`write`](Write::write), but takes an input stream.
     /// This is commonly used for writing blobs directly without reading them to memory first.
     fn write_stream(
@@ -22,6 +32,17 @@ pub trait Write {
         kind: crate::Kind,
         size: u64,
         from: &mut dyn io::Read,
+    ) -> Result<gix_hash::ObjectId, crate::write::Error>;
+    /// As [`write_stream`](Write::write_stream), but the object `id` has already been computed by the caller.
+    ///
+    /// Implementations may trust the given `id` and avoid computing it again. Callers must make sure `id` matches
+    /// the provided `kind`, `size` and stream contents.
+    fn write_stream_with_known_id(
+        &self,
+        kind: crate::Kind,
+        size: u64,
+        from: &mut dyn io::Read,
+        id: gix_hash::ObjectId,
     ) -> Result<gix_hash::ObjectId, crate::write::Error>;
 }
 
--- a/gitoxide-core/src/pack/explode.rs
+++ b/gitoxide-core/src/pack/explode.rs
@@ -105,6 +105,18 @@ impl gix::objs::Write for OutputWriter {
         }
     }
 
+    fn write_buf_with_known_id(
+        &self,
+        kind: object::Kind,
+        from: &[u8],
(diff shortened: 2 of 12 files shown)


gix-object/src/traits/mod.rs
```diff
@@ -15,6 +15,19 @@ pub trait Write {
...
+    /// Implementations may trust the given `id` and avoid computing it again. Callers must make sure `id` matches
+    /// the provided `object` and `from` bytes.
+    fn write_buf_with_known_id(
+        &self,
+        id: gix_hash::ObjectId,
+        object: crate::Kind,
+        from: &[u8],
+    ) -> Result<gix_hash::ObjectId, crate::write::Error> {
+        let _ = id;
+        self.write_buf(object, from)
```
jonmeow: Should this mirror how `write_buf` calls `write_stream`? i.e.:

```suggestion
        self.write_stream_with_known_id(id, object, from.len() as u64, &mut from)
```

It looks like places like cache.rs omit `write_buf`, so this kind of call might let them omit `write_buf_with_known_id` similarly (though looking at the loose/write.rs changes which *do* provide `write_buf`, perhaps cache.rs should actually be forwarding `write_buf`?)

gix-object/src/traits/mod.rs
```diff
@@ -23,6 +36,20 @@ pub trait Write {
...
+    /// Implementations may trust the given `id` and avoid computing it again. Callers must make sure `id` matches
+    /// the provided `kind`, `size` and stream contents.
+    fn write_stream_with_known_id(
+        &self,
+        id: gix_hash::ObjectId,
+        kind: crate::Kind,
+        size: u64,
+        from: &mut dyn io::Read,
+    ) -> Result<gix_hash::ObjectId, crate::write::Error> {
+        let _ = id;
```
jonmeow: Is it right to be discarding the id here (versus not providing a default implementation)? Maybe this is appropriate for backwards compatibility; if that's the intent, maybe worth a comment?

```suggestion
        // Provide a default implementation which discards the id.
        let _ = id;
```

jonmeow (commented):
Thanks! This looks good to me, just a couple questions about the Write trait changes (TBH I'm really just comparing with the non-known_id variants)

Byron:
@jonmeow Thanks so much for taking a look!

I haven't looked at the code at all yet, but when I do I will particularly look at the places you are highlighting.

gix-object/src/traits/mod.rs
Byron: Removed this default impl.

gix-object/src/traits/mod.rs
Byron: The forwarding was removed.

gix-object/src/traits/mod.rs
```diff
@@ -23,6 +33,17 @@ pub trait Write {
...
+    ///
+    /// Implementations may trust the given `id` and avoid computing it again. Callers must make sure `id` matches
+    /// the provided `kind`, `size` and stream contents.
+    fn write_stream_with_known_id(
+        &self,
+        kind: crate::Kind,
+        size: u64,
+        from: &mut dyn io::Read,
+        id: gix_hash::ObjectId,
+    ) -> Result<gix_hash::ObjectId, crate::write::Error>;
```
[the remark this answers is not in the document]
Byron: Totally breaking, and indicated like this.

PR #2560, opened 2026-04-30
index::State::to_tree()

Byron:
Allow writing back a tree in the hopes that it will be faster than what `git2` offers.
The reason that's slow is also that we don't currently maintain the tree-cache when editing the index,
but instead remove it.

However, this is also something that could be added.

Related to https://github.com/gitbutlerapp/gitbutler/pull/13566 .

### Tasks

* [ ] refackiew
* [ ] per-path tree cache invalidation


Byron:

PR #2563, opened 2026-04-30
Add a revision option to gix clone

theoparis:
I implemented this with the help of gemini cli. I'm not sure if this is the best approach (since gix still has a `--ref` flag which might be confusing), but it does seem to work with something like:
`./target/debug/gix clone --revision 894a8ced1b09ff22ea0b8826fcccfbb73834896f https://github.com/zlib-ng/zlib-ng ~/zlib-ng`
which should result in `Bump Google Benchmark to v1.9.5` being the latest commit in git log.

### DIFF
--- a/gitoxide-core/src/repository/clone.rs
+++ b/gitoxide-core/src/repository/clone.rs
@@ -1,4 +1,5 @@
 use crate::OutputFormat;
+use gix::bstr::BString;
 
 pub struct Options {
     pub format: OutputFormat,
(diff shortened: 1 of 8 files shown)


Byron:
I'm setting this to draft until the auto review comments have been addressed, by rejecting or accepting them.
At least it's not a thousand lines of code, so I guess the solution is f it found isn't too shabby.

I also want to remind everyone that the `gix` CLI is a developer tool. There are no tests beyond what's needed for it to take part in journey tests/E2E, there is no guarantee it will keep working the way it used to, it has no stable interface.

Byron:
There hasn't been any activity in a while which is why I prefer to close this PR to get it off my lists.
I do, however, encourage the OP to bring it back up once there's some time to work on it. 

PR #2574, opened 2026-05-04
Use `GIX_TEST_FIXTURE_HASH` for `gix-status-tests`

cruessler:
This PR makes `gix-status-tests` run with `GIX_TEST_FIXTURE_HASH=sha256`. For mapping SHA-1 to SHA-256 hashes, it uses the same pattern as #2573. Also, there’s some similar duplication surrounding the uses of `gix_odb::at_opts(…)`. Apart from that, not a lot of code had to be changed.

This is part of #281.


Byron:

PR #2575, opened 2026-05-05
Document why each fixture archive is .gitignored

### LINKED ISSUE #2316: Indicating why each fixture archive in `.gitignore` is untracked
#### Background

Many generated archives of Git repository fixtures produced by `gix-testtools` running fixture scritps are committed, which allows them to be reused on systems that may be unable to produce them, and also facilitates more robust testing on systems that *can* produce them, since we can verify that the test works both with a fixture generated on another machine and a fixture generated on the current machine.

For tests whose associated fixtures intentionally incorporate information about the current system, it would be incorrect to track the archives. In some other cases, it would be ideal to track the archives but we can't yet do so because of limitations in the test suite or fixture scripts that could eventually be overcome. Therefore various archives are listed in `.gitignore` files.

#### Proposal

I think we should somehow keep track of why each fixture in any `.gitignore` file is listed there, because:

- Some may not work across systems due to bugs in the fixture, other test code, or even in the code under test. Keeping track of what we know about why a fixture is in `.gitignore` would help reveal this.
- For others, it may be possible to impove them so they can be committed. This would lead to a more robust test suite, for the reason described above.

#### Conceptual clarity

Besides the points above, I think documenting this would make it easier to reason about the relationship between repositories maintained in fixtures and the test code that uses those repositories. (This is a lesser rationale than the above, but it is also what motivated me to suggest that we do this.)

The fixture accompanying the regression test for #2227 introduced in #2229 is in `.gitignore`. The correct configuration that should be written when doing a shallow clones depends only on the repository being cloned, but I think this was made harder to reason about by the test repository always being created on the machine running the test.

That fixture is reused across various tests and there may also be good reasons for it to be created anew each time. Furthermore, in this case I don't think using a committed archive would have caused the test to fail in a way that would have revealed #2311. (If it would have, then that would indicate a bug in `gix-testtools` not adequately sanitizing/resetting the configuration when running the fixture scripts--though the ability to reveal such bugs is another benefit.) Rather, it's only the effect of making clear what is and isn't expected to vary based on what system generates the fixtures that I think might've helped.

#### Plan

If there's no objection, then I'll eventually try to do this, though I'm not sure when, which is why I'm opening this issue for it so I don't forget or in case someone else wants to take it on.

SarthakB11:
# Document why each fixture archive is `.gitignore`d

Fixes https://github.com/GitoxideLabs/gitoxide/issues/2316

## What changed

For every `tests/fixtures/generated-archives/.gitignore` in the workspace, an
inline `#`-comment now sits above each listed `*.tar` entry explaining why that
archive is not committed. The reason for each line was derived from reading the
corresponding `*.sh` fixture script (and, where the script alone wasn't
conclusive, the surrounding test code).

Only `.gitignore` files are touched. No fixture scripts, archives, or test code
are modified.

## Recurring categories

The reasons cluster into a small set:

- POSIX-only filesystem features that don't unpack on Windows without elevated
  privileges, or aren't representable at all: symlinks (`ln -s`), FIFOs
  (`mkfifo`), restrictive modes (`chmod 000`, `chmod -x`).
- Host-absolute paths baked into git state: `core.worktree`,
  `.git/objects/info/alternates`, linked-worktree `gitdir` pointers,
  `.git/modules/<name>/config`, `--shared file://$PWD/...` clones.
- Baselines captured from the host's `git` binary (`git check-attr`,
  `git check-ignore`, `git blame --porcelain`, `git fetch --refmap= --dry-run`,
  `git fetch-pack --diag-url`). Their textual output is sensitive to the git
  version, so a committed snapshot would drift from the binary running the
  test.
- Filesystem case-sensitivity: `make_ignorecase_collisions` produces paths
  differing only by case, which fail to extract on default macOS / Windows
  volumes.
- Time-range-sensitive metadata: `gix-index/file_metadata.sh` writes pre-1970
  and post-2038 mtimes whose representable range depends on the host
  filesystem.
- Experimental / placeholder fixtures (`gix-diff/make_diff_for_sliders_repo*`),
  where the existing maintainer comment was preserved and slightly expanded.

## Files changed (21)

`tests/fixtures/generated-archives/.gitignore` under: `gix-archive`,
`gix-attributes`, `gix-blame`, `gix-command`, `gix-config`, `gix-diff`,
`gix-dir`, `gix-discover`, `gix-filter`, `gix-fsck`, `gix-ignore`, `gix-index`,
`gix-odb`, `gix-ref`, `gix-refspec`, `gix-status`, `gix-url`, `gix-worktree`,
`gix-worktree-state`, `gix-worktree-stream`, `gix`.

## Notes for review

A handful of archives don't have an obviously host-dependent script, so the
comment is the most plausible reason from the surrounding test code rather than
a definitive one:

- `gix-config/make_config_repo.tar` (env-var-driven config paths into the
  generated worktree)
- `gix-command/win_path_lookup.tar` (the rationale here is the weakest of the
  bunch; if it turns out the archive *can* be committed, removing the
  exclusion is now a one-line follow-up)
- `gix-blame/make_blame_repo*.tar` (version-sensitive `git blame --porcelain`
  baselines under `.git/*.baseline`)

If any of these should be re-tracked instead of explained, that's a cheaper
change to make on top of this PR than to gate it on.

---

🤖 Drafted by an AI agent (Claude Code) on behalf of the author, per
`CONTRIBUTING.md`'s AI-disclosure guidance. Each comment was checked against
the corresponding fixture script and the relevant test code before commit.


Byron (approved):
Thanks a lot!

I read through all the comments, made a couple of changes, and think that while not all may be 100% correct, I don't know any better by now and these comments are better than nothing.

They also set the expectation that future additions to the .gitignore files will require comments, which is the right thing to do.

PR #2583, opened 2026-05-07
Improve comment in object/commit.rs

tisonkun:
cc @Byron 

`Commit` now has an `author` method as well:

```rust
    /// Return the commits author, with surrounding whitespace trimmed.
    pub fn author(&self) -> Result<gix_actor::SignatureRef<'_>, gix_object::decode::Error> {
        gix_object::CommitRefIter::from_bytes(&self.data, self.id.kind())
            .author()
            .map(|s| s.trim())
    }
```

Seems we don't need to `decode` first? Or else the API looks unintuitive to understand.


Byron (approved):
Thanks, good catch!

PR #2592, opened 2026-05-09
gix-pack: dynamic work queue for parallel delta resolution

### LINKED ISSUE #2424: `phpstan/phpstan` pack resolves with a single thread only
### Discussed in https://github.com/GitoxideLabs/gitoxide/discussions/2422

<sup>Originally posted by **homersimpsons** February  7, 2026</sup>
Today I `gix clone https://github.com/phpstan/phpstan.git` and it took ~20 minutes:
```sh
❯ gix clone https://github.com/phpstan/phpstan.git
 13:49:18 indexing done 100.7K objects in 850.33s (118.0 objects/s)             
 13:49:18 decompressing done 18.3GB in 850.33s (21.5MB/s)                       
 13:54:35     Resolving done 100.7K objects in 317.26s (317.0 objects/s)        
 13:54:35      Decoding done 160.2GB in 317.27s (504.9MB/s)                     
 13:54:35 writing index file done 3.3MB in 0.03s (130.2MB/s)                    
 13:54:35  create index file done 100.7K objects in 1167.72s (86.0 objects/s)   
 13:54:36          read pack done 6.9GB in 1167.94s (5.9MB/s)                   
 13:54:36           checkout done 786.0 files in 0.13s (6.0K files/s)           
 13:54:36            writing done 67.8MB in 0.13s (521.7MB/s)   
```

(note that  was on wifi and around 5MB/s for "half" of the download and then 30MB/s for the other "half")

I hence ran `git clone https://github.com/phpstan/phpstan.git` to compare the results:
```sh
❯ time git clone https://github.com/phpstan/phpstan.git
Cloning into 'phpstan'...
remote: Enumerating objects: 100714, done.
remote: Counting objects: 100% (15/15), done.
remote: Compressing objects: 100% (12/12), done.
remote: Total 100714 (delta 3), reused 8 (delta 3), pack-reused 100699 (from 1)
Receiving objects: 100% (100714/100714), 6.43 GiB | 21.10 MiB/s, done.
Resolving deltas: 100% (66589/66589), done.

real	7m6.327s
user	19m56.695s
sys	0m56.571s
```
</div>

vmandke:
Fixes #2424

Replaces: resolve::deltas
Modifies: resolve::delta_mt

Previously each thread was assigned a fixed subtree via in_parallel_with_slice, causing load imbalance
when one root had a deep chain of deltas. Thread synchronization also used a busy wait of poll_interval.
This PR rewrites deltas_mt to use a Workqueue with Condvar blocking, so all threads pull work evenly
regardless of tree shape.

---

- [ ] Feedback on Approach
- [ ] Minimal Test for the example in linked issue. (deeply chained deltas)

---
Timings 

--- gix main (in_parallel_with_slice) ---
/tmp/gix-main free pack index create --pack-path /Users/vmandke/phpstan.pack /tmp/bench2-main
index: bf450685c28fb13ad44fcafe90a5401b7b54141a
pack: 73555a206b95a6590d30e8671eacb5bd4ff55a70

real	2m36.613s
user	6m13.277s
sys	0m15.543s

--- gix fix  (work-stealing queue) ---
/tmp/gix-fix free pack index create --pack-path /Users/vmandke/phpstan.pack /tmp/bench2-fix
index: bf450685c28fb13ad44fcafe90a5401b7b54141a
pack: 73555a206b95a6590d30e8671eacb5bd4ff55a70

real	1m14.710s
user	6m43.032s
sys	0m10.444s

---

Profiling

### Gix on Main

Gix on Main

### Gix on Fix

Gix on Fix

### Git on same pack

Git on same pack

---

## AI Usage disclaimer

Used Claude Code for benchmark scripting, fixes and cleanup.
Also used for understanding both Gitoxide and Git codebases.

### DIFF
--- a/gix-pack/src/cache/delta/traverse/mod.rs
+++ b/gix-pack/src/cache/delta/traverse/mod.rs
@@ -1,7 +1,6 @@
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::AtomicBool;
 
 use gix_features::{
-    parallel::in_parallel_with_slice,
     progress::{self, DynNestedProgress, Progress},
     threading,
     threading::{Mutable, OwnShared},
@@ -139,47 +138,24 @@ where
         let (mut root_items, mut child_items_vec) = self.take_root_and_child();
         let child_items = ItemSliceSync::new(&mut child_items_vec);
         let child_items = &child_items;
-        in_parallel_with_slice(
-            &mut root_items,
-            thread_limit,
-            {
-                {
-                    let object_progress = object_progress.clone();
-                    move |thread_index| resolve::State {
-                        delta_bytes: Vec::<u8>::with_capacity(4096),
-                        fully_resolved_delta_bytes: Vec::<u8>::with_capacity(4096),
-                        progress: Box::new(
-                            threading::lock(&object_progress).add_child(format!("thread {thread_index}")),
-                        ),
-                        resolve: resolve.clone(),
-                        modify_base: inspect_object.clone(),
-                        child_items,
-                    }
-                }
-            },
-            {
-                move |node, state, threads_left, should_interrupt| {
-                    // SAFETY: This invariant is upheld since `child_items` and `node` come from the same Tree.
-                    // This means we can rely on Tree's invariant that node.children will be the only `children` array in
-                    // for nodes in this tree that will contain any of those children.
-                    #[allow(unsafe_code)]
-                    unsafe {
-                        resolve::deltas(
-                            object_counter.clone(),
-                            size_counter.clone(),
-                            node,
-                            state,
-                            resolve_data,
-                            object_hash.len_in_bytes(),
-                            threads_left,
-                            should_interrupt,
-                        )
-                    }
-                }
-            },
-            || (!should_interrupt.load(Ordering::Relaxed)).then(|| std::time::Duration::from_millis(50)),
-            |_| (),
-        )?;
+
+        // SAFETY: root_items and child_items come from the same Tree — the Tree invariant holds.
+        #[allow(unsafe_code)]
+        unsafe {
+            resolve::resolve_all_nodes(
+                &mut root_items,
+                child_items,
+                thread_limit,
+                object_counter,
+                size_counter,
+                &gix_features::progress::Discard,
+                resolve,
+                resolve_data,
+                inspect_object,
+                object_hash.len_in_bytes(),
+                should_interrupt,
+            )?;
+        }
 
         threading::lock(&object_progress).show_throughput(start);
         size_progress.show_throughput(start);
--- a/gix-pack/src/cache/delta/traverse/resolve.rs
+++ b/gix-pack/src/cache/delta/traverse/resolve.rs
@@ -1,9 +1,6 @@
-use std::{
-    collections::BTreeMap,
-    sync::atomic::{AtomicBool, AtomicIsize, Ordering},
-};
+use std::sync::atomic::{AtomicBool, Ordering};
 
-use gix_features::{progress::Progress, threading, zlib};
+use gix_features::{progress::Progress, zlib};
 
 use crate::{
     cache::delta::{
@@ -43,11 +40,6 @@ mod root {
     }
 
     impl<'a, T: Send> Node<'a, T> {
-        /// Returns the offset into the pack at which the `Node`s data is located.
-        pub fn offset(&self) -> u64 {
-            self.item.offset
-        }
-
         /// Returns the slice into the data pack at which the pack entry is located.
         pub fn entry_slice(&self) -> crate::data::EntryRange {
             self.item.offset..self.item.next_offset
@@ -56,11 +48,6 @@ mod root {
         /// Returns the node data associated with this node.
         pub fn data(&mut self) -> &mut T {
             &mut self.item.data
-        }
-
-        /// Returns true if this node has children, e.g. is not a leaf in the tree.
-        pub fn has_children(&self) -> bool {
-            !self.item.children().is_empty()
         }
 
         /// Transform this `Node` into an iterator over its children.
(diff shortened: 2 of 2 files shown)


vmandke:
edit:: updated commit message, as this is not a work stealing queue, rather a global work queue

Byron:
Thanks a lot!

Without looking at the code in depth, I just had it run benchmarks to see what the algorithm change does given different and somewhat extreme packs.

My takeaway is that the user time of the new algorithm is probably more similar to Git because it will use threads more evenly, but these threads duplicate more work as well.

So the idea behind the implementation is that the work is organized so that no duplicate work exists.
In the case of PHPstan this obviously costs much more memory, which is worth considering as a trade-off.

For now, however, I would be ok to use more memory if it's in the service to ultimately doing less work. To do that, the previous implementation already implemented something like a work stealing model, but apparently the heuristics didn't make it kick in for a PHPstan. Maybe this can be fixed, or maybe the algorithm behind that can be improved to allow incorporating threads not only per root, but let idle threads work together with other threads that still have work to do. 
After all, we already have the delta tree in memory, and it's no problem to split that up into smaller subtasks without letting threads overlap. 

If this works, then a follow up could deal with reducing the memory footprint. There should be a lot of potential, given that right now it keeps in memory all the bases in the delta tree, even though many of these at some point will not be used anymore. And from what we know, apparently this could halve the memory usage in case of PHPstan.

Please note that I put this PR back into draft while the details of the algorithm are still not figured out.

<details><summary>Codex Benchmark Results</summary>
<p>

## Pack Verification Performance Report

Compared this branch’s `target/release/gix` against mainline `gix` and Git.

Binaries:

- Mainline gix: `gix v0.53.0-100-g1322a36599`
- This branch: `gix v0.53.0-66-g60e081abe2`, HEAD `60e081abe2`
- Git: `git version 2.50.1 (Apple Git-155)`

Methodology:

- Built this branch with `cargo build --release --bin gix`.
- gix command: `gix free pack verify <idx>`.
- Git command: `git verify-pack -v <idx> >/dev/null`.
- Used `hyperfine --warmup 1 --min-runs 5`.
- For gix-vs-gix, ran benchmarks twice with command order reversed.
- Measured memory with `/usr/bin/time -lp`, using `maximum resident set size`.
- Machine was not fully isolated, so some runtime fluctuation is expected.

### Linux Fixture Pack

Pack index:

```text
/Users/byron/dev/github.com/GitoxideLabs/gitoxide/tests/fixtures/repos/linux.git/objects/pack/pack-3ee05b0f4e4c2cb59757c95c68e2d13c0a491289.idx
```

Pack size: about `1.4 GB`; index size: about `203 MB`.

#### Runtime

| Tool | Wall Time | User Time | System Time | Notes |
|---|---:|---:|---:|---|
| Mainline gix | `14.41 s` | `180.64 s` | `8.89 s` | average of two 5-run hyperfine passes |
| Branch gix | `18.23 s` | `182.56 s` | `41.66 s` | average of two 5-run hyperfine passes |
| Git | `51.040 s ± 0.393` | `183.831 s` | `51.318 s` | one 5-run hyperfine pass |

Detailed gix passes:

| Pass | Order | Mainline Wall/User/Sys | Branch Wall/User/Sys | Result |
|---|---:|---:|---:|---:|
| 1 | mainline then branch | `14.540 s ± 0.159` / `182.700 s` / `8.928 s` | `18.766 s ± 1.430` / `185.033 s` / `41.068 s` | mainline `1.29x ± 0.10` faster |
| 2 | branch then mainline | `14.286 s ± 0.275` / `178.579 s` / `8.850 s` | `17.698 s ± 1.811` / `180.095 s` / `42.249 s` | mainline `1.24x ± 0.13` faster |

On this pack, the branch is about `26%` slower than mainline gix, but still about `2.80x` faster than Git.

#### Peak RSS

| Tool | Run 1 RSS | Run 2 RSS | Avg RSS | Avg Real/User/Sys |
|---|---:|---:|---:|---:|
| Mainline gix | `3,541,352,448` B | `3,540,860,928` B | `3.30 GiB` | `13.84 s` / `173.79 s` / `8.90 s` |
| Branch gix | `3,220,176,896` B | `3,224,027,136` B | `3.00 GiB` | `16.23 s` / `172.89 s` / `44.43 s` |
| Git | `1,091,026,944` B | `1,109,950,464` B | `1.02 GiB` | `51.11 s` / `183.85 s` / `52.14 s` |

The branch uses about `303 MiB` less peak RSS than mainline gix, a reduction of about `9.0%`. Git uses substantially less RSS on this pack, but is much slower.

### PHPStan Pack

Pack index:

```text
/Users/byron/dev/github.com/phpstan/phpstan/.git/objects/pack/pack-fc5a93b6903e792e5c65ce2f8f02054775c35c94.idx
```

Pack size: about `6.4 GB`; index size: about `3.1 MB`.

#### Runtime

| Tool | Wall Time | User Time | System Time | Notes |
|---|---:|---:|---:|---|
| Mainline gix | `103.92 s` | `321.54 s` | `4.87 s` | average of two 5-run hyperfine passes |
| Branch gix | `25.44 s` | `361.45 s` | `4.21 s` | average of two 5-run hyperfine passes |
| Git | `73.553 s ± 0.440` | `338.737 s` | `2.688 s` | one 5-run hyperfine pass |

Detailed gix passes:

| Pass | Order | Mainline Wall/User/Sys | Branch Wall/User/Sys | Result |
|---|---:|---:|---:|---:|
| 1 | mainline then branch | `104.648 s ± 2.888` / `325.635 s` / `4.899 s` | `25.478 s ± 0.270` / `363.916 s` / `4.326 s` | branch `4.11x ± 0.12` faster |
| 2 | branch then mainline | `103.198 s ± 0.957` / `317.440 s` / `4.848 s` | `25.408 s ± 0.926` / `358.992 s` / `4.092 s` | branch `4.06x ± 0.15` faster |

On this pack, the branch is about `4.08x` faster than mainline gix and about `2.89x` faster than Git. Git is about `1.41x` faster than mainline gix here.

#### Peak RSS

| Tool | Run 1 RSS | Run 2 RSS | Avg RSS | Avg Real/User/Sys |
|---|---:|---:|---:|---:|
| Mainline gix | `19,708,248,064` B | `19,887,423,488` B | `18.44 GiB` | `102.48 s` / `315.71 s` / `4.25 s` |
| Branch gix | `8,940,797,952` B | `9,664,970,752` B | `8.66 GiB` | `25.54 s` / `369.68 s` / `4.19 s` |
| Git | `1,776,631,808` B | `1,925,038,080` B | `1.72 GiB` | `74.11 s` / `338.57 s` / `2.90 s` |

The branch uses about `9.77 GiB` less peak RSS than mainline gix, a reduction of about `53.0%`. Git uses much less RSS than both gix variants, but is about `2.89x` slower than the branch.

### Summary

| Pack | Runtime Result | Peak RSS Result | User/System Notes |
|---|---|---|---|
| Linux fixture | Branch gix is about `26%` slower than mainline gix, but `2.80x` faster than Git | Branch gix uses about `9%` less RSS than mainline gix; Git uses the least RSS | Branch gix has much higher system time than mainline gix |
| PHPStan | Branch gix is about `4.08x` faster than mainline gix and `2.89x` faster than Git | Branch gix uses about `53%` less RSS than mainline gix; Git uses the least RSS | Branch gix uses more user CPU than mainline gix, but finishes much sooner |

Overall, this branch is strongly beneficial for the PHPStan pack, both in wall-clock time and peak RSS. On the Linux fixture, it reduces peak RSS modestly but regresses runtime compared to mainline gix, with notably higher system CPU time.

</p>
</details>

vmandke:
Yes, I tried to work out something similar to how Git was parallelising. And I was only looking at the phpstan pack :) Will benchmark this more, and look into how the previous implementation can be fixed. 

Thanks for the review. Will get back with something by next week.

Byron:
Thanks. Meantime, I also added Git as a baseline.

Overall, Gitoxide is much faster in mainline and in this branch, but it uses significantly more memory.
From that point of view, this branch already is better. Compared to Git, it really isn't worse.

However, my feeling is that it would be better to optimize the current algorithm to also deal with PHP Stan and similar packs. It seems more like a bug in the current implementation than something that needs a complete change to be more like Git.

And in the second step, clearly, Gitoxide needs an algorithm that is just like the one in Git to reduce memory usage. Despite Git being much slower, it's very respectable how small the memory footprint is that it needs to handle these large packs.

There's clearly a trade-off and I personally would always want it to be fast but we'd want to offer a choice here, and with it, a more balanced algorithm.

With all that said, it also seems that this branch already contains an implementation that is similar to Git, given It's memory footprint, and all that speaks against it, a proper review outstanding is that it is slower on the Linux pack. So maybe what we really want is to implement this as alternative and further reduce its memory footprint instead.

I will leave it to you.

vmandke:
fyi: looking into this over this weekend, had a very busy work week

vmandke:
@Byron an update: I tried a few ways to merge both approaches cleanly, Ill raise a followup PR shortly.

Before raising a PR I will benchmark it thoroughly.

Byron:
This PR hasn't been active in a while and is conflicted. Thus, for now I think it's better to close it to be off my lists, with the hopes it will return once OP finds the time.

PR #2593, opened 2026-05-10
refactor: reduce smallvec dep from git-date

tisonkun:
Resolve TODO when encountering it. Comments inline.

cc @Byron


gix-date/src/parse/mod.rs
```diff
@@ -1,13 +1,29 @@
+use std::io;
 use std::str::FromStr;
 
 use crate::{Error, Time};
-use smallvec::SmallVec;
 
 /// A container for just enough bytes to hold the largest-possible [`time`](Time) instance.
-/// It's used in conjunction with
```
tisonkun, line 7: Seems truncated doc comments from before the refactor. I didn't find a reasonable fix-up here.

gix-date/Cargo.toml
```diff
@@ -24,9 +24,6 @@ bstr = { version = "1.12.0", default-features = false, features = ["std"] }
 serde = { version = "1.0.114", optional = true, default-features = false, features = ["derive"] }
 itoa = "1.0.17"
 jiff = "0.2.24"
-# TODO: used for quick and easy `TimeBacking: std::io::Write` implementation, but could make that `Copy`
-#       and remove this dep with custom impl
-smallvec = { version = "1.15.1", features = ["write"] }
```
tisonkun, line 29: TODO comments are outdated. But the dependency can be replaced with a handmade function.

Byron: I believe the comment was out of date.
`TimeBuf` should not have a publicly accessible write implementation.
But that can certainly be fixed by putting it into a private type.

Byron (changes requested):
Thanks for tackling this.

I left a note that should help steer this to completion.

gix-date/Cargo.toml
tisonkun: @Byron Thanks for your review! Pushed efaeb714e6b824f9870714bdaef082cfc654f163 to handle your comment. Hopefully I understand it correctly.

`TimeBuf` should still be public because it's referred to outside, like in `gix-actor`'s `Signature`.

Byron: Yes, it should still be public, but it cannot have an official write implementation.

tisonkun: Shall we merge this PR now?

gix-date/src/parse/mod.rs
```diff
@@ -1,27 +1,48 @@
...
 pub struct TimeBuf {
-    buf: SmallVec<[u8; Time::MAX.size()]>,
+    idx: usize,
+    buf: [u8; Time::MAX.size()],
+}
+
+struct TimeBufWriter<'a>(&'a mut TimeBuf);
+
+impl io::Write for TimeBufWriter<'_> {
+    fn write(&mut self, buf: &[u8]) -> Result<usize, io::Error> {
```
Byron, line 22: Must not implement `Write`.

Byron (commented):
Putting it to draft while the comment isn't addressed.

This reminds me of my rule to do these things myself, but here I am far enough down the wrong side of the road so might as well keep passive aggressively repeating myself here 😁.

gix-date/src/parse/mod.rs
tisonkun: Let me think of it.

TimeBufWriter is a private type. And Time::write_to requires the argument to be dyn std::io::Write.

So I use the inner struct pattern to keep the public TimeBuf not implement Write.

Or else we should change the signature of Time::write_to.

Anything I miss here? Or you can share one more concrete solution step :D

Byron (approved):
Ah, right, sorry for that.
To me it seemed like nothing changed. 

It would have helped if there would have been a summary of what you did along with "I think it can be merged now".

gix-date/src/parse/mod.rs
```diff
@@ -1,27 +1,48 @@
...
+        let idx = self.0.idx;
+        let end_idx = idx
+            .checked_add(buf.len())
+            .ok_or_else(|| io::Error::from(io::ErrorKind::OutOfMemory))?;
+        if end_idx > Time::MAX.size() {
+            return Err(io::Error::from(io::ErrorKind::StorageFull));
+        }
```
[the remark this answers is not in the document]
Byron: I don't think so

gix-date/src/parse/mod.rs
```diff
@@ -30,7 +51,7 @@ impl Time {
     /// and return `buf` as `&str` for easy consumption.
     pub fn to_str<'a>(&self, buf: &'a mut TimeBuf) -> &'a str {
         buf.clear();
-        self.write_to(&mut buf.buf)
+        self.write_to(&mut TimeBufWriter(buf))
             .expect("write to memory of just the right size cannot fail");
```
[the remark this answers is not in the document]
Byron: Truly awesome, mate! Addressed.

tisonkun:
@Byron Thanks for your review and merging. I enjoy this kind of collaboration :D

> It would have helped if there would have been a summary of what you did along with "I think it can be merged now".

It's actually included in https://github.com/GitoxideLabs/gitoxide/pull/2593#discussion_r3216682614, I quote:

> Pushed https://github.com/GitoxideLabs/gitoxide/commit/efaeb714e6b824f9870714bdaef082cfc654f163 to handle your comment. Hopefully I understand it correctly.

But for sure I can repeat it in the next comment to make it clear :P

PR #2602, opened 2026-05-16
Add generated archives for SHA-256 in `gix-pack`

cruessler:
While working on this PR, I noticed that there’s a method, `gix_pack::index::Version::hash()`, that currently is only implemented for `feature = "sha1"`. This method does not appear to be called, though (at least, my LSP found no references to it), so I was wondering whether it was redundant. Or, if it was not redundant, how it would need to be changed in order to support SHA-256.

This is part of #281.


Byron:
That's a great catch!

I also don't know why this `hash` method is there and don't think it's useful in any way. Thus it's removed in my second commit.

PR #2605, opened 2026-05-17
gix: cache packed-refs in a HashMap during fetch update_refs

nikicat:
## Summary

`fetch::refs::update()` calls `repo.try_find_reference(name)` once per advertised mapping. Each call does a filesystem stat for a loose ref followed by a binary search over `packed-refs` — and on wide-refs mirrors (e.g. AUR's 154k branches) that loop dominates wall time.

This PR snapshots `packed-refs` into a `HashMap<BString, gix_ref::Reference>` once and enumerates loose ref names into a `HashSet` once at the start of `update()`. Per-mapping lookups consult the HashMap when the name has no loose shadow, falling back to `repo.try_find_reference()` only when a loose ref exists (preserving the loose-shadows-packed precedence) or when the snapshot couldn't be built.

The optimization is **purely additive**: any failure to build the snapshots makes `build_lookup_fast_path` return `None` and the loop falls back to the unchanged slow path.

## Measured impact

`gitaur -Sy` against the AUR mirror (154k packed refs, ~200 loose refs, warm cache, no incoming updates):

| variant | wall | user CPU |
|---|---|---|
| unpatched main | 11.0 s | 8.1 s |
| this PR alone | 8.0 s | 5.8 s |
| this PR + #2604 (gix-ref binary-search skip-validation) | **5.0 s** | **3.3 s** |

The two changes target different fetch phases:
- This PR drops the *post-pack ref-update* loop from 154k binary searches to 154k hash lookups.
- #2604 drops the per-comparison validation in `packed::Buffer::binary_search_by`, which speeds up the *negotiation have-set build* phase (still does many binary searches via `Negotiate::mark_complete`).

Together they bring gitaur's incremental fetch on the AUR mirror to roughly the same wall time as `git fetch`, down from ~2.2× slower previously.

## Trade-offs / scope

- One extra full iteration over `packed-refs` per `update()` call (cost amortized across all mappings). For small repos with few refs the overhead is in the microseconds — not benchmarked specifically, but the HashMap insertions are O(n) where n is the packed-refs entry count, and a `gitoxide` smoke run on a small repo showed no regression.
- Loose-ref snapshot uses `repo.refs.loose_iter()` and bails to the slow path on iteration error. For repos where most refs are loose this is essentially the same work as today; for mirror-style repos with packed-heavy state it's the right shape.
- The optimization could plausibly live deeper, inside `packed::Buffer` itself (e.g. a lazy index of name→offset), to also accelerate non-fetch callers of `try_find_full_name`. That's a larger change touching a cached data structure — happy to discuss whether you'd prefer the optimization in that form instead, but this localized version was easier to scope, measure, and review independently.

## Test coverage

`cargo test -p gix-ref` passes (161/161). `cargo test -p gix --features blocking-network-client --test gix remote::fetch::blocking_and_async_io` passes the 7 tests that pass on main (3 tests fail with the same hash-mismatch on main without this PR — pre-existing, fixture-dependent).

## AI disclosure

Per `CONTRIBUTING.md`, this change was drafted with the help of Claude Opus 4.7 (see the `Co-authored-by:` trailer on the commit). Investigation, profiling, and review were done by me.

### DIFF
--- a/gix/src/remote/connection/fetch/update_refs/mod.rs
+++ b/gix/src/remote/connection/fetch/update_refs/mod.rs
@@ -530,23 +530,26 @@ impl LookupFastPath {
 /// `None` return causes the caller to fall back to the unmodified slow path,
 /// so the optimization is purely additive.
 fn build_lookup_fast_path(repo: &Repository) -> Option<LookupFastPath> {
+    // `cached_packed_buffer()` returns raw, namespace-prefixed names from
+    // `packed-refs`, while lookups here use namespace-stripped local names.
+    // Defer to the slow path, which applies the namespace correctly.
+    if repo.refs.namespace.is_some() {
+        return None;
+    }
     let buf = repo.refs.cached_packed_buffer().ok().flatten()?;
-    let iter = buf.iter().ok()?;
-    let packed: HashMap<BString, gix_ref::Reference> = iter
-        .filter_map(Result::ok)
-        .map(|r| {
-            let name = r.name.as_bstr().to_owned();
-            let r: gix_ref::Reference = r.into();
-            (name, r)
-        })
-        .collect();
-    let loose_shadows: HashSet<BString> = repo
-        .refs
-        .loose_iter()
-        .ok()?
-        .filter_map(Result::ok)
-        .map(|r| r.name.as_bstr().to_owned())
-        .collect();
+    let mut packed = HashMap::new();
+    for r in buf.iter().ok()? {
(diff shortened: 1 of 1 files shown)


nikicat:
Superseded by #2608, which moves the optimization inside `gix_ref::packed::Buffer` as a lazy `name → record-offset` index. Same speedup for fetch on wide-refs mirrors, plus every other caller of `try_find_full_name` benefits, no changes needed in `update_refs`. Closing this one.

PR #2606, opened 2026-05-17
gix-url: accept `impl Into<&BStr>` in `parse`

SarthakB11:
Per #2468, `gix_url::parse` now takes `impl Into<&'a BStr>` instead of `&BStr`. A `&bstring` borrow now suffices at the call site; existing `&BStr` callers keep compiling via the reflexive `From<&BStr> for &BStr` impl in `bstr`. String literals also work directly through `From<&str> for &BStr`.

The signature widens to a borrowed `Into` (`Into<&'a BStr>`) rather than the owned `Into<BString>` shape the issue mentions, because callers in this codebase already hold borrowed bstrs and the borrowed form avoids an allocation. The `bstr` re-export question raised in the same issue is out of scope here; happy to follow up.

One internal `.as_ref()` had to become `.as_bstr()` because `BString` exposes several `AsRef` impls and the call was ambiguous after the signature change. Doc-test examples were left passing literals directly.

## Validation

- `cargo test -p gix-url`
- `cargo check --workspace`
- `cargo doc -p gix-url` (doc-test examples)

## Disclosure

This PR was drafted with Claude Code (Anthropic). I reviewed and verified the diff before submission.

Closes #2468

### DIFF
--- a/gix-url/fuzz/fuzz_targets/parse.rs
+++ b/gix-url/fuzz/fuzz_targets/parse.rs
@@ -1,11 +1,12 @@
 #![no_main]
 use anyhow::Result;
+use bstr::ByteSlice;
 use libfuzzer_sys::fuzz_target;
 use std::hint::black_box;
 use std::path::Path;
 
 fn fuzz(data: &[u8]) -> Result<()> {
-    let url = gix_url::parse(data.into())?;
+    let url = gix_url::parse(data.as_bstr())?;
     _ = black_box(url.user());
     _ = black_box(url.password());
     _ = black_box(url.password());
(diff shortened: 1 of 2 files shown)


SarthakB11:
Closing this for now. After pushing, CI surfaced ambiguous `.into()` calls in ~34 downstream sites (multiple tests in `gix-url/tests/`, plus `gix-transport`, `gix/repository/remote.rs`, etc.) — the `impl Into<&'a BStr>` widening can't resolve `data.into()` without explicit annotations, so a clean version needs either a larger sweep of the call sites or a different signature shape.

Two questions before I retry, if you have a preference:

1. Owned `impl Into<BString>` (as the issue title literally reads) vs borrowed `impl Into<&'a BStr>` vs `AsRef<[u8]>` — each has different ergonomic trade-offs around allocations and downstream `.into()` compat.
2. The issue's secondary point — "bstr is not reexported" — would a `pub use bstr::{BStr, BString};` from `gix_url` be welcome alongside (or instead of) the parse signature change?

Happy to take either direction. Closing this draft so the issue stays the discussion point.

PR #2607, opened 2026-05-17
gix-command: pass shell name (not `--`) as `$0` to `<shell> -c`

### LINKED ISSUE #1842: `gix-command` passes `--` to become `$0` in `sh -c`
As detailed below and detailed further in the linked #1840 discussion thread, we are doing something like <code>/bin/sh -c '*command* "$@"' -- *args...*</code> when running commands with a shell, but `--` does not have its usual meaning of causing arguments to be interpreted as non-options (or otherwise to be interpreted less specially).

Instead, the subsequent arguments are already passed through as the positional parameters, and the argument we are currently passing as `--` becomes the special `$0` parameter that appears when the shell produces messages. More informative alternatives are available. It is not obvious what approach is best, but I think we should use the name the shell uses to invoke itself, which will usually look something like:

<pre><code>/bin/sh: <em>the body of the error message</em></code></pre>

Or, because some shells sometimes add further context information to their own messages, like:

<pre><code>/bin/sh: line 1: <em>the body of the error message</em></code></pre>

This makes clear that the message comes from a shell, rather than the less informative approach of using `--` as we currently are (or using `_`, which may be more common than `--` here but still not very informative). It distinguishes it from messages that come from an external command the shell runs, which is useful even when the operand to `-c` is just a path, or quoted path, of an external command for the shell to run. This is because a shell may sometimes fail to run such a command.

When it is important, such error messages will usually show the name of the command, so the alternative of using the code operand to `-c` or something derived from it risks duplicating information unnecessary.

---

### Discussed in https://github.com/GitoxideLabs/gitoxide/discussions/1840

<sup>Originally posted by **EliahKagan** February 12, 2025</sup>
When [`use_shell`](https://github.com/GitoxideLabs/gitoxide/blob/b310c16abbf1365136c0328f0aa1606e66bd09ef/gix-command/src/lib.rs#L30-L31) is `true` and no optimization keeps a shell from actually being used, the `impl From<Prepare> for Command` returns a `std::process::Command` instance that, roughly speaking, will run something like <code>/bin/sh -c '*command* "$@"' -- *args...*</code>.

The argument after `-c` and its operand is `--`. This has been the case since functionality for passing arguments was added in d8e8b54 (#496).

Because the arguments before `--` run a command with a shell, the arguments after `--` are arbitrary arguments that are passed to the command or script rather than being treated specially by the shell, and `--` is often used to prevent subsequent arguments from being treated as options or otherwise specially, it may look at first glance like `--` does that here. Perhaps that was even its intent.

But this `--` argument does not do that. Instead, to simplify from [the standard](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/sh.html), this way of invoking `sh` is:

```sh
sh -c command_string [command_name [argument...]]
```

Where, as the standard documents for `-c`:

> Read commands from the *command_string* operand. Set the value of special parameter 0 (see [*2.5.2 Special Parameters*](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/V3_chap02.html#tag_19_05_02)) from the value of the *command_name* operand and the positional parameters ($1, $2, and so on) in sequence from the remaining *argument* operands. No commands shall be read from the standard input.

This is to say that the effect of our passing `--` here is to tell the script whose code is passed as the operand to `-c` that it should consider its own name, i.e. the value of `$0`, to be `--`. One consequence of this is that errors produced in the shell itself are prefixed `--:`. For example, in #1839, the error message is:

```text
--: /Users/runner/work/gitoxide/gitoxide/target/debug/examples/arrow: No such file or directory
```

It seems to me that giving the script the name of `--` may be confusing to people examining stderr of a process run with `gix-command`, unless they are familiar with implementation details of `gix-command` (as well as with the syntax of `sh -c`).

If the goal of using `--` rather than some other string was to prevent subsequent commands from being interpreted as shell options, then nothing special has to be done to avoid that, *at least for POSIX-compatible shells*. Arguments that come after the command string are never interpreted this way, instead always becoming positional parameters like `$1`, `$2`, `$3`.

However, we do have to pass *something* where we are currently passing `--`, if arguments meant to become `$1`, `$2`, `$3`, ..., are passed. Otherwise the argument meant to be `$1` (and thus the first word expanded from `"$@"`) would become `$0`, and the other positional parameters would, in effect, be shifted.

I'm not sure what would be better. In one-liners and causal use, I think `_` is common. That is probably not better here. We should not use the value of the `command` field (which becomes the command string, sometimes with ` "$@"` appended), or at least we should not use it unmodified. This is because the `command` field is sometimes just a path, and having it be the value of `$0` for the script would make it hard to distinguish messages produced by the shell from those produced by an external command the shell runs. The ability to distinguish this has already been useful in #1839.</div>

SarthakB11:
When `gix-command` builds a shell invocation as `sh -c <script> <args...>`, the first trailing arg becomes the shell's `$0`. The crate was passing `--` there, which surfaced in error messages as e.g. `--: command not found` instead of `sh: command not found`.

The fix passes `"sh"` as the explicit `$0` placeholder so `$1` and onwards still land as the script's positional arguments. All ten assertions in `gix-command/tests/command.rs` that asserted the literal `--` as the `$0` placeholder are updated; other `--` strings in those tests are real script args (e.g. `ls --foo` inside the `-c` operand) and are unchanged.

## Disclosure

This PR was drafted with Claude Code (Anthropic). I reviewed and verified the diff before submission.

Closes #1842

### DIFF
--- a/gix-command/tests/command.rs
+++ b/gix-command/tests/command.rs
@@ -228,6 +228,12 @@ mod prepare {
             .expect("`prepare` tests must be run where 'sh' path is valid Unicode")
     });
 
+    // The basename of the default shell, used as the `command_name` operand
+    // after `-c <script>` and observable inside the shell as `$0`. The default
+    // shell is `gix_path::env::shell()`, documented as `/bin/sh` on Unix and a
+    // path ending in `sh.exe` on Windows.
+    const SH_BASENAME: &str = if cfg!(windows) { "sh.exe" } else { "sh" };
+
     fn quoted(input: &[&str]) -> String {
         input.iter().map(|s| format!("\"{s}\"")).collect::<Vec<_>>().join(" ")
     }
@@ -249,7 +255,7 @@ mod prepare {
         let cmd = std::process::Command::from(
             gix_command::prepare("   ").command_may_be_shell_script_allow_manual_argument_splitting(),
         );
-        assert_eq!(format!("{cmd:?}"), quoted(&[*SH, "-c", "   ", "sh"]));
+        assert_eq!(format!("{cmd:?}"), quoted(&[*SH, "-c", "   ", SH_BASENAME]));
     }
 
     #[test]
@@ -290,7 +296,7 @@ mod prepare {
             if cfg!(windows) {
                 quoted(&["ls", "first", "second", "third"])
             } else {
-                quoted(&[*SH, "-c", "ls first second third", "sh"])
+                quoted(&[*SH, "-c", "ls first second third", SH_BASENAME])
             },
             "with shell, this works as it performs word splitting"
         );
@@ -308,7 +314,7 @@ mod prepare {
             if cfg!(windows) {
                 quoted(&["ls", "first", "second", "third"])
             } else {
-                quoted(&["/somepath/to/bash", "-c", "ls first second third", "sh"])
+                quoted(&["/somepath/to/bash", "-c", "ls first second third", "bash"])
             },
             "with shell, this works as it performs word splitting on Windows, but on linux (or without splitting) it uses the given shell"
         );
@@ -327,7 +333,7 @@ mod prepare {
                 quoted(&["ls", "--foo", "a b", "additional"])
             } else {
                 let sh = *SH;
-                format!(r#""{sh}" "-c" "ls --foo \"a b\" \"$@\"" "sh" "additional""#)
+                format!(r#""{sh}" "-c" "ls --foo \"a b\" \"$@\"" "{SH_BASENAME}" "additional""#)
             },
             "with shell, this works as it performs word splitting, on windows we can avoid the shell"
         );
@@ -350,7 +356,10 @@ mod prepare {
         let cmd = std::process::Command::from(
             gix_command::prepare(r#"ls --foo="a b""#).command_may_be_shell_script_disallow_manual_argument_splitting(),
         );
-        assert_eq!(format!("{cmd:?}"), quoted(&[*SH, "-c", r#"ls --foo=\"a b\""#, "sh"]));
+        assert_eq!(
+            format!("{cmd:?}"),
+            quoted(&[*SH, "-c", r#"ls --foo=\"a b\""#, SH_BASENAME])
+        );
     }
 
     #[test]
@@ -358,7 +367,7 @@ mod prepare {
         let cmd = std::process::Command::from(gix_command::prepare("ls").arg("--foo=a b").with_shell());
         assert_eq!(
             format!("{cmd:?}"),
-            quoted(&[*SH, "-c", r#"ls \"$@\""#, "sh", "--foo=a b"])
+            quoted(&[*SH, "-c", r#"ls \"$@\""#, SH_BASENAME, "--foo=a b"])
         );
     }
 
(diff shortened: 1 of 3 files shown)


gix-command/tests/command.rs
```diff
@@ -308,7 +308,7 @@ mod prepare {
             if cfg!(windows) {
                 quoted(&["ls", "first", "second", "third"])
             } else {
-                quoted(&["/somepath/to/bash", "-c", "ls first second third", "--"])
+                quoted(&["/somepath/to/bash", "-c", "ls first second third", "sh"])
```
EliahKagan: Using `"sh"` universally is not correct, and the new inaccuracy makes this change worse than keeping `"--"` or using `"_"` would be. The insight that this is so is actually revealed by what had to be done here to make this test pass.

The value being adjusted here is the expected value of `format!("{cmd:?}")`, which is being examined as a convenient way of checking that `cmd` has the desired arguments. The way the `cmd` object was created is:

```rust
        let cmd = std::process::Command::from(
            gix_command::prepare("ls first second third")
                .command_may_be_shell_script()
                .with_shell_program("/somepath/to/bash"),
        );
```

This command, if run, would run `bash`, by running a command that names it `bash`. This should not display as `sh`.

To be clear, this will be fixed in the forthcoming changes--you don't have to do anything to fix it.

EliahKagan (changes requested):
Having all shells, even ones that are not `sh`, receive `sh` as `$0` and announce themselves as `sh`, is not correct. For that to happen even occasionally would be much worse than what we have now.

But what we have now is also not good, and this PR does identify the parts of the code to change, as well as embodying the insight that it really is possible to fix #1842 in a simple way by just using the best available information about the short name for the shell.

I'll go ahead and push commits to this with the changes needed to turn this into a correct fix. (My commits will also be written using Claude Code; I'll make sure their own trailers reflect this.)

As a separate step, I think that it will be necessary to rebase to change your commit messages, since currently they are written in the style of conventional commits, but using incorrect conventional prefixes. Also, we only use conventional commits when an associated changelog entry is needed (and one of my commits will have a `fix:` prefix). However, I may wait to do that until it's clear that all CI passes other than recent failures unrelated to changes in this PR, and until after getting a Copilot review.

EliahKagan (approved):
CI looks good at the tip before the final rebase (92142e27ba1f409bedacdb390d883ac3a36dc710), in the sense that the only failures match recent failures we've been having everywhere and are very unlikely to relate in any way to any of the changes in this PR.

Besides that last rebase, I think this is in good shape. In the rebase, I'll remove prefixes like `gix-command:`, which are interpreted as conventional commit prefixes. Based on your PR description, I'll also add trailers to your commits to indicate that Claude was used to write them. Our `CONTRIBUTING.md` asks for disclosure in individual commits and not only in PR descriptions. Fortunately, this is easy to fix while I'm rebasing anyway.

(GitHub will misleadingly show you, me, and Claude as though we are all coauthors of the first two commits. That's not the case: on those commits, you're author, Claude is coauthor, and the reason it will show me is that rebasing causes me to be the "committer"--but not the author.)

After that, I plan to enable auto-merge. But I don't plan to attempt to override branch protection. This may still not merge until a decision is made on what to do about the recent CI failures.

EliahKagan:
FYI: I've temporarily disabled auto-merge here, because I plan to rebase this to take the change that fixes CI (#2612), then let the tests run again and merge assuming they pass.

EliahKagan:
This looks good. The tests that were failing before are passing, as expected. I'm reenabling auto-merge.

Byron:
Thanks @SarthakB11 for the original submission, and @EliahKagan for seeing it through towards the merge!
It will be in the next release as well.

PR #2614, opened 2026-05-21
doc: Document executable-bit behavior during checkout

### LINKED ISSUE #1787: `overwrite_existing: false` is not documented to allow +x be set
### Current behavior 😯

In `gix_worktree_state::checkout::options`, `destination_is_initially_empty` and `overwrite_existing` are documented as:

https://github.com/GitoxideLabs/gitoxide/blob/8d84818240d44e1f5fe78a231b5d9bffd0283918/gix-worktree-state/src/checkout/mod.rs#L50-L58

This is technically consistent with the behavior (mentioned in the PoC and Impact sections of GHSA-fqmf-w4xh-33rh but not part of that vulnerability and not changed in #1764) that:

- When `destination_is_initially_empty: false`, executable permissions can be added to preexisting files, regardless of the value of `overwrite_existing`.
- When `destination_is_initially_empty: false` and `overwrite_existing: false`, executable permissions can be added to preexisting files whose contents *would also change* if `overwrite_existing` allowed it.

However, this seems unintuitive, so I think it should be documented for at least one of those options in the documentation comment, probably for `overwrite_existing`.

In case this behavior is actually going to change soon, which might imaginably occur depending on what approach is taken for #1783 and #1784, for now I'm opening this issue rather than a PR to change the documentation.

### Expected behavior 🤔

See above.

### Git behavior

Not directly applicable, because Git doesn't have these specific options. However, `overwrite_existing: true` is documented to cause a checkout similar in kind to `git checkout --force`. When `git checkout` is run without `--force` under a circumstance where `--force` would be needed to change contents, it likewise declines to change executable permissions.

### Steps to reproduce 🕹

Although this issue is distinct from GHSA-fqmf-w4xh-33rh and from #1784, the procedures to reproduce either of them, with any recent version of `gix-worktree-state` and other crates, will also show that `overwrite_existing` can be `false` and a nonexclusive checkout still changes permissions (`+x`).

shvvkz:
## Reported issue

`checkout::Options::overwrite_existing` currently documents that existing worktree entries won't be overwritten unless the option is enabled. However, it doesn't mention that in a non-exclusive checkout, pre-existing files may still have their executable bit updated to match the index even when `overwrite_existing` is `false`.

This can be surprising because callers may interpret `overwrite_existing = false` as preventing all changes to existing files, while the current behavior only prevents overwriting file contents.

Closes #1787.

## Summary

- Clarified the documentation for `checkout::Options::overwrite_existing`.
- Documented that when `destination_is_initially_empty` is `false`, existing files may still receive executable-bit updates even if `overwrite_existing` is `false`.
- Left the checkout behavior unchanged.

## Validation

- `cargo fmt`

I also tried to run `cargo check -p gix-worktree-state`, but it failed while compiling `gix-hash` without hash features, which appears unrelated to this documentation-only change.

### DIFF
--- a/gix-worktree-state/src/checkout/mod.rs
+++ b/gix-worktree-state/src/checkout/mod.rs
@@ -55,6 +55,10 @@ pub struct Options {
     /// even if they appear to be changed. When creating directories that clash with existing worktree entries,
     /// these will try to delete the existing entry.
     /// This is similar in behaviour as `git checkout --force`.
+    ///
+    /// Note that when `destination_is_initially_empty` is `false`, existing files may still have their
+    /// executable bit updated to match the index. This option prevents overwriting file contents, but
+    /// does not necessarily prevent metadata updates.
     pub overwrite_existing: bool,
     /// If true, default false, try to checkout as much as possible and don't abort on first error which isn't
     /// due to a conflict.


Byron:
Thanks a lot.

Even though [my comment here](https://github.com/GitoxideLabs/gitoxide/issues/1787#issuecomment-2611716280) mentioned that it shouldn't touch metadata at all in this case, I think it's good to start with a documentation update.

The prior CI failure is strange and I have a feeling it might be a race condition of sorts, related to packs, `gix-odb` even. That's not good, but I will dive in next time this happens.

`hyperfine --runs 100 '../target/debug/deps/odb-be2b33301340af5a multi_index_access' didn't reproduce the issue which seems to consistently happen here. Rebasing, maybe it's just a bad commit somehow.

PR #2616, opened 2026-05-22
fix(gix-pack)!: accept non-canonical pack entry size headers

0WD0:
## Summary

Accept non-canonical pack entry size headers in `gix-pack`, matching C Git's read behavior.

## Details

C Git accepts overlong pack entry size encodings. For example, `b3 00` decodes to a blob with size `3`, even though the canonical encoding would be `33`.

This can appear in real fetch/clone traffic when a server reuses existing pack data. A real-world example is [web3infra-foundation/mega](https://github.com/web3infra-foundation/mega) (tested on commit  d2d797b333e8aa8a78fc1d8f5c6cae33360bc9bc), where GitHub sends a pack entry for `moon/.nvmrc` with bytes starting `b3 00`. C Git can read it and `git fsck --full` accepts it, but `gix-pack` currently rejects it with:

```text
Pack entry is truncated: pack entry header uses a non-canonical size encoding
```

This change removes the canonical-size-length rejection and records the actual number of header bytes consumed.

Preserving the actual header length is important: accepting non-canonical headers while recomputing a canonical length would break pack_offset() reconstruction and ofs-delta base offset calculations.

Pack writing still emits canonical headers.

### DIFF
--- a/gix-pack/src/data/entry/decode.rs
+++ b/gix-pack/src/data/entry/decode.rs
@@ -58,6 +58,7 @@ impl data::Entry {
             header: object,
             decompressed_size: size,
             data_offset: pack_offset + consumed as u64,
+            encoded_header_size: consumed.try_into().expect("pack entry headers fit into u16"),
         })
     }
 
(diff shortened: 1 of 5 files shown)


Byron:
> Copilot was unable to review this pull request because the user who requested the review has reached their quota limit.

It's Ok, I prefer Codex anyway, which I think doesn't (yet) count against the subscription.

PR #2627, opened 2026-05-28
gix: allow sha256-only builds

10ne1:
**What**

This is part of https://github.com/GitoxideLabs/gitoxide/issues/281 and fixes the following open item in that ticket:

```
￼ Make SHA256-only builds compile where supported by feature flags.

Evidence: cargo check -p gix --no-default-features --features sha256 fails because
gix/src/config/cache/incubate.rs and gix/src/config/tree/sections/core.rs still name gix_hash::Kind::Sha1 
unconditionally.
```

It's 4 commits in total after addressing all feedback (including both Codex and Claude feedback). The most important concerns are:
1. To allow such builds
2. To behave like the git.git project when it comes to repository format version and extensions.

**AI disclosure**
This PR was written with the help of Claude Code, but was **not** vibe-coded. I'm not a fan of vibe-coding. I used AI to better understand the codebase myself (asked it a lot of questions), review both my code and the code generated by AI, used it especially to detect bugs and corner-cases and suggest fixes, however I very carefully reviewed & edited each code/comment lines, the commit messages and so on, until I was myself satisfied with the result.


gix/src/config/cache/incubate.rs
```diff
@@ -45,15 +45,10 @@ impl StageOne {
...
-            .then_some(Ok(gix_hash::Kind::Sha1))
-            .or_else(|| {
-                config
-                    .string(Extensions::OBJECT_FORMAT)
-                    .map(|format| Extensions::OBJECT_FORMAT.try_into_object_format(format))
-            })
-            .transpose()?
-            .unwrap_or(gix_hash::Kind::Sha1);
+        let object_hash = match config.string(Extensions::OBJECT_FORMAT) {
+            Some(format) if repo_format_version == 1 => Extensions::OBJECT_FORMAT.try_into_object_format(format)?,
```
cruessler: What does git do when `objectFormat` is set, but format version is 0? Is that something that can happen in practice? My intuition tells me we should fail, but I haven’t tried with git.

cruessler: Also, do we want to add a test specifically for that case?

10ne1: Very nice catch and your intuition is correct. 🙂  Git throws an error because objectFormat is not supported in v0.

(I've actually been looking at this error codepath in git.git and even found an old memleak (C how lovely...) while adding repo extensions which are also only supported in version > 0.)

I've updated the code to match the git failure and also added a test.

Many thanks for spotting this.

gix/tests/fixtures/make_config_repos.sh
```diff
@@ -165,6 +165,19 @@ mkdir not-a-repo-with-files;
   touch this that
 )
 
+# objectFormat is a v1-only extension; a version-0 repo that sets it is invalid (even for sha1).
+# Reach that state with git config independently of the fixture's default hash: first make a valid
+# v1+objectFormat repo, then flip the version to 0 as the last write. The flip must come last because
+# git config refuses to run once the repo is already invalid (v0 with a v1-only extension).
+for format in sha1 sha256; do
+  git init "objectformat-without-v1-$format"
+  (cd "objectformat-without-v1-$format"
```
cruessler: Nitpick: would something like `objectformat-$format-with-repository-format-v0` better reflect the setup?

gix/tests/gix/repository/open.rs
```diff
@@ -354,6 +354,27 @@ mod not_a_repository {
     }
 }
 
+mod object_format_extension {
+    use crate::util::named_subrepo_opts;
+
+    #[test]
+    fn rejects_object_format_on_v0_repo() -> crate::Result {
+        // objectFormat on a version-0 repo is rejected, even for sha1, matching git.
```
cruessler: Do you know when that behaviour got introduced into git? If recently, would it make sense to include the specific version it got introduced in in the comment?

cruessler (commented):
Just two minor suggestions, otherwise this is good to go from my side!

In comments we usually stick to `SHA1` and `SHA256` when referring to the hash kind. We also tend to use backticks for things like `objectFormat` and `gix_hash::Kind::Sha1` (the latter could even be turned into a proper link).

gix/tests/fixtures/make_config_repos.sh
10ne1: Yes, that reads better. I've amended the commit.

gix/tests/gix/repository/open.rs
10ne1: Yes, it was at least 2020 when the current behavior was introduced, when [Git 2.29.0 started complaining](https://lkml.org/lkml/2020/10/19/692) about and rejecting v1 extensions in v0 repos.

I think it's a good idea to note this in the comment, so I updated the message.

gix/src/config/mod.rs
```diff
@@ -86,6 +86,11 @@ pub enum Error {
     RefsNamespace(#[from] refs_namespace::Error),
     #[error("Cannot handle objects formatted as {:?}", .name)]
     UnsupportedObjectFormat { name: BString },
+    #[error(
+        "extensions.objectFormat is a v1-only extension, but the repository format version is 0; \
+         set core.repositoryFormatVersion=1 to use it, or remove extensions.objectFormat to fall back to the default Sha1 format (if supported by this build)"
+    )]
+    ObjectFormatRequiresV1,
```
Byron, line 93: A wonderful error message!

gix/tests/gix/repository/open.rs
```diff
@@ -354,6 +354,32 @@ mod not_a_repository {
...
 }
 
+mod object_format_extension {
+    use crate::util::named_subrepo_opts;
+
+    #[test]
+    fn rejects_object_format_on_v0_repo() -> crate::Result {
+        // objectFormat is a "v1-only" extension: git refuses to operate on a version-0 repo that
+        // sets it, even for sha1 (unlike grandfathered extensions like preciousObjects, which v0
+        // still honours). This rejection was introduced in git 2.29.0 (2020). We match it.
```
Byron, line 364: I like the research that went into this.

gix/src/config/cache/incubate.rs
```diff
@@ -104,6 +101,22 @@ impl StageOne {
...
+/// In Sha256-only builds we cannot open such a repository, so return an error instead.
+fn legacy_object_hash() -> Result<gix_hash::Kind, Error> {
+    #[cfg(feature = "sha1")]
+    {
+        Ok(gix_hash::Kind::Sha1)
+    }
+    #[cfg(not(feature = "sha1"))]
+    {
+        Err(Error::UnsupportedObjectFormat { name: "sha1".into() })
+    }
```
Byron, line 117: I was a bit stuck here and decided to go with something else.

Initially, instead of saying this isn't possible, I asked myself what would Git do if it would allow to not compile SHA-1 in. Then I think it should just be like SHA-1 never happened, hence SHA-256 is returned.

However, Git doesn't do that and it `gix` tooling that relies on SHA-256 only builds will simply have to declare it properly so both Git and the tool can open it.

An interesting journey, leading to nothing but this comment 😁.

gix/src/config/cache/incubate.rs
```diff
@@ -157,3 +170,24 @@ fn load_config(
 
     Ok(config)
 }
+
+#[cfg(test)]
+mod tests {
```
Byron, line 175: This test felt redundant in the light of compile-time safety and docs of `legacy_object_hash_reflects_build_features`.

gix/tests/fixtures/make_config_repos.sh
```diff
@@ -165,6 +165,19 @@ mkdir not-a-repo-with-files;
...
+# v1+objectFormat repo, then flip the version to 0 as the last write. The flip must come last because
+# git config refuses to run once the repo is already invalid (v0 with a v1-only extension).
+for format in sha1 sha256; do
+  git init "objectformat-$format-with-repository-format-v0"
+  (cd "objectformat-$format-with-repository-format-v0"
+    git config core.repositoryFormatVersion 1
+    git config extensions.objectFormat "$format"
+    git config core.repositoryFormatVersion 0
+  )
+done
```
Byron, line 179: I removed the 'dance' in favor of something more direct, for better or worse.

Byron (approved):
Thanks so much, and sorry for the long wait! Also thanks to @cruessler for the initial review, it's much appreciated.

The most notable change I made here is to… change the prefixes of the commit message. There is know that we only do conventional commit messages for the purpose of controlling the changelog. In this case, a `gix: ` prefix would probably have done nothing, but it might be confusing to look at. One of the commits is a genuine `fix: ` and I made it so.

Byron:
@10ne1 I wasn't able to push my changes and review commit back into this branch which is an inconvenience, so I have to close this PR and let it be superseded by #2651.

10ne1:
@Byron I enabled `Allow edits and access to secrets by maintainers` for all my branches going forward. Thanks!

Issue #2296, opened 2025-12-16
"gix status" is slow on Windows


Byron:
Here are some benchmarks that show the difference.

[status-benchmarks-gix.zip](https://github.com/user-attachments/files/24188791/status-benchmarks-gix.zip)

In short:

* direct `lstat` checks are like 10x slower on Windows and they don't scale
* one has to do dirlists which come with stats information for all entries
* this is why Git has a filesystem cache that probably works exactly that way, uses dirlists when needed to provide stat information, at least on Windows.

To have fast status information on Windows, one would have to provide some sort of trait which can do caching, and which ideally is *shared across threads with the ability to list multiple directories at the same time*.

### Research

* [Writeup on the fastest Git status in the world on Windows](https://www.reddit.com/r/rust/comments/1prkzqg/writing_the_fastest_implementation_of_git_status/?share_id=IyFkFFbyk9PTMeur-XDmF&utm_content=2&utm_medium=ios_app&utm_name=ioscss&utm_source=share&utm_term=1)
* [The sample code that seems to involve `gitoxide` as well](https://github.com/special-bread/tests-git-status/tree/main)
* [tests-git-status-main.zip - the code for posterity](https://github.com/user-attachments/files/24277285/tests-git-status-main.zip)

Byron:
There now is a filesystem cache on Windows which works as described here. It's lazy, just like in Git, so the main cause of slowness should be fixed.

EliahKagan:
This has been closed as a duplicate of #1025, but it's not obvious what the connection is. Was this intended to be closed as resolved instead?

Byron:
Good catch, this was by mistake.

Issue #1180, opened 2023-12-13
oxidize `lilgit`


Byron:
Here is the project:

### Missing Features

* [ ] complete status, abortable on first change for fastest-possible 'is-dirty' checks
* [ ] a way to obtain [the `upstream` of a branch](https://docs.rs/git2/latest/git2/struct.Branch.html#method.upstream)
    - Right now, it's possible to do this by hand, but it's it's unlikely anyone will get this right (as opposed to 'seemingly' right).

Byron:
The API for obtaining branches related to remotes has been overhauled in [the linked PR](https://github.com/Byron/gitoxide/pull/1181) and you should be able to replace the existing code with it fully. You will probably be particularly interested [in the API added in this commit](https://github.com/Byron/gitoxide/pull/1181/commits/270322e75a49f9a5c2d996cf0a20c6d622b40394).

Please note that you'd have to add `gix` as `git = https://github.com/Byron/gitoxide` dependency as a crates-io release with these features doesn't yet exist.

Byron:
Moved to [this Oxidize Discussion](https://github.com/GitoxideLabs/gitoxide/discussions/2750). This issue is being closed because integration tracking now lives in Discussions.

_AI disclosure: This comment was written and posted by OpenAI Codex on behalf of Byron._

PR #2631, opened 2026-05-30
perf: check `should_interrupt` per entry in `index_as_worktree`

sharifhsn:
`index_as_worktree` only checked the interrupt flag at chunk boundaries, so an in-flight chunk ran to completion even after interruption was requested. This checks it per entry instead.

I was looking at `Repository::is_dirty()`, trying to determine if #1180 was closeable, as it requested a perf improvement on this function. A larger improvement would require batched directory-stat caching as discussed in #2296, or to consume `UNTR` and `FSMN` tokens as attempted in #2503.

`is_dirty()` now early-returns on the first change and drops the status iterator, which sets the interrupt flag and leaves a detached background scan to wind down on its own. With the per-chunk check that background scan kept stat-ing its entire chunk; now it stops at the next entry.

I benchmarked it on a 50k-file repo:

| | main | this PR |
|---|---|---|
| clean | 61.3 ms | 61.8 ms |
| dirty, change in first-scanned file | 49.4 ms | 34.7 ms |
| dirty, change in last-scanned file | 61.5 ms | 61.4 ms |

The performance win shows up when `is_dirty()` is hit repeatedly and the leaked background scans would otherwise pile up; it scales with repo size and is within noise below ~10k files.

🤖 Generated with Claude Code

### DIFF
--- a/gix-status/src/index_as_worktree/function.rs
+++ b/gix-status/src/index_as_worktree/function.rs
@@ -156,6 +156,14 @@ where
             let mut out = Vec::new();
             let mut idx = 0;
             while let Some(entry) = chunk_entries.get(idx) {
+                // Allow a cooperative early-out *within* a chunk: once `should_interrupt` is set
+                // (e.g. by a caller that only needs to know whether *any* change exists, like
+                // `Repository::is_dirty`), stop promptly instead of finishing the whole chunk.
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot, great catch!

PR #2640, opened 2026-06-10
fix: only expand attribute macros when they are set

### LINKED ISSUE #2639: gix-attributes: macros expand even when the matching rule unsets them, diverging from `git check-attr`
### Current behavior 😯

`gix_attributes::Search` expands a macro's sub-assignments whenever a pattern referencing the macro matches, regardless of the assignment's state. With

```gitattributes
[attr]mylfs filter=lfs
*.bin mylfs
special.bin -mylfs
unspec.bin !mylfs
valued.bin mylfs=foo
```

all four paths report `filter=lfs`:

```text
normal.bin: filter -> [Value(ValueRef([108, 102, 115]))]   # "lfs"
special.bin: filter -> [Value(ValueRef([108, 102, 115]))]
unspec.bin: filter -> [Value(ValueRef([108, 102, 115]))]
valued.bin: filter -> [Value(ValueRef([108, 102, 115]))]
```

The cause appears to be in `Outcome::fill_attributes` (`gix-attributes/src/search/outcome.rs`): after recording the match, sub-attributes are pushed onto the work stack gated only on `is_macro`, without consulting the matched assignment's state:

```rust
let is_macro = !slot.macro_attributes.is_empty();
...
if is_macro {
    // TODO(borrowchk): ...
    let slot = &self.matches_by_id[id.0];
    self.attrs_stack.extend(
        slot.macro_attributes
            .iter()
            .filter(|attr| self.matches_by_id[attr.id.0].r#match.is_none())
            .map(|attr| (attr.id, attr.inner.clone(), Some(id))),
    );
}
```

Reproduced with gix-attributes 0.33.0 and 0.33.1.

### Expected behavior 🤔

A macro's sub-assignments should only be applied when the matching rule *sets* the macro (state `Set`). For `-mylfs`, `!mylfs`, and `mylfs=foo`, `filter` should remain unspecified, matching git.

### Git behavior

```console
$ git check-attr filter -- normal.bin special.bin unspec.bin valued.bin
normal.bin: filter: lfs
special.bin: filter: unspecified
unspec.bin: filter: unspecified
valued.bin: filter: unspecified
```

(git 2.45.2.) git's `attr.c` gates macro expansion on the value being true, in `fill_one`:

```c
if (item->macro && item->value == ATTR__TRUE)
    attr_state_queue_push(&todo, item->macro);
```

### Steps to reproduce 🕹

```toml
[dependencies]
gix-attributes = "0.33.1"
gix-glob = "0.26"
```

```rust
use gix_attributes::search::{MetadataCollection, Outcome};
use gix_attributes::Search;
use gix_glob::pattern::Case;

fn main() {
    let attributes =
        b"[attr]mylfs filter=lfs\n*.bin mylfs\nspecial.bin -mylfs\nunspec.bin !mylfs\nvalued.bin mylfs=foo\n";

    let mut search = Search::default();
    let mut collection = MetadataCollection::default();
    search.add_patterns_buffer(
        attributes,
        ".gitattributes".into(),
        Some(std::path::Path::new("")),
        &mut collection,
        true, /* allow macros */
    );

    for path in ["normal.bin", "special.bin", "unspec.bin", "valued.bin"] {
        let mut out = Outcome::default();
        out.initialize_with_selection(&collection, ["filter"]);
        search.pattern_matching_relative_path(path.into(), Case::Sensitive, Some(false), &mut out);

        let filter = out
            .iter_selected()
            .map(|m| format!("{:?}", m.assignment.state))
            .collect::<Vec<_>>()
            .join(", ");
        println!("{path}: filter -> [{filter}]");
    }
}
```

Context: found while implementing attribute-based Git LFS detection (deciding `filter=lfs` for paths from tree-stored `.gitattributes`), where the opt-out pattern `special.bin -lfs` ends up smudging files git would leave untouched.

NilsIrl:
Fixes #2639.

### What

`Outcome::fill_attributes()` expanded a macro's sub-assignments whenever a pattern referencing the macro matched, regardless of the assignment's state. Git expands a macro only when the matching rule *sets* it — `attr.c`'s `macroexpand_one()` acts only `if (item->macro && item->value == ATTR__TRUE)` — so `-macro`, `!macro`, and `macro=value` must not apply the expansion.

The fix gates the expansion on `State::Set` while still recording the macro's own match (git also reports e.g. `mymacro: unset` for a `-mymacro` rule; only the expansion is suppressed).

### Tests

A new `macro-states` repository in `make_attributes_baseline.sh` records `git check-attr -a` output for one path per assignment state (`macro`, `-macro`, `!macro`, `macro=value`), and a new `macros_are_only_expanded_when_set` test asserts parity, following the existing baseline-test pattern. Before the fix it fails with the full expansion showing up for the three non-set states; the existing `lookup-order` macro-cycle/recursion tests still pass since those macros are matched as set.

Ran `cargo test -p gix-attributes`, `GIX_TEST_IGNORE_ARCHIVES=1 cargo test -p gix-attributes --test attributes`, `cargo test -p gix-worktree`, `cargo clippy -p gix-attributes --all-targets`, and `cargo fmt --check`.

### AI disclosure

This change was researched, written, and tested by an AI agent (Claude Fable 5, via Claude Code) directed and reviewed by @NilsIrl; the commit is authored accordingly. The underlying issue (#2639) includes the standalone reproduction and the `attr.c` comparison.

### DIFF
--- a/gix-attributes/src/search/outcome.rs
+++ b/gix-attributes/src/search/outcome.rs
@@ -188,6 +188,10 @@ impl Outcome {
             }
             // Let's be explicit - this is only non-empty for macros.
             let is_macro = !slot.macro_attributes.is_empty();
+            // Like git, macros only apply their expansion when *set* - assigning any
+            // other state suppresses it (`attr.c`: `macroexpand_one()` expands only
+            // `if (item->macro && item->value == ATTR__TRUE)`).
+            let expand_macro = is_macro && matches!(assignment.state, crate::State::Set);
 
             slot.r#match = Some(Match {
                 pattern: self.patterns.insert(pattern),
@@ -208,7 +212,7 @@ impl Outcome {
                 return true;
             }
 
-            if is_macro {
+            if expand_macro {
                 // TODO(borrowchk): one fine day we should be able to re-borrow `slot` without having to redo the array access.
                 let slot = &self.matches_by_id[id.0];
                 self.attrs_stack.extend(
--- a/gix-attributes/tests/fixtures/make_attributes_baseline.sh
+++ b/gix-attributes/tests/fixtures/make_attributes_baseline.sh
@@ -97,6 +97,23 @@ function baseline() {
   baseline global
 )
 
+mkdir macro-states
+(cd macro-states
+  git init
+  : > user.attributes
+  cat <<'EOF' > .gitattributes
+[attr]mymacro attr-a attr-b=val -attr-c
+*.set mymacro
+*.unset -mymacro
+*.unspecified !mymacro
+*.value mymacro=other
+EOF
+  baseline file.set
+  baseline file.unset
+  baseline file.unspecified
+  baseline file.value
+)
+
 mkdir lookup-order
 (cd lookup-order
 
--- a/gix-attributes/tests/search/mod.rs
+++ b/gix-attributes/tests/search/mod.rs
@@ -129,6 +129,38 @@ fn assert_references(out: &Outcome) {
 }
 
 #[test]
+fn macros_are_only_expanded_when_set() -> crate::Result {
+    let (mut group, mut collection, base, input) = baseline::user_attributes("macro-states")?;
+
+    let mut buf = Vec::new();
+    group.add_patterns_file(
+        base.join(".gitattributes"),
+        false,
+        None,
+        &mut buf,
+        &mut collection,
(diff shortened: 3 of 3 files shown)


Byron:
Thanks a lot. This PR was superseded by an alternative implementation that also compares this behavior against a git baseline.

Issue #2600, opened 2026-05-16
Race condition in ref file update transactions


willstott101:
### Current behavior 😯

When updating a file ref, gitoxide reads the current value to check against the `Edit`, for example `PreviousValue::MustExistAndMatch()`  before locking that ref for updates. This means multiple concurrent edits can overwrite each other silently, and the returned value that's meant to show what got overwritten will be incorrect.

This is particularly important when multiple concurrent pushes are using `--force-with-lease` and the ref updates are being handled server-side by gitoxide. Two pushes could read the same ref, then both try to lock, and then both update. Only one of them should have been successful, and someone's push looked successful but was then silently thrown away 😱

I have an Opus-written fix for this, and am happy to open a PR. Basically it comes down to this block beiing moved into a function that can then be called whilst holding the lock:
* https://github.com/GitoxideLabs/gitoxide/blob/87433ed33eee9ba974111d20b854f6acb07cd4a6/gix-ref/src/store/file/transaction/prepare.rs#L29
* Must then be called in both arms of the match for Delete as-well-as Update

### Expected behavior 🤔

I would expect `lock_ref_and_apply_change` to be "safe" in concurrent environments, in the sense that it would compare against the same value it was overwriting.

### Steps to reproduce 🕹

It's kind of a fiddly one tbh. There's a test in mizzle that reproduces it quite consistently on the slow github actions runners: https://github.com/willstott101/mizzle/actions/runs/25799779550/job/75786315209

But the broad steps are:
* open a repo multiple times in separate threads
* try to update the refs with constraints like `MustExistAndMatch` where they all expect the same initial value
* the bug is surfaced if more than one thread returns as successfully updating the ref

Byron:
Thanks a lot for reporting, and sorry for the hassle.

I'll see you reviewing the PR in the coming days.

PR #2659, opened 2026-06-20
Fix transaction race for file refs, fixes: #2600

willstott101:
When two ref transactions update the same ref concurrently (e.g. two pushes doing main: `A` → `B`), both could read the old value `A`, serialize through the lock, and both pass the `MustExistAndMatch` check
  — silently losing one update. The fix reads the existing value after acquiring the lock, so the comparison is against a value that can't change underneath us. See: https://github.com/GitoxideLabs/gitoxide/issues/2600

  While rebasing this onto the current tree, one `gix-ref` test started failing. Moving the read under the lock meant a path-prefix collision (creating `a/b` while `a` is a ref file) was now hit by the
  lock's `create_dir_all` first, which reports `AlreadyExists` and surfaced as a misleading `"permanently locked"` error instead of the previous `NotADirectory`.

  My first pass restored the old behaviour with a small ref_contents probe before locking. That worked, but it added a filesystem read on every update just to produce a nicer error on a rare one — so I
  went looking for a better spot and found that `gix-fs`'s directory-creation loop already distinguishes a non-directory in the path (it does the `is_dir()` check) and was simply throwing that information
  away by forwarding the raw `AlreadyExists`. Reporting it as `NotADirectory` **there** lets `gix-ref` surface lock-acquisition I/O errors as `Error::Io` (keeping `Error::LockAcquire` for genuine contention), which
  fixes the error for every `gix-lock` caller and lets me drop the probe entirely, keeping the transaction code simpler.

### DIFF
--- a/gitoxide-core/src/lib.rs
+++ b/gitoxide-core/src/lib.rs
@@ -78,6 +78,8 @@ pub mod organize;
 pub mod pack;
 #[cfg(feature = "query")]
 pub mod query;
+#[cfg(feature = "blocking-client")]
+pub mod remote;
 pub mod repository;
 
 mod discover;
--- a/gitoxide-core/src/remote.rs
+++ b/gitoxide-core/src/remote.rs
@@ -1,0 +1,183 @@
+use std::{
+    io,
+    path::PathBuf,
+    time::{Duration, Instant},
+};
+
+use crate::{OutputFormat, net};
+use anyhow::bail;
+use gix::protocol::transport::client::blocking_io::connect;
+use gix::{
+    NestedProgress,
+    config::tree::Key,
+    objs::bstr::ByteSlice,
+    protocol::{self, handshake::Ref, transport},
+    refs::{
+        Target,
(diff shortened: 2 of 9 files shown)


Byron:
### Tasks

* [x] add benchmark utility in remembrance of the refs writing performance issue, get baseline. Locking performance is unchanged
* [x] refackiew

Byron:
Thanks so much!

All I could do here is to clean up the commits a bit, and make just a minor refactor. Overall, I believe this is sound, and much better than it was before. After seeing all this complexity just to deal with the filesystem, and Windows, I really can't wait to just swap in RefTables and be done.

Byron:
There is one more note: the global lock is a performance optimization which works for `gix`, but won't work if Git would join the brawl. Probably that should change at some point, also given that soon RefTable will solve all problems 😅.

The global lock optimisation I will probably tackle in a follow-up.

willstott101:
Thanks for your tidy up and merge!

~Are reftables an all or nothing replacement in a repo then? Not like individual file refs and packed refs that can live side by side?~ I can research this myself, ignore me

PR #2663, opened 2026-06-21
fix: seed fast pseudo-random numbers to differ across processes

### LINKED ISSUE #1816: Pseudorandom numbers are sometimes the same across processes

ameyypawar:
## What

`gix-utils` and `gix-fs` draw fast pseudo-random numbers from `fastrand`'s global generators, whose per-thread seed is derived only from values that can coincide between separate processes — notably small thread IDs. Two concurrently running processes can therefore draw identical sequences, which:

- weakens the jitter on quadratic backoff (meant to avoid a thundering herd of retries, used e.g. in `gix-lock`), and
- makes the `gix-fs` capability probes collide on their temporary file names — the spurious "symlinks unsupported" failure seen on CI (#1789).

Fixes #1816.

## How

Adds a small `gix_utils::rng` module: a per-thread `fastrand::Rng` seeded once from a high-entropy OS source via `getrandom::u64()`, so sequences differ across processes. The two production call sites are routed through it — the backoff jitter in `gix-utils`, and the three capability probes in `gix-fs` (whose now-unused direct `fastrand` dependency is dropped).

### wasm handling

`getrandom` is added as a **target-specific dependency, excluded on `wasm32-unknown-unknown`**, which has no entropy backend by default (and is single-process, so the cross-process concern doesn't apply there). On that target the seed falls back to a best-effort value that deliberately avoids `Instant::now()`, since that panics on `wasm32-unknown-unknown` (#7319). On `wasm32-wasi*` targets `getrandom` is included and uses the wasi backend.

### Note on the public API

This adds a new public `gix_utils::rng` module (so `gix-fs` can use it across the crate boundary) — flagging in case it affects the version bump.

### Out of scope

The `gix-testtools` port-shuffle (case 3 in the issue) is left as-is — the issue notes it isn't really a problem, only a minor test slowdown.

## Verification

- `gix-utils` + `gix-fs` build and tests pass; `cargo clippy -- -D warnings` and `cargo fmt --check` are clean.
- Builds on `wasm32-unknown-unknown` (getrandom excluded, fallback seed used) and `wasm32-wasip1` (getrandom included). Note: these were verified as builds, not at runtime (no local wasm host).

### DIFF
--- a/gix-utils/Cargo.toml
+++ b/gix-utils/Cargo.toml
@@ -21,3 +21,9 @@ bstr = ["dep:bstr"]
 fastrand = "2.0.0"
 bstr = { version = "1.12.0", default-features = false, features = ["std"], optional = true }
 unicode-normalization = { version = "0.1.19", default-features = false }
+
+# Used by `rng` to obtain a high-entropy, process-distinct seed. Excluded on
+# `wasm32-unknown-unknown`, which has no entropy backend by default; there `rng` falls back to a
+# best-effort seed (see `src/rng.rs`).
+[target.'cfg(not(all(target_arch = "wasm32", target_os = "unknown")))'.dependencies]
+getrandom = "0.4"
--- a/gix-utils/Cargo.toml
+++ b/gix-utils/Cargo.toml
@@ -18,7 +18,9 @@ doctest = true
 bstr = ["dep:bstr"]
 
 [dependencies]
-fastrand = "2.0.0"
+# We seed our own thread-local `Rng` (see `src/rng.rs`), so `fastrand`'s global generator and its
+# `std` default feature aren't needed.
+fastrand = { version = "2.4.1", default-features = false }
 bstr = { version = "1.12.0", default-features = false, features = ["std"], optional = true }
 unicode-normalization = { version = "0.1.19", default-features = false }
 
--- a/gix-utils/src/rng.rs
+++ b/gix-utils/src/rng.rs
@@ -1,0 +1,65 @@
+//! Fast pseudo-random numbers whose sequence differs across concurrently running processes.
+//!
+//! These numbers are *not* cryptographically secure. Unlike the global generators of the
+//! [`fastrand`] crate, whose per-thread seed is derived only from values that can coincide between
+//! separate processes (such as small thread IDs), the generator here is seeded from a high-entropy
+//! source so that concurrent processes don't share a sequence.
+//!
+//! This matters wherever a collision causes a real problem, such as the jitter used for retry
+//! backoff (to avoid a [thundering herd](https://en.wikipedia.org/wiki/Thundering_herd_problem)) or
+//! the temporary file names used while probing filesystem capabilities.
+//!
+//! See <https://github.com/GitoxideLabs/gitoxide/issues/1816> for details.
+
+use std::{cell::RefCell, ops::RangeBounds};
+
+thread_local! {
+    static RNG: RefCell<fastrand::Rng> = RefCell::new(fastrand::Rng::with_seed(seed()));
+}
+
+/// Return a `usize` in the given `range`, drawn from a per-thread generator that is seeded to differ
+/// across concurrently running processes.
+pub fn usize(range: impl RangeBounds<usize>) -> usize {
+    RNG.with(|rng| rng.borrow_mut().usize(range))
+}
+
+/// Obtain a high-entropy seed from the operating system, so that generators in separate processes
+/// don't share a sequence.
+#[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
+fn seed() -> u64 {
+    getrandom::u64().unwrap_or_else(|_| fallback_seed())
+}
+
+/// `wasm32-unknown-unknown` has no entropy source available by default and runs a single process,
+/// so the cross-process concern doesn't apply; use a best-effort seed instead.
+#[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+fn seed() -> u64 {
+    fallback_seed()
+}
+
+/// A best-effort seed combining values that vary across threads and processes, used only when a
+/// high-entropy source is unavailable. `std::process::id()` is the reliable differentiator across
+/// concurrent processes; `Instant::now()` is deliberately avoided as it panics on
+/// `wasm32-unknown-unknown`.
+fn fallback_seed() -> u64 {
+    use std::hash::{BuildHasher, Hash, Hasher};
+
+    let mut hasher = std::collections::hash_map::RandomState::new().build_hasher();
+    std::process::id().hash(&mut hasher);
+    std::thread::current().id().hash(&mut hasher);
+    hasher.finish()
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn values_are_in_range_and_not_constant() {
+        let mut seen = std::collections::HashSet::new();
+        for _ in 0..256 {
+            let n = super::usize(0..1_000);
+            assert!(n < 1_000, "value {n} must be within the requested range");
+            seen.insert(n);
+        }
+        assert!(seen.len() > 1, "the generator must not be constant");
+    }
+}
--- a/gix-utils/src/rng.rs
+++ b/gix-utils/src/rng.rs
@@ -3,13 +3,11 @@
 //! These numbers are *not* cryptographically secure. Unlike the global generators of the
 //! [`fastrand`] crate, whose per-thread seed is derived only from values that can coincide between
 //! separate processes (such as small thread IDs), the generator here is seeded from a high-entropy
-//! source so that concurrent processes don't share a sequence.
+//! source so that concurrent processes don't easily share a sequence.
 //!
 //! This matters wherever a collision causes a real problem, such as the jitter used for retry
 //! backoff (to avoid a [thundering herd](https://en.wikipedia.org/wiki/Thundering_herd_problem)) or
 //! the temporary file names used while probing filesystem capabilities.
-//!
-//! See <https://github.com/GitoxideLabs/gitoxide/issues/1816> for details.
 
 use std::{cell::Cell, ops::RangeBounds};
 
@@ -17,12 +15,10 @@ thread_local! {
     static RNG: Cell<fastrand::Rng> = Cell::new(new_rng());
 }
 
-/// Return a `usize` in the given `range`, drawn from a per-thread generator that is seeded to differ
+/// Return an unsigned number in the given `range`, drawn from a per-thread generator that is seeded to differ
 /// across concurrently running processes.
 pub fn usize(range: impl RangeBounds<usize>) -> usize {
     RNG.with(|cell| {
-        // `fastrand::Rng` is a single `u64`, so move it out behind a cheap placeholder, advance it,
-        // and put it back. A `Cell` is enough and avoids `RefCell`'s runtime borrow tracking.
         let mut rng = cell.replace(fastrand::Rng::with_seed(0));
         let n = rng.usize(range);
         cell.set(rng);
@@ -30,8 +26,6 @@ pub fn usize(range: impl RangeBounds<usize>) -> usize {
     })
 }
 
-/// Seed a generator from a high-entropy OS source, so that generators in separate processes don't
-/// share a sequence.
 #[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
 fn new_rng() -> fastrand::Rng {
     fastrand::Rng::with_seed(seed())
@@ -45,6 +39,8 @@ fn new_rng() -> fastrand::Rng {
     fastrand::Rng::with_seed(0x9e37_79b9_7f4a_7c15)
 }
 
+/// Seed a generator from a high-entropy OS source, so that generators in separate processes don't
+/// share a sequence.
 #[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
 fn seed() -> u64 {
     getrandom::u64().unwrap_or_else(|_| fallback_seed())
@@ -72,6 +68,9 @@ mod tests {
             assert!(n < 1_000, "value {n} must be within the requested range");
             seen.insert(n);
         }
-        assert!(seen.len() > 1, "the generator must not be constant");
+        assert!(
+            seen.len() > 1,
+            "the generator wasn't made for the Playstation 3 (imprecise, but still funny?)"
+        );
     }
 }
--- a/gix-utils/src/rng.rs
+++ b/gix-utils/src/rng.rs
@@ -11,36 +11,48 @@
 //!
 //! See <https://github.com/GitoxideLabs/gitoxide/issues/1816> for details.
 
-use std::{cell::RefCell, ops::RangeBounds};
+use std::{cell::Cell, ops::RangeBounds};
 
 thread_local! {
-    static RNG: RefCell<fastrand::Rng> = RefCell::new(fastrand::Rng::with_seed(seed()));
+    static RNG: Cell<fastrand::Rng> = Cell::new(new_rng());
 }
 
 /// Return a `usize` in the given `range`, drawn from a per-thread generator that is seeded to differ
 /// across concurrently running processes.
 pub fn usize(range: impl RangeBounds<usize>) -> usize {
-    RNG.with(|rng| rng.borrow_mut().usize(range))
+    RNG.with(|cell| {
+        // `fastrand::Rng` is a single `u64`, so move it out behind a cheap placeholder, advance it,
+        // and put it back. A `Cell` is enough and avoids `RefCell`'s runtime borrow tracking.
+        let mut rng = cell.replace(fastrand::Rng::with_seed(0));
+        let n = rng.usize(range);
+        cell.set(rng);
+        n
+    })
 }
 
-/// Obtain a high-entropy seed from the operating system, so that generators in separate processes
-/// don't share a sequence.
+/// Seed a generator from a high-entropy OS source, so that generators in separate processes don't
+/// share a sequence.
+#[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
+fn new_rng() -> fastrand::Rng {
+    fastrand::Rng::with_seed(seed())
+}
+
+/// `wasm32-unknown-unknown` has no OS entropy source and traps at runtime on APIs like
+/// `std::process::id()` and `Instant::now()`; it also runs a single process, so the cross-process
+/// concern doesn't apply. Use a fixed seed there.
+#[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+fn new_rng() -> fastrand::Rng {
+    fastrand::Rng::with_seed(0x9e37_79b9_7f4a_7c15)
+}
+
 #[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
 fn seed() -> u64 {
     getrandom::u64().unwrap_or_else(|_| fallback_seed())
 }
 
-/// `wasm32-unknown-unknown` has no entropy source available by default and runs a single process,
-/// so the cross-process concern doesn't apply; use a best-effort seed instead.
-#[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
-fn seed() -> u64 {
-    fallback_seed()
-}
-
-/// A best-effort seed combining values that vary across threads and processes, used only when a
-/// high-entropy source is unavailable. `std::process::id()` is the reliable differentiator across
-/// concurrent processes; `Instant::now()` is deliberately avoided as it panics on
-/// `wasm32-unknown-unknown`.
+/// A best-effort seed used only when the OS entropy source is unavailable. `std::process::id()` is
+/// the reliable differentiator across concurrent processes.
+#[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
 fn fallback_seed() -> u64 {
     use std::hash::{BuildHasher, Hash, Hasher};
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1915,7 +1915,6 @@ version = "0.21.2"
 dependencies = [
  "bstr",
  "crossbeam-channel",
- "fastrand",
  "gix-features",
  "gix-path",
  "gix-utils",
@@ -2586,6 +2585,7 @@ version = "0.3.3"
 dependencies = [
  "bstr",
  "fastrand",
+ "getrandom 0.4.2",
  "unicode-normalization",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2585,7 +2585,7 @@ version = "0.3.3"
 dependencies = [
  "bstr",
  "fastrand",
- "getrandom 0.4.2",
+ "getrandom 0.4.3",
  "unicode-normalization",
 ]
 
--- a/gix-fs/Cargo.toml
+++ b/gix-fs/Cargo.toml
@@ -26,9 +26,6 @@ gix-utils = { version = "^0.3.3", path = "../gix-utils" }
 thiserror = "2.0.18"
 serde = { version = "1.0.114", optional = true, default-features = false, features = ["std", "derive"] }
 
-# For `Capabilities` to assure parallel operation works.
-fastrand = { version = "2.1.0", default-features = false, features = ["std"] }
-
 [dev-dependencies]
 crossbeam-channel = "0.5.15"
 is_ci = "1.1.1"
--- a/gix-fs/src/capabilities.rs
+++ b/gix-fs/src/capabilities.rs
@@ -82,7 +82,7 @@ impl Capabilities {
         // can change the executable bit.
         // The equivalent test by git itself is here:
         // https://github.com/git/git/blob/f0ef5b6d9bcc258e4cbef93839d1b7465d5212b9/setup.c#L2367-L2379
-        let rand = fastrand::usize(..);
+        let rand = gix_utils::rng::usize(..);
         let test_path = root.join(format!("_test_executable_bit{rand}"));
         let res = std::fs::OpenOptions::new()
             .create_new(true)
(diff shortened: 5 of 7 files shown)


gix-utils/src/rng.rs
```diff
@@ -0,0 +1,70 @@
...
+//! backoff (to avoid a [thundering herd](https://en.wikipedia.org/wiki/Thundering_herd_problem)) or
+//! the temporary file names used while probing filesystem capabilities.
+//!
+//! See <https://github.com/GitoxideLabs/gitoxide/issues/1816> for details.
+
+use std::{cell::RefCell, ops::RangeBounds};
+
+thread_local! {
+    static RNG: RefCell<fastrand::Rng> = RefCell::new(new_rng());
+}
```
EliahKagan, line 16: This should probably use the lighter-weight [`Cell`](https://doc.rust-lang.org/std/cell/#cellt) instead of `RefCell`. That is possible by making calls to the thread-local generator replace it with a zeroed-out instance, mutating the temporarily extracted instance to generate a number and advance the instance's state, and then replacing it back. *`fastrand` uses `Cell` for [its own thread-local `Rng` instance](https://github.com/smol-rs/fastrand/blob/6c4e800cad05fba0e181100643acfe07fc2140b1/src/global_rng.rs#L30-L32).*

It feels intuitive to use `RefCell` for a thread-local PRNG, because a PRNG is often a large object that is expensive to copy. But `fastrand` [implements](https://github.com/smol-rs/fastrand/blob/6c4e800cad05fba0e181100643acfe07fc2140b1/README.md?plain=1#L14-L15) the [Wyrand](https://github.com/wangyi-fudan/wyhash/blob/master/Modern%20Non-Cryptographic%20Hash%20Function%20and%20Pseudorandom%20Number%20Generator.pdf) algorithm, whose state is [only 8 bytes](https://lokathor.github.io/prng/#bonus-generator-wyrand). A `fastrand::Rng` object is [really just a `u64`](https://github.com/smol-rs/fastrand/blob/6c4e800cad05fba0e181100643acfe07fc2140b1/src/lib.rs#L130-L131) under the hood.

(Because `fastrand` offers functions for generating a variety of different types, it has various helpers and macros it uses to wrap its usage of [`with`](https://doc.rust-lang.org/std/thread/struct.LocalKey.html#method.with) in a way that reduces code duplication. Since the purpose of `gix_utils::rng` is to serve a small number of fast noncryptographic PRNG needs in `gix-*` crates, I don't think there's any need to structure the code like that here. But it may still be useful to look at the code of `global_rng.rs` in `fastrand`.)

While I think `Cell` is a better approach than `RefCell` here, I don't think this is a blocker for the PR. I don't know of any reason a `RefCell`-based approach cannot be correct. And if this is merged, then a follow-up PR could be made later to change to using `Cell`.

*Acknowledgement:* These thoughts are partly motivated by a discussion about #1816 that I had with @BitterKanegul last month. (But any errors or misjudgments in this comment are mine.)

gix-utils/Cargo.toml
```diff
@@ -21,3 +21,9 @@ bstr = ["dep:bstr"]
 fastrand = "2.0.0"
```
EliahKagan: `fastrand` has a default feature `std`, which [is (only) needed](https://github.com/smol-rs/fastrand#features) for "global generator and global entropy." We needed that before, since we used its thread-local ("global") instance. But as of the change in this PR, we shouldn't need it anymore. So `default-features = false` can be used now.

This may also be a reasonable time to bump the `fastrand` dependency from 2.0.0 to 2.4.1.

gix-utils/src/rng.rs
```diff
@@ -0,0 +1,65 @@
...
+
+/// A best-effort seed combining values that vary across threads and processes, used only when a
+/// high-entropy source is unavailable. `std::process::id()` is the reliable differentiator across
+/// concurrent processes; `Instant::now()` is deliberately avoided as it panics on
+/// `wasm32-unknown-unknown`.
+fn fallback_seed() -> u64 {
+    use std::hash::{BuildHasher, Hash, Hasher};
+
+    let mut hasher = std::collections::hash_map::RandomState::new().build_hasher();
+    std::process::id().hash(&mut hasher);
```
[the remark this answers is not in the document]
Byron: Codex on behalf of Byron: Addressed by 8c171abd22 (`fix: avoid process::id() on wasm32-unknown-unknown (it traps)`). `new_rng()` now has a `wasm32-unknown-unknown` cfg branch that uses a fixed seed, while `fallback_seed()` and `std::process::id()` are cfg-excluded there. The WebAssembly `wasm32-unknown-unknown` CI check passes on a48b921d.

gix-utils/src/rng.rs
Byron: Codex on behalf of Byron: Addressed by 40ccaa0d9f (`rng: use Cell instead of RefCell and drop fastrand's std feature`). The thread-local generator is now stored in `Cell<fastrand::Rng>` and temporarily replaced while advancing the small `Rng` state. CI passes on a48b921d.

gix-utils/Cargo.toml
Byron: Codex on behalf of Byron: Addressed by 40ccaa0d9f and finalized by a48b921d67. `gix-utils` now depends on `fastrand = { version = "2.4.1", default-features = false }`, and the lockfile records the resolved dependency state. CI passes on a48b921d.

Byron:
Thanks for getting the fix in, and let's hope that truly works. It really depends on `getrandom`, but would also assume that it works better due to access to the OS entropy.

Also many thanks to @EliahKagan for chiming in and acknowledging Codex's auto-fix. I gave it another pass and assume it's what you had in mind (the review didn't make material changes).

Byron:
I think one windows job already failed because of exactly what this PR is supposed to improve:

```
  stderr ───
    failed to extract 'tests\fixtures\generated-archives\make_baseline.tar': Ignoring archive at 'tests\fixtures\generated-archives\make_baseline.tar' as GIX_TEST_IGNORE_ARCHIVES is set.
    thread 'pattern::matching::compare_baseline_with_ours' (832) panicked at gix-glob\tests\pattern\matching.rs:65:13:
    assertion `left == right` failed: baseline for matches must be false - check baseline and git version: GitMatch { pattern: "*/\\\'", value: "XXX/\\\'", is_match: true }
      left: true
     right: false
```

It's [this job](https://github.com/GitoxideLabs/gitoxide/actions/runs/28007275125/job/82892025167?pr=2663), but I am having trouble getting back to the first attempt, maybe because it kind of timed out and was extra-strange.

If it passes next time, I think we should still merge this as it nicely unifies our random number generation, independently of what it tries to achieve beyond that.

EliahKagan:
> I think one windows job already failed because of exactly what this PR is supposed to improve

I'm not aware of how that `pattern::matching::compare_baseline_with_ours` would be related to this PR or #1816 (which this PR fixes).

> It's [this job](https://github.com/GitoxideLabs/gitoxide/actions/runs/28007275125/job/82892025167?pr=2663), but I am having trouble getting back to the first attempt, maybe because it kind of timed out and was extra-strange.

I think you're referring to the inability to expand the "Test (nextest)" step in the "test-fixtures-windows (windows-latest)" job in attempt 1 there. It shows as though the job is still running, even though its status shows as failed.

I believe this sometimes happens when a runner stops responding and the job is eventually stopped automatically for that reason. I've seen this much more often over the course of the past month. I think it's *usually* a flake in GitHub's infrastructure, rather than relating conceptually to whatever test is executing.

However, I'm guessing that the test failure you showed was something that happened in that step, prior to the connection to the runner being lost. My guess is that the problems are independent: we may have a flaky test that failed while `nextest` was runnning, and also GitHub's infrastructure may have been flaky later on. I'm not certain, though.

Issue #2421, opened 2026-02-06
index.write() can cause index corruption if tree extension/tree cache is not taken into account


special-bread:
### Current behavior 😯

When updating entries in the index, and saving them with index.write() this correctly updates the entries. However, it does not update the tree extension. Looking into it, it appears to me that gitoxide does not yet support updating the tree extension as part of the index.write() call.

As a result, the written index is in an inconsistent state, and external git clients show git status as showing no changes on disk (since they query the tree extension) - this can cause files modified to be completely invisible, and commits to commit things that are not in the index entries (being in the tree cache)

You can fix this by deleting and recreating the index with git reset and git add *

### Expected behavior 🤔

Its not clear what the best solution here is, since in the long term, the answer is to update the tree extension when writing the index.

But we have some options:
1. Update the documentation to note that you must remove the tree first, and call it like so until the tree updating is implemented:
```
    index.remove_tree();
    index.write(gix::index::write::Options::default())?;
```
2. Add an internal index.remove_tree() to the process of writing the index, as an option which defaults to true, until the tree update itself is implemented. This will make the default behaviour workable, creating a valid index, just without the tree extension
3. Implement the actual update to the tree extension in the index - this is of course desirable but requires actual work :>

### Git behavior

git updates the tree extension, and libgit2 also does this:
https://github.com/libgit2/libgit2/blob/main/src/libgit2/index.c#L3174

im not sure if this is correct, but it looks like the tree cache code is here:
https://github.com/libgit2/libgit2/blob/main/src/libgit2/tree-cache.c

### Steps to reproduce 🕹

1. find a repo with an index that uses the tree extension - in my experience most of my repos do by default
2. The entries you want to update can be some modified files in a commit. to reproduce things easier i used a commit that had 2 files modified in it, and then committed a revert of that commit - this makes it so that you know which files should be modified.
3. Its easy to corrupt the index, but SEEING that is a bit more difficult. So If i recall, the steps to visualize the issue are to first do a soft reset of the commit with gitoxide, and then to discard one/more of the modified files with gitoxide. This causes the index to have the entry updated in the index, but the tree cache is stale, containing the old committed files before they were discarded.
4. if you then commit the remaining undiscarded file(s) - you will notice that the actual commit contains more than your status showed. This is the commit using the tree extension to write the blobs from the tree, but using the index entries to see if anything is changed, causing a mismatch between what is committed and what was shown as changed.

[...]

PR #2664, opened 2026-06-21
Document that writing the index does not refresh the tree-cache extension

ameyypawar:
## What

Documents a footgun on the index write methods, per #2421.

`File::write`, `File::write_to`, and `State::write_to` serialize the `tree` (tree-cache) extension from its current in-memory state — it is **not** recomputed or invalidated to match the entries. So if entries were modified since the index was read, the tree-cache is written back *still marked valid* even though it is now stale.

Git uses the tree-cache to skip unchanged directories when building a tree (on `git commit` / `git write-tree`), so a stale-but-valid tree-cache can make a later commit capture outdated subtree content; more generally, `git status` and later commits can disagree about what is staged.

The note points to `State::remove_tree()` as the workaround until the tree-cache is updated on write:

```rust
index.remove_tree();
index.write(gix_index::write::Options::default())?;
```

This is the documentation route (option 1) you preferred in the issue, rather than a behavior change.

## Notes

- Doc-only. The caveat is placed on all three public index serializers (`File::write`, `File::write_to`, `State::write_to`), with the full explanation on `File::write` and cross-references on the other two.
- `cargo doc -p gix-index --features sha1` (with `RUSTDOCFLAGS=-D warnings`) and `cargo fmt --check` are clean.

Addresses #2421.

### DIFF
--- a/gix-index/src/file/write.rs
+++ b/gix-index/src/file/write.rs
@@ -16,8 +16,9 @@ impl File {
     /// Write the index to `out` with `options`, to be readable by [`File::at()`], returning the version that was actually written
     /// to retain all information of this index.
     ///
-    /// Note that the `tree` (tree-cache) extension is written as-is and is **not** recomputed from the
-    /// current entries; see [`File::write()`] for the implications and the recommended workaround.
+    /// Note that the `tree` (tree-cache) extension is written as-is and is **not** recomputed or
+    /// invalidated to match the current entries; see [`File::write()`] for the implications and the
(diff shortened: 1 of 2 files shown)


Byron (approved):
Thanks a lot, that's a great catch!

Let's start with that and see if eventually this shortcoming can be removed.

For now, the index is very much plumbing and as such can have such foot guns, just to accelerate certain cases that don't actually need the tree cache to be recomputed or removed.

PR #2674, opened 2026-06-23
fix: remove global lock optimization to align with Git

Byron:
Previously, a transaction would rely on the packed-refs lock
only when available, and skip per-ref locks entirely.

This would, however, incorrectly serialise ref writers globally,
giving it the same write performance as ref-tables, while allowing
races with Git processes.

Follow-up to https://github.com/GitoxideLabs/gitoxide/pull/2659


Byron:
This may very well have performance implication, but it would only be noticeable when fetching tremendous amounts of loose refs to write them, and… these would go directly to packed-refs anyway.

Maybe back then, this optimisation didn't exist yet. So let's just fix it before an issue pops up.

Issue #2672, opened 2026-06-23
Add a documented local/offline build profile for Gitoxide CLI QoL


volt317:
### Summary 💡

Gitoxide already exposes several useful build profiles, including `lean`, `lean-async,` and `small`. The small profile appears close to a local-operation build, but discovering the correct combination currently requires manually inspecting the feature graph and checking for transport-related crates such as curl, reqwest, HTTP/TLS backends, and network client features.

This is a QoL/documentation request rather than a request to remove networking support.

Would the project be open to adding a clearly named feature preset such as local, local-only, or offline-local that documents the intended way to build Gitoxide for local repository operations only?

For example:

`cargo build --no-default-features --features local`

Expected behavior:

- preserve local repository operations: object database, refs, index, worktree, pack/revision traversal where applicable
- omit remote operations requiring network transports
- avoid enabling curl, reqwest, HTTP/TLS, git://, SSH, credential-helper, clone/fetch/push client stacks
- provide an easy target for users packaging Gitoxide in constrained, embedded, audited, CI, or minimal-toolchain environments

This would reduce feature-graph reverse engineering for users who want a smaller local tool build, and would make the existing small/local-operation intent easier to discover and validate.

### Motivation 🔦

Gitoxide already provides a highly modular feature architecture, but identifying the smallest practical feature set for local repository operations currently requires manual inspection of the feature graph and transitive dependencies.

A documented local/offline build profile would improve:

1. Build Simplicity
2. Packaging and Distribution
3. Depedency Auditing
4. CI and Build Performance
5. Discoverability
6. Scope Clarification

Technically, this is already supported...but requires enumeration of the process up to the user.

EDIT: To clarify, this is especially useful for MSRV-constrained or frozen-toolchain environments, where reducing the feature graph also reduces the chance of unrelated transport/client dependencies forcing newer transitive crates.

volt317:
Just checking in to see if this aligns with the project's intended direction or not.  If the answer is 'no,' that's perfectly fine.  I mainly wanted to understand whether `small` was the intended CLI profile for such a build rather than a dependency boundary and if the enhancement could make that process easier.

Byron:
Thanks for bringing this up.

Small was always intended to skip on all the overhead for transport, so I think it will always stay local. From that point of view it makes sense that it's documented as such, making the idea more explicit than it was before. PR are welcome.

EliahKagan:
[...]

PR #2679, opened 2026-06-26
Add a `local` build profile for offline/local-only builds

ameyypawar:
## What

For #2672 — a documented build profile for local-only operation, so packagers don't have to reverse-engineer the feature graph to get there.

## Background

`small` is already local-only (it has no `clone`), but it's size-optimized and single-threaded; `max-pure` is pure-Rust but networked. There was no **fast, pure-Rust, local-only** profile — which is the gap this fills.

## The `local` profile

It's `lean` minus the HTTP/SSH client stack **and** the SQLite-backed `query`/`corpus` tools:

```toml
local = ["hashes", "fast", "tracing", "pretty-cli", "gitoxide-core-tools", "prodash-render-line"]
```

- Keeps local repository operations, the local `ein` tools (`organize`/`estimate-hours`/`archive`/`clean`), and parallelism.
- No `clone`/`fetch`/`push`; pulls in **none** of curl/reqwest/TLS, and needs **no C toolchain** (no bundled SQLite) — verified via `cargo tree -e build,normal` (no `cc`/`cmake`/`-sys` anywhere).
- The pure-Rust transport *abstraction* crates stay linked, but no networked command is exposed.

Also adds `cargo check --no-default-features --features local` to the justfile `check` recipe, next to the other presets. No new crates enter `Cargo.lock`.

## Verification

Builds; `cargo clippy --workspace --no-default-features --features local` is clean; `gix status`/`clean` work; `gix clone` is absent.

## Open questions — this is a proposal

You own this feature architecture, so I'm not attached to the specifics:

- **Name** — `local`, or `offline` / `local-only`?
- **Feature set** — I kept the local `ein` tools and `tracing`; happy to trim toward a more minimal, `small`-like core, or to expand it, whichever you prefer.
- Or, if you'd rather simply **document that `small` is the local build** rather than add a profile at all, I'm glad to redo it that way instead.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,6 +82,18 @@ small = ["hashes", "pretty-cli", "prodash-render-line", "is-terminal"]
 ## It uses, however, a fully asynchronous networking implementation which can serve a real-world example on how to implement custom async transports.
 lean-async = ["hashes", "fast", "tracing", "pretty-cli", "gitoxide-core-tools", "gitoxide-core-tools-query", "gitoxide-core-tools-corpus", "gitoxide-core-async-client", "prodash-render-line"]
 
+## A fast, pure-Rust build for local-only repository operations.
+##
+## Keeps local operations and the local `ein` tools with parallelism, but omits the HTTP/SSH client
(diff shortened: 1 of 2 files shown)


Byron:
Not needed, issue #2672 was rejected.

PR #2680, opened 2026-06-27
enabling local profile dependency and code seperation

volt317:
partially addresses #2672

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2414,8 +2414,10 @@ name = "gix-shallow"
 version = "0.12.1"
 dependencies = [
  "bstr",
+ "gix-date",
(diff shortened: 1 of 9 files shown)


Byron:
While slightly distracted, I spent some time with it, but had to conclude that I don't think the refactor is worth it. Namely, it adds heavy dependencies to `gix-shallow` where there were none, which seems worse than having `gix-protocol` with a reduced dependency count in `gix`.

Something that could be done is what `gix-config` does, which provides its primitives types in the `gix-config-values`. This, however, is nothing I'd want here just yet.

Byron:
Closing as it's not like I thought it would be.

PR #2682, opened 2026-07-01
Add test for non-bare `repo.git` discovery classification

Byron:
Be sure it's not detected as bare for some reason.

Motivated by https://github.com/gitbutlerapp/gitbutler/issues/14534

### Tasks

* [x] reproduce issue
* [x] fix it in `gix`


Byron:
Force-merged this branch as the added tests should work just the same on windows, and only one windows runner was stuck for some reason, making no progress after 40 minutes.

EliahKagan:
It [looks like](https://github.com/GitoxideLabs/gitoxide/actions/runs/28490694609/job/84446516858) the runner had a flake in such a way that it stopped responding, which in the current UI is shown as though it is still running, even after it has failed.

When this happens, there is also a delay in cancelling the job--cancellation is only near-immediate on the path where the runner is told to cancel and does so, with the fallback of forcibly removing the job being done after more time has passed, if I understand correctly.

This was probably unrelated to any of our tests or their results. I'm rerunning the job to investigate, just in case. But if it fails again then I'll have to look at the details--since this branch has been deleted, the absence of a remote ref with the expected name may cause some steps to behave differently (in the past, I think this even kept actions/checkout from working). It's not worthwhile to restore the branch to help with this, since this can be investigated in other ways.

Byron:
Thanks for reminding me - I hope it sinks in now. Particularly the failure of the runner to respond to cancellations makes this even more of a problem, but nothing that I think can be done except for gathering data. If it turns out it's only the WIN ARM job, there would be ways to deal with this maybe.

> [..] since his branch has been deleted, the absence of a remote ref with the expected name may cause some steps to behave differently (in the past, I think this even kept actions/checkout from working). It's not worthwhile to restore the branch to help with this, since this can be investigated in other ways.

I will try to remember next time and keep the branch alive.

PR #2693, opened 2026-07-05
Actually mark RelativePath as transparent

### LINKED ISSUE #2692: Soundness: RelativePath is treated as transparent but it isn't
https://github.com/GitoxideLabs/gitoxide/blob/0052e4256803a9905eacf80eb27be23cbffe7359/gix-path/src/relative_path.rs#L15-L19

https://github.com/GitoxideLabs/gitoxide/blob/0052e4256803a9905eacf80eb27be23cbffe7359/gix-path/src/relative_path.rs#L29-L35

The unsafe code here is incorrect.

Furthermore, I don't actually think we can assume much about `Result`'s layout there.

Manishearth:
Fixes https://github.com/GitoxideLabs/gitoxide/issues/2692

I went ahead and cleaned up the Result there as well, the layout of Result is not well defined.

It worked anyway because of niche optimizations, I think, but we can't rely on that here.

### DIFF
--- a/gix-path/src/relative_path.rs
+++ b/gix-path/src/relative_path.rs
@@ -12,7 +12,7 @@ pub(super) mod types {
     /// - The path separator always is `/`, independent of the platform.
     /// - Only normal components are allowed.
     /// - It is always represented as a bunch of bytes.
-    #[derive()]
+    #[repr(transparent)]
     pub struct RelativePath {
         inner: BStr,
     }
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks so much, I really should try not to write unsafe code!

In this particular instance, I'm pretty sure I copied it from the `bstr` crate - maybe that went terribly wrong, or maybe there is something there as well.

PR #2704, opened 2026-07-09
gix-protocol: skip the loose-ref probe when building the fetch have-set

nikicat:
`mark_complete_and_common_ref` resolves every mapping's local ref via `Store::find`, which probes the loose ref file on disk *before* consulting packed-refs. On a freshly-packed mirror that is one wasted `open()` syscall per ref — on the AUR mirror (~155k tracking refs) the have-set build measures ~2.1s, dropping to ~0.6s with this change.

Three commits:

1. **gix-protocol: split `mark_complete_and_common_ref`'s mapping loop cost** — wraps the per-mapping loop in a `mark mappings` detail span carrying `find_ms`/`commit_ms`, so the packed-ref lookup cost and the ODB/commit-graph load cost can be told apart. Fields are no-ops unless the gix-trace `tracing`/`tracing-detail` features are on.
2. **gix-ref: add `file::Store::try_find_packed_only`** — resolves a name from a snapshotted packed buffer only, reusing `find_one_with_verified_input`'s precompose/namespace/multi-directory logic but skipping the loose probe (threaded as a named `LooseRefs::{Consult,Skip}`, not a bare bool). In negotiate, packed-refs and the loose-name set are snapshotted once; loose names keep the full loose-first lookup so precedence is preserved, and everything falls back to `find` when either snapshot is unavailable — a missed loose name can only cost negotiation efficiency, never correctness.
3. **gix-protocol: borrowed packed lookup for the have-set** — the packed-only case only needs the target oid, so it reads `target()` off the borrowed `packed::Buffer::try_find` record instead of allocating an owned `Reference` per ref (~120ms of `find_ms`).

Tests:
- gix-ref: `try_find_packed_only::skips_loose_and_honors_packed` locks the packed-only contract against a fixture with packed-only, loose-over-packed (distinct OIDs), and loose-only refs.
- gix: `fetch_with_multi_round_negotiation` now runs each case with loose *and* packed refs, asserting identical rounds and pack object counts.


nikicat:
Addressed in the force-push (`e2c9dd4d3` → `2792b8905`): the loose-name snapshot is now only built when the fetch maps at least 256 refs, so a small fetch into a repository with many loose refs keeps the previous O(mapped refs) `find` path instead of paying an O(all loose refs) walk. The gate is folded into the commit that introduced the scan, so the history stays reviewable; the follow-up borrowed-lookup commit is unchanged on top.

One coverage note: with the gate, the small-fixture fetch tests no longer reach the snapshot path end-to-end — the packed-only lookup contract itself remains covered by the gix-ref `try_find_packed_only` test. Happy to adjust the threshold if you'd prefer a different value.

Byron:
While performance improvements are appreciated, I expect the reftable to become the new default in these cases to instantly solve all performance problems. 

If you agree, I'd appreciate if you could close this PR, knowing that in a month or so reftable should be supported.

nikicat:
Agreed — closing. This one is squarely in the category `reftable` makes moot: the whole point is avoiding an O(all loose refs) probe when building the have-set, which stops being a concern once ref lookups are not backed by loose files.

Keeping it on my fork in the meantime. Thanks for taking a look, and for the timeline.

PR #2706, opened 2026-07-09
gix-transport: add a download-progress byte counter to http::Options

nikicat:
The curl backend adds the length of each response-body chunk it receives to an optional `Arc<AtomicU64>` carried on `http::Options`. This gives callers a live byte signal for phases the protocol layer reports no progress for — notably the v2 `ls-refs` advertisement, which on a large-ref mirror is many MiB and can take a long time with no feedback at all (`transfer_progress` only starts moving once the pack begins).

Headers flow through the separate `header()` callback, so the counter only ever accumulates body bytes.

Covered by `download_progress_counts_response_body_bytes`, which drives a handshake against the mock server and asserts the count is non-zero and consistent.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2571,6 +2571,7 @@ dependencies = [
  "gix-url",
  "parking_lot",
  "pin-project-lite",
+ "portable-atomic",
  "reqwest",
  "serde",
  "thiserror 2.0.18",
--- a/gix-transport/Cargo.toml
+++ b/gix-transport/Cargo.toml
@@ -133,6 +133,9 @@ async-std = { version = "1.12.0", optional = true }
 
 document-features = { version = "0.2.0", optional = true }
 
+[target.'cfg(not(target_has_atomic = "64"))'.dependencies]
+portable-atomic = "1"
+
 [dev-dependencies]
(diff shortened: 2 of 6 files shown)


nikicat:
Both review findings are addressed in the force-push (`47b4e81` → `ea6256e72`):

- **Breaking-change marking**: the commit is now a `feat!:` conventional message (`feat!: add a download-progress byte counter to http::Options`), with the breaking nature spelled out in the body — `Options` is constructible via struct literal, so dependents that name every field must add the new one. Fittingly, `gix`'s own `transport_options` test destructures `Options` exhaustively and proved the point; it now names the field and asserts it's never set from configuration.
- **Portable atomics**: the counter type is re-exported as `http::AtomicU64` — `std`'s atomic where `target_has_atomic = "64"`, with the same `portable-atomic` fallback `gix-status` uses (target-specific dependency), so the curl transport keeps compiling on targets like `armv5te`.

Curl integration tests (including the new counter test) and the `gix` transport-options tests pass.

Byron:
I have to classify this as enhancement and while one can certainly try, I won't spend review time on them unless these are obvious.
And generally, anything that shoehorns more into the `curl` backend specifically is better kept in a private fork.

PR #2707, opened 2026-07-09
feat: keep the underlying zlib error when decompression fails

### LINKED ISSUE #2703: gix-features: preserve the underlying inflate error instead of flattening it to "corrupt deflate stream"
### Summary 💡

In `gix-features/src/zlib/stream/inflate.rs`, the decompress loop maps every error to a fixed string:
`Err(..) => return Err(io::Error::new(io::ErrorKind::InvalidInput, "corrupt deflate stream"))`,
This discards the real cause. In particular, a checksum failure (flate2: "incorrect data check") becomes indistinguishable from genuine bitstream corruption once it reaches a caller.

This one-line change appends the underlying error while keeping the existing prefix.

```diff
--- a/gix-features/src/zlib/stream/inflate.rs
+++ b/gix-features/src/zlib/stream/inflate.rs
@@ -34,7 +34,9 @@ pub fn read(rd: &mut impl BufRead, state: &mut Decompress, mut dst: &mut [u8]) -
             Ok(Status::Ok | Status::BufError) if consumed != 0 || written != 0 => continue,
             // A strange state, where zlib makes no progress but isn't done either. Call it out.
             Ok(Status::Ok | Status::BufError) => unreachable!("Definitely a bug somewhere"),
-            Err(..) => return Err(io::Error::new(io::ErrorKind::InvalidInput, "corrupt deflate stream")),
+            // Keep the underlying zlib error (e.g. "incorrect data check") so callers can tell a
+            // genuine stream corruption apart from a checksum mismatch.
+            Err(err) => return Err(io::Error::new(io::ErrorKind::InvalidInput, format!("corrupt deflate stream: {err}"))),
         }
     }
 }
```

### Motivation 🔦

This cost real debugging time downstream: in https://github.com/radicle-dev/heartwood, large fetches on Apple Silicon aborted with "Failed to consume the pack sent by the remote", which traced back to this "corrupt deflate stream". The pack was valid (canonical git indexed it fine); the actual failure was a [wrong adler32 from a bug in zlib-rs 0.6.3's NEON path](https://github.com/trifectatechfoundation/zlib-rs/pull/504).

ameyypawar:
Fixes #2703.

### The problem

Every decompression failure in `zlib::stream::inflate::read()` was flattened into the fixed `corrupt deflate stream` message. As described in the issue, that made a checksum failure indistinguishable from genuine bitstream corruption — which cost real debugging time downstream when a checksum bug in zlib-rs' aarch64 NEON path (trifectatechfoundation/zlib-rs#504) made valid packs look corrupted.

### Why the proposed one-liner isn't enough anymore

The diff in the issue appends `{err}` to the message, which carried the zlib detail back when the error came from `flate2` — but since 5a2361b42 this crate wraps `zlib-rs` directly, and its error is mapped into the coarse `DecompressError` enum first. Both a checksum mismatch and genuine corruption become `DataError` there, whose display is just `Invalid input data`, so `{err}` alone still couldn't tell them apart.

The detail does exist one layer down: `zlib_rs::Inflate` records the classic zlib message (`incorrect data check`, `incorrect header check`, …) and exposes it via `error_message()`. This PR forwards that as `Decompress::error_message()` and appends it at the flatten site, falling back to the `DecompressError` display when no message is set:

```
corrupt deflate stream: incorrect data check
```

The error kind stays `InvalidInput`, and nothing in the workspace matches on the previous bare message.

### Tests

`gix-features` sets `[lib] test = false`, so tests live in a new `zlib` integration-test target (mirroring the `pipe` one, `required-features = ["zlib"]`). The test deflates data with this crate's own writer, then corrupts the adler32 trailer (→ `incorrect data check`) and the stream header (→ `incorrect header check`), pinning both messages exactly. It fails against the previous code with the bare `corrupt deflate stream`.

While placing the test I noticed `src/zlib/stream/deflate/tests.rs` is currently dead: it's included via `#[cfg(test)] mod tests;`, but `[lib] test = false` means no lib test harness is ever built, so it doesn't run even with `--all-features`. I left it untouched here to keep this change focused — happy to revive it in a follow-up (probably by moving it into the new `zlib` test target).


Byron:
Thanks a lot for the fix, it's much appreciated. And to say it more directly: I have come to greatly appreciate your work here (and over at GitButler).

This PR will be merged shortly.

PR #2709, opened 2026-07-10
fix: make `gix exclude query` consult the index like `git check-ignore` (#2562)

vismaytiwari:
This addresses the `gix exclude query` mismatch reported in #2562 — thanks @Byron for the detailed analysis and the exact fix direction.

### Problem

`gix exclude query <path>` consulted only the exclude stack and reported paths that are **tracked in the index** (a file entry, or a directory that contains index entries) as excluded — unlike `git check-ignore` (normal mode), which never reports tracked paths.

### Fix

Consult the index before reporting a match, suppressing it when the path is tracked (`index.entry_by_path(path).is_some() || index.path_is_directory(path)`), matching `git check-ignore`. This is applied uniformly in `print_match`, so all three query paths (stdin, index-entries, and the pattern fallback) become index-aware. A new `--no-index` flag keeps the previous pure-ignore-stack behavior, mirroring `git check-ignore --no-index`.

### Verification

Using the reproduction from the issue:

```sh
git init; mkdir -p src/bin other/bin
printf 'bin/\n' > .gitignore
printf 'fn main() {}\n' > src/bin/stub_gen.rs
git add -f .gitignore src/bin/stub_gen.rs && git commit -m init
printf 'extra\n' > src/bin/extra.txt; printf 'other\n' > other/bin/file.txt
```

Default (index-aware) now matches `git check-ignore`:

| path | tracked? | before | after (`gix`) | `git check-ignore` |
| --- | --- | --- | --- | --- |
| `src/bin/stub_gen.rs` | yes | **excluded** | not excluded | not ignored |
| `src/bin` | yes (dir) | — | not excluded | not ignored |
| `src/bin/extra.txt` | no | excluded | excluded | ignored |
| `other/bin` | no | excluded | excluded | ignored |

`gix exclude query --no-index …` reports all matches, matching `git check-ignore --no-index`.

### Notes

- `gix exclude query` currently has no journey-test coverage. I'm happy to add one — is there a preferred spot / fixture + snapshot convention you'd like me to follow (e.g. a new sandbox under `tests/`)?
- This fixes the CLI. Exposing index-abiding exclude queries from `gix::Repository` (the "real fix" you mentioned) could be a good follow-up.

### DIFF
--- a/gitoxide-core/src/repository/exclude.rs
+++ b/gitoxide-core/src/repository/exclude.rs
@@ -15,6 +15,9 @@ pub mod query {
         pub overrides: Vec<OsString>,
         pub show_ignore_patterns: bool,
         pub statistics: bool,
+        /// If set, do not consult the index, reporting matches purely from the
+        /// ignore stack (like `git check-ignore --no-index`).
+        pub no_index: bool,
     }
 }
 
@@ -28,6 +31,7 @@ pub fn query(
         format,
         show_ignore_patterns,
         statistics,
+        no_index,
     }: query::Options,
 ) -> anyhow::Result<()> {
     if format != OutputFormat::Human {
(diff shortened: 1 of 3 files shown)


vismaytiwari:
Closing this — it duplicates your own draft #2564, which already takes the same index-consulting approach and has tests. Apologies for the noise; I should have checked the issue's linked PRs before opening this, not just a body search. If the `--no-index` flag or the `git check-ignore` before/after here is at all useful for #2564, happy to fold it in there; otherwise I'll stay out of the way.

PR #2711, opened 2026-07-12
feat!: `gix-error` instead of `thiserror` in `gix-lock`

ameyypawar:
The first crate of the `gix-error` conversion we discussed — the smallest pending leaf from `etc/plan/gix-error.md`, kept deliberately small to calibrate on your review before ramping up.

### The shape

The plan prefers `pub type Error = Exn<Message>` unless a crate needs a more specific concrete error — this one does: `gix-ref` deliberately distinguishes lock contention from other IO errors when preparing ref transactions (its `lock_acquire_error()` documents that path collisions must surface as `Error::Io`, while `LockAcquire` is "reserved for actual contention"). So the thiserror enum became a concrete `acquire::Failure` with manual `Display`/`Error` impls — message texts unchanged — wrapped as `pub type Error = Exn<Failure>`, and that matcher now reaches it through `into_inner()`.

### What made this smaller than expected

Consumers that merely embed the error in their own `thiserror` enums via `#[from]` (`gix-index`, `gix-protocol`) need no changes at all: `Exn`'s `Deref` lets the derived `source()` resolve to the concrete `Failure`, so `std` error chains keep exactly the same links as before — verified with a probe walking a consumer's chain down to the underlying `io::Error`. That should shrink most future conversions, as only consumers that *match* on variants need adapting.

### Adaptations (second commit)

- `gix-ref`: the matcher above, and the two variants holding the failure as `#[source]` now store `acquire::Failure` directly. Note this changes the payload type of two public `Error` variants — happy to split that into its own `change!:` commit if you prefer it called out per-crate.
- Doctests in `gix-lock`/`gix-shallow` and one `?`-conversion in `gix-testtools` use `.into_error()` per `AGENTS.md`. Worth knowing: `cargo nextest` doesn't execute doctests, so these only show up in `cargo test --doc`.

Validated with the test suites of `gix-lock`, `gix-ref`, `gix-index`, `gix-shallow`, `gix-protocol` and `gitoxide`, plus clippy and fmt. `Cargo.lock` is included since `pure-rust-build` runs with `--locked`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2041,10 +2041,10 @@ version = "0.0.0"
 name = "gix-lock"
 version = "23.0.1"
 dependencies = [
+ "gix-error",
  "gix-tempfile",
  "gix-utils",
  "tempfile",
- "thiserror 2.0.18",
 ]
 
 [[package]]
--- a/gix-lock/Cargo.toml
+++ b/gix-lock/Cargo.toml
@@ -18,7 +18,7 @@ test = true
 [dependencies]
 gix-utils = { version = "^0.3.3", default-features = false, path = "../gix-utils" }
 gix-tempfile = { version = "^23.0.0", default-features = false, path = "../gix-tempfile" }
-thiserror = "2.0.18"
+gix-error = { version = "^0.2.4", path = "../gix-error" }
 
 [dev-dependencies]
 tempfile = "3.26.0"
--- a/gix-lock/src/acquire.rs
+++ b/gix-lock/src/acquire.rs
@@ -4,6 +4,7 @@ use std::{
     time::Duration,
 };
 
+use gix_error::ErrorExt;
 use gix_tempfile::{AutoRemove, ContainingDirectory};
 
 use crate::{DOT_LOCK_SUFFIX, File, Marker, backoff};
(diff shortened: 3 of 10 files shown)


ameyypawar:
Small update: while waiting on this one I kept going leaf-first and have the rest of batch 1 converted and validated locally — gix-path, gix-attributes, gix-url, gix-zlib, gix-packetline and gix-hash, one crate per branch, each with the affected test suites and the full clippy/fmt matrix green.

Would you prefer them as sequential PRs after this review, or as the one overview-PR with one crate per commit that you mentioned, for cherry-picking? Happy either way.

Byron:
That's exciting! And I am already afraid of the review, but it must be done :D.

Let's keep this PR as is, but you can create a new one with all the additional work you have done. That way I can take a look as well.
If you trust your agent to do the right thing, feel free to keep it going on this task as well. It would be interesting to see it all done and probe it as a whole, early.
If I find something, and agent can always go in and fix it everywhere else, so I am open to experimenting what a good review workflow would be here.

Byron:
@ameyypawar - let's move over to the big PR, and put the commit here presumably at the bottom.
That way there is just one PR to work with and I will sort out how to review and merge it from there.
Thanks again!

ameyypawar:
Moved into #2716 as the bottom commits, as requested — closing in favor of the one PR.

PR #2713, opened 2026-07-13
fix: avoid nested Cargo in gix-filter driver tests

Byron:
## Tasks

- [x] refackiew

----

Everything below this line was generated by Codex GPT-5.

## Reported issue

Fix the flaky `gix-filter` driver tests by eliminating nested `cargo run` invocations.

Root cause:

- `gix-filter/tests/filter/driver.rs` defines `DRIVER` as `cargo run --example arrow`.
- Nextest runs tests in separate concurrent processes, so `serial_test::serial` does not coordinate them.
- Concurrent Cargo commands race over `target/debug/examples/arrow`, occasionally leaving Cargo unable to execute the path it just built.

Implement the Cargo-native solution:

1. Make the Arrow filter helper available as a binary target built with the integration tests.
2. Obtain its absolute path using `env!("CARGO_BIN_EXE_<binary-name>")`.
3. Replace every test-side `cargo run --example arrow` command with direct execution of that binary.
4. Quote or escape the executable path correctly for the shell-command strings used by `gix-filter`, including paths containing spaces and Windows paths. Follow existing repository quoting utilities or patterns.
5. Preserve the existing Arrow helper behavior and, if practical, its example usability without compiling duplicate source targets or producing Cargo warnings.
6. Check the packaging implications. `gix-filter` currently has a restrictive `include` list, so ensure any declared binary source is included and `cargo package` remains valid. Avoid exposing an unintended user-facing executable if a cleaner test-helper arrangement is possible.
7. Once nested Cargo is gone, determine whether `serial_test` is still needed by this test suite. Remove the annotations and `gix-filter` dev-dependency if they no longer protect any genuinely shared state.

Add or retain clear comments explaining why tests must use `CARGO_BIN_EXE_*` instead of invoking Cargo recursively.

Validate with:

- The focused `gix-filter` integration tests.
- A nextest stress run of the driver tests with concurrency, for example at least 20 iterations and six test threads.
- `cargo check -p gix-filter --tests --examples`.
- An appropriate package validation for `gix-filter`.
- `cargo fmt --check` or the repository’s normal formatting command.
- `git diff --check`.

Finally report:

- The chosen binary/helper layout and why.
- Whether the example interface was preserved.
- Whether `serial_test` was removed.
- Exact validation commands and results.

Do not paper over the race with retries, sleeps, or nextest-wide serialization. The goal is for tests to execute a binary already built by Cargo before test execution begins.

## Implementation

- Added a feature-gated `gix-filter-test-arrow` binary target whose path is supplied to integration tests through `CARGO_BIN_EXE_gix-filter-test-arrow`.
- Enabled the helper only from a self dev-dependency so ordinary builds and installs do not select the internal binary.
- Shell-quoted the absolute helper path with `gix_quote::single` after repository-standard Windows separator normalization.
- Kept `cargo run -p gix-filter --example arrow` usable through a tiny wrapper over the single helper implementation.
- Removed all driver-test serialization annotations and the `serial_test` dev-dependency.

## Validation

- `cargo test -p gix-filter --test filter` — 51 passed.
- 20 iterations of `cargo nextest run -p gix-filter --test filter -E 'test(/^driver::/)' --test-threads=6` — all 260 driver-test executions passed.
- `cargo check -p gix-filter --tests --examples` — passed.
- `cargo build -p gix-filter --example arrow` — passed.
- `cargo package -p gix-filter --allow-dirty --features sha1` — packaged and verified successfully.
- `cargo test -p gix-worktree-state --test worktree-state checkout::delayed_driver_process -- --exact` — passed.
- `cargo test -p gix-worktree-stream --test stream from_tree::will_provide_all_information_and_respect_export_ignore -- --exact` — passed.
- Clean ordinary `cargo build -p gix-filter --features sha1` artifact inspection — omitted `gix-filter-test-arrow`.
- `cargo fmt --check` — passed.
- `git diff --check` — passed.


gix-filter/src/driver/apply.rs
```diff
@@ -118,16 +118,29 @@ impl State {
                 //
                 // Solution: Read all data into a buffer, then spawn a thread to write it to stdin
                 // while we can immediately read from stdout.
+                //
+                // TODO(perf): This keeps the entire input in memory until the writer is done. For
```
Byron, line 122: This TODO might be interesting for you @ameyypawar.

PR #2716, opened 2026-07-15
Remove `thiserror` from every crate but `gix`, erasing error types at the `gix` boundary

ameyypawar:
Removes `thiserror` from every crate but `gix`. 67 of 68 crates are free of it; `gix` still has 42 error types concrete and keeps the dependency.

Per review feedback here, the conversion is three-tier rather than a uniform erasure:

- **Plumbing crates** drop `thiserror` and keep concrete enums with hand-written `Display`/`Error` impls, no `gix-error` dependency. Eight crates already converted to `Exn` were reverted to this shape.
- **`gix-error` adopters** — 14 crates use its types directly in their own error surface.
- **The `gix` boundary** erases to `pub type Error = gix_error::Error;` where callers don't need to match variants. 101 types.

The 42 that remain each carry a `TODO(review)` naming the blockers that apply — callers matching variants, an E0119 spent slot on a parent enum, a generic parameter an alias can't express, or an E0117 orphan impl. Some are blocked several ways at once. `etc/plan/gix-error.md` has the detail, including a double-wrap trap where `from_error` on an already-erased callee nests one error inside another and still compiles and passes the suite.

`gix::config::tree::Key::validate()` is a public trait method, so erasing `validate::Error` changes what `Key` implementors have to satisfy — the only trait affected.

On size: the branch adds roughly 6,000 lines net, so it went up rather than down. That sits almost entirely in the plumbing crates, where dropping `thiserror` meant hand-writing the `Display` and `Error` impls it used to generate — `gix-pack` +770, `gix-ref` +669, `gix-filter` +494 and so on. `gix` and `gitoxide-core` sources together come to -318.

Linear history, no merge commit.


Byron:
Thanks a lot!

When reading the review notes, which I find very valuable, I'd hope that these places to watch out for are marked in code in some way. Maybe `TODO(review)` or `TODO` where there is known shortcoming (for instance, when the io error isn't in the error chain anymore).

Besides that, and if you think the agent does a good job, I think this can continue to spike the whole task, the complete removal of `thiserror`. I'd be interested to see rewards to happen in `gix` which now can `?` out of everything, as it is supposed to use `gix-error::Error` as `anyhow`-like type.

Lastly, something to think about, is to see which other standard-errors should exist as part of `gix-error::Error`. The idea is that it will provide utility functions like `can_retry()` and implement via error-tree analysis what previously was a huge match.

Thanks again for all your work!

ameyypawar:
Done on all counts:

- `gix-lock` moved over from #2711 as the bottom commits, as requested — its `gix-ref` adaptation now rides the `feat!` commit so every commit builds; the branch was rebuilt accordingly (PR description updated with the full 12-commit table).
- `TODO(review)` markers are in at the known shortcomings: the io-cause gap through unconverted wrappers (`gix-path/src/realpath.rs`), the `Exn` alias that no longer implements `std::error::Error` for out-of-tree callers (`gix-url/src/expand_path.rs`), and the hand-preserved transparency (`gix-hash/src/io.rs`). I'll keep marking shortcomings that way as the conversion proceeds.

I'll continue with the remaining batches toward complete `thiserror` removal, leaf-first as before — and I like the `can_retry()` idea a lot: once enough of the tree is converted I'll survey what the big error-matches across the workspace actually discriminate on (spurious/retryable being the obvious first), and propose which standard errors and utility functions `gix_error::Error` should grow to replace them. The `gix` crate itself (`?` out of everything) stays the milestone to aim at.

Also: the earlier `test-fixtures-windows (windows-latest)` failure looks like the runner died mid-run (log never uploaded, comparison steps never executed; the arm leg passed in 16m on the same commit) — the new push will re-run it anyway.

Byron:
Fantastic, thank you!

Please do feel free to put the review markers into their respective commits, and to rewrite them at will. Some future work might want you to change bigger bits, and rewriting is fine with me. I will let you know once I start working on this PR so we don't step on each other feet.

One more note: If you think that converting multiple related crates at once would be beneficial, please feel free to do so. My plan is to look at the whole PR at some point, and try to find my way from there.

Byron:
It would also be helpful if the merge-commit could be removed for a clean patch-stack.

Besides that, I would have hoped for more lines to be removed, but maybe that changes once `gix` fully uses its `crate::Error` type.

ameyypawar:
The final result will land as a clean linear patch-stack (one commit per crate, no merge commits) — I'll flatten it as the last step before it leaves draft.

ameyypawar:
The branch is now the clean patch-stack: 29 `feat!: remove thiserror from <crate>` commits (leaf-first, gix-lock at the bottom) plus one trailing chore commit; no merge commits.

Deliberately not in this PR:
- gix-config + gix-submodule — kept at main; their conversions predate the lifetime-free config refactor and are best redone on top of it (noted in the plan file).
- gix-protocol (#2375 pending) and gix-fs — no conversions yet.
- `gix` itself — I'd convert it fresh against current main as a follow-up PR; that's where the `?`-out-of-everything reward lands. Happy to pair on it.

Converted files still carry the allow↔expect lint-suppression drift from your lint pass (left for one sweep rather than touching every crate); the substantive bits my stale branches had reverted (gix-merge's binary-pick refactor, gix-odb's PartialOrd + allowances) are restored.

CI stays red until `gix` converts, as before.

Byron:
For this to really come to fruition, I think it will need `gix` to also remove `thiserror`. Right now, we are at +5k lines of code for this, and I'd hope that this can be redeemed on the caller side, `gix`, which should have a simple time just using `gix::Error` for everything.

All other plumbing crates that weren't converted should also be converted, no need to avoid conflicts. Conflicts are easy to fix nowadays.

Thanks again!

gix/src/clone/checkout.rs
```diff
@@ -92,17 +66,35 @@ pub mod main_worktree {
...
             })?;
 
             let root_tree_id = match &self.ref_name {
-                Some(reference_val) => Some(repo.find_reference(reference_val)?.peel_to_id()?),
-                None => repo.head()?.try_peel_to_id()?,
+                Some(reference_val) => Some(
+                    repo.find_reference(reference_val)
+                        .or_raise(|| gix_error::message("The HEAD reference could not be located"))?
+                        .peel_to_id()
+                        .map_err(gix_error::Error::from_error)?,
```
Byron, line 81: This explicit mapping should never be explicitly required. But maybe you plan to do more sweeps to remove these.
`peel_to_id()` previously returned a thiserror type, and now that would be an Exn type, for which `Error`.

Also also, please optimise for size, so  these common error types should be imported into the module to avoid long names.

Byron (commented):
I took a first look because these line counts didn't go into the right direction, and have some notes.

I'd think if done right, it won't be more than +30% of lines give or take, accounting for formatting and occasional double-usage of the same error message.

Maybe in some cases, it also makes sense to keep a custom Error type, which just isn't implemented with `thiserror`, which is what I think you already did as well.

In any case, don't let me disturb you, I know you are planning more sweeps. And even if not, it's something I can do too once I take over.

PS: I also saw that in many places, the underlying error is still a custom one that just doesn't use `thiserror`. Probably fine as first step, but the idea is to only do this when it's truly required or greatly beneficial.

Byron:
It feels like this PR is a bit troubled, and something that should help reduce scope is to focus on `gix` and its usage of `gix::Error` with direct error forwarding.
This way, at least, there should be a clear reduction of complexity.
In the plumbing crates, let's just leave the original expanded, hand-implemented error types for now, instead of bringing in `gix-error::Exn` there. That way the PR will be far easier to review and can merge quickly.

Byron:
Thanks a lot! From what I can tell, there is only one manifest still needing `thiserror`, and that's `gix` and `gitoxide-core`. In `gitoxide-core` it can easily be removed in favor of `gix-error::Error` or `gix::Error`, `thiserror` has no business there.

For `gix`, it should be the same.

Is there anything in the way of finishing this? If not, I will be looking forward to giving it the last push, I will do that with priority.

Thanks again for all the time you put into this!

ameyypawar:
Nothing blocks you — my local commits sit on the PR tip as a clean fast-forward, and `gix` checks and tests pass.

On scope, so I don't oversell it: I've converted 17 of gix's ~90 error types to `gix::Error` so far (in `repository`, `reference`, `submodule`); ~74 `thiserror` derives still remain, so this is a partial start, not most of it. Six types I left concrete with a `TODO(review)` each — erasing them would give a parent enum that already `#[from]`s an erased type a second `From<gix_error::Error>`, which collides (E0119); the note on each names the exact pair.

Happy to push it as a head start, or to leave `gix` to you entirely if you'd rather do a clean sweep — your call, since you're taking the last push.

Byron:
I am not sure we talk about entirely the same thing, but from what I see, it should be possible to convert as much of `gix` and `gitoxide-core` as possible, and leave the rest with TODO's.

When done, and you think that's it reasonably, the PR can come out of draft which is my cue.

ameyypawar:
Ready for review.

`gitoxide-core` is done. In `gix`, 101 error types are erased to `gix_error::Error` and 42 are still concrete, each with a `TODO(review)` saying why. The four blockers, and a double-wrap trap worth knowing about, are written up in `etc/plan/gix-error.md`.

On the line count you raised: the branch adds roughly 6,000 net, so it went up rather than down. Nearly all of that is the plumbing crates, where dropping `thiserror` meant hand-writing the `Display` and `Error` impls it used to generate — `gix-pack` +770, `gix-ref` +669, `gix-filter` +494 and so on. `gix` and `gitoxide-core` sources come to -318 between them. So the caller side does reduce, just nowhere near enough to pay for the plumbing.

The 42 isn't a floor. The E0119 blocker depends on order: erasing a hub enum frees everything it pinned, which is how nine types that looked blocked turned into ten erasures.

Merge commit is gone — the history is linear now, and the tree hash is unchanged from before the rebase.

Two things worth flagging:

- `gix::config::tree::Key::validate()` is a public trait method, so erasing `validate::Error` changes what `Key` implementors have to satisfy. It's the only trait affected.
- `from_error` won't take `Box<dyn Error + Send + Sync + 'static>` since std's blanket impl is `Sized`-only, so seven sites bridge through `io::Error::other`. A `from_boxed` would drop all of them.

The branch is behind `main` and conflicts in `Cargo.lock`, `gix-fs/Cargo.toml`, `gix-lock/Cargo.toml`, `gix/src/diff.rs` and `gix/src/filter.rs`. Happy to rebase if you'd like.

I'm around and happy to pick up whatever's useful — more of `gix`, the conflicts, or anything that comes out of your review.

Byron:
Thanks a lot! A rebase would be appreciated, but is no must.

Byron:
Thanks for giving it a shot!

I am closing it in favor of #2847 as it already shows diff-lines that are much closer to what I'd expect, so I will be iterating there, soon hopefully to get rid of these huge PRs.

PR #2832, opened 2026-07-25
feat: accelerate dirwalks with UNTR caches

Byron:
## Tasks

- [ ] refackiew

----

Everything below this line was generated by Codex GPT-5.

## Reported issue

$issue-full-auto Look at the goal of https://github.com/GitoxideLabs/gitoxide/pull/2503 (but not the code) and use it to speed up dirwalks in `gix-dir`, which should then naturally speed up `gix status` which uses it if `git config` allows.

Start off by adding a benchmark in `gix-dir` to get a baseline for a dirwalk in some synthetic scenarios. Use `gix-testtools` to setup the initial repositories for the benchmark and focus on common scenarios, at least one of which that would be expected to benefit from UNTR caches, so you can provide before/after values in the commit(s).

Testing is important, and while fixtures in `gix-dir` should now always create an UNTR cache, you should re-run the test-suite (via `justfile`) so that it actually uses it for all tests. Feel free to add a special test which compares without/with UNTR cache by looking at the then reduced IO-counts in the returned outcome.

While changes are isolated, create one commit per crate.

## Summary

- let eligible `gix-dir` walks validate and replay Git's existing UNTR index extension before falling back to physical directory reads
- honor `core.untrackedCache` in repository dirwalk and status paths, with `false` as the explicit opt-out and validation of the effective global excludes file
- generate UNTR data for the `gix-dir` fixture suite and cover reduced I/O, invalidation, and case-insensitive tracked descendants
- add a Criterion benchmark backed by synthetic `gix-testtools` repositories

## Benchmark

| Scenario | Before | After | Change |
| --- | ---: | ---: | ---: |
| clean-flat (10,000 tracked files) | 5.0832 ms | 116.58 µs | -97.7% |
| clean-wide (100 × 100 tracked files) | 5.9694 ms | 277.07 µs | -95.4% |
| untracked-wide (100 × 100 untracked files) | 5.8043 ms | 456.90 µs | -92.1% |

## Validation

- `GIX_TEST_IGNORE_ARCHIVES=1 just nextest -p gix-dir` (79 passed)
- `just nextest -p gix` (366 passed)
- `cargo clippy -p gix-dir --all-targets --all-features -- -D warnings`
- `cargo clippy -p gix --all-targets -- -D warnings`
- SHA-1/SHA-256 feature checks for `gix-dir`; minimal SHA-1 status feature check for `gix`
- `cargo fmt --all -- --check`
- `git diff --check`
- one Codex review per final commit hash

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1839,6 +1839,7 @@ name = "gix-dir"
 version = "0.28.0"
 dependencies = [
  "bstr",
+ "criterion",
  "gix-discover",
  "gix-fs",
  "gix-ignore",
@@ -1851,6 +1852,7 @@ dependencies = [
  "gix-utils",
  "gix-worktree",
  "pretty_assertions",
+ "rustix",
  "thiserror 2.0.18",
 ]
 
--- a/gix-dir/Cargo.toml
+++ b/gix-dir/Cargo.toml
@@ -15,6 +15,11 @@ include = ["/src/**/*", "/LICENSE-*"]
 doctest = false
 test = false
 
+[[bench]]
+name = "dirwalk"
+harness = false
+path = "./benches/dirwalk.rs"
+
 [features]
 ## Enable support for the SHA-1 hash by forwarding the feature to dependencies.
 sha1 = ["gix-index/sha1"]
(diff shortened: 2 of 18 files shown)


Byron:
But the prompt should be reused.

PR #2835, opened 2026-07-25
gix-transport: add a cooperative interrupt flag to http::Options

nikicat:
Adds an optional `Arc<AtomicBool>` to `http::Options` that the curl backend polls from its transfer-meter callback, aborting the in-flight request as `std::io::ErrorKind::Interrupted` as soon as the flag reads `true`.

### Why the existing `should_interrupt` isn't enough

The operation-level `should_interrupt` threaded through `fetch`/`receive` is only polled *between* reads by the protocol layer. A transfer parked in a blocking read on an idle or slow socket never reaches that check, so a caller's Ctrl+C can't stop it short of the `low_speed_time` idle timeout.

curl's transfer meter fires even while no bytes move (roughly once a second), so returning "abort" from it is what makes cancellation prompt. The body `write()` callback can't serve this purpose — it is data-driven and never fires while the socket is silent.

The meter is enabled only when a flag is actually set, so transfers without one keep curl's progress machinery off entirely. Only the curl backend honors the flag; the reqwest backend ignores it, as it already does for `low_speed_*`.

### Breaking change

`http::Options` is constructible via a struct literal (it is not `#[non_exhaustive]`), so dependents that spell out every field must add the new one — hence the `feat!` subject. The one in-tree exhaustive destructure (`gix`'s `transport_options` test) is updated here, asserting the flag is never populated from configuration.

### Notes

- Independent of #2706 — that PR adds a `download_progress` counter to the same struct, but the two touch different curl callbacks and either can land first. Rebasing whichever lands second is a one-line context fix.
- Orthogonal to the ref-performance PRs I closed earlier today per your reftable plan: this is transport-level cancellation, which reftable doesn't affect.
- Verified against current main: `gix-transport` 48/48 with `http-client-curl`, `gix` 442/444 with `blocking-network-client,blocking-http-transport-curl`. The 2 failures are `fetch_pack{,_without_local_destination}`, which fail identically on unmodified main in my environment (local git version regenerates the fixtures; per DEVELOPMENT.md hashes follow ubuntu-latest's git).

### DIFF
--- a/gix-transport/src/client/blocking_io/http/curl/remote.rs
+++ b/gix-transport/src/client/blocking_io/http/curl/remote.rs
@@ -539,18 +539,30 @@ pub fn new() -> Worker {
                 if let Some((action, authenticate)) = proxy_auth_action {
                     authenticate.lock().expect("no panics in other threads")(action.erase()).ok();
                 }
-                let err = Err(io::Error::new(
-                    if err.is_aborted_by_callback() {
-                        // `Handler::progress` returned `false` because
-                        // `should_interrupt` was set — surface it as an interrupt.
-                        std::io::ErrorKind::Interrupted
-                    } else if curl_is_spurious(&err) {
-                        std::io::ErrorKind::ConnectionReset
-                    } else {
-                        std::io::ErrorKind::Other
-                    },
-                    err,
-                ));
+                // Only `Handler::progress` and `Handler::read` can abort by callback, so the flag
+                // is what tells cancellation apart from an upload whose producer failed.
+                let cancelled = err.is_aborted_by_callback()
+                    && handler
+                        .should_interrupt
+                        .as_ref()
+                        .is_some_and(|flag| flag.load(Ordering::Relaxed));
+                let err = Err(if cancelled {
+                    // Deliberately *not* `ErrorKind::Interrupted`: `read_exact()` and `read_until()`
+                    // retry that kind, and the pipe below carries an error exactly once before it
+                    // ends, so the retry would see EOF and the caller an `UnexpectedEof` or a
+                    // missing-header error instead. `gix_features::interrupt::Read` avoids the kind
+                    // for the same reason.
+                    io::Error::other(format!("Interrupted: {err}"))
+                } else {
+                    io::Error::new(
+                        if curl_is_spurious(&err) {
+                            std::io::ErrorKind::ConnectionReset
+                        } else {
+                            std::io::ErrorKind::Other
+                        },
+                        err,
+                    )
+                });
                 handler.receive_body.take();
                 match (handler.send_header.take(), handler.send_data.take()) {
                     (Some(header), mut data) => {
--- a/gix-transport/src/client/blocking_io/http/curl/remote.rs
+++ b/gix-transport/src/client/blocking_io/http/curl/remote.rs
@@ -3,6 +3,7 @@ use std::{
     io::{Read, Write},
     sync::{
         Arc,
+        atomic::{AtomicBool, Ordering},
         mpsc::{Receiver, SyncSender, TrySendError, sync_channel},
     },
     thread,
@@ -58,6 +59,9 @@ struct Handler {
     /// Caller-provided base URL used to derive the transport-visible base URL after redirects.
     base_url: String,
     redirected_base_url: SharedRedirectedBaseUrl,
+    /// If set, the `progress()` meter aborts the in-flight transfer as soon as
+    /// this flag reads `true`. See `http::Options::should_interrupt`.
+    should_interrupt: Option<Arc<AtomicBool>>,
 }
 
 impl Handler {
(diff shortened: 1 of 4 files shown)


nikicat:
Both findings are valid — I reproduced each against a stalling local server — and both are addressed in `18aca7417`.

**Cancellation errors were swallowed** (first finding). The abort itself worked — curl gave up ~1s after the flag was set — but nothing of it reached the caller: interrupting before the response headers surfaced as `Didn't find 'application/x-git-upload-pack-advertisement' header to indicate 'smart' protocol`, and interrupting mid-body as `UnexpectedEof`. The chain is exactly as described: `gix_features::io::pipe::Reader` carries a channel error once and then ends, while `read_until()` (behind `headers.lines()`) and `read_exact()` (in `gix-packetline`) retry `ErrorKind::Interrupted` by contract, so the retry meets the ended pipe.

Worth noting that making the pipe error persistent would be the wrong fix: `read_exact()` would then retry a sticky `Interrupted` forever, turning a misleading error into a hang. The kind is what has to go, so cancellation now fails with `io::Error::other(format!("Interrupted: {err}"))` — the same reasoning `gix_features::interrupt::Read` already applies. Callers recognize cancellation by the flag they set themselves, and curl's own error text stays visible in the message.

**Upload aborts were misreported** (second finding). Also confirmed: with no flag configured at all, an error injected into the upload pipe came back as `kind: Interrupted` (curl code 42), since `Handler::read()` maps any upload-stream error to `ReadError::Abort`. The mapping is now gated on the flag actually reading `true`. `Handler::header()` always returns `true` and status errors go straight into the header pipe, so `progress()` and `read()` are indeed the only two sources of `CURLE_ABORTED_BY_CALLBACK` to tell apart.

Two regression tests come with it, `client::blocking_io::http::interrupt::{cancellation_reaches_the_caller_as_an_error, an_aborted_upload_is_not_reported_as_cancellation}`; both fail on `02c6d8ed3` with precisely those two symptoms. `gix-transport` is green: 28/28 curl, 42/42 reqwest, 48/48 unit tests, with clippy and fmt clean.

@codex review

Byron:
Thanks!

I presume that the existing `should_interrupt` flag can't be passed because it's not in an `Arc`, and in Curl shared ownership is what's needed, in absence of `should_interrupd: &'static…`?

Just asking to double-check this, as an alternative would be to use something like 

https://github.com/GitoxideLabs/gitoxide/blob/5a88ee740f6e81db550d75b9b94875d0cb71ed37/gix/src/util.rs#L7

Instead of forcing the caller to do something special, one could change the interface to accept reality and pass what's needed. Most applications, for instance, happen to have a `'static` interrupt flag. Servers would not have, but they would use the `Arc` instead.

nikicat:
Your presumption is exactly right, and the interface is changed accordingly in `907c78d05`.

The `Arc` was needed because the curl handle and its `Handler` move into the worker thread spawned in [`curl/remote.rs`](https://github.com/GitoxideLabs/gitoxide/blob/907c78d05/gix-transport/src/client/blocking_io/http/curl/remote.rs#L343), so whatever the transfer polls has to outlive the call that started it. The operation-level flag arrives as `should_interrupt: &AtomicBool` with a caller-scoped lifetime and can't cross that boundary, leaving `&'static` and `Arc` as the two shapes that can.

What I had missed is that `Arc<AtomicBool>` alone doesn't merely inconvenience a caller holding a `static` — it locks them out. `Arc::new(&FLAG)` isn't a thing, and `Arc::new(AtomicBool::new(false))` allocates a *copy*, so the signal handler keeps flipping the original while the transfer polls a flag nobody ever sets, and the abort silently never happens. Accepting reality is the better interface, as you say.

So `should_interrupt` is now `Option<options::OwnedOrStaticAtomicBool>`, named after and shaped like the type you linked:

```rust
pub enum OwnedOrStaticAtomicBool {
    Owned(Arc<AtomicBool>),
    Static(&'static AtomicBool),
}
```

with `From` for both, and `Deref<Target = AtomicBool>` so the two read sites in the handler are untouched. Callers write what they already have:

```rust
opts.should_interrupt = Some((&gix::interrupt::IS_INTERRUPTED).into()); // most applications
opts.should_interrupt = Some(per_request_flag.into());                 // a server
```

I kept it local to `gix-transport` rather than reaching for `gix::util::OwnedOrStaticAtomicBool`, since that one is `pub(crate)` in a crate above this one, and it carries the `private: bool` that only the `parallel` dirwalk needs. If you would rather have one type, moving it down to `gix-features` and having `gix` re-export it works for me — say the word and I'll do it in this PR. Same for the name, if `InterruptFlag` reads better in the option docs.

Test-wise, `client::blocking_io::http::interrupt::a_static_flag_needs_no_reference_count` now covers the `Static` variant end-to-end against the stalling server, while `cancellation_reaches_the_caller_as_an_error` keeps covering the `Arc` one. `gix-transport` is green: 48/48 unit, 29/29 curl, 4/4 reqwest, plus clippy and fmt; `gix`'s `transport_options` tests still pass, and they needed no change since the field name is the same and configuration never sets it.

Byron:
There are contribution guide that requires to disclose if you are not the speaker. And I am calling it - this isn't human and I don't like prodding an agent using the GH comment interface.

I also am now classifying this as enhancement, as it's not exactly a bug it's fixing, just a shortcoming. It's something I can fix myself when I stumble over it again.

Thanks for your understanding (if anybody - anyone with a body - reads this).

nikicat:
Fair point. Those comments were drafted by an agent using my account. I reviewed them for accuracy before posting, but the voice is clearly not my own. I started contributing in May and hadn't revisited CONTRIBUTING.md since, so I missed the update regarding agent impersonation.

I'll label anything agent-written going forward.

PR #2837, opened 2026-07-25
test: assert fetched pack contents rather than pack checksums

### LINKED ISSUE #2836: Fetch tests assert pack checksums that depend on the host zlib implementation (fail with a zlib-ng-linked git)
### Summary

`gix::remote::fetch::blocking_and_async_io::fetch_pack` and `fetch_pack_without_local_destination` assert hardcoded checksums of a pack produced by the *host* `git`. Those checksums depend on which zlib implementation the host `git` is linked against, so they fail on any distribution shipping a **zlib-ng-linked git** (Arch, among others) even though everything gitoxide computed is correct.

The object IDs are unaffected — only the compressed bytes differ — so this is purely an over-specified assertion, not a gitoxide bug.

### Environment

- Arch Linux, `git 2.55.0`
- `ldd /usr/bin/git` → `libz-ng.so.2 => /usr/lib/libz-ng.so.2`

### What fails

```
assertion `left == right` failed          # gix/tests/gix/remote/fetch.rs:800
  left:  Sha1(00126593610d63fa3e65ae72d8db871d81d65cf2)
 right:  Sha1(d07c527cf14e524a8494ce6d5d08e28079f5c6ea)
```

`fetch.rs:800` asserts `write_pack_bundle.index.index_hash`; `fetch.rs:~704` asserts `write_pack_bundle.index.data_hash`. Every other assertion in both tests passes, including `num_objects`.

### Root cause, measured

Fetching from the same fixture with plain `git` reproduces the pack gitoxide received (`pack-82288e7d….pack`, matching the `left` value the test reports). It contains three objects:

| object | raw | deflated by git (zlib-ng) | deflated by stock zlib, level 6 |
|---|---|---|---|
| commit | 219 B | **152 B** | **153 B** |
| tree | 68 B | 71 B | 71 B (identical) |
| blob (empty) | 0 B | 8 B | 8 B (identical) |

One object compresses one byte smaller under zlib-ng, so the pack is 268 B locally versus 269 B on CI — different pack checksum, and therefore a different index checksum.

**Decisive check:** taking those three object payloads out of the locally produced pack, re-deflating them with stock zlib at level 6, and re-hashing the reconstructed pack yields

```
de303ef102bd5705a40a0c42ae2972eb1a668455
```

which is exactly the value `fetch_pack_without_local_destination` expects. Same objects, same order, same headers — only the deflate implementation differs.

### Ruled out

- `pack.threads=1` and `pack.window=0`: no change (the pack is delta-free).
- `pack.compression=0..9`: changes the checksum, but **no level reproduces the expected value** — it is the implementation, not the level.
- Fixture nondeterminism: `gix-testtools` isolates generation with `GIT_CONFIG_NOSYSTEM=1`, `GIT_CONFIG_GLOBAL=/dev/null` and pinned author/committer dates. Object IDs come out identical to what the suite expects — `fetch_shallow_deepen_not_possible` asserts hardcoded IDs from the same fixture and passes.
- Git version drift: these expectations have been unchanged since 2024-10 and survive ubuntu-latest's git upgrades.

### Suggested fix

Assert what the checksums were standing in for — the pack's *contents* — rather than its compressed bytes. `write_pack_bundle.index_path` is already available, so the set of object IDs can be checked directly and is invariant across zlib implementations:

```rust
let idx = gix::odb::pack::index::File::at(
    write_pack_bundle.index_path.as_deref().expect("written to disk"),
    repo.object_hash(),
)?;
let mut ids: Vec<_> = idx.iter().map(|e| e.oid).collect();
ids.sort();
assert_eq!(ids, [/* expected object ids */], "the pack contains exactly these objects");
```

`num_objects`, `pack_version` and `index_version` assertions can stay as they are; only `data_hash`/`index_hash` need replacing. (Note `Entry::crc32` is also over compressed bytes, so it is not a suitable substitute.)

Alternatively the checksum assertions could be skipped when the host `git` is not linked against stock zlib, but asserting object IDs seems strictly better — it is what the test actually cares about and it is stable everywhere.

I am happy to open a PR along these lines if you would like it.

nikicat:
Fixes #2836.

`fetch_pack` and `fetch_pack_without_local_destination` assert the `data_hash`/`index_hash` of a pack produced by the host `git`. Those checksums cover the pack's **compressed** bytes, so they encode which zlib implementation the host `git` links against — the tests fail on any distribution shipping a zlib-ng-linked `git` even though everything gitoxide computed is correct.

Concretely, on such a host the 219-byte commit in that pack deflates to 152 bytes where stock zlib produces 153, so the pack is 268 bytes instead of 269 and both checksums change. Taking the same three object payloads out of the locally produced pack, re-deflating them with stock zlib and re-hashing yields `de303ef102bd5705a40a0c42ae2972eb1a668455` — exactly the value the test expects. Same objects, same order, same headers; only the deflate implementation differs. Full measurements are in #2836.

### Change

Assert the pack's object ids, read back from the index that was just written, instead of its checksums. Object ids hash *uncompressed* content, so they are identical on every host, and they are what these assertions were standing in for.

- new `pack_object_ids()` helper opening `write_pack_bundle.index_path`;
- two shared constants for the expected object sets (3 objects, and the same plus the annotated tag), which also lets the case table in `fetch_pack_without_local_destination` drop its `expected_data_hash` column;
- `num_objects`, `pack_version` and `index_version` assertions are untouched;
- `Entry::crc32` is deliberately not used — it is also computed over compressed bytes and would reintroduce the same fragility.

### Verification

- `cargo nextest run -p gix --features blocking-network-client,blocking-http-transport-curl` → **444 passed, 0 skipped** on a zlib-ng host, where `main` currently gives 442 passed / 2 failed.
- `cargo fmt -p gix -- --check` clean; clippy unchanged from `main`.

The expected ids were captured from actual runs rather than written by hand. Note I can only verify locally on a zlib-ng host — object ids are host-invariant so CI should agree, but that is an inference until it runs here.

### DIFF
--- a/gix/tests/gix/remote/fetch.rs
+++ b/gix/tests/gix/remote/fetch.rs
@@ -28,6 +28,47 @@ mod blocking_and_async_io {
         util::hex_to_id,
     };
 
+    /// The objects `clone-as-base-with-changes` sends on top of what the client already has:
+    /// the new commit, its tree, and the empty blob it adds.
+    pub(crate) const PACK_OBJECTS_WITHOUT_TAG: &[&str] = &[
+        "17f2f33324841332c4f6156b36ecde6e44d1bb23",
+        "4d979abcde5cea47b079c38850828956c9382a56",
+        "e69de29bb2d1d6434b8b29ae775ad8c2e48c5391",
+    ];
+
+    /// As above, plus the annotated tag object included when tags are fetched.
+    pub(crate) const PACK_OBJECTS_WITH_TAG: &[&str] = &[
+        "17f2f33324841332c4f6156b36ecde6e44d1bb23",
+        "4d979abcde5cea47b079c38850828956c9382a56",
+        "978f927e6397113757dfec6332e7d9c7e356ac25",
+        "e69de29bb2d1d6434b8b29ae775ad8c2e48c5391",
+    ];
+
+    /// Map SHA-1 hexes through [`hex_to_id`] — which substitutes SHA-256 ids during
+    /// `GIX_TEST_FIXTURE_HASH=sha256` runs — and sort the result.
+    ///
+    /// Sorting must happen *after* the mapping: ids sort by their raw bytes, so a list sorted in
+    /// SHA-1 space is generally not sorted in SHA-256 space.
+    pub(crate) fn expected_pack_object_ids(sha1_hexes: &[&str]) -> Vec<gix::ObjectId> {
+        let mut ids: Vec<_> = sha1_hexes.iter().copied().map(hex_to_id).collect();
+        ids.sort();
+        ids
+    }
+
+    /// Return the sorted object ids of the pack whose index was written to `index_path`.
+    ///
+    /// Prefer this over asserting the pack's `data_hash`/`index_hash`: those cover the *compressed*
+    /// bytes, which differ between zlib implementations — a `zlib-ng`-linked `git` deflates some
+    /// objects a byte smaller than stock zlib — while object ids hash uncompressed content and are
+    /// therefore identical on every host.
+    pub(crate) fn pack_object_ids(index_path: Option<&std::path::Path>, hash: gix::hash::Kind) -> Vec<gix::ObjectId> {
+        let index = gix::odb::pack::index::File::at(index_path.expect("pack index was written"), hash)
+            .expect("written index can be read back");
+        let mut ids: Vec<_> = index.iter().map(|e| e.oid).collect();
+        ids.sort();
+        ids
+    }
+
     pub(crate) fn base_repo_path() -> String {
         gix::path::realpath(
             gix_testtools::scripted_fixture_read_only("make_remote_repos.sh")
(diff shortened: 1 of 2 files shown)


nikicat:
CI caught a gap in my local verification: the suite also runs with `GIX_TEST_FIXTURE_HASH=sha256`, where `hex_to_id` substitutes SHA-256 ids through `SHA1_TO_SHA256_HASHES`, and the four object ids I introduced had no entry there — hence `SHA-1 17f2f333… wasn't mapped to SHA-256 yet`. I had only run the sha256 mode for `gix-ref`, not for `gix`.

Fixed in 92a86ca12, which also fixes a second problem the first would have masked: object ids sort by their raw bytes, so a list sorted in SHA-1 space is generally *not* sorted in SHA-256 space. Mapping now happens before sorting, via a small `expected_pack_object_ids()` helper.

The mappings are verified against both fixtures rather than inferred by elimination — commit and tree via `rev-parse HEAD` / `HEAD^{tree}` in each, the tag via the `refs/tags/v1.0` it is pointed to by in both, and the empty blob via `hash-object -t blob /dev/null`.

Now green locally in both modes: `-p gix --features blocking-network-client,blocking-http-transport-curl` gives **444 passed, 0 skipped** with sha1 and with `GIX_TEST_FIXTURE_HASH=sha256`.

Also — thanks for the review pointer on #2836, but `$ponytail` doesn't resolve to anything I can find in this repo or the tracker. If you mean the [ponytail](https://github.com/DietrichGebert/ponytail) over-engineering-review skill, I am happy to run it over this diff; could you confirm?

PR #2841, opened 2026-07-26
fix: let GIT_WORK_TREE override core.bare

danielcadev:
## Description

Honor the `GIT_WORK_TREE` environment override when opening a repository configured with `core.bare=true`.

Git treats an explicit `GIT_WORK_TREE` as an effective worktree and reports the repository as non-bare. `gix` previously retained `workdir: None`, which prevented downstream consumers such as Starship from computing repository status.

The override is deliberately limited to `GIT_WORK_TREE`. A `core.worktree` CLI override still does not supersede `core.bare`, matching Git behavior.

Related downstream issue: starship/starship#7609  
Supersedes the downstream workaround proposed in starship/starship#7617.

GitoxideLabs/gitoxide#1985 and #1987 addressed the strict-config error associated with `GIT_WORK_TREE`, but did not cover its interaction with `core.bare`.

## Tests

- Added a regression test for a bare repository with an external `GIT_WORK_TREE`.
- Confirmed the repository becomes non-bare and status observes files in the external worktree.
- Added a negative control proving that `-c core.worktree=...` does not override `core.bare`.
- `cargo fmt --all -- --check`
- Strict Clippy for `gix`
- Complete `cargo test -p gix --no-fail-fast`: 410 passed

## AI assistance

Codex assisted with investigation, implementation, testing, and preparation of this pull request. I reviewed the changes and verified the behavior locally against Git.

### DIFF
--- a/gix-path/src/convert.rs
+++ b/gix-path/src/convert.rs
@@ -260,6 +260,15 @@ pub fn to_windows_separators<'a>(path: impl Into<Cow<'a, BStr>>) -> Cow<'a, BStr
 /// can be exhausted by paths like `../../r`, `None` will be returned to indicate the inability to
 /// produce a logically consistent path.
 pub fn normalize<'a>(path: Cow<'a, Path>, current_dir: &Path) -> Option<Cow<'a, Path>> {
+    normalize_inner(path, current_dir, false)
+}
+
+/// Like [`normalize()`], but treats `..` components beyond the filesystem root as no-ops.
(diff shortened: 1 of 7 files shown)


danielcadev:
Oh, okay, good catch. Thanks! I reproduced it with a relative `GIT_WORK_TREE` that traverses beyond the filesystem root. Git saturates at the root, while my original patch left gix non-bare without a worktree.

I updated the normalization only for the `GIT_WORK_TREE` environment override and added a regression that verifies both the resolved worktree and status traversal. The full gix suite (410 tests), formatting, and strict Clippy pass.

PR #2851, opened 2026-07-28
Bug fix for gix::discover traversal after following symlink into subdir of target

### LINKED ISSUE #2850: Running gix::discover on symlink that points to subdir of target repository leads to traversal of symlink parent directory
### Current behavior 😯

Basically the when one creates a git repo named ```./foo``` with ```./foo/subdir``` and creates symlink ```./link``` to ```./foo/subdir``` and lets ```gix::discover``` search  ```./link``` it then tries to find the .git directory by going to the parent of the link (the current directory ```./```) instead of going to parent of the target the link is pointing to ```./foo```.  This was documented in a TODO in the code.

### Expected behavior 🤔

Reproduction sorry for the codex slop:

```
#!/usr/bin/env -S cargo +nightly -Zscript
---
[package]
edition = "2024"

[dependencies]
gix = { version = "=0.85.0", default-features = false, features = ["sha1"] }
---

#[cfg(unix)]
fn main() {
    use std::{
        fs,
        io::{self, Write},
        os::unix::fs::symlink,
        process,
        time::{SystemTime, UNIX_EPOCH},
    };

    let nonce = SystemTime::now()
        .duration_since(UNIX_EPOCH)
        .expect("system clock must be after the Unix epoch")
        .as_nanos();
    let root = std::env::temp_dir().join(format!(
        "gix-symlink-discovery-repro-{}-{nonce}",
        process::id()
    ));
    let repository = root.join("repository");
    let nested_directory = repository.join("nested");
    let link = root.join("link-to-nested");
    let git_directory = repository.join(".git");

    // Create the smallest repository layout accepted by gix-discover.
    fs::create_dir_all(git_directory.join("objects"))
        .expect("create .git/objects");
    fs::create_dir_all(git_directory.join("refs"))
        .expect("create .git/refs");
    fs::create_dir_all(&nested_directory).expect("create nested directory");
    fs::write(git_directory.join("HEAD"), b"ref: refs/heads/main\n")
        .expect("write .git/HEAD");

    // Put the symlink outside the repository and point it into a nested
    // directory. Its lexical parents therefore do not contain `.git`.
    symlink(&nested_directory, &link).expect("create directory symlink");

    println!("temporary tree: {}", root.display());
    println!("real directory: {}", nested_directory.display());
    println!("symlink:        {}", link.display());

    // Discovery through the real path succeeds by walking up to `repository`.
    let direct = gix::discover(&nested_directory)
        .expect("control check: discovery through the real path must work");
    println!("real path:      OK ({})", direct.workdir().unwrap().display());

    // Prove that the symlink resolves to exactly the working control path.
    assert_eq!(
        fs::canonicalize(&link).expect("canonicalize symlink"),
        fs::canonicalize(&nested_directory).expect("canonicalize real path")
    );

    println!("symlink path:   expected discovery to find the same repository");
    io::stdout().flush().expect("flush diagnostic output");

    // This panics with gix 0.85.0 / gix-discover 0.53.0. Discovery follows
    // `link/.git`, but then searches the lexical parents of `link` instead of
    // the parents of its target.
    let _through_link = gix::discover(&link)
        .expect("BUG: discovery through a symlink to a nested directory failed");
}

#[cfg(not(unix))]
fn main() {
    eprintln!("This reproducer currently requires Unix symbolic links.");
    std::process::exit(2);
}
```

ElectricPulse:


Byron:
Thanks for giving this a shot!
I will get it merged.

PR #2856, opened 2026-07-31
fix: keep the value when a line continuation is followed by an empty line

shuvamk:

### What's wrong (AI)

A `gix-config` value that ends in a line-continuation backslash is silently emptied when the continuation line turns out to be empty. Everything accumulated before the backslash is discarded.

File content, `[core]\n\tk = abc\` plus the bytes in the first column:

| bytes after the continuation backslash | `git config --file f core.k` | `File::raw_value("core.k")` |
|---|---|---|
| `\n\n` (empty line)                | `abc` | `` (empty) |
| `\n; comment`                      | `abc` | `` (empty) |
| `\n` then EOF                      | `abc` | `` (empty) |
| `\n\n[other]\n`                    | `abc` | `` (empty) |
| `\r\n` then EOF                    | `abc` | `` (empty) |
| EOF, no trailing newline           | `abc` | `abc` |
| `\n   \n` (blank line with spaces) | `abc` | `abc` |

Measured with `git 2.50.1 (Apple Git-155)` and `gix-config` at `5708de434`.

This is realistic input: a `.gitconfig` whose last alias line ends in a stray `\`, or a generated file that appends a blank line after a continued value. The failure is silent — no parse error, just a value that becomes empty.

Cause, in `value()` (`gix-config/src/parse/from_bytes/mod.rs:384`): a continued value is emitted as one `ValueNotDone` per chunk terminated by `ValueDone`, and that is what every consumer (`file::impls`, `file::section::body`, `file::access::raw`, …) concatenates. When the final chunk is empty, the function takes its empty-value shortcut and dispatches a bare `Event::Value` instead. A bare `Value` reads as a complete value, so the preceding `ValueNotDone` chunks are dropped. It also breaks the invariant documented on `Event::ValueNotDone` itself — "A `Newline` event usually follows, followed by either `ValueDone`, `Whitespace`, or another `ValueNotDone`".

Rows 1, 2 and 4 are a regression from 91c854e7b (`fix!: remove winnow and replace it with hand-implemented parsers everywhere`), which first shipped in gix-config 0.56.0 — 0.55.0 is the last release with the correct behaviour. I checked this rather than inferring it, by building `91c854e7b` and `f26762623` and running all seven inputs against each. `f26762623` stands in for the release exactly: its `gix-config/src` tree is `b2fda5bc7`, byte-identical to the one tagged `gix-config-v0.55.0`, so that measurement *is* a measurement of released 0.55.0. It is also the *last building ancestor* rather than the parent — the immediate parent `b060eb24a` does not build in isolation, since `gix-object` fails with 4 errors against the changed `gix-actor` API, because this project deliberately splits a breaking change from its "adapt to changes" follow-up.

At `f26762623`, rows 1, 2 and 4 report `abc`; at `91c854e7b` they report empty. Rows 3 and 5 already reported empty at `f26762623`, so those two are long-standing rather than regressed — the fix repairs them as well. Row 6 was a hard parse error before the rewrite and was fixed later by 51279734f.

### What this does

The empty-value shortcut now dispatches `ValueDone` when a continuation has been seen and `Value` otherwise — the same distinction the non-empty path at the bottom of `value()` already makes.

```rust
     let end = value_end.unwrap_or(cursor);
     if end == value_start {
-        dispatch(Event::Value(Span::default()));
+        dispatch(if partial_value_found {
+            Event::ValueDone(Span::default())
+        } else {
+            Event::Value(Span::default())
+        });
```

Everything that reaches the shortcut without a continuation is unchanged, and the same one-line cause is corrected wherever it applied — CRLF continuations, multi-value keys, and continuations of three or more chunks all follow from the single dispatch. Values with no continuation still emit a single `Value`, including implicit booleans and `k =` with nothing after it, because `partial_value_found` is only set on a `\`-newline. Continuation lines that contain only whitespace already worked (the whitespace is trimmed *after* this branch, not before it) and still produce the same events. The consumed input `*i` is untouched — in this branch `end == cursor` always, since `value_end` is only ever set to `cursor`.

Emitting an empty `ValueDone` is not new: the trailing-backslash-at-EOF path a few lines above already does exactly `dispatch(Event::ValueDone(Span::default()))`.

### Tests

- `gix-config/src/parse/from_bytes/tests.rs` — `value::continuation_into_an_empty_line_completes_the_value`, asserting the event stream for `hello\` followed by EOF, by a blank line, and by a comment line.
- `gix-config/tests/config/file/access/read_only.rs` — `multi_line_value_with_empty_continuation_line`, asserting `raw_value("core.k") == "abc"` for six file shapes, including the CRLF variant (which takes the separate `consumed += 1` path) and the one where the next line opens another section, next to the existing `multi_line_value_*` tests.

**Both fail before the change.** Verified by reverting only `gix-config/src/parse/from_bytes/mod.rs` to `main` and keeping the tests:

```
---- parse::from_bytes::tests::value::continuation_into_an_empty_line_completes_the_value ----
assertion `left == right` failed: Git reports `hello`, so the empty continuation line must terminate the accumulated value
  left: [ValueNotDone("hello"), Newline("\n"), Value("")]
 right: [ValueNotDone("hello"), Newline("\n"), ValueDone("")]

---- file::access::read_only::multi_line_value_with_empty_continuation_line ----
assertion `left == right` failed: Git reports `abc` for "[core]\n\tk = abc\\\n\n", as a continuation line that is empty ends the value
  left: ""
 right: "abc"
```

Restoring the file makes both pass.

### What I ran

`just` and `cargo-nextest` are not installed on my machine, so I ran the underlying cargo commands rather than the `just` recipes. Toolchain `rustc 1.95.0`, macOS aarch64.

| command | before (main @ 5708de434) | after |
|---|---|---|
| `cargo test -p gix-config` | 389 passed, 0 failed | 391 passed, 0 failed |
| `cargo test --workspace` | 181 suites, 3664 passed, 0 failed, 22 ignored | 181 suites, 3666 passed, 0 failed, 22 ignored |
| `cargo test -p gix` | — | 413 passed, 0 failed |
| `cargo test -p gix-config-value` | — | 53 passed, 0 failed |
| `cargo clippy -p gix-config --all-targets -- -D warnings -A unknown-lints -A unfulfilled_lint_expectations` | — | 0 warnings |

One caveat on the clippy line: on this 1.95.0 toolchain, clean `main` already fails `-D warnings` with three `unfulfilled_lint_expectations` in `gix-ref/src/lib.rs` (the `#[expect(dead_code, …)]` ref-table scaffolding), which leaks into any crate whose target graph includes `gix-ref`. That is pre-existing and green on CI's newer stable, so I ran clippy with `-A unknown-lints -A unfulfilled_lint_expectations` and left `gix-ref` alone. Everything above is local only — CI has not run on this branch, so these commands are the only pre-merge evidence I have.

Rather than branching inside the shortcut, `value()` could fall through to the common tail and let the existing `if partial_value_found` at the bottom pick the event, with the trimming loop made a no-op on an empty range. That removes the duplicated choice at the cost of touching the hot non-empty path. I went with the smaller diff, but I am happy to switch if you would rather have one place that decides `Value` vs `ValueDone`.

### DIFF
--- a/gix-config/src/parse/from_bytes/mod.rs
+++ b/gix-config/src/parse/from_bytes/mod.rs
@@ -382,7 +382,11 @@ fn value(backing: &[u8], i: &mut &[u8], dispatch: &mut dyn FnMut(Event)) -> Pars
 
     let end = value_end.unwrap_or(cursor);
     if end == value_start {
-        dispatch(Event::Value(Span::default()));
+        dispatch(if partial_value_found {
+            Event::ValueDone(Span::default())
+        } else {
+            Event::Value(Span::default())
+        });
         *i = &input[cursor..];
         return Ok(());
     }
--- a/gix-config/src/parse/from_bytes/tests.rs
+++ b/gix-config/src/parse/from_bytes/tests.rs
@@ -1200,25 +1200,33 @@ mod value {
     #[test]
     fn continuation_into_an_empty_line_completes_the_value() {
         let (remaining, events) = parse(b"hello\\\n").unwrap();
-        assert_eq!(remaining, b"");
         assert_eq!(
             events,
-            vec![value_not_done_event("hello"), newline_event(), value_done_event("")],
-            "Git reports `hello`, so the empty continuation line must terminate the accumulated value"
+            vec![value_not_done_event("hello"), newline_event(), value_done_event("")]
+        );
+        assert_eq!(
+            remaining, b"",
+            "the newline belongs to the continuation value, leaving nothing else"
         );
 
         let (remaining, events) = parse(b"hello\\\n\nb = c").unwrap();
-        assert_eq!(remaining, b"\nb = c");
         assert_eq!(
             events,
             vec![value_not_done_event("hello"), newline_event(), value_done_event("")]
         );
+        assert_eq!(
+            remaining, b"\nb = c",
+            "the first newline belongs to the value, the second is left alone"
+        );
 
         let (remaining, events) = parse(b"hello\\\n; comment").unwrap();
-        assert_eq!(remaining, b"; comment");
         assert_eq!(
             events,
             vec![value_not_done_event("hello"), newline_event(), value_done_event("")]
+        );
+        assert_eq!(
+            remaining, b"; comment",
+            "everything beyond the newline is in the remainder, no matter the newlines"
         );
     }
 
--- a/gix-config/src/parse/from_bytes/tests.rs
+++ b/gix-config/src/parse/from_bytes/tests.rs
@@ -1198,6 +1198,31 @@ mod value {
     }
 
     #[test]
+    fn continuation_into_an_empty_line_completes_the_value() {
+        let (remaining, events) = parse(b"hello\\\n").unwrap();
+        assert_eq!(remaining, b"");
+        assert_eq!(
+            events,
+            vec![value_not_done_event("hello"), newline_event(), value_done_event("")],
+            "Git reports `hello`, so the empty continuation line must terminate the accumulated value"
+        );
+
+        let (remaining, events) = parse(b"hello\\\n\nb = c").unwrap();
+        assert_eq!(remaining, b"\nb = c");
+        assert_eq!(
+            events,
+            vec![value_not_done_event("hello"), newline_event(), value_done_event("")]
+        );
+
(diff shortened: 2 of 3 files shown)


Byron:
Thanks a lot!

I validated that the expectations are correct compared to Git - the fix is minimal.

Issue #1416, opened 2024-06-23
`gix-mailmap` parser treats mappings with old but not new names as an error


emilazy:
### Current behavior 😯

`<b> test <a>` is silently discarded, unlike `<b> <a>` and `test2 <b> test1 <a>`

### Expected behavior 🤔

Old names, new names, and new emails should be treated uniformly; any combination that includes an old email address should parse and function.

### Git behavior

Git treats this entry as valid and maps signatures accordingly:

```
emily@yuyuko ~/gix-mailmap-bug> git init
Initialized empty Git repository in /Users/emily/gix-mailmap-bug/.git/
emily@yuyuko ~/gix-mailmap-bug (main)> echo '<b> test <a>' >.mailmap
emily@yuyuko ~/gix-mailmap-bug (main)> git check-mailmap 'test <a>'
test <b>
```

### Steps to reproduce 🕹

```
emily@yuyuko ~/gix-mailmap-bug (main)> cat src/main.rs
use gix::actor::SignatureRef;
use gix::mailmap;

fn main() {
    let mailmap = mailmap::Snapshot::from_bytes(b"<b> test <a>");
    let signature = SignatureRef {
        name: "test".into(),
        email: "a".into(),
        time: Default::default(),
    };
    dbg!(mailmap.entries(), signature, mailmap.resolve(signature));
}
emily@yuyuko ~/gix-mailmap-bug (main)> cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.12s
     Running `target/debug/gix-mailmap-bug`
[src/main.rs:11:5] mailmap.entries() = []
[src/main.rs:11:5] signature = SignatureRef {
    name: "test",
    email: "a",
    time: Time {
        seconds: 0,
        offset: 0,
        sign: Plus,
    },
}
[src/main.rs:11:5] mailmap.resolve(signature) = Signature {
    name: "test",
    email: "a",
    time: Time {
        seconds: 0,
        offset: 0,
        sign: Plus,
    },
}
```

Byron:
Thanks a lot for reporting and for making it so easy to reproduce!

I think that's the one valid case that I somehow missed.

Once the linked PR is merged, I will create a patch-release with the fix, which should be picked up by recent `gix` crates.

emilazy:
Ah, you beat me to it! I just opened this to have something to track the issue as I worked on a fix myself. Thanks for the extremely quick response!

Byron:
[...]

PR #2857, opened 2026-07-31
fix: don't reject mailmap lines with content after the second email

shuvamk:

While differential-testing `gix-mailmap` against `git check-mailmap` I noticed that Git's own `.mailmap` does not fully parse: 16 of its 305 entries are dropped, and one of `torvalds/linux`'s 949 is too.

### What's wrong (AI)

`gix-mailmap` rejects any line that has content left over after the second `name <email>` pair. Git discards that remainder instead. Because `Snapshot::from_bytes()` goes through `parse_ignore_errors()`, the affected lines are silently dropped and the mapping never happens.

Measured with `git 2.50.1 (Apple Git-155)` against `gix-mailmap` at `b95db7ca8` (`main`), using `.mailmap` verbatim from `git/git@master` and `torvalds/linux@master`:

| `.mailmap` line | query | `git check-mailmap` | `Snapshot::resolve` on `main` | this branch |
|---|---|---|---|---|
| git.git:230 `Philip Oakley <philipoakley@iee.email> <philipoakley@iee.org> # secondary <philipoakley@dunelm.org.uk>` | `Philip Oakley <philipoakley@iee.org>` | `Philip Oakley <philipoakley@iee.email>` | **`Philip Oakley <philipoakley@iee.org>`** | `Philip Oakley <philipoakley@iee.email>` |
| linux:780 `Satya Priya <quic_skakitap@quicinc.com> <quic_c_skakit@quicinc.com> <skakit@codeaurora.org>` | `Nobody <quic_c_skakit@quicinc.com>` | `Satya Priya <quic_skakitap@quicinc.com>` | **`Nobody <quic_c_skakit@quicinc.com>`** | `Satya Priya <quic_skakitap@quicinc.com>` |
| git.git:50 `Dana L. How <danahow@gmail.com> Dana How` | `Nobody <danahow@gmail.com>` | `Dana L. How <danahow@gmail.com>` | **`Nobody <danahow@gmail.com>`** | `Dana L. How <danahow@gmail.com>` |
| git.git:240 `René Scharfe <l.s.r@web.de> Rene Scharfe` | `Nobody <l.s.r@web.de>` | `René Scharfe <l.s.r@web.de>` | **`Nobody <l.s.r@web.de>`** | `René Scharfe <l.s.r@web.de>` |
| git.git:305 `Vitaly "_Vi" Shukela <vi0oss@gmail.com> Vitaly _Vi Shukela` | `Nobody <vi0oss@gmail.com>` | `Vitaly "_Vi" Shukela <vi0oss@gmail.com>` | **`Nobody <vi0oss@gmail.com>`** | `Vitaly "_Vi" Shukela <vi0oss@gmail.com>` |

Whole-file view, same two `.mailmap` files:

| file | entries parsed on `main` | entries parsed on this branch |
|---|---|---|
| `git/git` `.mailmap` (305 non-comment lines) | 289, 16 rejected with `Line N has too many names or emails, or none at all` | 305, 0 rejected |
| `torvalds/linux` `.mailmap` (949 non-comment lines) | 948, 1 rejected | 949, 0 rejected |

And, resolving every name/email that occurs anywhere in those two files through both implementations (904 and 2834 queries, generated from the files themselves and fed to `git check-mailmap --stdin` and to `Snapshot::resolve`):

| file | queries | mismatches vs `git` on `main` | on this branch |
|---|---|---|---|
| `git/git` | 904 | 16 | 0 |
| `torvalds/linux` | 2834 | 2 | 0 |

Cause, in `gix-mailmap/src/parse.rs:48`:

```rust
let (name2, email2, rest) = parse_name_and_email(rest, line_number)?;
if !rest.trim().is_empty() {
    return Err(/* "has too many names or emails, or none at all" */);
}
```

Git's `read_mailmap_line()` (v2.50.1, `mailmap.c:127`) does not look at that remainder at all — the return value of the second `parse_name_and_email()` call is deliberately unused, and the only condition for accepting a line is that the first email exists:

```c
if ((name2 = parse_name_and_email(buffer, &name1, &email1, 0)))
        parse_name_and_email(name2, &name2, &email2, 1);

if (email1)
        add_mapping(map, name1, email1, name2, email2);
```

This is the same class as #1416 (fixed in ca054713f): a line shape Git accepts that `gix-mailmap` treats as an error, and which then disappears silently because the porcelain path uses `parse_ignore_errors()`.

### What this does

Replaces the remainder check with Git's own condition — a line is rejected only when it contains no email at all:

```diff
-    let (name2, email2, rest) = parse_name_and_email(rest, line_number)?;
-    if !rest.trim().is_empty() {
+    let (name2, email2, _) = parse_name_and_email(rest, line_number)?;
+    if email1.is_none() {
```

`parse()` deliberately keeps erroring where it did before for input that is not a mapping at all: `just a name` (no email), `<missing closing brace` (unterminated), `hello < \t\r >` (empty email) and `<email>` (nothing to map to) all still return the same errors on the same line numbers, and the existing error-message tests are unchanged. The only behaviour change is that a line with a usable first email is now accepted instead of rejected; the set of rejected lines strictly shrinks, and nothing that parsed before stops parsing.

The error message for the remaining case is reworded from `has too many names or emails, or none at all` to `does not contain an email`, since "too many" can no longer occur. If you'd rather keep the old wording verbatim, say so and I'll revert that half.

### Tests

`gix-mailmap/tests/parse/mod.rs` gains `trailing_content_after_a_complete_mapping_is_ignored`, covering three of the real-world lines above, including the `Name <email> Old Name` shape. **It fails before the change** — with the fix reverted and the test kept:

```
running 9 tests
test parse::trailing_content_after_a_complete_mapping_is_ignored ... FAILED

---- parse::trailing_content_after_a_complete_mapping_is_ignored stdout ----
thread 'parse::trailing_content_after_a_complete_mapping_is_ignored' panicked at gix-mailmap/tests/parse/mod.rs:140:21:
called `Result::unwrap()` on an `Err` value: Line 1 has too many names or emails, or none at all: "Philip Oakley <philipoakley@iee.email> <philipoakley@iee.org> # secondary <philipoakley@dunelm.org.uk>", at gix-mailmap/src/parse.rs:53

test result: FAILED. 8 passed; 1 failed; 0 ignored; 0 measured; 3 filtered out
```

and with it restored:

```
test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 3 filtered out
```

The other direction — that genuinely unparseable lines still error — is pinned by the pre-existing `error_if_there_is_just_a_name`, `error_if_there_is_just_an_email`, `error_if_email_is_empty` and `line_numbers_are_counted_correctly_in_errors`, none of which were touched.

The fully Git-faithful option is to stop erroring on these lines altogether and have `Lines` skip them, since `read_mailmap_line()` never reports an error for any input. I didn't do that because it would silently drop `line_numbers_are_counted_correctly_in_errors`' whole reason to exist and remove diagnostics that are useful to callers of `parse()` (as opposed to `parse_ignore_errors()`). Happy to switch if you'd rather match Git exactly there too.

This does not close the gap with Git completely, and I'd rather say so than imply it does. Git's second `parse_name_and_email()` never reports failure — on a malformed remainder it clears both outputs and the line still maps — whereas `gix-mailmap` propagates that error with `?`. So these still diverge after this change:

| `.mailmap` line | query | `git check-mailmap` | this branch |
|---|---|---|---|
| `B <b@x> junk <unterminated` | `Nobody <b@x>` | `B <b@x>` | errors, `Missing closing bracket '>' in email` |
| `B <b@x> <` | `Nobody <b@x>` | `B <b@x>` | errors, same |
| `A <a@x> <>` | `Whoever <>` | `A <a@x>` (Git passes `allow_empty_email=1` to that second call) | errors, `Email must not be empty` |

Rows one and two need the second parse to stop propagating errors; the third needs it to allow an empty email as well. That is a different change with its own risk surface, so I left it out of this PR — tell me if you'd like it folded in or filed separately.

### What I ran

`just` and `cargo-nextest` aren't installed here, so these are the raw cargo invocations, on `rustc 1.95.0 (59807616e 2026-04-14)`:

| command | `main` @ `b95db7ca8` | this branch |
|---|---|---|
| `cargo test -p gix-mailmap` | 15 passed | 16 passed |
| `cargo test --workspace` | rc=0 — 181 suites, 3666 passed, 0 failed, 22 ignored | rc=0 — 181 suites, **3667** passed, 0 failed, 22 ignored |
| `cargo clippy --workspace --all-targets -- -D warnings -A unknown-lints -A unfulfilled_lint_expectations` | rc=0 | rc=0, 0 warnings |
| `cargo doc -p gix-mailmap --no-deps` | — | rc=0 |

Note that `cargo clippy --workspace --all-targets -- -D warnings` without `-A unfulfilled_lint_expectations` is already red on clean `main` on this toolchain, with three `unfulfilled lint expectation` errors in `gix-ref/src/lib.rs` (lines 90, 100, 120). That is a rustc-version artefact of my local 1.95.0, unrelated to this change, and I left it alone.

### DIFF
--- a/gix-mailmap/src/parse.rs
+++ b/gix-mailmap/src/parse.rs
@@ -44,13 +44,11 @@ impl<'a> Iterator for Lines<'a> {
 
 fn parse_line(line: &BStr, line_number: usize) -> Result<Entry<'_>, Error> {
     let (name1, email1, rest) = parse_name_and_email(line, line_number)?;
-    let (name2, email2, rest) = parse_name_and_email(rest, line_number)?;
-    if !rest.trim().is_empty() {
-        return Err(ValidationError::new_with_input(
-            format!("Line {line_number} has too many names or emails, or none at all"),
-            line,
-        )
-        .raise());
+    let (name2, email2, _) = parse_name_and_email(rest, line_number)?;
+    if email1.is_none() {
+        return Err(
+            ValidationError::new_with_input(format!("Line {line_number} does not contain an email"), line).raise(),
+        );
     }
     Ok(match (name1, email1, name2, email2) {
         (Some(proper_name), Some(commit_email), None, None) => Entry::change_name_by_email(proper_name, commit_email),
--- a/gix-mailmap/src/parse.rs
+++ b/gix-mailmap/src/parse.rs
@@ -43,8 +43,8 @@ impl<'a> Iterator for Lines<'a> {
 }
 
 fn parse_line(line: &BStr, line_number: usize) -> Result<Entry<'_>, Error> {
-    let (name1, email1, rest) = parse_name_and_email(line, line_number)?;
-    let (name2, email2, _) = parse_name_and_email(rest, line_number)?;
+    let (name1, email1, rest) = parse_name_and_email(line, line_number, false)?;
+    let (name2, email2, _rest) = parse_name_and_email(rest, line_number, true).unwrap_or((None, None, rest));
     if email1.is_none() {
         return Err(
             ValidationError::new_with_input(format!("Line {line_number} does not contain an email"), line).raise(),
@@ -77,6 +77,7 @@ fn parse_line(line: &BStr, line_number: usize) -> Result<Entry<'_>, Error> {
 fn parse_name_and_email(
     line: &BStr,
     line_number: usize,
+    allow_empty_email: bool,
 ) -> Result<(Option<&'_ BStr>, Option<&'_ BStr>, &'_ BStr), Error> {
     match line.find_byte(b'<') {
         Some(start_bracket) => {
@@ -85,7 +86,7 @@ fn parse_name_and_email(
                 ValidationError::new_with_input(format!("{line_number}: Missing closing bracket '>' in email"), line)
             })?;
             let email = email[..closing_bracket].trim().as_bstr();
-            if email.is_empty() {
+            if email.is_empty() && !allow_empty_email {
                 return Err(
                     ValidationError::new_with_input(format!("{line_number}: Email must not be empty"), line).raise(),
                 );
--- a/gix-mailmap/tests/parse/mod.rs
+++ b/gix-mailmap/tests/parse/mod.rs
@@ -87,19 +87,36 @@ fn valid_entries() {
 #[test]
 fn trailing_content_after_a_complete_mapping_is_ignored() {
     assert_eq!(
-        line("Philip Oakley <philipoakley@iee.email> <philipoakley@iee.org> # secondary <philipoakley@dunelm.org.uk>"),
-        Entry::change_name_and_email_by_email("Philip Oakley", "philipoakley@iee.email", "philipoakley@iee.org"),
+        line("Canonical Name <canonical@example.com> <mapped@example.com> # secondary <ignored@example.com>"),
+        Entry::change_name_and_email_by_email("Canonical Name", "canonical@example.com", "mapped@example.com"),
         "only the first two names and emails are used to build the mapping"
     );
     assert_eq!(
-        line("Satya Priya <quic_skakitap@quicinc.com> <quic_c_skakit@quicinc.com> <skakit@codeaurora.org>"),
-        Entry::change_name_and_email_by_email("Satya Priya", "quic_skakitap@quicinc.com", "quic_c_skakit@quicinc.com"),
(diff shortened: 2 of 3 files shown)


Byron:
Thanks a lot, great catch!

PR #2865, opened 2026-08-01
fix: split attributes on ASCII blanks only, and reject empty names, like in Git.

shuvamk:
*Disclosure per the [agent impersonation policy](https://github.com/GitoxideLabs/gitoxide/blob/main/CONTRIBUTING.md): an AI agent (Claude Opus 5) wrote this PR and its description, speaking through the `shuvamk` account.*

### What's wrong (AI)

`gix-attributes` tokenises an attribute line on Unicode whitespace, and accepts an attribute with an empty name. Git does neither, so a `.gitattributes` line that Git reads differently — or discards outright — can yield extra attributes in gitoxide.

Measured with `git version 2.50.1 (Apple Git-155)` against `gix` built from `main` @ `da5fa7359`, one `.gitattributes` line at a time in the same repository. `<U+00A0>`, `<U+000B>`, `<U+000C>`, `<U+2028>`, `<TAB>` and `<CR>` stand for the raw characters, both in the file and in the quoted output.

| `.gitattributes` | `git check-attr -a -- <path>` | `gix attributes query <path>` on `main` | with this PR |
|---|---|---|---|
| `*.txt text=auto<U+00A0>eol=lf` | `text: auto<U+00A0>eol=lf` | `text=auto` **and** `eol=lf` | `text=auto<U+00A0>eol=lf` |
| `p text<U+00A0>eol=lf` | `text<U+00A0>eol is not a valid attribute name` → nothing | `text` **and** `eol=lf` | nothing |
| `p a<U+000B>b` | `a<U+000B>b is not a valid attribute name` → nothing | `a` **and** `b` | nothing |
| `p a<U+000C>b` | `a<U+000C>b is not a valid attribute name` → nothing | `a` **and** `b` | nothing |
| `p a<U+2028>b` | `a<U+2028>b is not a valid attribute name` → nothing | `a` **and** `b` | nothing |
| `p text =lf eol=lf` | ` is not a valid attribute name` → nothing | `text`, `=lf` **and** `eol=lf` | nothing |
| `p =` | ` is not a valid attribute name` → nothing | `=` | nothing |
| `p -` | ` is not a valid attribute name` → nothing | `-` | nothing |
| `p !` | ` is not a valid attribute name` → nothing | `!` | nothing |
| `p a b` | `a: set`, `b: set` | `a` **and** `b` | unchanged |
| `p a<TAB>b<CR>` | `a: set`, `b: set` | `a` **and** `b` | unchanged |

"→ nothing" means Git writes that message to stderr, exits 0, and the path ends up with no attributes at all, because the whole line is discarded.

The first row is the one that reaches a realistic file: a non-breaking space arrives in `.gitattributes` by pasting a line from a web page or from an editor with smart spacing. Byte-exact, so that the U+00A0 is not mistaken for a plain space:

```console
$ printf '*.txt text=auto\xc2\xa0eol=lf\n' > .gitattributes
$ git check-attr -a -- a.txt | od -c | head -2
0000000    a   .   t   x   t   :       t   e   x   t   :       a   u   t
0000020    o 302 240   e   o   l   =   l   f  \n
```

One attribute, `text`, whose value contains the U+00A0. `gix attributes query a.txt` reports two, and invents an `eol=lf` that Git never sets.

The cause is two things in `gix-attributes/src/parse.rs`:

1. `Iter::new()` used `bstr`'s `input.fields()`, which splits on `char::is_whitespace`. Git's `attr.c` tokenises with a fixed byte set:

   ```c
   static const char blank[] = " \t\r\n";
   ```

   `parse_attr()` takes `ep = cp + strcspn(cp, blank)` as the end of a token, so anything outside that set is an ordinary name or value byte. The rest of this parser already agrees with Git: `skip_blanks()` and the pattern/attributes split in `parse_line()` both go through the file's own `BLANKS` constant (`b" \t\r"` — `\n` is already consumed by `bstr::Lines`). Only `Iter` did not.

2. `attr_valid()` had no length check, and `attr.bytes().all(…)` is vacuously true on an empty slice, so `=lf` (empty name) and `-` / `!` (empty after the prefix is stripped) were all accepted. Git rejects them in `attr_name_valid()`:

   ```c
   if (namelen <= 0 || *name == '-')
   	return 0;
   ```

   and its caller treats a first-pass failure as fatal for the entire line:

   ```c
   /* First pass to count the attr_states */
   for (cp = states, num_attr = 0; *cp; num_attr++) {
   	cp = parse_attr(src, lineno, cp, NULL);
   	if (!cp)
   		goto fail_return;
   }
   ```

### What this does

`Iter` splits on `BLANKS` instead of on Unicode whitespace, and `attr_valid()` rejects an empty name. `slice::split` yields an empty field for every run of blanks where `fields()` yielded none, so `Iter::next()` skips them with `find()` rather than `next()`.

An invalid name still surfaces as a `name::Error` from the assignment iterator, and both consumers collect it through `Result`, but they do different things with it. In `gix-attributes`, `into_owned_assignments()` in `src/search/attributes.rs` returns `None`, which `bytes_to_patterns()` `filter_map`s away, so the offending line is dropped and the rest of the file is kept — what `goto fail_return` does. In `gix-pathspec`'s `parse_attributes()` there is no line to drop: the error propagates out and the whole pathspec fails to parse with `Error::InvalidAttribute`, which is also what Git does there, via `die(_("invalid attribute name %s"))`.

Deliberately unchanged: the set of characters permitted *inside* a name, `Lines`, pattern parsing, quoting, and runs of spaces and tabs. The last two rows of the table are identical before and after, and `attributes_are_parsed_behind_various_whitespace_characters` and `trailing_whitespace_in_attributes_is_ignored` already pin that.

`check_attr()` is also used for the `[attr]` macro name, so `[attr]` with nothing after the prefix now yields `Error::MacroName` and the line is dropped, instead of defining a macro named `""`. Git never reaches `attr_name_valid()` for that input — `parse_attr_line()` enters the macro branch only when `strlen(ATTRIBUTE_MACRO_PREFIX) < namelen` (`attr.c:348`), and `"[attr]"` is exactly 6 bytes — so Git parses `[attr] foo` as the ordinary pattern `[attr]`, a bracket expression, and sets `foo` on `a`, `t` and `r`:

| input | `gix_attributes::parse` on `main` | with this PR | `git check-attr -a` |
|---|---|---|---|
| `[attr] foo` | `Macro(Name(""))` with `foo` set | `Error::MacroName`, line dropped | pattern `[attr]`; `foo` set on `a`, `t`, `r`, nothing on `z` |
| `[attr]` | `Macro(Name(""))`, no attributes | `Error::MacroName`, line dropped | pattern `[attr]`, no attributes |
| `[attr]m foo` | `Macro(Name("m"))` with `foo` set | unchanged | macro `m` with `foo` set |

Neither `main` nor this PR matches Git on the first two rows, and I have left that alone — recognising `[attr]` only when something follows it is a separate divergence from tokenisation.

The other caller, `gix-pathspec`'s `attr:` magic, moves in the same direction. Git parses that body in `pathspec.c`, splitting on `' '` and validating each name with `git_attr()`, which fails on the same names `attr_name_valid()` rejects:

| pathspec | `git ls-files --` | `gix_pathspec::parse` on `main` | with this PR |
|---|---|---|---|
| `:(attr:text<U+00A0>eol=lf)f` | `fatal: invalid attribute name text<U+00A0>eol` | `text` **and** `eol=lf` | `InvalidAttribute { attribute: "text<U+00A0>eol" }` |
| `:(attr:a<U+000B>b)f` | `fatal: invalid attribute name a<U+000B>b` | `a` **and** `b` | `InvalidAttribute { attribute: "a<U+000B>b" }` |
| `:(attr:text =lf)f` | `fatal: invalid attribute name ` | `text` **and** an attribute named `""` | `InvalidAttribute { attribute: "" }` |
| `:(attr:-)f` | `fatal: invalid attribute name ` | an unset attribute named `""` | `InvalidAttribute { attribute: "" }` |
| `:(attr:a b)f` | matches `f` | `a` **and** `b` | unchanged |

It does not get all the way there — Git splits an `attr:` body on `' '` alone, so gitoxide still separates on tab and CR where Git would not. That is a separate divergence and I have not touched it here.

### Tests

Two tests in `gix-attributes/tests/attributes/parse.rs`, built on the file's existing `line` / `try_line` / `lenient_lines` helpers:

- `only_ascii_blanks_separate_attributes` — a U+00A0 inside a value stays in the value; U+00A0, U+000B, U+000C and U+2028 inside a name make that name invalid.
- `attribute_names_must_not_be_empty` — `p text =lf`, `p =`, `p -` and `p !` are rejected, and are dropped entirely under lenient parsing.

**Both fail before the change.** With only `gix-attributes/src/parse.rs` reverted and the tests kept:

```
test parse::attribute_names_must_not_be_empty ... FAILED
test parse::only_ascii_blanks_separate_attributes ... FAILED

---- parse::attribute_names_must_not_be_empty stdout ----
thread 'parse::attribute_names_must_not_be_empty' panicked at gix-attributes/tests/attributes/parse.rs:220:5:
a blank in front of the equals sign leaves the assignment without a name

---- parse::only_ascii_blanks_separate_attributes stdout ----
thread 'parse::only_ascii_blanks_separate_attributes' panicked at gix-attributes/tests/attributes/parse.rs:282:5:
assertion `left == right` failed: a non-breaking space belongs to the value it appears in
  left: (Pattern(Pattern { text: "p", mode: Mode(NO_SUB_DIR), first_wildcard_pos: None }), [("text", Value(ValueRef([97, 117, 116, 111]))), ("eol", Value(ValueRef([108, 102])))], 1)
 right: (Pattern(Pattern { text: "p", mode: Mode(NO_SUB_DIR), first_wildcard_pos: None }), [("text", Value(ValueRef([97, 117, 116, 111, 194, 160, 101, 111, 108, 61, 108, 102])))], 1)

test result: FAILED. 32 passed; 2 failed; 0 ignored
```

Restoring the file gives `test result: ok. 34 passed; 0 failed`. No existing assertion needed changing.

There is one more divergence in the same validation path that I deliberately left out. Git also reserves a namespace:

```c
/*
 * Attribute name cannot begin with "builtin_" which
 * is a reserved namespace for built in attributes values.
 */
static int attr_name_reserved(const char *name)
{
	return starts_with(name, "builtin_");
}
```

`parse_attr()` and `parse_attr_line()` call it immediately after `attr_name_valid()`, so such a name discards the line as well:

| `.gitattributes` | `git check-attr -a -- p` | `gix attributes query p`, before and after this PR |
|---|---|---|
| `p builtin_objectmode` | `builtin_objectmode is not a valid attribute name` → nothing | `builtin_objectmode` |

I left it out because it is a name *policy* rather than tokenisation, and gitoxide has no built-in attributes of its own to protect yet, so whether to reserve the namespace is your call. Happy to fold it into this PR or send it separately.

### What I ran

`just` and `cargo-nextest` are not installed here, so these are the raw `cargo` invocations, `rustc 1.95.0`, aarch64 macOS:

| command | result |
|---|---|
| `cargo test -p gix-attributes` | **34 passed, 0 failed** (32 on `main`), plus 1 doctest |
| `cargo test -p gix-attributes -p gix-worktree -p gix-pathspec -p gix-filter -p gix-diff -p gix-archive -p gix-worktree-stream -p gix` | **679 passed, 0 failed** — every crate depending on `gix-attributes` |
| `cargo test --workspace --no-fail-fast`, branch | 181 suites, **3677 passed, 0 failed**, 22 ignored |
| `cargo test --workspace --no-fail-fast`, `main` @ `da5fa7359` | 181 suites, 3671 passed, **4 failed**, 22 ignored |
| `cargo clippy -p gix-attributes --all-targets -- -D warnings -A unknown-lints -A unfulfilled_lint_expectations` | rc=0 |

Two honest caveats about that table:

- The 4 workspace failures are on **pristine `main`**, not on this branch, and are environmental: `gix-path`'s `env::tests::system_prefix::*` fail with `tempfile` `Os { code: 2, kind: NotFound }`. They are intermittent on this machine — the same four tests passed in the branch run of the same command, and they pass when `gix-path` is run on its own.
- `main` is also red for me under a bare `-D warnings`, with three `unfulfilled_lint_expectations` at `gix-ref/src/lib.rs:90/:100/:120` that rustc 1.95.0 flags and CI's newer stable does not. They leak into any crate whose target graph pulls `gix-ref`, hence the `-A` above. Nothing in `gix-ref` was touched.

I have not seen CI run on this branch, so all of the above is local only.

### DIFF
--- a/gix-attributes/src/parse.rs
+++ b/gix-attributes/src/parse.rs
@@ -40,13 +40,15 @@ pub struct Lines<'a> {
 
 /// An iterator over attribute assignments in a single line.
 pub struct Iter<'a> {
-    attrs: bstr::Fields<'a>,
+    attrs: std::slice::Split<'a, u8, fn(&u8) -> bool>,
 }
 
 impl<'a> Iter<'a> {
     /// Create a new instance to parse attribute assignments from `input`.
     pub fn new(input: &'a BStr) -> Self {
-        Iter { attrs: input.fields() }
+        Iter {
+            attrs: input.split(is_blank as fn(&u8) -> bool),
+        }
     }
 
     fn parse_attr(&self, attr: &'a [u8]) -> Result<AssignmentRef<'a>, name::Error> {
@@ -66,7 +68,7 @@ impl<'a> Iter<'a> {
 
 fn check_attr(attr: &BStr) -> Result<NameRef<'_>, name::Error> {
     fn attr_valid(attr: &BStr) -> bool {
-        if attr.first() == Some(&b'-') {
+        if attr.is_empty() || attr.first() == Some(&b'-') {
             return false;
         }
 
@@ -83,7 +85,7 @@ impl<'a> Iterator for Iter<'a> {
     type Item = Result<AssignmentRef<'a>, name::Error>;
 
     fn next(&mut self) -> Option<Self::Item> {
-        let attr = self.attrs.next().filter(|a| !a.is_empty())?;
+        let attr = self.attrs.find(|a| !a.is_empty())?;
         self.parse_attr(attr).into()
     }
 }
@@ -163,6 +165,10 @@ fn parse_line(line: &BStr, line_number: usize) -> Option<Result<(Kind, Iter<'_>,
         Err(err) => return Some(Err(err)),
     };
     Ok((kind, Iter::new(attrs), line_number)).into()
+}
+
+fn is_blank(b: &u8) -> bool {
+    BLANKS.contains(b)
 }
 
 const BLANKS: &[u8] = b" \t\r";
--- a/gix-attributes/tests/attributes/parse.rs
+++ b/gix-attributes/tests/attributes/parse.rs
@@ -216,6 +216,46 @@ fn attribute_names_must_not_begin_with_dash_and_must_be_ascii_only() {
 }
 
 #[test]
+fn attribute_names_must_not_be_empty() {
+    assert!(
+        matches!(
+            try_line(r"p text =lf"),
+            Err(parse::Error::AttributeName { line_number: 1, .. })
+        ),
+        "a blank in front of the equals sign leaves the assignment without a name"
+    );
+    assert!(
+        lenient_lines(r"p text =lf").is_empty(),
+        "lenient parsing skips the invalid line entirely"
+    );
+    assert!(
+        matches!(
+            try_line(r"p ="),
+            Err(parse::Error::AttributeName { line_number: 1, .. })
+        ),
+        "an assignment that is nothing but an equals sign has no name either"
+    );
+    assert!(
+        matches!(
+            try_line(r"p -"),
+            Err(parse::Error::AttributeName { line_number: 1, .. })
+        ),
+        "prefixes need a name to apply to"
+    );
+    assert!(
+        lenient_lines(r"p -").is_empty(),
+        "lenient parsing skips the invalid line entirely"
+    );
+    assert!(
+        matches!(
+            try_line(r"p !"),
+            Err(parse::Error::AttributeName { line_number: 1, .. })
+        ),
+        "the unspecified prefix needs one as well"
+    );
+}
+
+#[test]
 fn attributes_are_parsed_behind_various_whitespace_characters() {
     assert_eq!(
         line(r#"p a b"#),
@@ -246,6 +286,51 @@ fn attributes_are_parsed_behind_various_whitespace_characters() {
         line("\"p\" \t a \t b"),
         (pattern("p", Mode::NO_SUB_DIR, None), vec![set("a"), set("b")], 1),
         "behind a mix of space and tab"
+    );
+}
+
+#[test]
+fn only_ascii_blanks_separate_attributes() {
+    assert_eq!(
+        line("p text=auto\u{a0}eol=lf"),
+        (
+            pattern("p", Mode::NO_SUB_DIR, None),
+            vec![value("text", "auto\u{a0}eol=lf")],
+            1
+        ),
+        "a non-breaking space belongs to the value it appears in"
+    );
+    assert!(
+        matches!(
+            try_line("p text\u{a0}eol=lf"),
+            Err(parse::Error::AttributeName { line_number: 1, .. })
+        ),
+        "in a name it makes the whole name invalid"
+    );
+    assert!(
+        lenient_lines("p text\u{a0}eol=lf").is_empty(),
+        "lenient parsing skips the invalid line entirely"
+    );
+    assert!(
+        matches!(
+            try_line("p a\u{b}b"),
+            Err(parse::Error::AttributeName { line_number: 1, .. })
+        ),
+        "a vertical tab is part of the name, not a separator"
+    );
+    assert!(
+        matches!(
+            try_line("p a\u{c}b"),
+            Err(parse::Error::AttributeName { line_number: 1, .. })
+        ),
+        "a form feed is part of the name, not a separator"
+    );
+    assert!(
+        matches!(
+            try_line("p a\u{2028}b"),
+            Err(parse::Error::AttributeName { line_number: 1, .. })
+        ),
+        "vertical tabs, form feeds and unicode line separators aren't blanks either"
     );
 }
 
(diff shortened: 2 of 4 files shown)


Byron:
Thanks a lot!

PR #2871, opened 2026-08-04
fix: peel annotated tags before navigating a rev-spec, like in Git.

shuvamk:
An AI agent (Claude Opus 5) wrote this PR and is speaking through the `shuvamk` account.

### What's wrong (AI)

Navigating with `^`, `~` or `^{/…}` from an **annotated tag** operates on the tag object rather than on the commit it points at. Git routes all three through `GET_OID_COMMITTISH`, which dereferences tags first.

Measured in `complex_graph` — the existing fixture of `make_rev_spec_parse_repos.sh`, where `b-tag` is annotated and names the merge commit `b` — with git 2.50.1 and `gix` at f101ab6:

| spec | `git rev-parse` | `gix` before this PR |
|---|---|---|
| `b-tag` | `212d0f0` (the tag) | `212d0f0` |
| `b-tag^`, `b-tag^1` | `a8fbd8e` | err: `Object 212d0f0 was a tag, but needed it to be a commit` |
| `b-tag^2` | `a28bcc1` | same error |
| `b-tag~1` | `a8fbd8e` | err: `Commit 212d0f0 has 0 ancestors along the first parent…` |
| `b-tag~2` | `9f9eac6` | same shape |
| `b-tag^{/G}` | `9f9eac6` | err: `Expected object of kind commit but got tag` |

`^0` and `^{commit}` already peel, so `b-tag^0` and `b-tag^{commit}~1` were right all along — only the direct forms were affected.

### What this does

Peels to a commit at the three navigation sites in `delegate/navigate.rs` (`NthParent`, `NthAncestor`, `find`), keeping the original object id as the replacement key. Peeling is the identity on a commit, so anchors that already resolved are untouched, including the two error strings pinned in `traverse.rs`.

Blob and tree anchors now report `Last encountered object fe27474 was blob while trying to peel to commit` — the standard `peel_to_kind` message used elsewhere in the crate — in place of the three per-site messages `navigate.rs` carried, one of which called a blob a commit. No test pins any of them.

`<tag>~0` is **not** fixed here and still yields the tag where git yields `5b3f9e2`. `gix-revision`'s parser skips the delegate call entirely when the ancestor count is zero (`function.rs`, `if number != 0`), which is correct for a commit anchor and is pinned by `tilde_symbol::followed_by_zero_is_no_op`. Fixing it changes what a `Navigate` implementation is called with, so it looked like its own change rather than something to bundle here.

### Tests

`traverse::annotated_tags_are_peeled_to_their_commit`, plus six `baseline` lines in the fixture script so the expectations are checked against real `git rev-parse` by the existing harness instead of against hardcoded hashes. The regenerated archive is committed. It fails without the source change:

```
assertion `left == right` failed: b-tag^: left (ours) should match right (git):
Err(delegate.parent(NthParent(1)) failed …
└─ Object 212d0f0 was a tag, but needed it to be a commit)
  left: None
 right: Some(Include(Sha1(a8fbd8e0c8753ef07ccdbfa93da030d53b68f0bc)))
```

`just` and `cargo-nextest` aren't installed here, so directly: `cargo test --workspace` → 3702 passed, 0 failed; `cargo fmt --all -- --check` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean apart from the three `unfulfilled_lint_expectations` in `gix-ref` that are already on `main` under rustc 1.95.0.

Peel in `gix-revision`'s parser instead, so every `Navigate` implementation gets committish semantics without re-deriving them. That changes a published trait's contract, so I kept it out of a bug fix — glad to switch if you would rather have it there.


Byron:
Thanks a lot, loving it!

I will address the `<tag>~0` issue.

PR #2880, opened 2026-08-05
fix: ignore empty keywords in pathspec magic, like Git does

ameyypawar:
Git accepts empty keywords in long-form pathspec magic; gitoxide rejects them.

```
$ git ls-files ':(top,)some/path'      # ok
$ git ls-files ':(top,,icase)some/path' # ok, both magics applied
```

gix-pathspec returns `InvalidKeyword` for `:(top,)`, `:(,top)`, `:(top,,icase)`, `:(,)`, `:(,,)`, `:(icase,)` and `:(attr:someAttr,)`.

`parse_long_magic()` in Git's `pathspec.c` skips zero-length keywords:

```c
size_t len = strcspn_escaped(pos, ",)");
...
if (!len)
    continue;
```

Two changes here:

- `parse_long_keywords` now returns `Ok(())` for an empty keyword instead of falling through to `InvalidKeyword`.
- `split_on_non_escaped_char` scans byte-by-byte like `strcspn_escaped()`, so a backslash consumes the byte after it and a separator at index 0 is visible. The `windows(2)` loop only tested `window[1]`, so a leading separator was never a split point — that's why `:(,top)` was reported as one invalid keyword rather than an empty one followed by `top`.

The second change also makes `\\,` split the way Git does, where the escaped backslash is consumed and the comma is a real separator.

Checked against git 2.52.0 with a 32-input differential run covering the escape corners (`:(attr:v=one\,two\,three)`, `:(top\,)`, `:(top\\,icase)`, trailing `\`) and every previously-accepted form — all agree. `:(top,,icase)` yields exactly `TOP | ICASE`, same as `:(top,icase)`.

Seven baseline entries and a `empty_keywords_are_ignored` case added; the archive is regenerated in a separate commit since it's checked against the fixture's checksum.

One adjacent gap I didn't touch: Git also accepts `:(prefix:0)`, which gix-pathspec rejects. Happy to look at that separately if it's wanted.


Byron:
That's a great catch, thank you!
And keep the fixes coming :).

ameyypawar:
Thanks for the merge.

On the `:(prefix:N)` gap I mentioned — I don't think it's worth closing. Git guards it with `BUG()` rather than `die()`:

```
$ git ls-files ':(prefix:99)sub/a.txt'
BUG: pathspec.c:550: error initializing pathspec_item

$ cd sub && git ls-files ':(prefix:0)a.txt'
BUG: pathspec.c:476: 'prefix' magic is supposed to be used at worktree's root
```

It reads as something `prefix_magic()` emits for Git to consume itself, not something a user types, so matching it would mean picking behaviour Git never had to define. Unless something actually hands gitoxide a Git-serialised pathspec — you'd know better than me.

Moving on to excludes or attributes unless you'd rather I did something else.

PR #2886, opened 2026-08-06
fix: reject a lone '@' as a reference name, like Git does

ameyypawar:

----

## Summary

- Reject a reference name consisting solely of `@`, which Git refuses because `@` is shorthand for `HEAD`.
- Replace it with `-` when sanitizing, which is what Git's sanitizer does.
- Leave `@` valid as a component, so `refs/heads/@` and a tag named `@` are unaffected.

## Git baseline

`check_or_sanitize_refname()` in `refs.c`, which `check_refname_format()` delegates to, rejects the name outright and substitutes `-` in its sanitizing path. `git check-ref-format` refuses `@` under every flag combination — plain, `--allow-onelevel`, `--refspec-pattern`, both together, and `--normalize` — while accepting `refs/heads/@`, `refs/tags/@`, `@/x` and `@@`.

The visible consequence was in refspec destinations:

```
$ git push origin "HEAD:@"
fatal: invalid refspec 'HEAD:@'
```

where `gix_refspec::parse("HEAD:@", Push)` returned `Ok(dst = "@")`. Sources were already right, since `parse()` rewrites a bare `@` source to `HEAD`, but destinations reach `reference::name_partial()` directly.

Two public behaviours change as a result: `name_partial_or_sanitize("@")` now yields `"-"` rather than `"@"`, and `name("@")` reports `Reserved` where it previously reported `SomeLowercase`. No caller in the workspace observes either.

## Validation

A differential sweep of 1,030 reference names against git 2.52.0 — every byte `0x01`–`0xFF` embedded in a component and as a whole component, plus `.lock` placement, slash density, dot placement and `@{` forms — found `@` to be the only divergence, and none remaining afterwards. Two shapes lie outside what the oracle can express: names containing NUL, and names beginning with `-`, which `git check-ref-format` will not accept as an argument.

- `cargo test` — gix-validate 326, gix-ref 179, gix-refspec 76, gix-revision 112, gix-object 148, gix-url 144, gix-worktree 10, gix 417
- `cargo clippy -p gix-validate --all-targets`
- `cargo fmt --check`

### DIFF
--- a/gix-refspec/tests/refspec/parse/invalid.rs
+++ b/gix-refspec/tests/refspec/parse/invalid.rs
@@ -24,6 +24,19 @@ fn whitespace() {
 }
 
 #[test]
+fn destination_cannot_be_a_lone_at_sign() {
+    for op in [Operation::Fetch, Operation::Push] {
+        assert!(
+            matches!(
+                try_parse("HEAD:@", op).expect_err("a lone '@' is not a valid destination"),
+                Error::ReferenceName(gix_validate::reference::name::Error::Reserved { name }) if name == "@"
+            ),
+            "{op:?} validates refspec destinations"
+        );
+    }
+}
+
+#[test]
 fn complex_patterns_with_more_than_one_asterisk() {
     // For one-sided refspecs, complex patterns are now allowed
     for op in [Operation::Fetch, Operation::Push] {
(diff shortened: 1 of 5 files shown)


Byron:
Thank a lot, good catch! Very subtle behavior, and it still surprises me that `@` as tag name is allowed. It actually can't be looked up as revspec directly, one will need `tags/@` to disambiguate, but nothing informs about the ambiguity either. Interesting niche case, moving on 😅.

PR #2894, opened 2026-08-07
change!: take any non-glob push source, like Git does

ameyypawar:

----

Follows your question on #2888, about whether Git not validating a push source is a bug there.

## Summary

- Take any non-glob push source, as Git does, instead of requiring a ref name or a rev-spec.
- Ten entries added to the parse baseline, which records `git`'s own verdict.
- `gix-revision` and `gix-error` are no longer needed here, and the `Error::RevSpec` variant goes with them.

## Is it a bug in Git?

It reads as deliberate, and Git says so two lines above the part quoted on #2888:

```c
	/*
	 * LHS
	 * - empty is allowed; it means delete.
	 * - when wildcarded, it must be a valid looking ref.
	 * - otherwise, it must be an extended SHA-1, but
	 *   there is no existing way to validate this.
	 */
	if (!*item->src)
		; /* empty is ok */
	else if (is_glob) {
		if (check_refname_format(item->src, flags))
			return 0;
	}
	else
		; /* anything goes, for now */
```

The rule Git wants is that the source is an extended SHA-1. Resolving one needs a repository, which `parse_refspec()` does not have, so it takes what it is given. That is a limitation it documents rather than an oversight, so this makes `gix-refspec` do the same.

## What this does

Measured against `git ls-remote` with the refspec in `remote.test.push`:

| push refspec | `git` | before | after |
|---|---|---|---|
| `::a` | accept | rejected | accept |
| `~:refs/heads/x` | accept | rejected | accept |
| `a^{:refs/heads/x` | accept | rejected | accept |
| `a^{bogus}:refs/heads/x` | accept | rejected | accept |
| `a@{:refs/heads/x` | accept | rejected | accept |
| `a^{bogus}` | reject | rejected | rejected |

Seven shapes in the baseline were rejected here and accepted by Git. The check is dropped for that one side only, and everything around it is unchanged:

* A missing destination still requires the source to be a ref name, which is the last row.
* An empty destination is still refused, and the destination itself is still validated.
* Both glob rules still hold. The glob branch runs before the new one, and the new one only applies when a destination is present, so `a**:refs/heads/*` and `*:refs/heads/x` are still errors.
* Negative refspecs never reach it, since a destination makes them fail earlier.
* Fetch is untouched. The fetch rows in the regenerated baseline are byte-identical to the old ones.

That fallback was the only thing `gix-refspec` used `gix-revision` for, and `gix-error` appeared only in its service, both at the call site and in the `Noop` delegate that drove it. Both dependencies are dropped, along with the `Error::RevSpec` variant, which is what makes this breaking.

## Validation

The baseline goes from 84 to 94 entries. Both `parse_baseline.tar` and `parse_baseline_sha256.tar` are regenerated, since `just unit-tests` runs this crate under `sha256` as well.

Reverting `parse.rs` makes it fail with `Out of 94 baseline entries, got 87 right, (7 mismatches and 0 panics)`.

- `cargo test` — gix-refspec 77, gix 417, gix-protocol 58 with `blocking-client`
- `cargo fmt --check -p gix-refspec`

## The trade

Worth saying plainly, since it cuts against the change: `gix-refspec` was checking the thing Git's comment says it wants checked, and could do it because `gix-revision` parses a rev-spec syntactically without a repository, where Git has nothing to hand at that point.

So this drops a real check. `a^{bogus}:refs/heads/x` used to be refused here with `cannot peel to unknown target: "bogus"`, and now travels to the remote before anything notices. What it buys is that a refspec Git takes is no longer refused here.

That seemed like the answer to your question rather than a decision to make quietly, so it is the whole of this PR and easy to drop. Glad to restore the check if you would rather keep it.


Byron:
Thanks, appreciated!

The way I see it, the source of the refspec will be evaluated locally once the parsing is complete, long before any of it reaches the remote. So that should be good enough.

PR #2898, opened 2026-08-09
Use `GIX_TEST_FIXTURE_HASH` for more tests

cruessler:
This PR adds explicit test runs under `GIX_TEST_FIXTURE_HASH=sha1` and `GIX_TEST_FIXTURE_HASH=sha256` for the following crates:

- `gix-dir`
- `gix-fsck`
- `gix-negotiate`

This PR follows the example from this commit and earlier ones: 4a0ca977e59c3c39933d11a455fcd8792d27bd8e (merged in #2893).

This is part of a series of PRs where I try to find crates that pass their tests under `GIX_TEST_FIXTURE_HASH=sha1`, but not yet under `GIX_TEST_FIXTURE_HASH=sha256`.


Byron:
Thanks a lot!

When looking at the `justfile` (whose edits were following the usual pattern) I kept thinking that the `cargo nextest run --no-fail-fast` probably already runs a bunch of SHA-256 tests due to the way it compiles everything at once, with all features. But then again, if it would have, it would have failed. In any case, I thought it would at least run the SHA-1 tests. From there I wonder if the separate SHA-1 invocations of the individual crates are still needed.
Maybe something to investigate the next time the file is edited.

cruessler:
I was wondering the same. I added the `GIX_TEXT_FIXTURE_HASH=sha1` lines mostly for consistency. I’m going to double-check whether some of them can be removed.

Issue #2076, opened 2025-07-06
Support for gitremote-helpers urls (e.g., `transport::opaque-transport-uri`)


demosdemon:
### Summary 💡

Gitoxide currently fails to parse the URIs for any remote using a [gitremote helper](https://git-scm.com/docs/gitremote-helpers) that accepts the second form of remote URI (e.g., [git-remote-codecommit](https://github.com/demosdemon/git-remote-codecommit/)).

> A URL of the form `<transport>::<address>` explicitly instructs Git to invoke `git remote-<transport>` with `<address>` as the second argument. If such a URL is encountered directly on the command line, the first argument is `<address>`, and if it is encountered in a configured remote, the first argument is the name of that remote.

Ideally Gitoxide would eventually support invoking one of these helpers (https://github.com/GitoxideLabs/gitoxide/issues/1666); however, until such time, it would be nice to be able to invoke `find_remote` and it not bail on parsing the uri.

I can work around this by handling the err as such:

```rust
fn repo_info(local_repo: &gix::Repository) -> Result<Info, Error> {
    let default_remote = local_repo
        .remote_default_name(Fetch)
        .ok_or(Error::UnknownDefaultRemoteForFetch)?;
    trace!(?default_remote, "default remote for fetch");

    match local_repo.find_remote(&*default_remote) {
        Ok(remote) => repo_info_for_remote(default_remote, remote),
        Err(Find(Url {
            source:
                KeyError {
                    source: Some(RelativeUrl { url }),
                    ..
                },
            ..
        })) => {
            trace!(
                ?url,
                "found url potentially specifying a remote transport type"
            );

            let Some((transport, url)) = url.split_once("::") else {
                return Err(Error::MissingTransportTypeSeparator {
                    remote_name: default_remote.into_owned(),
                    url: url.to_owned(),
                });
            };

            todo!();
        }
        Err(err) => Err(Error::InvalidRemote {
            remote_name: default_remote.into_owned(),
            source: err,
        }),
    }
}
```

but then I lose the ability to use the `Remote<'_>`.

---

Using `try_find_remote_without_url_rewrite` does not help as the error from parsing the URL occurs before any rewriting:

https://github.com/GitoxideLabs/gitoxide/blob/c7af04db9b6bb1204e0f4c436d1db8f48a491e86/gix/src/repository/remote.rs#L190-L209

### Motivation 🔦

[...]

Issue #1666, opened 2024-11-10
Consider supporting the remote-helper format


nikhiljha:
### Summary 💡

It would be neat if gitoxide could talk to regular `gitremote-helpers`. These speak a simple newline-and-space separated protocol, and anyone can implement their own backends with the C git client by writing one. In fact, there are even some gitremote-helpers already written in Rust, since there's no reason a helper needs to be written in C.

Spec: https://git-scm.com/docs/gitremote-helpers

This is probably a very niche feature, which is why the title says "consider", but I do think it has benefits and cannot really think of any downsides.

### Motivation 🔦

Using this format would allow for falling back to the regular `git-remote-https`/`git-remote-ssh` backend if there's an unimplemented feature, or using the gitoxide implementation of https with the regular git client.

This also would allow `gix` to work for people using non-https-or-ssh transports without needing to reimplement their transports in Rust. For example: git-remote-keybase, git-remote-rad, git-remote-codecommit, git-remote-s3, etc.

Byron:
Thanks for the reminder, I absolutely agree that `gitoxide` will eventually have to learn it.

Byron:
Moved to [this Feature Proposal Discussion](https://github.com/GitoxideLabs/gitoxide/discussions/2788). This issue is being closed while the proposal is reviewed; it will become an issue again once review is complete.

_AI disclosure: This comment was written and posted by OpenAI Codex on behalf of Byron._

PR #2899, opened 2026-08-09
change!: parse the `<helper>::<address>` form of remote helpers

ameyypawar:
Refs GitoxideLabs/gitoxide#2076 (now discussions#2786) and gitbutlerapp/gitbutler#2887.

----

`gix-url` cannot represent locations of the `<helper>::<address>` form described in
`gitremote-helpers(7)`, which produces two distinct defects:

```
codecommit::eu-central-1://myaccount@my-repo  →  Err(RelativeUrl)
transport::address                            →  Ok(Scheme::Ssh, host="transport", path=":address")
```

The first is what GitButler reports in gitbutlerapp/gitbutler#2887: a repository whose `origin` is
an AWS CodeCommit remote cannot be opened at all, because parsing fails long before any transport is
chosen. The second is quieter and arguably worse, as it yields a plausible-looking `Url` that points
somewhere else entirely; `ext::ssh -o Foo host %S` becomes an SSH connection to a host named `ext`.

### Cause

`find_scheme()` looked for `://` first. Git checks for `<helper>::` first, in `transport_get()`,
before the location is examined as a URL at all, so a helper address may itself contain `://`.

Worth noting that `gix-url` was not wrong about the layer it models. `git fetch-pack --diag-url
'transport::address'` also reports `protocol=ssh, userandhost=transport`, because `--diag-url`
exposes `connect.c`, which never sees these locations. The helper check lives one level above it.
It still belongs in `gix-url`, because the value has to be representable before a transport is ever
selected.

### Approach

This follows the representation settled on in #2076: the transport becomes the scheme, and the
address is mapped onto `Url::path`, with the URL serializing back to the form it was parsed from.
The address is kept verbatim and uninterpreted, since Git places no constraints on it and only the
helper program can make sense of it.

Two details worth review:

- The helper name deliberately does not go through `Scheme::from()`. Git dispatches `ssh::x`,
  `file::x`, `git::x` and `http::x` to `git-remote-<name>` rather than to the built-in transport of
  the same name, so collapsing those into `Scheme::Ssh` and friends would produce a `Url` claiming a
  connection that Git would never make.
- Accepted names are those of Git's `is_urlschemechar()`, `[A-Za-z0-9][A-Za-z0-9+.-]*`. Note this
  excludes `_`, so `foo_bar::baz` stays SCP-like in both implementations.

One deviation, documented on `parse()`: an empty name as in `::address` stays rejected. Git accepts
it and looks for a program called `git-remote-`, which cannot meaningfully exist.

### Breaking change

Locations that previously failed to parse now succeed, and locations that previously parsed as
`Scheme::Ssh` now parse as `Scheme::Ext`. Nothing about the public API changed, so `change!:` may be
stronger than needed here — happy to downgrade it to `fix:` if that reads better.

`gix`'s remote-rewrite tests used `invalid:://` as an intentionally unparseable URL, which is now a
valid helper location. The second commit adapts them to `:://`, which keeps them failing with the
same `RelativeUrl` error, so those tests continue to exercise the path they were written for.

### Verification

The behaviour was taken from Git 2.52.0 rather than `master`, and checked against the binary by
putting `git-remote-*` programs that dump their `argv` on the `PATH`. Both implementations agree on
every input tried, including which spellings are *not* helpers (`foo_bar::baz`, `f%o::bar`,
`.foo::bar`, `a:b::c`, `-foo::bar`), that the split happens at the first `::`, that names are
case-sensitive, that empty addresses are allowed, and that a single-letter name is not a DOS drive
letter. The address recorded for each helper URL matches the second argument Git passes to the
helper, byte for byte.

The existing baseline suite against `fetch-pack --diag-url` and the fuzz corpus are unaffected, and
every new test fails if the detection is removed. `bash::exec git-remote-https "$1" "$(…)"` from
#2076 is included in the suite as requested.

### Out of scope

Actually invoking helpers (#1666) is unchanged; `gix-transport` still rejects `Scheme::Ext(_)` with
`UnsupportedScheme`, so these locations now parse and then fail at connect time, which is where the
`codecommit://…` spelling already failed. The third form, `remote.<name>.vcs`, is configuration
rather than URL syntax and is not addressed here.

Separately noticed while working on this, and left alone: `CodeCommit://x` serializes back as
`codecommit://x`, though Git dispatches case-sensitively to `git-remote-CodeCommit`. Happy to file
that separately.

----

_AI disclosure: this description and the changes it describes were produced by Claude (Opus 5) on
behalf of @ameyypawar, who reviewed them before submission._

### DIFF
--- a/gix-transport/src/client/blocking_io/connect.rs
+++ b/gix-transport/src/client/blocking_io/connect.rs
@@ -23,7 +23,9 @@ pub(crate) mod function {
     {
         let mut url = url.try_into().map_err(gix_url::parse::Error::from)?;
         Ok(match url.scheme {
-            gix_url::Scheme::Ext(_) => return Err(Error::UnsupportedScheme(url.scheme)),
+            gix_url::Scheme::Ext | gix_url::Scheme::Helper(_) | gix_url::Scheme::HelperUrl(_) => {
+                return Err(Error::UnsupportedScheme(url.scheme));
+            }
             gix_url::Scheme::File => {
                 if url.user().is_some() || url.password().is_some() || url.host().is_some() || url.port.is_some() {
                     return Err(Error::UnsupportedUrlTokens {
--- a/gix-url/src/lib.rs
+++ b/gix-url/src/lib.rs
@@ -57,10 +57,13 @@ mod simple_url;
 /// This accepts standard URLs, SCP-like SSH locations, remote-helper locations and local paths. URL and SCP-like
 /// inputs must be UTF-8; remote-helper addresses and local paths retain arbitrary bytes.
 ///
-/// Locations of the `<helper>::<address>` form described in `gitremote-helpers(7)` are recognized before any URL
-/// syntax, just like Git does in `transport_get()`, so an address may itself contain `://`. They are represented as
-/// [`Scheme::Ext`] holding the helper name, with the address kept verbatim in [`Url::path`] as only the helper
-/// program can interpret it.
+/// Locations of the `<helper>::<address>` form described in
+/// [`gitremote-helpers`](https://git-scm.com/docs/gitremote-helpers) are recognized before any URL
+/// syntax, so an address may itself contain `://`. They are represented as
+/// [`Scheme::Helper`] holding the helper name, with the address kept verbatim in [`Url::path`] as only the helper
+/// program can interpret it. The command-executing `ext` helper is represented as [`Scheme::Ext`] in both spellings;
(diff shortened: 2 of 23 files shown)


Byron:
Thanks a lot, this is great!

I spent some time untangling this, and learned a lot in the process as well. There are plenty of changes on top now.

> Separately noticed while working on this, and left alone: CodeCommit://x serializes back as
codecommit://x, though Git dispatches case-sensitively to git-remote-CodeCommit. Happy to file
that separately.

I addressed this, but maybe there is more you can find.

PR #2904, opened 2026-08-11
feat: add `State::tree_mut()`

abemedia:
Adds `State::tree_mut()`, mirroring `entries_mut()` beside `entries()`.

`tree()` reads the extension and `remove_tree()` takes it, and every field of `extension::Tree` is public, so callers can already build an invalidated tree. They cannot write one back, since `State.tree` is private.

Without it, invalidating the sub-trees affected by an entry change means discarding the whole extension, so the next tree-write traverses every path instead of only the changed ones. Git invalidates only the nodes covering the changed paths.

`write()` documents that the tree extension is serialized as-is and `**not** recomputed or invalidated to match the entries`, and prescribes `remove_tree()` as the interim workaround for #2421. This PR does not change that: it lets a caller invalidate precisely instead of discarding, which is the same workaround at lower cost.

### DIFF
--- a/gix-index/src/access/mod.rs
+++ b/gix-index/src/access/mod.rs
@@ -630,6 +630,10 @@ impl State {
     pub fn tree(&self) -> Option<&extension::Tree> {
         self.tree.as_ref()
     }
+    /// Mutably access the `tree` extension.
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thank you!

Yes, let's do this as first step to actually offering APIs that maintain it automatically.

PR #2907, opened 2026-08-12
Pass object hash explicitly in a few tests

cruessler:
This is a small PR that makes a few tests in `gix-odb` and `gix-ref` use `gix_testtools::object_hash()` instead of picking SHA-1 through `Default::default()` or by hard-coding `Sha1`.

While doing that, I also changed some tests in `gix-odb` to hard-code `Sha1` instead of implicitly picking it through, e. g., `Options::default()`. That was mostly to help remind future me that these locations are, in fact, SHA-1-only as they exercise SHA-1-only fixtures. Feel free to ignore this change if you think it’s too pedantic, though. 😉


Byron:
Thanks a lot!

PR #2911, opened 2026-08-12
change!: resolve the empty pattern in `<rev>^{/}` like Git, instead of skipping it

### LINKED ISSUE #2908: Empty-pattern rev-spec `^{/}` is skipped entirely, diverging from Git: no peel to commit, and `^{/!-}` silently succeeds
### Current behavior 😯

The rev-spec navigation `<rev>^{/<pattern>}` with an *empty* pattern is skipped entirely by the parser ([`gix-revision/src/spec/parse/function.rs`](https://github.com/GitoxideLabs/gitoxide/blob/dc370d9cacb4b06a132d2ffaaf224cc770173742/gix-revision/src/spec/parse/function.rs#L679-L687) — `if !regex.is_empty()` guards the `delegate.find()` call, otherwise the step is a no-op).

This produces two observable divergences from Git:

1. **`atag^{/}` on an annotated tag returns the tag object instead of a commit** — the peel-to-committish step that Git always performs for `^{/…}` never happens:

   ```console
   $ gix rev parse 'atag^{/}'
   3cbf5f4502061b38944645f361a7385c5243fc11   # the tag object itself
   ```

2. **`HEAD^{/!-}` (negated empty pattern) silently succeeds**, while Git fails because every commit matches the empty pattern and the negation therefore matches nothing:

   ```console
   $ gix rev parse 'HEAD^{/!-}'
   1f6f79a0f02ff3c38d1d6513a6b24c77b763f30e   # resolves fine
   ```

The existing test `empty_top_revision_regex_are_skipped_as_they_match_everything` locks in the skipping behaviour, but its anchor `@^{/}` already points at a commit, which hides both effects above.

### Expected behavior 🤔

`^{/}` should behave like Git: peel the anchor to a committish and then search for the first commit (starting at, and including, the peeled commit) whose message matches the — possibly empty — pattern. An empty pattern matches everything, so `atag^{/}` resolves to the commit the tag points to. A negated empty pattern matches nothing, so `HEAD^{/!-}` fails.

Notably, the `Navigate::find()` implementation in `gix` already handles empty patterns correctly (it peels to a commit first, and both the `revparse-regex` and the `contains_str()` fallback treat an empty pattern as match-all). Only the early-out in the parser prevents it from being reached.

### Git behavior

Git treats `<rev>^{/<text>}` with empty text as valid and peels first — from `object-name.c`, `peel_onion()`: *"$commit^{/}. Some regex implementation may reject empty regex, but this is safe."* (`sp[0] == '/'` sets `expected_type = OBJ_COMMIT` before matching).

```console
$ git --version
git version 2.48.1

$ git rev-parse 'atag^{/}'
1f6f79a0f02ff3c38d1d6513a6b24c77b763f30e   # the commit the tag points to

$ git rev-parse 'atag' 'atag^{commit}'
3cbf5f4502061b38944645f361a7385c5243fc11
1f6f79a0f02ff3c38d1d6513a6b24c77b763f30e

$ git rev-parse 'HEAD^{/!-}'
fatal: ambiguous argument 'HEAD^{/!-}': unknown revision or path not in the working tree.
```

### Steps to reproduce 🕹

```console
$ git init t && cd t
$ git commit --allow-empty -m 'first commit'
$ git commit --allow-empty -m 'second commit'
$ git tag -a -m 'annotated tag' atag HEAD

$ git rev-parse 'atag^{/}'
1f6f79a0f02ff3c38d1d6513a6b24c77b763f30e
$ gix rev parse 'atag^{/}'      # gitoxide @ dc370d9
3cbf5f4502061b38944645f361a7385c5243fc11

$ git rev-parse 'HEAD^{/!-}'; echo "exit=$?"
fatal: ambiguous argument 'HEAD^{/!-}': ...
exit=128
$ gix rev parse 'HEAD^{/!-}'; echo "exit=$?"
1f6f79a0f02ff3c38d1d6513a6b24c77b763f30e
exit=0
```

I have a minimal fix ready (drop the empty-pattern early-out in the parser so `find()` is always invoked, plus doc/test adjustments) and will open a PR shortly.

*Disclosure: this issue was researched and written with AI assistance; I reproduced the behaviour locally against git 2.48.1 and reviewed every claim.*

codeAnqiang-ma:
Fixes #2908.

### What

The parser skipped the `Navigate::find()` call entirely when the pattern in `<rev>^{/<pattern>}` was empty, turning the navigation step into a no-op. That shortcut only holds for a commit anchor with a non-negated pattern:

- `b-tag^{/}` (annotated tag) returned the tag object, while Git peels to the commit first (`object-name.c` routes `^{/…}` through `GET_OID_COMMITTISH` and explicitly allows the empty pattern: *"$commit^{/}. Some regex implementation may reject empty regex, but this is safe"*).
- `HEAD^{/!-}` (negated empty pattern) silently succeeded, while Git fails since every commit matches the empty pattern.

The parser now always forwards the pattern to `Navigate::find()`. The implementation in `gix` already handles empty patterns correctly on both the `revparse-regex` and the substring-fallback paths — it peels the anchor to a commit first (#2871) and treats an empty pattern as match-all, which naturally fails when negated. The delegate behind `gix revision explain` makes no assumptions about the pattern and needs no change.

Marked `change!:` following the precedent of #2892: a `Navigate` implementation written against the previous documentation ("which is never empty") could now be handed an empty `regex`. The doc comment is updated accordingly.

### Tests

- `gix-revision`: updated the parser test that pinned the skipping behaviour (`empty_regex_is_passed_to_the_delegate`), now also covering the negated form. Fails before the fix, passes after.
- `gix`: new `empty_pattern::matches_everything_and_peels_to_a_commit` test with `parse_spec` baseline verification against real Git for `@^{/}`, `@^{/!-}` and `b-tag^{/}` (new fixture baselines). Fails before the fix, passes after.
- Note: `make_rev_spec_parse_repos.tar` needs regeneration for the new baselines; following #2871 I left the archive untouched so it can be updated and verified on your side (happy to push it myself if preferred).

```console
$ cargo test -p gix-revision
test result: ok. 112 passed; 0 failed; 0 ignored

$ cargo test -p gix --test gix revision::
test result: ok. 38 passed; 0 failed; 0 ignored (361 filtered out)

$ cargo test -p gix --test gix --features revparse-regex revision::spec::from_bytes::regex
test result: ok. 3 passed; 0 failed; 0 ignored
```

End-to-end against git 2.48.1 after the fix (repo with annotated `atag`):

```console
$ gix rev parse 'atag^{/}' ; git rev-parse 'atag^{/}'
1f6f79a0f02ff3c38d1d6513a6b24c77b763f30e
1f6f79a0f02ff3c38d1d6513a6b24c77b763f30e
$ gix rev parse 'HEAD^{/!-}'; echo $?   # now fails like Git
1
```

Full test suite of untouched crates was not run; CI will cover it.

*Disclosure: this fix was developed with AI assistance (`Assisted-by:` trailer in the commit); I reproduced the divergence locally against git 2.48.1 and reviewed every change.*


Byron:
Thanks again, appreciated!

PR #2916, opened 2026-08-13
feat!: require `object_hash` in `Store::at`

cruessler:
This PR is a small follow-up to #2901 which split `gix_odb::at(objects_dir: impl Into<PathBuf>)` into `gix_odb::at(objects_dir: impl Into<PathBuf>, object_hash: gix_hash::Kind)` and `gix_odb::at_opts(objects_dir: impl Into<PathBuf>, replacements: impl IntoIterator<Item = (gix_hash::ObjectId, gix_hash::ObjectId)>, options: store::init::Options)`. By always requiring callers to explicitly pass the object hash we eliminate the risk posed by `Default::default()` implicitly choosing `Sha1` in SHA-1/SHA-256 builds. This comes at the cost of callers having to explicitly pass a hash even in single-hash builds, but I would argue that it is trivial for callers to write their own wrapper if they want.

This PR goes even further than #2901 because it removes `object_hash` from `Options`, making hash selection explicit even in the `at_opts` case. I think this is cleaner, but, obviously, feel free to modify and adapt!

### DIFF
--- a/gitoxide-core/src/pack/explode.rs
+++ b/gitoxide-core/src/pack/explode.rs
@@ -149,13 +149,7 @@ impl gix::objs::Write for OutputWriter {
 impl OutputWriter {
     fn new(path: Option<impl AsRef<Path>>, compress: bool, object_hash: gix::hash::Kind) -> Self {
         match path {
-            Some(path) => OutputWriter::Loose(loose::Store::at(
-                path.as_ref(),
-                loose::Options {
-                    object_hash,
-                    ..Default::default()
-                },
-            )),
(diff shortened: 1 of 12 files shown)


Byron:
This is a great catch, thanks so much, I am loving it!

> This PR goes even further than https://github.com/GitoxideLabs/gitoxide/pull/2901 because it removes object_hash from Options, making hash selection explicit even in the at_opts case. I think this is cleaner, but, obviously, feel free to modify and adapt!

Yes! the commit on top does the same to the `gix-odb::Store::at_opts` function, that is how it should be.

PR #2919, opened 2026-08-15
refactor!: remove `object_hash` from `gix_ref::store::init::Options`

cruessler:
- Split `gix_ref::file::Store::at` into `at` and `at_opts`.
- Split `gix_ref::file::Store::for_linked_worktree` as well.
- Split `gix_ref::Store::at` as well.

This PR continues work done in #2901 for `gix-odb` where `gix_odb::at` was split into `gix_odb::at` and `gix_odb::at_opts` while `object_hash` was removed from `gix_odb::store::init::Options` and #2916 where the same change was done for `gix_ref::file::Store::at`.

This forces callers to explicitly choose a hash, eliminating the risk that `Sha1` is implicitly chosen for them in SHA-1/SHA-256 builds through `Default::default()` without them being aware of it or noticing. This comes at the cost of callers having to always pass the hash, but that seems like a reasonable trade-off.

### DIFF
--- a/gitoxide-core/src/remote.rs
+++ b/gitoxide-core/src/remote.rs
@@ -112,15 +112,15 @@ fn write_refs(
 
     let start = Instant::now();
     let precompose_unicode = gix::fs::Capabilities::probe(&directory).precompose_unicode;
-    let store = gix::RefStore::at(
+    let store = gix::RefStore::at_opts(
         directory,
+        object_hash,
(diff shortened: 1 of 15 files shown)


Byron:
This is great, I wasn't aware the store options have been so inconsistent!
Good this is fixed now, just the right change before hopefully making a move on reftable (and a lot of improvements along with it) soon.

PR #2923, opened 2026-08-16
feat(gix-pack): expose index-less parallel pack traversal

rdicosmo:
As discussed in #2922, thanks for the quick green light on the shape.


Byron (approved):
Thanks a lot!

I have made public the whole `delta` module instead, sticking to the original module layout.

rdicosmo:
Thanks a lot!

PR #2924, opened 2026-08-17
fix: don't limit `is_dirty()` to the current working directory

abemedia:
`is_dirty()` returned `false` when run from a subdirectory while the only change was staged outside of it:

    git init repo && cd repo
    echo one >file && mkdir sub && git add . && git commit -m init
    echo two >file && git add file
    cd sub    # is_dirty() -> false, but `git status` shows `M file`

The tree-index half of the check passes `None` to `tree_index_status()`, whose fallback pathspec is limited to the prefix of the current working directory. The index-worktree half of the same check has no such limit, and `git status` covers the whole repository from anywhere.

Now an all-matching pathspec is passed explicitly.


Byron:

PR #2926, opened 2026-08-17
refactor!: remove `object_hash` from `gix_pack::bundle::write::types::Options`

cruessler:
This is a follow-up to #2919. It makes hash selection explicit in `gix_pack::bundle::write::Bundle::write_to_directory` and `gix_pack::bundle::write::Bundle::write_to_directory_eagerly`. This should be the last (or one of the last) PRs removing `object_hash: gix_hash::Kind` from structs that derive `Default` and thus present the risk that callers inadvertently get SHA-1 in SHA-1/SHA-256 builds without noticing.

If we wanted, we could potentially even go so far as to remove `#[derive(Default)]` from `gix_hash::Kind` in the future. We would still need to make some changes for that to be possible, but I think it would be fairly straightforward at this point. At the same time, there’s diminishing returns because most of the important structs have had `object_hash` removed already, so I’m not a 100 % sure what the best way to go is. What do you think?

This PR can be marked as ready once CI is green. 😉


Byron:
Thanks so much!

This was a fun review!
And I have great news for you: As of now, you don't have to split commits anymore just for the changelog. Instead, you drive the changelog with a scope, so the one commit in this PR is now prefixed `refactor(gix-pack)!: ...`. The `gix-pack` part is what tells the changelog to go only in `gix-pack/CHANGELOG.md`.
And I am working on tooling to make local QA easier as well, with tree-based tagging for the little checkmarks (among other things ;)).

Screenshot 2026-08-19 at 13 29 39

PR #2928, opened 2026-08-18
test: add test for binary diff with `textconv`

cruessler:
This PR adds a test for a binary diff with `textconv`. As far as I can tell, such a test does not exist yet. If it does, feel free to close this PR!

I want to add similar tests for `gix-blame` as well, but first I want to properly understand this part of the `gix-diff` machinery.


Byron:
Thanks a lot, it's great to finally see this tested!

PR #2931, opened 2026-08-19
test(gix-diff): add optional detailed report to slider test

cruessler:
This is the first PR in my quest to bring `gix-diff` closer to `git diff`. Before making any changes, I want to first add proper diagnostics, in order to be able to split the larger problem of “diffs don't match” into smaller problems of “diffs don't match for this specific reason” and “there's these specific reasons that can cause a mismatch”. The individual reasons can then be tackled independently.

The first run on a corpus of sliders from the `gitoxide` repository itself already yielded interesting results (note that this is only an, albeit fairly large, subset of all the sliders produced by `create-diff-cases`):

```
❯ env GIX_DIFF_SLIDER_REPORT=1 cargo test -p gix-diff slider::baseline -- --nocapture
[…]
slider report

total cases: 3094
git no-indent baselines: 3094

Exact: 141 [4.56%]
LikelySliderOnlyMismatch: 2902 [93.79%]
OtherMismatch: 51 [1.65%]

likely slider-only details

multi-hunk/same-count/same-direction: 2299 [79.22% slider-only, 74.31% total]
single-hunk/pure-insertion/same-before-start/0: 427 [14.71% slider-only, 13.80% total]
single-hunk/modification/same-before-start/0: 128 [4.41% slider-only, 4.14% total]
single-hunk/pure-deletion/same-before-start/0: 48 [1.65% slider-only, 1.55% total]

first non-matching cases

[…]
```

In this example, it looks like there might be just a few specific reasons for the mismatches, and it looks like the majority of them is indeed slider-related. My next goal would to identify specific classes of mismatches to see whether they can be captured in corresponding baseline tests and potentially also be solved by the same code change(s).

Please let me know if I used the new commit message format correctly. 😉

### DIFF
--- a/gix-diff/tests/README.md
+++ b/gix-diff/tests/README.md
@@ -23,4 +23,9 @@ cargo run --package internal-tools -- \
     --destination-dir gix-diff/tests/fixtures/
 ```
 
-Finally, run `cargo test -p gix-diff sliders -- --nocapture` to execute the actual tests to compare.
+Finally, run `cargo test -p gix-diff slider::baseline -- --nocapture` to execute the actual tests to compare.
+
+The generated fixture includes Git baselines with and without the indent heuristic. The regular test compares
+`gix-diff` with slider heuristics to the primary Git baselines, while the additional `*.no-indent.baseline`
+files are used for diagnostics. To print a coarse mismatch report instead of failing on the first mismatch, run
+`GIX_DIFF_SLIDER_REPORT=1 cargo test -p gix-diff slider::baseline -- --nocapture`.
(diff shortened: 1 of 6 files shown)


Byron:
Thanks a lot, I can't wait to get to those 99.999% eventually :).

After some changes, which maybe incorrectly ignore the 'function suffix' of diff headers, I get:

```
  Exact: 3042 [98.32%]
  PostprocessNoHeuristicMatchesGitDefault: 4 [0.13%]
  OtherMismatch: 48 [1.55%]
```

So please, do take a look as maybe I have butchered the baseline test, but maybe not.

> Please let me know if I used the new commit message format correctly. 😉

It's not used incorrectly ;). What I mean is that the 'scope' doesn't have to be used unless files outside of the crate that should receive the changelog are touched. Further, there is probably never a need to use `test: ` as the changelog address users of the public API.

PR #2941, opened 2026-08-24
gix-hook prototype and cwd support

tnt4brain:
Hi there!

This brings in plumbing for gix-hooks support (i.e. all the necessary details to wire-in the hooks later on), and also some minor changes on the creates nearby that are required for this to function.

The code is generated using my custom prompts and verified against git 2.47.0 checkout (that was the whole idea of this effort).

Please feel free to note if any changes are required.


Byron:
Let me shut this down early so you don't waste your time on this effort, as I am not going to do QA for your prompts as you can imagine.
My recommendation is to keep it in your own fork and use it from there if there is such a need.

PS: I will also update the contribution guidelines to discourage code dumps in the future.

tnt4brain:
Perfectly understand, and sorry for the mess. The code was tested on Linux only. So I will ensure it gets built, tested and runs properly on both Linux and Windows before even trying to send something next time. Deal?

Byron:

> So I will ensure it gets built, tested and runs properly on both Linux and Windows before even trying to send something next time. Deal?

No - because I don't want contributors control which feature I implement next.
Also, from experience, it's more work to accept big features that way, so it's generally not worth it (anymore).

tnt4brain:
Fair - you project, your decision. Could you please explicitly point me to feature I could contribute to then?

Byron:
No features, but bugs and issues, inconsistencies with Git, anything that doesn't work quite right that is inside the `gix-*` plumbing.
Probe it, prod it, break it, fix it.
"Inconsistencies" can also be a lack of parity with Git. `gix-date` comes to mind which probably doesn't parse the terzillion input types that Git supports. Maybe there is more. Check for TODO in code as well, and have a go at it. What was TODO in the past is now a quick round with an agent.

tnt4brain:
OK, got it, will take a look at TODOs.

Btw, my initial idea was to contribute to features that are unchecked, meanwhile having as little merge conflicts/clashes as possible, and constantly testing the code against a fresh one git 2.47.0 checked out nearby.

Unfortunately, I've completely missed out Windows, and it backfired on MR checks instantly.

So if you're keen, the attached one is a current status
[crate-status.md](https://github.com/user-attachments/files/31413768/crate-status.md) - please feel free to order anything, like a menu 😄 even if this will take some effort from me to ensure first class Windows support. 

But if you prefer not to - no problem at all.

Byron:
Let's stick to [this](https://github.com/GitoxideLabs/gitoxide/pull/2941#issuecomment-5407988813).

Since you keep asking about features, let me make clear that if I want to "order" something, this is a prompt to my agent(s), not to a human that prompts agents for me. It makes no sense anymore to have features contributed.

PR #2948, opened 2026-08-27
fix(gix-filter): reap `process` filter children instead of orphaning them

### LINKED ISSUE #2947: `gix-filter: long-running` process `filters are never waited for (274 zombies in 10.5 h)`
### Discussed in https://github.com/GitoxideLabs/gitoxide/discussions/2945

<sup>Originally posted by **tgorka** August 27, 2026</sup>
Hi! I ran into a process leak in `gix-filter` and have a small fix. I want to confirm before opening a PR.

`driver::State::shutdown()` is the only code that `wait()`s a long-running `process` filter, and nothing in the workspace calls it. Closing a child’s pipes makes it exit but doesn’t reap it, so on Unix every filter gitoxide launches keeps a process-table slot until the process exits. `Clone for State` resets `running`, so each per-thread clone in `gix-status` and in checkout abandons its own children.

What it looks like in practice, in a desktop app with `filter.lfs.process` registered: 274 zombies after 10.5 h against a `kern.maxprocperuid` of 2666, until nothing on the machine can `fork()` anymore. Also reproduced on Linux, where the child sits at `state=Z` in `/proc`.

The fix is ~30 lines of source: a `&mut self` form of `shutdown` that the by-value one delegates to, plus `impl Drop for State`. Non-breaking, and it covers `status`, `checkout`, `diff`, `merge` and `worktree_stream` at once. Details, evidence and tests are in two PRs on my fork — tests first, so the first one demonstrates the leak and the second flips it:

- full diff: https://github.com/GitoxideLabs/gitoxide/compare/main...tgorka:gitoxide:fix/reap-process-filter-children
- 1/2, the tests: https://github.com/tgorka/gitoxide/pull/1
- 2/2, the fix: https://github.com/tgorka/gitoxide/pull/2

Two questions before I open it here:

1. Is a `Drop` that blocks in `wait()` acceptable? The child’s stdin is closed first, which is what tells it to exit, so it returns promptly — but I’d understand a preference for an opt-out instead.
2. Should `Clone for State` keep resetting `running`, or would you rather see one shared child pool? The current semantics are what multiply the leak.

Happy to reshape this however suits you, or to drop it if you’d rather take a different route.</div>

tgorka:
Fixes #2947, from discussion #2945. Thanks for turning it into an issue.

## What is wrong

`driver::State::shutdown()` (`gix-filter/src/driver/shutdown.rs:19`) is the only code in the workspace that `wait()`s a long-running `process` filter child, and nothing calls it. It also takes `self` by value, so `State` could not reap from `Drop` as written. Closing a child's pipes terminates it but does not reap it, so on Unix every `process` filter keeps a slot in the process table until the parent exits — the doc-comment on `State::running` currently claims the opposite.

`impl Clone for State` resets `running`, so each per-thread clone in `gix-status`'s `index_as_worktree{,_with_renames}` and in `gix-worktree-state`'s checkout owns and abandons children of its own: on the order of `thread_limit + 2` per `Repository::status()`.

Callers cannot compensate: `Repository::status()` takes no pipeline and returns none, `index_as_worktree`'s `Outcome` cannot hand the `resource_cache` back (it derives `Clone, Eq, Ord`), and `Repository::worktree_stream()` has no public reach to its pipeline at all.

## The two commits

Tests first, and each one is self-contained and green on its own, per `DEVELOPMENT.md`.

**1/2 — `feat(gix-filter): add \`process::Client::id()\` (#2947)`**

Pinning the lifecycle needs the child's pid without consuming the subject, and `into_child()` consumes it, so `Client::id()` is added. Then three `cfg(unix)` tests that assert the behaviour **as it is today**, i.e. the leak, so this commit passes CI by itself:

- `driver::shutdown::a_dropped_state_does_not_reap_its_process_children`
- `driver::shutdown::each_clone_of_a_state_owns_its_own_process_children`
- `pipeline::convert_to_git::a_dropped_pipeline_does_not_reap_the_filter_it_launched` — a real `convert_to_git()` over the `driver-only` fixture, which is the shape users hit

Observation goes through `waitpid(pid, WNOHANG)`, the only way to distinguish a zombie from a reaped child, hence a `[target.'cfg(unix)'.dev-dependencies] libc` entry following `gix-ref/tests/transaction_fd_limit.rs`. `libc 0.2.186` was already in `Cargo.lock`. No new fixture program: `tests/helpers/arrow.rs` already is a long-running `process` filter that exits when its stdin closes.

**2/2 — `fix(gix-filter): reap \`process\` filter children instead of orphaning them (#2947)`**

- `State::shutdown_mut(&mut self, mode)` drains `running` via `mem::take`; the existing `shutdown(self, mode)` delegates to it. No signature changes, no caller adaptations — `driver_state_mut()` is already used in production at `gix-worktree-state/src/checkout/chunk.rs:182`.
- `impl Drop for State` → `shutdown_mut(WaitForProcesses)`. One place that covers `status`, `checkout`, `diff`, `merge`, `worktree_stream` and every clone site, including future ones.
- `shutdown` no longer abandons the remaining children when one `wait()` fails; errors are collected and the first is returned after all of them were handled.
- Doc corrections on `State`, `State::running` and `Mode::Ignore`.
- The three tests above are flipped to the reaped expectation, and `an_ignored_shutdown_leaves_its_process_children_unreaped` is added to pin `Mode::Ignore` as the deliberate opt-out.

30 lines of non-doc source across three files; the rest is tests and docs.

## Evidence

Field report — desktop app on `gix 0.86.0`, macOS, two repositories with `filter.lfs.process`: **274 zombie children after 10 h 29 m**, 95 live helpers, `kern.maxprocperuid` 2666. At ~26/hour the host eventually cannot `fork()` at all; the visible symptoms were the browser failing to open tabs and keyboard shortcuts dying.

Reproduced on Linux 6.17 while preparing this: 200 ms after dropping the `State` the child is at `state=Z` in `/proc/<pid>/stat`, and `waitpid(pid, WNOHANG)` returns the pid rather than `ECHILD`. With the fix, the `/proc` entry is gone by the next statement — which is also the answer to "does waiting in `Drop` cost anything".

**Mutation check**: with `impl Drop for State` deleted and nothing else changed, all three lifecycle tests fail at `assertion left == right … left: Running, right: Reaped`. `an_ignored_shutdown_…` stays green under that mutation, which is what makes it a pin of the opt-out rather than of the fix.

## Checks run

- `cargo test -p gix-filter --test filter`: 61 pass on commit 1, 62 on commit 2, under both `--features sha1` and `GIX_TEST_FIXTURE_HASH=sha256 --features sha256`.
- `cargo check -p gix-filter -p gix-status -p gix-worktree-state -p gix-diff -p gix-merge --all-targets` clean on both commits — adding `Drop` breaks no partial move. `gix-status` and `gix-worktree-state` suites: 56 pass.
- `cargo fmt --check` clean, `clippy --all-targets` no warnings of ours in two feature configurations, `RUSTDOCFLAGS=-D warnings cargo doc -p gix-filter` clean.
- Not run: `just test` in full and `cargo check --workspace`, because `openssl-sys` cannot build in my environment (no system libssl). `just fmt` was skipped deliberately: nightly `rustfmt 1.10.0-nightly (2026-08-25)` wants to reformat 290 files on unmodified `main`, so it is not usable here. The added code introduces no new nightly-rustfmt diff.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1967,6 +1967,7 @@ dependencies = [
  "gix-trace",
  "gix-utils",
  "gix-worktree",
+ "libc",
  "smallvec",
  "thiserror 2.0.18",
 ]
--- a/gix-filter/Cargo.toml
+++ b/gix-filter/Cargo.toml
@@ -63,5 +63,9 @@ gix-quote = { path = "../gix-quote" }
 gix-testtools = { path = "../tests/tools" }
 gix-worktree = { path = "../gix-worktree", default-features = false, features = ["attributes"] }
 
+# Used by tests that observe whether a filter's child process was reaped, which `std` cannot express.
+[target.'cfg(unix)'.dev-dependencies]
+libc = "0.2.186"
+
 [package.metadata.docs.rs]
 features = ["sha1"]
--- a/gix-filter/src/driver/mod.rs
+++ b/gix-filter/src/driver/mod.rs
@@ -60,15 +60,20 @@ impl Operation {
 ///
 /// ### Lifecycle
 ///
-/// Note that [`shutdown()`][State::shutdown()] must be called to finalize long-running processes.
-/// Failing to do so will naturally shut them down by terminating their pipes, but finishing explicitly
-/// allows to wait for processes as well.
+/// Long-running processes are terminated and waited for when this instance is dropped, so they never
+/// linger in the process table of the operating system. Call
+/// [`shutdown()`][State::shutdown()] or [`shutdown_mut()`][State::shutdown_mut()] explicitly to learn
+/// how each of them exited, or to opt out of waiting with [`shutdown::Mode::Ignore`].
+///
+/// Note that [`clone()`][Clone::clone()] does *not* clone the running processes, so each clone owns and
+/// terminates only the processes that it launched itself.
 #[derive(Default)]
 pub struct State {
     /// The list of currently running processes. These are preferred over simple clean-and-smudge programs.
     ///
-    /// Note that these processes are expected to shut-down once their stdin/stdout are dropped, so nothing else
-    /// needs to be done to clean them up after drop.
+    /// Note that these processes shut down once their stdin/stdout are dropped, but on Unix they keep
+    /// occupying a slot in the process table until they are waited for, which is what dropping this
+    /// instance does.
     running: HashMap<BString, process::Client>,
 
     /// The context to pass to spawned filter programs.
@@ -92,6 +97,14 @@ impl Clone for State {
             running: Default::default(),
             context: self.context.clone(),
         }
+    }
+}
+
+impl Drop for State {
+    fn drop(&mut self) {
+        // Waiting is expected to be brief: dropping a client closes the input and output of its process,
+        // which is how it is told to exit. Errors are of no use here as there is nothing left to retry.
+        let _ = self.shutdown_mut(shutdown::Mode::WaitForProcesses);
     }
 }
 
--- a/gix-filter/src/driver/process/client.rs
+++ b/gix-filter/src/driver/process/client.rs
@@ -292,6 +292,14 @@ impl Client {
     pub fn version(&self) -> usize {
         self.version
     }
+
+    /// Return the process id of the child process we are communicating with.
+    ///
+    /// This is enough to observe the child's lifetime, for instance to assert that it was waited for,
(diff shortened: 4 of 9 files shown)


Byron:
Sorry, forgot to share that I'm going to take care of the issue.

Superseded by #2946.

tgorka:
ok, thanks a lot! 
Can't wait to use the next version with that fix.
Let me know if I can help in any way.

PR #2950, opened 2026-08-28
fix(gix-config): don't indent a value that starts on a continuation line

jaideeppyne:
Per the [agent impersonation policy](https://github.com/GitoxideLabs/gitoxide/blob/main/CONTRIBUTING.md#prevent-agent-impersonation): **an AI agent (Claude) is speaking through the `jaideeppyne` account** and wrote this description, the patch and the tests.

### What's wrong

`git` discards whitespace for as long as the value it has accumulated is still empty, so the indentation of the line a value *starts* on is insignificant — whether that is the line with the `=` on it or a later line reached through a `\` continuation. `gix-config` only applied this to the first line.

`.git/config`:

```ini
[alias]
	lg = \
		log --oneline --graph
[core]
	pager = \
	    less -FRX
```

| key | `git config --get` | `File::raw_value()` before | after |
|---|---|---|---|
| `alias.lg` | `log --oneline --graph` | `\t\tlog --oneline --graph` | `log --oneline --graph` |
| `core.pager` | `less -FRX` | `\t    less -FRX` | `less -FRX` |

Measured with `git 2.50.1 (Apple Git-155)` against `gix-config` at `e3a6fa151`.

The failure is silent: the alias still "works", it just runs with leading tabs glued to the command. `git-config(1)` says whitespace surrounding `name`, `=` and `value` is discarded, that the backslash and end-of-line characters are discarded, and that a value needing leading whitespace **must** be double-quoted — so unquoted indentation can never be part of the value. `config.c: parse_value()` implements exactly that by skipping whitespace while `cs->value.len == 0`, and `\`+`\n` only `continue`s, so the skipping carries across continuations.

Cause, in `value()` (`gix-config/src/parse/from_bytes/mod.rs`): the continuation arm sets `value_start` to the first byte of the next line, so that line's indentation lands inside the `ValueNotDone`/`ValueDone` span that every consumer concatenates.

### What this does

`value()` now emits that indentation as `Event::Whitespace` for as long as every chunk seen so far has been empty. That is the sequence `Event::ValueNotDone` already documents — "A `Newline` event usually follows, followed by either `ValueDone`, `Whitespace`, or another `ValueNotDone`" — and it was simply never produced.

Because the bytes stay in an event, `to_bstring()` still round-trips the file unchanged, and every consumer (`file::impls`, `file::section::body`, `file::access::raw`, `file::mutable::*`) already ignores `Whitespace` when concatenating value chunks.

Indentation that follows actual content is untouched, matching `git`: `k = abc\` + `\n\tdef` stays `abc\tdef`, and the existing multi-line-alias test — which begins on the `=` line — keeps its interior whitespace.

### Tests

- `parse::from_bytes::tests::value_continuation::indentation_of_the_line_a_value_starts_on_is_insignificant` (LF and CRLF) and `…::indentation_is_insignificant_until_the_value_has_content` (a whitespace-only continuation line keeps the value empty, so the *next* line's indentation is insignificant too; and the negative case).
- `file::access::read_only::multi_line_value_starting_on_a_continuation_line_is_not_indented`, next to `multi_line_value_with_empty_continuation_line`, covering five continuation shapes plus the negative case. Every expected value was taken from real `git config --get` output, not derived.

All three fail on `main` and pass with the patch; the rest of the `gix-config` suite (396 tests) is unchanged. As a broader check, a 1262-file generated config corpus compared against `git config --list -z` went from 113 divergences to 1 (`[core]\ra = b\r`, CR-only line endings — pre-existing and unrelated).

### Known related gap, deliberately not addressed

`a = ""  x` still yields `  x` where `git` yields `x`: same rule, but the whitespace sits inside a chunk rather than at its start, so the span-based parser cannot drop it without unquoting during parsing. Happy to look at that separately if it's worth it.

### DIFF
--- a/gix-config/src/parse/from_bytes/mod.rs
+++ b/gix-config/src/parse/from_bytes/mod.rs
@@ -1,6 +1,9 @@
 use bstr::{BStr, ByteSlice};
 
-use crate::parse::{Comment, Error, Event, MaybeDecoded, Span, error::ParseNode, section};
+use crate::{
+    parse::{Comment, Error, Event, MaybeDecoded, Span, error::ParseNode, section},
+    value::is_git_whitespace,
+};
 
 type ParseResult<T> = Result<T, ()>;
 
@@ -293,9 +296,6 @@ fn config_value(backing: &[u8], i: &mut &[u8], dispatch: &mut dyn FnMut(Event)) 
     if let Some(rest) = i.strip_prefix(b"=") {
         *i = rest;
         dispatch(Event::KeyValueSeparator);
-        if let Ok(whitespace) = take_spaces1(i) {
-            dispatch(Event::Whitespace(Span::new(backing, whitespace)));
-        }
         value(backing, i, dispatch)
     } else {
         dispatch(Event::Value(Span::default()));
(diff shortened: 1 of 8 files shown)


Byron:
Thanks a lot!

PR #2973, opened 2026-09-07
fix(gix-config-value): accept git's 12-bit hex color shorthand

rawsun007:
Written by Claude Opus 5 in Claude Code, through Roshan's account, per the identification rule in CONTRIBUTING.md. He has not read the diff line by line yet.

`git-config(1)` documents two hex forms: "24-bit RGB values as hex, like #ff0ab3, or 12-bit RGB values like #f1b, which is equivalent to the 24-bit color #ff11bb". Only the 24-bit form parsed, because the hex branch required exactly six digits, so `git` accepted `#f1b` and `gix` rejected it. Same parity gap as #2967, one type over.

Both lengths are handled now, doubling each shorthand digit. `Name::Rgb` is unchanged, so a shorthand parses to the long form and renders back as six digits: `#f1b` displays as `#ff11bb`. Git calls the two equivalent, but it is a normalisation, so say the word if you would rather keep the input spelling.

One existing assertion changed: `name::invalid` asserted `Name::from_str("#fff").is_err()`, which pinned the old behaviour. It is replaced by the lengths `git` really rejects.

Found by parsing ~110 colour specs with both `git` and `Color::try_from` and diffing the verdicts; `#RGB` was the only disagreement. Expected values recorded from `git config --type=color` on git 2.50.1. `cargo test -p gix-config-value` passes 56; reverting only `src/color.rs` fails exactly the one new test. `cargo clippy` reports only the pre-existing removed-lint warning that every crate in the workspace emits.

### DIFF
--- a/gix-config-value/src/color.rs
+++ b/gix-config-value/src/color.rs
@@ -115,9 +115,10 @@ impl TryFrom<BString> for Color {
 /// Discriminating enum for names of [`Color`] values.
 ///
 /// `git-config` supports the eight standard colors, their bright variants, an
-/// ANSI color code, or a 24-bit hex value prefixed with an octothorpe/hash.
-/// Color names and the `bright` prefix are matched case-insensitively, and
-/// `bright` may only precede one of the eight standard colors.
+/// ANSI color code, or a hex value prefixed with an octothorpe/hash. The hex value
+/// is either 24-bit, like `#ff11bb`, or the 12-bit shorthand `#f1b`, which stands
+/// for the same color. Color names and the `bright` prefix are matched
+/// case-insensitively, and `bright` may only precede one of the eight standard colors.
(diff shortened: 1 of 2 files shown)


rawsun007:
Both Codex findings were correct and both cited real rules in AGENTS.md, which I had read only for its AI clause. Fixed in 104cc54: the `.unwrap()` is now `.expect("the shorthand parses, as asserted above")`, and every new `assert*!` carries a description naming the input and the invariant. 56 tests still pass and the mutation check still fails exactly the one new test.

The test comments recording where the expected values came from I have kept, since Test Best Practices asks for clear-text breadcrumbs about what makes a fixture special.

Posted by Claude Opus 5 running in Claude Code, through Roshan's account, per CONTRIBUTING.md.

Byron (approved):
Great catch, thanks!

PR #2984, opened 2026-09-09
fix(gix-config-value)!: expand `~` and `~user` like `git` does

justonemorenight:
Written with assistance from the Zed coding agent through @justonemorenight's account, per the identification rule in `CONTRIBUTING.md`.

Git expands `~` and `~user` paths via `interpolate_path()` in `path.c`. In `gix-config-value`, `Path::interpolate()` previously required a trailing slash for the current user (`~/`) and required a slash for named users (`self.starts_with(b"~") && self.contains(&b'/')`). Consequently:
- A standalone `~` was left unexpanded as a literal `"~"`.
- A standalone `~user` without a trailing slash was left unexpanded as a literal `"~user"`.
- A nonexistent user without a slash (`~nonexistent`) silently returned the literal instead of producing a user lookup error.

### Parity with Git

Recorded from `git -c t.k=<input> config --type=path t.k` on git 2.50.1:

| input | git | before | after |
|---|---|---|---|
| `~` | `/home/user` | `"~"` | `/home/user` |
| `~/` | `/home/user/` | `/home/user` | `/home/user` |
| `~/foo` | `/home/user/foo` | `/home/user/foo` | `/home/user/foo` |
| `~user` | `/home/user` | `"~user"` | `/home/user` |
| `~user/` | `/home/user/` | `/home/user` | `/home/user` |
| `~user/foo` | `/home/user/foo` | `/home/user/foo` | `/home/user/foo` |
| `~nonexistent` | error | `"~nonexistent"` | error |

### Changes
- Expand `~` to `home_dir` when standalone, matching `~/`.
- In `interpolate_user`, allow paths without `/` (e.g. `~user`) to expand directly to the user's home directory.
- Update `tests/value/path.rs`: replace `tilde_alone_does_not_interpolate` with `tilde_alone_substitutes_current_user`, and add tests for `~user` and missing user error cases.

Marked breaking (`!`): `~` and `~user` without trailing slashes now interpolate instead of returning literal strings.

### Verification
- `cargo test -p gix-config-value` (56 tests + 2 doctests pass)
- `cargo test -p gix-config` (366 tests pass)
- `cargo clippy -p gix-config-value` (clean)
- `cargo fmt --check` (clean)

### DIFF
--- a/gix-config-value/src/path.rs
+++ b/gix-config-value/src/path.rs
@@ -147,10 +147,10 @@ impl From<&str> for Path {
 impl Path {
     /// Interpolates this path into a path usable on the file system.
     ///
-    /// If this path starts with `~/` or `~user/` or `%(prefix)/`
-    ///  - `~/` is expanded to the value of `home_dir`. The caller can use the [dirs](https://crates.io/crates/dirs) crate to obtain it.
+    /// If this path starts with `~/` or `~` or `~user` or `%(prefix)/`
+    ///  - `~` or `~/` is expanded to the value of `home_dir`. The caller can use the [dirs](https://crates.io/crates/dirs) crate to obtain it.
     ///    If it is required but not set, an error is produced.
-    ///  - `~user/` to the specified user’s home directory, e.g `~alice` might get expanded to `/home/alice` on linux, but requires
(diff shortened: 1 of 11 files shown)


Byron:
Thanks for the hint!

PR #2986, opened 2026-09-09
fix(gix-config-value): expand ~user with no path after it

rawsun007:
written by Claude Opus 5 in Claude Code, through Roshan's account, per the identification rule in CONTRIBUTING.md. he has not read the diff line by line yet.

third type in this crate after #2967 and #2980, same method: run the inputs through `git config --type=path` and diff.

on git 2.50.1, everything matched except the tilde with no `/` after it:

| input | git | before |
| --- | --- | --- |
| `~root` | that user's home | `~root`, unchanged |
| `~root/` | home, trailing slash kept | already right |
| `~root/x` | home joined with `x` | already right |
| `%(prefix)` | `%(prefix)`, unchanged | already right |
| `~nosuchuser/x` | error | already right |

`git` takes everything past the `~` as the user name when there is no `/`, so the fix is to stop requiring one. `interpolate_user` now takes the whole remainder as the name in that case and returns the home directory as-is.

one thing i deliberately did not change, because it looks like your decision rather than an oversight. `git` also expands a bare `~` to the home directory, and this crate returns it unchanged, which `tilde_alone_does_not_interpolate` asserts. that test goes through `interpolate_without_context`, so what it really pins is `~` with no `home_dir` in the `Context`, where returning the literal is a reasonable lenience and erroring would be the alternative. note `~/x` does error there, so the two are inconsistent today. happy to make `~` expand whenever `home_dir` is set and keep the literal when it is not, which would close the gap without touching that test, if that is the behaviour you want.

`cargo test -p gix-config-value` is 59 passing. reverting only `src/path.rs` fails exactly the new test. `cargo fmt` applied; clippy reports only the pre-existing removed-lint warning the workspace emits.

### DIFF
--- a/gix-config-value/src/path.rs
+++ b/gix-config-value/src/path.rs
@@ -194,7 +194,7 @@ impl Path {
                 err,
             })?;
             Ok(home_path.join(val))
-        } else if self.starts_with(b"~") && self.contains(&b'/') {
+        } else if self.starts_with(b"~") && self.len() > 1 {
             self.interpolate_user(home_for_user.ok_or(interpolate::Error::Missing {
                 what: "home for user lookup",
             })?)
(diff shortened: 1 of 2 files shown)


Byron:
You must be running the same workflow: Duplicate of https://github.com/GitoxideLabs/gitoxide/pull/2984

rawsun007:
my fault, i should have searched for an open pr before starting. #2984 predates mine by hours and is the better change anyway, it does the bare `~` half i left alone and threads it through `gix-config` and `gix`, which i had not looked at. sorry for the triage time.

codex was right on both counts too, for the record: the new assertion had no description and i did not update the rustdoc, and i had already been pulled up on the assertion rule in #2973.

Assisted-by: Claude Opus 5 (Claude Code).

PR #2993, opened 2026-09-14
fix: make [[:blank:]] and [[:space:]] match what Git matches

youdie006:
`AGENTS.md` says to use git itself as the reference implementation, so I did: 12 POSIX classes x 254 bytes through `git check-ignore --no-index --stdin -z`, compared against `gix_glob::wildmatch`. 6 mismatches out of 3048, all in these two classes.

| class | Git | `gix-glob` before |
|---|---|---|
| `[[:blank:]]` | `{SP, TAB}` | `is_ascii_whitespace()` = `{SP, TAB, LF, FF, CR}` — too wide |
| `[[:space:]]` | `{SP, TAB, LF, CR}` | `t_ch == b' '` = `{SP}` — too narrow |

Neither maps to the predicate of the same name: Git's `isspace()` reads its own `sane_ctype` table, which excludes vertical tab and form feed, so it is narrower than both C `isspace` and Rust's `is_ascii_whitespace`. `[[:print:]]` at `gix-glob/src/wildmatch.rs:341` is already hand-written as `(0x20u8..=0x7e).contains(&t_ch)` for the same kind of reason; `b"space"` was the only class left as a single-byte equality.

Reached in production from `gix-ignore/src/search.rs:145` through `gix_glob::Pattern::matches_repo_relative_path` (`gix-glob/src/pattern.rs:164`), so a `.gitignore` line of `*[[:space:]]*` does not ignore a tab-containing filename here while `git check-ignore` does.

<details>
<summary>Verification</summary>

Base `c609062db5`. The oracle was validated against 8 known answers first (`digit`/`5` and `/x`, `alpha`/`Q` and `/7`, `upper`/`A` and `/a`, `xdigit`/`f` and `/g`) before any number was trusted, and the 12-row table was reproduced a second way — real files on disk plus `git status --ignored --porcelain -z` — so it is not an artefact of how `--stdin` handles control bytes. Before: 6 mismatches. After: 0.

Every row below ran exactly one test; each anchor's occurrence count was asserted as 1 before writing.

| row | `wildmatch.rs` md5 | result |
|---|---|---|
| fix | `cdf10d14…` | ok |
| base | `673c2b32…` | FAILED |
| revert `blank` only | `b06752c6…` | FAILED |
| revert `space` only | `6881e765…` | FAILED |
| over-correct `space` +VT | `77927c70…` | FAILED |
| under-correct `space` -CR | `ec29687e…` | FAILED |
| over-correct `blank` +LF | `bfe08d43…` | FAILED |
| under-correct `blank` -TAB | `b14dcc90…` | FAILED |
| fix restored | `cdf10d14…` | ok |

Both classes are mutated in both directions, so the test does not stop one short in either.

Gates: `cargo fmt --all -- --check` exit 0; `cargo clippy -p gix-glob --all-targets --all-features -- -D warnings -A unknown-lints --no-deps` exit 0; `cargo doc -p gix-glob` exit 0; `cargo test -p gix-glob` 41 + 1 passed; and the same on the declared MSRV 1.88.0.

I could not run the `gix-ignore` suite: `cargo check -p gix-hash` fails on this machine at the unmodified base as well (identical exit on base and on this branch), and `gix-hash` is not in `gix-glob`'s dependency tree, so it is unrelated to this change.

</details>

One thing I left alone: `gix-glob/tests/wildmatch/mod.rs:287-296` has a `// TODO: reproduce these` block of five rows dropped from the `t3070-wildmatch.sh` port. I have not looked into those.

---

Disclosure, per CONTRIBUTING: I am an AI agent (Claude) working through this account. The measurements above are from runs on this machine.


Byron:
Good catch!

PR #2999, opened 2026-09-15
Add configurable CycloneDX and SPDX SBOM generation

Byron:
## Tasks

- [x] refackiew

----

Everything below this line was generated by `Codex GPT-6`.

Adds `just sbom-install`, `just sbom`, and `just sbom-test` to generate CycloneDX 1.5 and SPDX 2.3 JSON using Cargo-installed tools and Rust code in `jtt`.

`just sbom` inventories all workspace members with all features and platforms. `--package` selects one crate with its default features on the host platform, with support for `--features`, `--no-default-features`, `--all-features`, `--target`, and `--output-dir`. For example:

```sh
just sbom --package gitoxide --no-default-features --features small
just sbom --package gix --features blocking-http-transport-reqwest-rust-tls
```

Resolution uses a temporary workspace and a copy of `Cargo.lock` so unrelated members cannot enable extra HTTP/TLS backends. `cargo tree` supplies the final graph, respecting platform-specific feature activation and host build dependencies. Inventories include runtime and build dependencies, exclude dev-only dependencies, and preserve original package identities, including binary-only workspace members. Both output files are published after generation and conversion succeed.

`cargo deny` cannot export these formats. The pinned tools are `cargo-cyclonedx` 0.5.9 and `sbom-tools` 0.2.0; the latter converts the same CycloneDX inventory to SPDX. Conversion reports metadata loss, such as CycloneDX properties, on stderr. These are Cargo dependency inventories; they do not scan system libraries or compiled binaries.

## Validation

- `just sbom-test`: isolated offline fixtures cover library/binary features, target filtering, package/version parity, valid references, output paths with spaces, source lockfile preservation, and failed conversion preserving previous output.
- Full `gix-testtools` tests and doctests with `sha1,sbom` and `GIX_TEST_IGNORE_ARCHIVES=1`.
- Focused Clippy with `--no-deps` and `-D warnings`, `cargo machete`, stable Rust and justfile formatting, `just check-mode`, and script-mode help.
- Real inventories for all 71 workspace members, `gitoxide` with `small` and `max-pure`, and `gix` without HTTP, with Reqwest/Rustls, and with Curl/OpenSSL. Both formats have matching package versions and valid references; backend selection and the unchanged source lockfile were verified.

Clippy without `--no-deps` hits an existing `collapsible_if` warning in `gix-tempfile/src/registry.rs` with the installed Rust toolchain. The focused check passes.

## Reported issue

> Add tooling to the justfile to create an SBOM in both formats. Using cargo deny would be good if possible, and generally, `cargo` based tooling.

> Ideally it's possible to specify a crate and its features as well, particularly relevant for the `gix` binary, but also for the `gix` crate which has vastly different dependencies depending on which HTTP backend is chosen, if at all, or which TLS implementation.

> Can we avoid python please? If custom tooling is needed, rather use `jtt` in this repository.


Byron:
CC @EliahKagan just as FYI. It will be a while until I try and review this.

Byron:
Let's get this merged, also improves the justfile

PR #3008, opened 2026-09-21
perf(gix-hash): use sha1dc for sha-1 object ids

srijs:
Swaps the SHA-1 backend of `gix-hash` to [`sha1dc`](https://crates.io/crates/sha1dc), which detects the same collision attacks but uses native sha-1 and vector instructions where available.

## Performance

Apple M4 (4P/6E), against a bare clone of `git/git`: 421,292 objects, 305 MiB pack, 7.74 GiB of uncompressed object content, median object 10 KiB. Both binaries built from one tree and one target directory with only the backend toggled; fastest of 5 rounds, rotating between builds each round.

| operation | `sha1-checked` | `sha1dc` | |
|---|---:|---:|---:|
| `gix free pack verify` (all cores) | 2.36s | 1.09s | **2.17x** |
| `gix free pack verify` (1 thread) | 12.59s | 6.07s | **2.07x** |
| `gix free pack index create` (all cores) | 4.28s | 2.81s | **1.52x** |
| `gix free pack index create` (1 thread) | 14.69s | 7.82s | **1.88x** |


Byron:
Thanks. so. much!

It is miraculous, and propels `gitoxide` way ahead of Git once again, but with SHA-1 collision detection enabled this time!

I have one issue to wait for, which follows. If you think this should be fixed, I am happy to merge once the patch was released.

### Astra review of `sha1dc`

> Offline tests and extensive reference comparisons passed for scalar, NEON and SSE2 paths, but big-endian ARM acceleration has a correctness defect. Production readiness remains conditional: fix or exclude that path; native SHA-NI/AVX2 execution and the exact locked-dependency build were not validated here.

- [P2] Keep big-endian AArch64 on the scalar compression path — /Users/byron/dev/github.com/srijs/sha1dc/src/block/backend/armv8.rs:40-41
  On `aarch64_be` targets with SHA-1 acceleration enabled, this unconditional byte reversal produces an incorrect message schedule. Cross-compiling the existing `msg::<0>` and `spill::<0>` functions turns input bytes `01 23 45 67` into `0x67452301`, whereas the scalar implementation produces `0x01234567`. Collision filtering, recompression and mitigation consequently consume incorrect words. Restrict this backend to little-endian targets until an endian-correct implementation is validated.

Here is a patch I had it create, making the big endian target conditional.

<details><summary>Patch</summary>
<p>


```patch
diff --git a/src/block/backend.rs b/src/block/backend.rs
index 77d8ce6..771e167 100644
--- a/src/block/backend.rs
+++ b/src/block/backend.rs
@@ -10,7 +10,8 @@
 use crate::BLOCK_SIZE;
 use crate::block::rounds;
 
-#[cfg(target_arch = "aarch64")]
+// The ARMv8 backend's message loads assume a little-endian target.
+#[cfg(all(target_arch = "aarch64", target_endian = "little"))]
 mod armv8;
 mod scalar;
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
@@ -22,7 +23,11 @@ mod sha_ni;
 /// a system call, so a `no_std` build uses `target_feature` only. Such a build
 /// needs the features on the command line, for example
 /// `-C target-feature=+sha2`.
-#[cfg(any(target_arch = "x86", target_arch = "x86_64", target_arch = "aarch64"))]
+#[cfg(any(
+    target_arch = "x86",
+    target_arch = "x86_64",
+    all(target_arch = "aarch64", target_endian = "little")
+))]
 fn has_sha1_instructions() -> bool {
     #[cfg(all(feature = "std", any(target_arch = "x86", target_arch = "x86_64")))]
     {
@@ -54,7 +59,7 @@ fn has_sha1_instructions() -> bool {
 enum Repr {
     #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
     ShaNi,
-    #[cfg(target_arch = "aarch64")]
+    #[cfg(all(target_arch = "aarch64", target_endian = "little"))]
     Armv8,
     Scalar,
 }
@@ -66,11 +71,15 @@ pub(crate) struct Backend(Repr);
 
 impl Backend {
     pub(crate) fn new() -> Self {
-        #[cfg(any(target_arch = "x86", target_arch = "x86_64", target_arch = "aarch64"))]
+        #[cfg(any(
+            target_arch = "x86",
+            target_arch = "x86_64",
+            all(target_arch = "aarch64", target_endian = "little")
+        ))]
         if has_sha1_instructions() {
             #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
             return Self(Repr::ShaNi);
-            #[cfg(target_arch = "aarch64")]
+            #[cfg(all(target_arch = "aarch64", target_endian = "little"))]
             return Self(Repr::Armv8);
         }
 
@@ -101,7 +110,7 @@ impl Backend {
                 // SAFETY: `Repr::ShaNi` means the features were checked.
                 unsafe { sha_ni::compress_spill(state, block, m1) };
             }
-            #[cfg(target_arch = "aarch64")]
+            #[cfg(all(target_arch = "aarch64", target_endian = "little"))]
             Repr::Armv8 => {
                 // SAFETY: `Repr::Armv8` means the features were checked.
                 unsafe { armv8::compress_spill(state, block, m1) };
@@ -166,7 +175,7 @@ mod tests {
         match backend.0 {
             #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
             Repr::ShaNi => true,
-            #[cfg(target_arch = "aarch64")]
+            #[cfg(all(target_arch = "aarch64", target_endian = "little"))]
             Repr::Armv8 => true,
             Repr::Scalar => false,
         }
@@ -177,7 +186,11 @@ mod tests {
     #[test]
     fn hardware_is_used_when_the_target_guarantees_it() {
         let guaranteed = cfg!(any(
-            all(target_arch = "aarch64", target_feature = "sha2"),
+            all(
+                target_arch = "aarch64",
+                target_endian = "little",
+                target_feature = "sha2"
+            ),
             all(
                 any(target_arch = "x86", target_arch = "x86_64"),
                 target_feature = "sha",
@@ -194,6 +207,12 @@ mod tests {
         }
     }
 
+    #[cfg(all(target_arch = "aarch64", target_endian = "big"))]
+    #[test]
+    fn big_endian_aarch64_uses_scalar() {
+        assert!(matches!(Backend::new().0, Repr::Scalar));
+    }
+
     #[test]
     fn hardware_agrees_with_scalar_compression() {
         let hardware = Backend::new();
```


</p>
</details> 



### My own review

I ran Astra on this PR which was uneventful and look at all changes in detail myself.
It's all unventful, while only making the code simpler.

### Benchmarks

Here is my own standard benchmark, with a clear 2x speedup.

```
‚ùØ hyperfine -w1 -M2 'gix free pack verify -a less-time /Users/byron/dev/github.com/phpstan/phpstan/.git/objects/pack/pack-fc5a93b6903e792e5c65ce2f8f02054775c35c94.idx' './target/release/gix free pack verify -a less-time /Users/byron/dev/github.com/phpstan/phpstan/.git/objects/pack/pack-fc5a93b6903e792e5c65ce2f8f02054775c35c94.idx'
Benchmark 1: gix free pack verify -a less-time /Users/byron/dev/github.com/phpstan/phpstan/.git/objects/pack/pack-fc5a93b6903e792e5c65ce2f8f02054775c35c94.idx
  Time (mean ¬± œÉ):     23.520 s ¬±  0.021 s    [User: 348.231 s, System: 4.384 s]
  Range (min ‚Ä¶ max):   23.505 s ‚Ä¶ 23.535 s    2 runs

Benchmark 2: ./target/release/gix free pack verify -a less-time /Users/byron/dev/github.com/phpstan/phpstan/.git/objects/pack/pack-fc5a93b6903e792e5c65ce2f8f02054775c35c94.idx
  Time (mean ¬± œÉ):     11.463 s ¬±  0.149 s    [User: 169.851 s, System: 2.906 s]
  Range (min ‚Ä¶ max):   11.358 s ‚Ä¶ 11.569 s    2 runs

Summary
  ./target/release/gix free pack verify -a less-time /Users/byron/dev/github.com/phpstan/phpstan/.git/objects/pack/pack-fc5a93b6903e792e5c65ce2f8f02054775c35c94.idx ran
    2.05 ¬± 0.03 times faster than gix free pack verify -a less-time /Users/byron/dev/github.com/phpstan/phpstan/.git/objects/pack/pack-fc5a93b6903e792e5c65ce2f8f02054775c35c94.idx
```

### Codex


Apple M4 Max (12P/4E), 64 GB RAM. Single-run timings; Linux uses the completed warm-up pair.

- **PHPStan:** 100,714 objects, 6.43 GiB pack, 162.57 GiB uncompressed.
- **Linux:** 10,922,421 objects, 1.78 GiB pack.

| operation | installed `gix` (`sha1-checked`) | branch `gix` (`sha1dc`) | speedup |
|---|---:|---:|---:|
| PHPStan: `gix free pack verify -a less-time` (all cores) | 28.32s | 12.42s | **2.28√ó** |
| PHPStan: `gix free pack verify -a less-time` (1 thread) | 280.90s | 140.23s | **2.00√ó** |
| Linux: `gix free pack index create` (all cores) | 34.63s | 22.56s | **1.53√ó** |

Linux indexing used `-p <linux.pack>` with output discarded; both binaries reproduced the existing index checksum. Its single-thread case wasn‚Äôt measured.

These compare the requested existing binaries: baseline `6b0ae90bab` and branch `0d91e711ea`, with different compiler versions too.

Byron (approved):
Approving to indicate intent, but switching to draft until it's clear if `sha1dc` needs fixing.

srijs:
Thanks for the review. I’ll take a look in a minute and will get back to you!

srijs:
Okay, so yes, this is a bug _iff_ your aim is to support big-endian aarch64 targets.

Why do I make this so conditional? All `aarch64_be-*` targets in Rust are Tier 3 (https://doc.rust-lang.org/rustc/platform-support.html#tier-3). So the Rust project itself doesn't do any automated tests for these, and in fact, while trying to see if I could produce a fixed version, there seems to be at least one bug in `core::arch::aarch64` related to bit ordering in intrinsics.

Do you know if there's any users with `aarch64_be` using `gitoxide` today?

Happy to go either way. I can add the guarding in, but also full disclosure, given these are Tier 3 targets there don't seem to be pre-built `std` or `core` libraries available and so it gets quite difficult to test in CI, which makes me less confident this won't regress again in the future.

Byron:
Thanks for looking into this!

If there is such a thing called 'over-correctness' then I think the patch above yields exactly that. Maybe it's defensive correctness, as that will be correct even once some ominous `aarch64_be` gets its own `std` builds.

To me it sounds like it won't be a problem in quite some while, and its early solution (if the patch is one) is squarely in upstream territory. My gut feeling still is to apply that patch, but I also admit that I'd do so out of fear 😅.

In any case, thanks so much!!
This is awesome.
Thanks to you, `gitoxide` is fast again.

